PR #188, opened 2014-12-17
Add getUiViewForToken and getExternalWebSession methods to hack-session

jparyani:
`getUiViewForToken` is pretty straightforward. I just made `getProxyForApiToken` accessible from proxy.js and called that directly. I also added `getApiEndpoint` because I realized I need a way to tell inside an app when a request is hitting the api server and I should be using `getUiViewForToken`. Should this perhaps be hidden from apps and handled seamlessly by the backend?

`getExternalWebSession` was a little more involved. Mostly it was a reimplementation of parts of sandstorm-http-bridge.c++ in node.js. Do you think we should be passing along cookies, or should they be dropped?

### DIFF
--- a/shell/server/drivers/external-ui-view.js
+++ b/shell/server/drivers/external-ui-view.js
@@ -22,20 +22,11 @@ var Http = Npm.require("http");
 var Https = Npm.require("https");
 var ApiSession = Capnp.importSystem("sandstorm/api-session.capnp").ApiSession;
 
-var removeTrailingSlash = function (path) {
-  if (path.indexOf("/", path.length - 1) === -1) {
-    return path;
-  }
-
-  return path.slice(0, path.length - 1);
-};
-
-WrappedUiView = function (token, proxy, path) {
+WrappedUiView = function (token, proxy) {
   // TODO(someday): handle the fact that these proxies will be garbage collected every 2 minutes,
   // even if it's in use.
   this.token = token;
   this.proxy = proxy;
-  this.path = path;
 };
 
 WrappedUiView.prototype.newSession = function (userInfo, context, sessionType, sessionParams, retryCount) {
@@ -53,8 +44,8 @@ WrappedUiView.prototype.newSession = function (userInfo, context, sessionType, s
     // This will allow the caller to request only a subset of the permissions
     // granted by the token, which is useful to protect against bugs.
     var session = self.proxy.uiView.newSession(
-        self.proxy.userInfo, context, sessionType, sessionParams).session.castAs(ApiSession);
-    return {session: new Capnp.Capability(new WrappedApiSession(session, self.path), ApiSession)};
+        self.proxy.userInfo, context, sessionType, sessionParams).session;
+    return {session: session};
   }).catch(function (error) {
     return self.proxy.maybeRetryAfterError(error, retryCount).then(function () {
       return self.newSession(userInfo, context, sessionType, sessionParams, retryCount + 1);
@@ -62,36 +53,11 @@ WrappedUiView.prototype.newSession = function (userInfo, context, sessionType, s
   });
 };
 
-WrappedApiSession = function (session, path) {
-  this.session = session;
-  this.path = removeTrailingSlash(path);
-};
-
-WrappedApiSession.prototype._makePath = function (path) {
-  return this.path + path;
-};
-
-WrappedApiSession.prototype.get = function (path, context) {
-  return this.session.get(this._makePath(path), context);
-};
-
-WrappedApiSession.prototype.post = function (path, content, context) {
-  return this.session.post(this._makePath(path), content, context);
-};
-
-WrappedApiSession.prototype.put = function (path, content, context) {
-  return this.session.put(this._makePath(path), content, context);
-};
-
-WrappedApiSession.prototype.delete = function (path, context) {
-  return this.session.delete(this._makePath(path), context);
-};
-
-getWrappedUiViewForToken = function (token, path) {
+getWrappedUiViewForToken = function (token) {
   var proxyPromise = getProxyForApiToken(token);
 
   return proxyPromise.then(function (proxy) {
-    return {view: new WrappedUiView(token, proxy, path)};
+    return {view: new WrappedUiView(token, proxy)};
   });
 };
 
@@ -122,7 +88,6 @@ ExternalWebSession = function (url, grainId, options) {
   this.host = parsedUrl.hostname;
   this.port = parsedUrl.port;
   this.protocol = parsedUrl.protocol;
-  this.path = removeTrailingSlash(parsedUrl.path);
   this.grainId = grainId;
   this.options = options || {};
 };
@@ -161,16 +126,12 @@ var responseCodes = {
   505: {type: "serverError"}
 };
 
-ExternalWebSession.prototype._makePath = function (path) {
-  return this.path + path;
-};
-
 ExternalWebSession.prototype._requestHelper = function (method, path, context, content, contentType) {
   var session = this;
   return new Promise(function (resolve, reject) {
     var options = _.clone(session.options);
     options.headers = options.headers || {};
-    options.path = session._makePath(path);
+    options.path = path;
     options.method = method;
     if (contentType) {
       options.headers["content-type"] = contentType;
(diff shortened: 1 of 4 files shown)


src/sandstorm/hack-session.capnp
```diff
@@ -104,6 +105,15 @@ interface HackSessionContext @0xe14c1f5321159b8f
   revokeApiToken @5 (tokenId :Text);
   # Revoke (delete) a previously-generated token.
 
+  getUIViewForToken @6 (token :Text) -> (session: Grain.UiSession);
```
kentonv: I think this should return UiView. It should be up to the caller to call newSession() and choose the appropriate WebSession.Params (currently you seem to be filling in invalid dummy values).

Also, nitpick: The 'I' should be lowercase, i.e. `getUiViewForToken`.

shell/server/mail.js
```diff
@@ -377,18 +378,23 @@ HackSessionContextImpl.prototype.getPublicId = function() {
   }).bind(this));
 };
 
+var findRequestMethod = function (url) {
+  if (url.indexOf('https://') === 0) {
```
kentonv: Nitpick: It's better to write: `url.lastIndexOf('https://', 0)`

Otherwise when the url doesn't start with `https://` you'll needlessly search the whole string.

Also, even more nitpicky: can we try to stick to double-quotes? I know, hard to change habits.

kentonv: (Just realized this is copied code, but probably should be fixed anyway.)

shell/server/mail.js
```diff
@@ -535,6 +541,23 @@ HackSessionContextImpl.prototype.revokeApiToken = function (tokenId) {
   }).bind(this));
 }
 
+HackSessionContextImpl.prototype.getUIViewForToken = function (token) {
+  var proxyPromise = getProxyForApiToken(token);
+
+  return proxyPromise.then(function (proxy) {
+    // Just pass in a dummy host so that getSession doesn't complain
+    return {session: proxy.getSession({headers: {host: 'none'}})};
```
kentonv: Hmm, what keeps this proxy from being cleaned up after a few minutes?

src/sandstorm/hack-session.capnp
```diff
@@ -104,6 +105,15 @@ interface HackSessionContext @0xe14c1f5321159b8f
...
+  # Return a UIView corresponding to the token's grain
+
+  getApiEndpoint @7 () -> (endpointUrl :Text);
+  # Return the API's endpoint url
+
+  getExternalWebSession @8 (domain: Text) -> (session :WebSession.WebSession);
```
kentonv: This should perhaps take a WebSession.Params as input, in order to set UserAgent and such.

shell/server/mail.js
```diff
@@ -584,3 +607,221 @@ var getSmtpPool = function () {
...
+  var session = this;
+  return new Promise(function (resolve, reject) {
+    options.headers = options.headers || {};
+
+    // TODO(someday) mimic a browser's user agent?
+    // options.headers['user-agent'] = 'curl/7.38.0';
+
+    // set accept header
+    if ('accept' in context) {
+      options.headers.accept = context.accept; // TODO(soon): fix this
```
kentonv: "fix this"? Can you describe what's broken (or just fix it :) )?

shell/server/mail.js
```diff
@@ -584,3 +607,221 @@ var getSmtpPool = function () {
...
+          // ???
+          err = new Error("Invalid status code " + resp.statusCode + " received in response.");
+          err.nature = "localBug";
+          reject(err);
+          break;
+      }
+    });
+
+    req.on('error', function (e) {
+      e.nature = "networkFailure";
```
kentonv: Needs update for new KJ exception taxonomy.

shell/server/mail.js
```diff
@@ -584,3 +607,221 @@ var getSmtpPool = function () {
...
+    req.setTimeout(15000, function () {
+      req.abort();
+      err = new Error("Request timed out.");
+      err.nature = "localBug";
+      err.durability = "overloaded";
+      reject(err);
+    });
+
+    if (content) {
+      // TODO(someday): support streaming
```
kentonv: Outbound streaming is handled using a completely different set of method (`putStreaming` / `postStreaming`), so this comment probably doesn't belong here.

kentonv:
> getUiViewForToken is pretty straightforward. I just made getProxyForApiToken accessible from proxy.js and called that directly. I also added getApiEndpoint because I realized I need a way to tell inside an app when a request is hitting the api server and I should be using getUiViewForToken. Should this perhaps be hidden from apps and handled seamlessly by the backend?

Actually, the correct thing is probably for `getUiViewForToken` to take both the host and token parts of the webkey. It should then automatically loopback when the host matches the local API host and fall back to an HTTP session when remote. Of course, that HTTP session needs to send the appropriate `Authorization` header for the remote server being Sandstorm.

Maybe what we really want is to replace all three methods that you've added with a single method:

    getUiViewForEndpoint (url :Text) -> (view :UiView);

Then:
- If the URL is a local webkey, pass back a capability directly.
- If the URL is a remote webkey (i.e. a different Sandstorm server), set up a WebSession that sends the `Authorization` header correctly. You'll need to wrap it in a dummy `UiView` that only supports `WebSession`. (Some day, we'll connect via Cap'n Proto instead, and actually return the remote `UiView` verbatim.)
- If the URL is not a webkey at all, treat it like `getExternalWebSession`. We should actually allow URLs with a path, BTW -- just prefix each request's path with it.

With this approach, apps don't actually have to detect if the address is a webkey at all.

> getExternalWebSession was a little more involved. Mostly it was a reimplementation of parts of sandstorm-http-bridge.c++ in node.js. Do you think we should be passing along cookies, or should they be dropped?

Passing cookies is fine, as long as we don't do something silly like automatically pull them out of the user's browser's cookie jar (not that we have any way to do that).

jparyani:
This has now been refactored into a `getUiViewForEndpoint` method as you suggested.

shell/server/drivers/external-ui-view.js
```diff
@@ -0,0 +1,319 @@
...
+        self.proxy.userInfo, context, sessionType, sessionParams).session.castAs(ApiSession);
+    return {session: new Capnp.Capability(new WrappedApiSession(session, self.path), ApiSession)};
+  }).catch(function (error) {
+    return self.proxy.maybeRetryAfterError(error, retryCount).then(function () {
+      return self.newSession(userInfo, context, sessionType, sessionParams, retryCount + 1);
+    });
+  });
+};
+
+WrappedApiSession = function (session, path) {
```
kentonv: Hmm it didn't occur to me that allowing a path in the API token implies that we need to wrap ApiSession. That's sad because we have to update this code any time ApiSession (or WebSession) gets a new method, and you're actually not even wrapping some of the existing methods now. Also sad for performance.

I don't think we have an actual use case for including a path in an API token, so how about we just disallow it and simplify this?

shell/server/drivers/external-ui-view.js
```diff
@@ -0,0 +1,319 @@
...
+  if (sessionType !== ApiSession.typeId) {
+    throw new Error("SessionType must be ApiSession.");
+  }
+
+  retryCount = retryCount || 0;
+  var self = this;
+
+  return this.proxy.keepAlive().then(function () {
+    // ignore the passed userInfo and instead use the one associated with this token
+    // TODO(someday): handle permissions from the viewInfo
```
kentonv: More specifically:

// TODO(someday): Merge / intersect the two userInfo objects (especially permissions) rather than only taking the token's user info. This will allow the caller to request only a subset of the permissions granted by the token, which is useful to protect against bugs.

shell/server/hack-session.js
```diff
@@ -323,3 +323,20 @@ HackSessionContextImpl.prototype.getIpNetwork = function () {
...
+    var apiHost = ROOT_URL.protocol + "//" + makeWildcardHost("api");
+    var urlProtoAndHost = parsedUrl.protocol + "//" + parsedUrl.host;
+    var token = parsedUrl.hash.slice(1); // Get rid of # which is always the first character
+    if (urlProtoAndHost === apiHost) {
+      return getWrappedUiViewForToken(token, parsedUrl.path);
+    } else {
+      return {view: new ExternalUiView(url, this.grainId, token)};
+    }
+  } else {
+      return {view: new ExternalUiView(url, this.grainId)};
```
kentonv: indentation

jparyani:
Done.

PR #189, opened 2014-12-20
Initial implementation of IpNetwork/IpInterface for hack-session

jparyani:


shell/server/mail.js
```diff
@@ -535,6 +538,256 @@ HackSessionContextImpl.prototype.revokeApiToken = function (tokenId) {
   }).bind(this));
 }
 
+function IpInterfaceImpl () { }
+
+IpInterfaceImpl.prototype.listenTcp = function (portNum, port) {
+  return new Promise(function (resolve, reject) {
+    var resolved = false;
+    var server = Net.createServer(function (connection) {
+      connection.done = connection.end;
```
kentonv: This is a little too clever for me. :)

Can we explicitly define an object here meant to implement the Cap'n Proto interface?

I worry about things like we add a new method to ByteStream which happens to conflict with a method in node's connection type, and then suddenly things inexplicably don't work.

shell/server/mail.js
```diff
@@ -535,6 +538,256 @@ HackSessionContextImpl.prototype.revokeApiToken = function (tokenId) {
...
+
+};
+
+function BoundUdpPortImpl(server, address, port) {
+  this.server = server;
+  this.address = address;
+  this.port = port;
+}
+
+BoundUdpPortImpl.prototype.send = function(message, returnPort) {
```
kentonv: Comment that we're assuming that `returnPort` is always the one corresponding to `this.server`, but that we should really be maintaining a map, but that that requires capability equality which we don't have yet.

shell/server/mail.js
```diff
@@ -535,6 +538,256 @@ HackSessionContextImpl.prototype.revokeApiToken = function (tokenId) {
...
+    // start with lower bits of address and build the output in reverse
+    // this ensures that we can place a colon every 4 characters
+    out += hex[hex.length - 1 - i];
+    if ((i + 1) % 4 === 0) {
+      out += ':';
+      ++numColons;
+    }
+  }
+
+  // need to fill out colons on the upper end of the address til we have 7
```
kentonv: I think a double-colon can be used to mean "fill with as many zeroes as necessary" in an ipv6 address... though I might be making that up.

kentonv:
Can we put all this IP code in a separate file? `mail.js` is getting a little overloaded. :) We should also refactor the non-mail stuff out of mail.js, in a separate change...

jparyani:
All your comments are now fixed.

PR #199, opened 2015-01-06
Fix replyto

### LINKED ISSUE #197: Mail with replyTo bounces erroneously
I've got a sandstorm set up to get mail forwarded from Gmail, and it works fine unless the inbound email has a Reply-To: header set. In which case I get:

    Delivery to the following recipient failed permanently:
    ...
    The error that the other server returned was:
    550 Type error in field: replyTo

I noticed instead of using the mailparser mail.replyTo object, sandstorm emits:

    replyTo: mail.headers['reply-to'] || {}

Changing that line to 

    replyTo: mail.replyTo || {}

Seems to result in a different error:

    550 No field named: 0

So I figure I'm a good bit out of my depth here.

I can reliably reproduce this by sending myself a message to any Google Group.

(aside: Hi Kenton! Not sure if you remember me from our Google days when I was working on OAuth, but I set up my Sandstorm server over the holiday—neat stuff!)

jparyani:
Fixes #197 

Unfortunately we got the protocol wrong a bit for reply-to, so we now bounce emails that have more than 1 address in the reply-to header.

### DIFF
--- a/shell/server/mail.js
+++ b/shell/server/mail.js
@@ -22,7 +22,6 @@ var Future = Npm.require("fibers/future");
 var EmailRpc = Capnp.importSystem("sandstorm/email.capnp");
 var HackSessionContext = Capnp.importSystem("sandstorm/hack-session.capnp").HackSessionContext;
 var Supervisor = Capnp.importSystem("sandstorm/supervisor.capnp").Supervisor;
-var EmailMessage = EmailRpc.EmailMessage;
 var EmailSendPort = EmailRpc.EmailSendPort;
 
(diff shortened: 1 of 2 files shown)


kentonv:
It looks like this will bounce any e-mail with multiple reply-tos. That seems a little harsh. Note that the bounce is potentially going to go back to the e-mail sender who is not necessarily a Sandstorm user. Can we instead just drop the additional reply-tos for now, maybe logging an error so we can get a sense of whether this is at all common?

jparyani:
Ok this is changed to now only log an error.

PR #212, opened 2015-01-24
Create /appdemo/ to launch a demo instance of an app

### LINKED ISSUE #211: Support a "Try it on Sandstorm" button
I want to create a "Try it on Sandstorm" button that app maintainers, for example, MediaGoblin, will enjoy putting on their websites to let visitors test out the app quickly.

Details & discussion here:

https://alpha.sandstorm.io/grain/TfJdgMQXDArRiFYg7vub9w

paulproteus:
The URL structure is /appdemo/:packageId (which is the sha256sum of the
package).

For now, we ignore any query string data provided. Crucially, we ignore a
?url= parameter, which /install/ uses -- since the app demo only works for
apps that are already installed, need no ?url parameter.

This also extracts the code to launch and enter a grain, given either a
package ID or an action ID, into a helper function.

Close #211

### DIFF
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -487,6 +487,56 @@ $KEY</textarea></p>
   </div>
 </template>
 
+<template name="appdemo">
+  <div id="topbar">{{> homeLink}} App demo</div>
+
+  <div id="demo" class="main-content">
+    <div class="centered-box">
+      <h2></h2>
+      {{#if allowDemo}}
+        {{#if isSignedUp}}
+          {{#if isDemoUser}}
+            <p>You are currently logged in as a one-hour demo user.</p>
+            <p><button onclick="browseHome()">Continue the demo</button></p>
+          {{else}}
+            <p>You are already logged in as a real user.</p>
+          {{/if}}
+        {{else}}
+          {{#if loggingIn}}
+            <p>Logging in...</p>
+          {{else}}
+            <p><input id="demo-display-name" type="hidden" value="">
+               <button id="createDemoUser">Try {{ appName }} »</button></p>
+          {{/if}}
+        {{/if}}
+
+        <h3>What is this?</h3>
+        <p>If you click above, you'll open a temporary instance of {{
+          appName }} on the Sandstorm demo server.</p>
+
+	<p>Sandstorm is an open source hosting platform for personal
+          instances of web apps. Users can upload and install
+          arbitrary software. In addition to improving privacy and
+          control,
+          <a href="https://blog.sandstorm.io/news/2014-07-21-open-source-web-apps-require-federated-hosting.html">
+          this is the only way to make Open Source web apps viable.</a></p>
+        <p>Notes:</p>
+        <ul>
+          <li>Demo accounts last one hour.</li>
+          <li>Demo accounts can only install apps from Sandstorm.io, but this
+            is an artificial limitation. Real users can upload whatever code they want, as
+            each app server runs in a secure sandbox.</li>
+          <li>We only have one machine. It may get slow or crashy under excessive load.</li>
+          <li>We know the UI design needs work. This is a tech demo. ;)</li>
+          <li><a href="https://sandstorm.io">Read more at Sandstorm.io.</a></li>
+        </ul>
+      {{else}}
+        <p>Sorry, this server does not allow demo accounts.</p>
+      {{/if}}
+    </div>
+  </div>
+</template>
+
 <template name="demo">
   <div id="topbar">{{> homeLink}} Demo</div>
 
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -488,57 +488,11 @@ $KEY</textarea></p>
 </template>
 
 <template name="appdemo">
-  <div id="topbar">{{> homeLink}} App demo</div>
-
-  <div id="demo" class="main-content">
-    <div class="centered-box">
-      <h2></h2>
-      {{#if allowDemo}}
-        {{#if isSignedUp}}
-          {{#if isDemoUser}}
-            <p>You are currently logged in as a one-hour demo user.</p>
-            <p><button onclick="browseHome()">Continue the demo</button></p>
-          {{else}}
-            <p>You are already logged in as a real user.</p>
-          {{/if}}
-        {{else}}
-          {{#if loggingIn}}
-            <p>Logging in...</p>
-          {{else}}
-            <p><input id="demo-display-name" type="hidden" value="">
-               <button id="createDemoUser">Try {{ appName }} »</button></p>
-          {{/if}}
-        {{/if}}
-
-        <h3>What is this?</h3>
-        <p>If you click above, you'll open a temporary instance of {{
-          appName }} on the Sandstorm demo server.</p>
-
-	<p>Sandstorm is an open source hosting platform for personal
-          instances of web apps. Users can upload and install
-          arbitrary software. In addition to improving privacy and
-          control,
-          <a href="https://blog.sandstorm.io/news/2014-07-21-open-source-web-apps-require-federated-hosting.html">
-          this is the only way to make Open Source web apps viable.</a></p>
-        <p>Notes:</p>
-        <ul>
-          <li>Demo accounts last one hour.</li>
-          <li>Demo accounts can only install apps from Sandstorm.io, but this
-            is an artificial limitation. Real users can upload whatever code they want, as
-            each app server runs in a secure sandbox.</li>
-          <li>We only have one machine. It may get slow or crashy under excessive load.</li>
-          <li>We know the UI design needs work. This is a tech demo. ;)</li>
-          <li><a href="https://sandstorm.io">Read more at Sandstorm.io.</a></li>
-        </ul>
-      {{else}}
-        <p>Sorry, this server does not allow demo accounts.</p>
-      {{/if}}
-    </div>
-  </div>
+  {{> demo}}
 </template>
 
 <template name="demo">
-  <div id="topbar">{{> homeLink}} Demo</div>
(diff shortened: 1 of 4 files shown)


paulproteus:
Notes on this pull request:

* I've manually tested it. Is there automated testing I should be doing?

* This is the first Meteor code I've ever written, beyond doing the tutorial _yesterday_. Let me know if there are Meteor style things I could improve on.

* If some of my comments are unnecessary, just say so, and I'll make 'em go away.

Feedback very welcome.

paulproteus:
Oh, and this uses a new `/appdemo/` URL path.

It doesn't have any smart handling for if you're not a demo user, but rather are a real user.

I could probably move it to `/install/` if we want. And/or I could redirect the visitor to `/install/` if we want. For now, though, we only plan to link to this via people's app pages, and it'll always go straight to demo.sandstorm.io, so it's not clear that we should bother having that feature.

shell/client/shell.html
```diff
@@ -487,6 +487,56 @@
...
+	<p>Sandstorm is an open source hosting platform for personal
+          instances of web apps. Users can upload and install
+          arbitrary software. In addition to improving privacy and
+          control,
+          <a href="https://blog.sandstorm.io/news/2014-07-21-open-source-web-apps-require-federated-hosting.html">
+          this is the only way to make Open Source web apps viable.</a></p>
+        <p>Notes:</p>
+        <ul>
+          <li>Demo accounts last one hour.</li>
+          <li>Demo accounts can only install apps from Sandstorm.io, but this
```
kentonv: Maybe we should change this text to "only install approved apps" (both here and the original text).

shell/client/shell.html
```diff
@@ -487,6 +487,56 @@
...
+            <p><input id="demo-display-name" type="hidden" value="">
+               <button id="createDemoUser">Try {{ appName }} »</button></p>
+          {{/if}}
+        {{/if}}
+
+        <h3>What is this?</h3>
+        <p>If you click above, you'll open a temporary instance of {{
+          appName }} on the Sandstorm demo server.</p>
+
+	<p>Sandstorm is an open source hosting platform for personal
```
kentonv: You should factor out all this common text into a separate template which both the regular demo page and the app demo page can include. (I think the template include syntax is something like {{> templateName}}, but not sure.)

shell/shared/demo.js
```diff
@@ -84,7 +84,21 @@ if (Meteor.isServer) {
         // Log them in on this connection.
         return Accounts._loginMethod(this, "createDemoUser", arguments,
             "demo", function () { return { userId: userId }; });
+      },
+
+      getAppNameFromAppId: function(packageId) {
```
kentonv: This should be a Meteor.publish() instead of a method, since it's just providing access to some of the database. That way:
- If the app name changes the page will update (not really important in this case, but a good habit).
- You can block rendering while you wait for the subscription.
- You don't have to use a session variable to store the result.

shell/shared/demo.js
```diff
@@ -162,3 +176,34 @@ Router.map(function () {
     }
   });
 });
+
+Router.map(function () {
+  this.route("appdemo", {
+    path: "/appdemo/:packageId",
+    waitOn: function () {
+      return Meteor.subscribe("packageInfo", this.params.packageId);
```
kentonv: Hmm, do you actually use this subscription?

shell/shared/demo.js
```diff
@@ -162,3 +176,34 @@ Router.map(function () {
...
+    },
+    onAfterAction: function() {
+      /* Stash the packageId into the Meteor session, so that, after
```
kentonv: Please use line (//) comments rather than block (/**/) comments. (Part of the style guide soon to be written! ;) )

shell/shared/demo.js
```diff
@@ -162,3 +176,34 @@ Router.map(function () {
...
+	 creating a new demo user, we know which app to start a grain
+	 for.
+      */
+      Session.set("packageId", this.params.packageId);
```
kentonv: Hmm, I don't think you need this. Instead, have your data() function include `packageId` in the object it returns. You can access that from your template helper as `this.params.packageId`, I think.

Generally session variables are a way to record actions that the user did which affect the rendering but aren't recorded anywhere else, e.g. "clicked the menu button to expand the menu".

shell/shared/shell.js
```diff
@@ -330,48 +379,70 @@ if (Meteor.isClient) {
...
+    "click .action-required button": function (event) {
+      event.currentTarget.parentNode.parentNode.style.display = "none";
+    },
+  });
 
-        packageId = action.packageId;
-        command = action.command;
-        actionTitle = action.title;
-      }
+  Template.appdemo.helpers({
```
kentonv: Can we keep appdemo template helpers in `demo.js`?

shell/shared/demo.js
paulproteus: Well, I noticed earlier that if I didn't do this, the app name would come up empty. I can re-test to find out if it's needed.

shell/shared/demo.js
paulproteus: For all comments? Or just for one-line comments?

shell/shared/shell.js
paulproteus: +1

shell/shared/demo.js
paulproteus: I can use `packageId` in the template that way, but I couldn't figure out how to use it from Javascript code except by adding it to the session.

shell/shared/demo.js
paulproteus: Interesting. I'm not sure I fully understand how I'd implement this but am tentatively enthusiastic.

shell/client/shell.html
paulproteus: +1

shell/client/shell.html
paulproteus: +1

shell/shared/shell.js
```diff
@@ -330,48 +379,70 @@ if (Meteor.isClient) {
...
-      Meteor.call("newGrain", packageId, command, title, function (error, grainId) {
-        if (error) {
-          console.error(error);
-        } else {
-          Router.go("grain", {grainId: grainId});
+
+      Accounts.callLoginMethod({
+        methodName: "createDemoUser",
+        methodArguments: [displayName],
+        userCallback: function (err) {
```
kentonv: A somewhat more meteor-magic way to do this would be to have /appdemo/:packageId go directly to installing and opening the app if the user is already logged in. Then you don't need a login callback here, because once the login is complete the page will automatically re-render because Meteor.userId() is a reactive data source. :)

paulproteus:
Some other notes:

* Let's stick to /appdemo/

* Let's use appId not packageId as the input to /appdemo/

* Let's make sure to install the package whose package version is highest

paulproteus:
I'm going to close this for now, pending fixing the above.

shell/shared/demo.js
kentonv: All comments.

paulproteus:
Relevant for fixing this:

```
> Packages.findOne({_id: "302a1a51e2dd1dc75455f184b08484c0"}).manifest.actions[0].title
{ defaultText: 'New WordPress Site' }
```

shell/client/shell.html
paulproteus: FWIW, what I've done is turned the `demo` template into something more configurable, and have the `appdemo` template merely load in the entire `demo` template.

shell/shared/demo.js
paulproteus: I was able to totally remove this instead.

shell/shared/shell.js
paulproteus: I will attempt to do this.

paulproteus:
@kentonv requesting your review!

I've manually tested this. The one thing I haven't manually tested is that it works properly in the face of packages with multiple versions.

I'm honestly not super excited about the code to handle packages with multiple versions (and picking the newest one), nor about the fact that I don't think we have a test suite that's useful for verifying that, but I can test it manually.

paulproteus:
I can confirm that my sorting isn't really working:

```
> Packages.findOne({appId: appId}, {sort: {"manifest.appVersion": -1}})
{ _id: '359d2e296aaba05dd1eb87836d438bd3',
  appId: 'mkqyz71cc96e8gmq2h4dhnfm56k90t2055muv33h2ptw1q6h8n4h',
  error: null,
  manifest: 
   { minApiVersion: 0,
     maxApiVersion: 0,
     actions: [ [Object] ],
     continueCommand: { argv: [Object], environ: [Object] },
     appVersion: 5,
     minUpgradableAppVersion: 0 },
  progress: 1,
  status: 'ready' }
> Packages.insert({_id: 'party6', _appId: 'mkqyz71cc96e8gmq2h4dhnfm56k90t2055muv33h2ptw1q6h8n4h', error: null, manifest: { minApiVersion: 0, maxApiVersion: 0, actions: [ {ignored: true} ], appVersion: 6}, progress: 1, status: 'ready'})
'party6'
> Packages.findOne({appId: appId}, {sort: {"manifest.appVersion": -1}})
{ _id: '359d2e296aaba05dd1eb87836d438bd3',
  appId: 'mkqyz71cc96e8gmq2h4dhnfm56k90t2055muv33h2ptw1q6h8n4h',
  error: null,
  manifest: 
   { minApiVersion: 0,
     maxApiVersion: 0,
     actions: [ [Object] ],
     continueCommand: { argv: [Object], environ: [Object] },
     appVersion: 5,
     minUpgradableAppVersion: 0 },
  progress: 1,
  status: 'ready' }
```

paulproteus:
Nope -- huge thanks to @jparyani for showing me that I was INSERTing the wrong thing!

In manual testing, this does find the right Package!

PR #216, opened 2015-01-26
Handle query and hash parameters being passed to grains

jparyani:


shell/shared/grain.js
```diff
@@ -411,7 +411,11 @@ Router.map(function () {
         result.appOrigin = document.location.protocol + "//" + makeWildcardHost(session.hostId);
         setCurrentSessionId(session._id, result.appOrigin, grainId);
         result.sessionId = session._id;
-        result.path = encodeURIComponent("/" + (self.params.path || ""));
+        var currentPath = Iron.Location.get().path;
```
kentonv: Any particular reason to use this rather than just window.location?

jparyani: No, not really. You want me to change it?

kentonv: I think it makes sense to pull the path and hash from the same place for clarity. In fact if you just start with `window.location` as a string, it'll already include `hash`, right?

jparyani: I don't think I want to use window.location as a string since I need to parse out the pathname and hash anyways. It seems much better to use window.location.pathname and window.location.hash directly

kentonv: Fair enough.

paulproteus: For what it's worth:

https://bugzilla.mozilla.org/show_bug.cgi?id=483304

`window.location.hash` differences between Firefox and everyone else.

shell/shared/grain.js
```diff
@@ -411,7 +411,11 @@ Router.map(function () {
...
+        var rootPath = "/grain/" + grainId;
+        var grainPath = currentPath.slice(rootPath.length);
+        result.path = encodeURIComponent(grainPath);
+        result.hash = window.location.hash || "";
```
paulproteus: I... don't know if this will lead to escaping problems in browsers. I think my brain capacity has been exceeded.

jparyani:
Fixed.

I believe the issue that @paulproteus brought up with the fragment is actually not a problem. It seems that firefox behaves a bit differently than every other browser and will unescape characters in the fragment when accessing `window.location.hash`, but it seems to be fine passing these unescaped characters back in as the fragment (even though technically some of the characters might not be allowed according to spec).

paulproteus:
Let me just say I'm super​ excited to hear that the issue I brought up
won't be a problem.

PR #218, opened 2015-01-27
Create /appdemo/ to launch a demo instance of an app

paulproteus:
See commit log.

Sorry that I force pushed; it means that GitHub won't unify this with https://github.com/sandstorm-io/sandstorm/pull/212 .

### DIFF
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -487,6 +487,56 @@ $KEY</textarea></p>
   </div>
 </template>
 
+<template name="appdemo">
+  <div id="topbar">{{> homeLink}} App demo</div>
+
+  <div id="demo" class="main-content">
+    <div class="centered-box">
+      <h2></h2>
+      {{#if allowDemo}}
+        {{#if isSignedUp}}
+          {{#if isDemoUser}}
+            <p>You are currently logged in as a one-hour demo user.</p>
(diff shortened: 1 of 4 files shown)


paulproteus:
Currently rebasing + making one last change. Also I dare not click "Close pull request" anymore.

paulproteus:
@kentonv -- I tried to implement your suggestion from before:

> A somewhat more meteor-magic way to do this would be to have /appdemo/:packageId go directly to installing and opening the app if the user is already logged in. Then you don't need a login callback here, because once the login is complete the page will automatically re-render because Meteor.userId() is a reactive data source. :)

but I got stuck because that means that `/appdemo/:appId` would be a GET request with the side-effect, for any logged-in user, of creating a grain, which feels like cross-site-request-forgery territory. Plus I coded an infinite loop where I was creating grains continuously, but that's an implementation issue. The CSRF issue is more troubling.

If you can think of a way that's reasonably clean to implement the meteor-magic suggestion you had, I'm all ears. For now I submit this to your review as-is.

paulproteus:
@kentonv Please re-review!

paulproteus:
Cool!!! This is my first pull request merged in 2015, and the only one I've written so far that touches the Meteor/C++ code rather than build/install/etc. scripts.

Hooray!

PR #223, opened 2015-01-28
clarification build & hack frontend

JamborJan:
Added some clarification on how to build sandstorm from sources and hack the front end.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ Please install the following:
 * `unzip`
 * `strace`
 * `curl`
+* `sudo`
 * ImageMagick
 * [Clang compiler](http://clang.llvm.org/) version 3.4 or better
 * [Meteor](http://meteor.com)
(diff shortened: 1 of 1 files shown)


garply:

kentonv:
Hmm, thanks for the suggestions, but I think what we really want to do here is improve `install.sh` to ask more relevant questions. It should ask:

* Do you intend to develop Sandstorm apps? (If so -> install to start as root)
* Do you intend to hack on Sandstorm itself? (If so -> have files owned by your user account)

That way people don't have to carefully read docs to get things right. :)

I'll put this on my TODO list.

paulproteus:
Just to repeat/restate/clarify here:

I guess this suggests there are two "development" modes for Sandstorm, and
the install script should let you pick which kind of development mode you
mean.
>
>

JamborJan:
Perfect from my point of view, I absolutely agree that the setup process should be running the install.sh and done. No other paperwork required. All questions asked by during the process. 

I was not sure if sandstorm is willing to follow such an approach or not that's why I made some small changes to the document.

Ps: the script should also check all dependencies e.g. is sudo installed, if not it should ask if it can install them. Maybe that's more complicated as it depends on the Linux distribution.

kentonv:
We've finally revamped the installer script (thanks @paulproteus), though now that I look back at this PR, I realize that part of the suggestion is that if you plan to hack on Sandstorm itself (not just apps) then you really want to install it chowned to *you*. In particular, `shell/run-dev.sh` requires this.

OTOH, if the permissions aren't right, `run-dev.sh` will actually tell you how to fix it, so arguably isn't not that important that people be aware of this at install time.

@paulproteus should we add an installer flag for this, or do you think the error output of `run-dev.sh` makes it not necessary?

zenhack:
`run-dev.sh` doesn't exist anymore (well, it just prints an error about being deprecated and exits), and the replacement, `sandstorm dev-shell`, works fine with a root-installed sandstorm, as long as the user is part of the `sandstorm` group. Is this still relevant?

PR #225, opened 2015-01-28
Draft SturdyRef design.

kentonv:
Draft design for how SturdyRefs (persistent capabilities) will work in Sandstorm.

Required reading:
* [persistent.capnp](https://github.com/sandstorm-io/capnproto/blob/master/c++/src/capnp/persistent.capnp)
* [Cap'n Proto RPC level 2](https://capnproto.org/rpc.html#protocol-features)


kentonv:
Comments addressed.

dwrensha:
Ok, this persistence stuff looks good to me. I still don't have a good grasp on how user permissions will get attached to shared capabilities, but it sounds like that should be a separate conversation.

PR #226, opened 2015-01-29
Add tcp bridge to sandstorm-http-bridge

jparyani:
This is currently controlled by an option in `bridgeConfig` named `enableIpBridge`. We should probably also add something to the web UI to warn users when an app requests this kind of access (on install?).

### DIFF
--- a/src/sandstorm/sandstorm-ip-bridge.c++
+++ b/src/sandstorm/sandstorm-ip-bridge.c++
@@ -113,9 +113,9 @@ private:
   }
 
 public:
-  AcceptedConnection(kj::Own<kj::AsyncIoStream>&& connectionParam, TcpPort::Client && _port)
+  AcceptedConnection(kj::Own<kj::AsyncIoStream>&& connectionParam, TcpPort::Client && portParam)
       : connection(kj::refcounted<RefcountedAsyncIoStream>(kj::mv(connectionParam))),
-        port(_port), errorHandler(kj::addRef(*connection)), taskSet(errorHandler),
+        port(kj::mv(portParam)), errorHandler(kj::addRef(*connection)), taskSet(errorHandler),
         numOutstandingWrites(0) {
     auto request = port.connectRequest();
     request.setDownstream(kj::heap<Downstream>(kj::addRef(*connection)));
--- a/src/sandstorm/sandstorm-ip-bridge.c++
+++ b/src/sandstorm/sandstorm-ip-bridge.c++
@@ -99,7 +99,9 @@ private:
     // multiple times, so only abort the first. This will have a side-effect
     // of causing tryRead to fail, and erroring out messageLoop() below.
     if (!aborted) {
-      connection->stream->abortRead();
+      try {
+        connection->stream->abortRead();
+      } catch (...) { }
       aborted = true;
     }
   }
--- a/src/sandstorm/sandstorm-ip-bridge.c++
+++ b/src/sandstorm/sandstorm-ip-bridge.c++
@@ -14,12 +14,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// This program is useful for including in Sandstorm application packages where
-// the application itself is a legacy HTTP web server that does not understand
-// how to speak the Cap'n Proto interface directly.  This program will start up
-// that server and then redirect incoming requests to it over standard HTTP on
-// the loopback network interface.
-
 #include "sandstorm/sandstorm-ip-bridge.h"
 
 #include <kj/debug.h>
--- a/src/sandstorm/sandstorm-ip-bridge.c++
+++ b/src/sandstorm/sandstorm-ip-bridge.c++
@@ -15,6 +15,7 @@
 // limitations under the License.
 
 #include "sandstorm/sandstorm-ip-bridge.h"
+#include "sandstorm/util.h"
 
 #include <kj/debug.h>
 #include <netinet/ip.h>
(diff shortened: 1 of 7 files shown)


src/sandstorm/sandstorm-ip-bridge.c++
```diff
@@ -0,0 +1,181 @@
...
+  struct AcceptedConnection {
+    class Downstream final: public ByteStream::Server {
+    private:
+      kj::Own<RefcountedAsyncIoStream> connection;
+
+    public:
+      explicit Downstream(kj::Own<RefcountedAsyncIoStream> && connection) :
+        connection(kj::mv(connection)) {}
+
+      kj::Promise<void> write(WriteContext context) {
```
kentonv: The ByteString interface is documented to allow multiple write()s at a time (to avoid starving the stream between one write() returning and the next starting), but currently AsyncIoStream doesn't like overlapping writes. I would like to fix `AsyncIoStream` to allow this (mostly because it fits better with Win32's system calls, though there's not much benefit on Linux), but haven't gotten around to that yet.

For now, `Downstream` should have a `Promise<void>` member which is the promise returned by the last call to write(), and before issuing a new write() you should wait for the previous one to complete. You'll also need to `fork()` these promises so that you have something to return to the direct caller.

In `done()`, you'll want to make sure not to call `shutdownWrite()` until the last `write()` has completed.

src/sandstorm/sandstorm-ip-bridge.c++
```diff
@@ -0,0 +1,181 @@
...
+  public:
+    explicit RefcountedAsyncIoStream(kj::Own<kj::AsyncIoStream>&& stream)
+        : stream(kj::mv(stream)) {}
+    kj::Own<kj::AsyncIoStream> stream;
+  };
+
+  static const int BUFFER_SIZE = 8192;
+
+  // TCP handling
+  struct AcceptedConnection {
```
kentonv: I recommend using `struct` only if all members are public; otherwise use `class`.

kentonv: This could probably use more comments explaining the purpose of the class.

src/sandstorm/sandstorm-ip-bridge.c++
```diff
@@ -0,0 +1,181 @@
...
+
+#include "sandstorm/sandstorm-ip-bridge.h"
+
+#include <kj/debug.h>
+#include <netinet/ip.h>
+#include <linux/netfilter_ipv4.h>
+#include <unistd.h>
+
+namespace sandstorm {
+  namespace ipbridge {
```
kentonv: This (and its contents) shouldn't be indented.

src/sandstorm/sandstorm-ip-bridge.c++
```diff
@@ -0,0 +1,181 @@
...
+  public:
+    explicit AcceptedConnection(kj::Own<kj::AsyncIoStream>&& connectionParam, TcpPort::Client port)
+        : connection(kj::refcounted<RefcountedAsyncIoStream>(kj::mv(connectionParam))), port(port) {
+      auto request = port.connectRequest();
+      request.setDownstream(kj::heap<Downstream>(kj::addRef(*connection)));
+      upstream = kj::heap<ByteStream::Client>(request.send().getUpstream());
+    }
+
+    kj::Promise<void> start() {
+      return messageLoop();
```
kentonv: Hmm, start() seems kind of redundant if it just calls another public function and returns its result.

src/sandstorm/sandstorm-ip-bridge.c++
```diff
@@ -0,0 +1,181 @@
...
+        });
+      });
+    }
+  };
+
+  TcpPort::Client getTcpClient(kj::Own<kj::AsyncIoStream>& connection, HackSessionContext::Client& session) {
+    auto request = session.getIpNetworkRequest().send().getNetwork().getRemoteHostRequest();
+
+    struct sockaddr destaddr;
+    uint size = sizeof(destaddr);  // TODO(soon): make static?
```
kentonv: I would argue against making it static. If you do then you're assuming that `connection` is always the same object which seems unexpected from the function signature.

kentonv: I may have misread -- on second reading I'm not even sure what you're suggesting should be static. But in general, I feel static is almost always evil. :)

src/sandstorm/sandstorm-ip-bridge.c++
```diff
@@ -0,0 +1,181 @@
...
+    return serverPort.accept().then([&](kj::Own<kj::AsyncIoStream>&& connection) {
+      auto connectionState = kj::heap<AcceptedConnection>(kj::mv(connection),
+                                                          getTcpClient(connection, session));
+      auto promise = connectionState->start();
+      taskSet.add(promise.attach(kj::mv(connectionState)));
+      return runTcpBridge(serverPort, taskSet, session);
+    });
+  }
+
+  // UDP handling
```
kentonv: Comment that this code is incomplete and doesn't currently work.

src/sandstorm/sandstorm-ip-bridge.h
```diff
@@ -0,0 +1,39 @@
...
+#ifndef SANDSTORM_IP_BRIDGE_H_
+#define SANDSTORM_IP_BRIDGE_H_
+
+#include <kj/async-io.h>
+#include <sandstorm/hack-session.capnp.h>
+#include <sandstorm/util.capnp.h>
+#include <sandstorm/ip.capnp.h>
+
+namespace sandstorm {
+  namespace ipbridge {
```
kentonv: Shouldn't be indented.

src/sandstorm/sandstorm-ip-bridge.c++
```diff
@@ -0,0 +1,181 @@
...
+    kj::Promise<void> messageLoop() {
+      return connection->stream->tryRead(buffer, 1, BUFFER_SIZE)
+            .then([this] (size_t size) -> kj::Promise<void> {
+        if (size < 1) {
+          return upstream->doneRequest().send().then([] (auto args) {});  // EOF
+        }
+
+        auto request = upstream->writeRequest();
+        request.setData(kj::ArrayPtr<capnp::byte>(buffer, size));
+        return request.send().then([this] (auto args) {
```
kentonv: You should queue the `write()` call to a `TaskSet` (one owned by the `AcceptedConnection`) and then immediately do another `read()`, so that you can take advantage of doing multiple simultaneous `write()`s to reduce round trips.

But ideally you should do no more than, say, 4 simultaneous writes before you pause and wait for one of them to return, so that there is some backpressure if the app isn't accepting the data fast enough.

kentonv:
Generally good code! Biggest issue right now is supporting multiple simultaneous writes (in both directions) for the purpose of avoiding unnecessary round trip latency. This is a mostly theoretical issue for single-machine Sandstorm (where there is basically zero latency between processes) but could matter later for Blackrock. Otherwise, just some minor style nitpicks.

jparyani:
Forgot to comment, but your suggested changes have all been added.

jparyani:
Done

jparyani:
Done

jparyani:
Actually, wait on merging this. It now needs a rebase, and I'm running into some issues with porting Transmission that I should fix before this merges.

jparyani:
This is ready to merge. The only major change from the last code review is that I added a function called getFirstSessionContext for fetching a hackSessionContext for IpBridge and Dns related functionality that isn't tied to a user.

kentonv:
I'm concerned that `getFirstSessionContext()` is not just ugly but may actually not work at all in the presence of background processing -- there may be no session context available.

I hate to delay this further, but I think perhaps what we need to do here is go ahead and implement sturdyrefs. Then the app can request the IP networking capability from HackSessionContext while the user is actually present, `save()` it, and use the saved capability going forward. `restore()` goes through `SandstormApi` which in turn calls `SandstormCore` -- no sessions involved.

We're actually pretty close on SturdyRefs, so maybe it isn't so bad to go ahead and finish that now?

zenhack:
This has several problems that make me think the PR isn't worth salvaging at this point:

- It has conflicts.
- It builds on hack session, but now that the powerbox exists we shouldn't ever be adding new things to hack session.
- From the comments it sounds like this needs a refactor to use SturdyRefs?
- It seems to assume the supervisor is doing stuff with iptables, but iirc that got removed at some point, so the logic that depends on that is probably broken.
- It looks like it's WIP anyway?

I feel like given all of the bitrot it might be easier to just start from scratch.

ocdtrekkie:
Indeed. It sounds like Jason was building this to package an app to use it (Transmission), but that was never released, and we don't want to enable new networking features that aren't built the way we want them built in the future. (Presumably temporary hack methods make a lot more sense when you're a startup trying to reach critical mass, our long-running community project need not do that though.)

PR #229, opened 2015-01-30
add fields for content-encoding header on requests

dwrensha:
Today I learned that the `Content-Encoding` header is allowed in *requests*. See [RFC 7231](http://tools.ietf.org/html/rfc7231#section-3.1.2.2).

For example, sometimes Git [sets this header to "gzip"](https://github.com/git/git/blob/master/remote-curl.c#L590). This [Apache documentation](http://httpd.apache.org/docs/2.2/mod/mod_deflate.html) hints that WebDAV clients do this too.

Until this is fixed, both GitWeb and GitLab can sometime emit the following error on `git fetch` or `git pull`:
"fatal: protocol error: bad line length character"


dwrensha:
I have  few pending adjustments for this. Please hold on review...

dwrensha:
OK, ready now for review.

Issue #215, opened 2015-01-25
Ist not possible to scroll / move in mobile view


JamborJan:
Tested with iOS8 on iPhone 5s

No matter which app / grain is opened its not possible to scroll or move the page.

E.g. GitLab has a mobile view but it's not possible to scroll down. In other apps with no mobile view like ethercalc it's not possible to move the sheet in any direction.

dwrensha:
Ugh. I'm seeing this too on Safari on iOS8.

I tried changing our CSS [here](https://github.com/sandstorm-io/sandstorm/blob/e71362d6543c7ce1ed5ce3f2b6549218c3dbca0f/shell/client/shell.css#L228) to this:
```
#app.main-content {
   overflow: auto;
   -webkit-overflow-scrolling:touch;
}
```
and that made scrolling seem to at least partially work. But GitLab's infinite scrolling is still broken after this change; it fails to load more content when you reach the bottom.

Maybe someone who knows more about CSS than I do can help here?

kentonv:
We need to do a lot more work on mobile, but I'm in favor of any small changes that help even if they don't fix everything.

(I don't know much about mobile CSS, personally.)

JamborJan:
I do some tests and send a PR when I find a proper solution.

JamborJan:
I'm still struggeling with my build server so I cannot test with that.

Is it possible to change the CSS on my server without doing the build process?

I had an issue in the past on mobile devices where ```cursor: pointer;``` helped e.g. when labels where not click-able etc. maybe that helps here too. Would test it if I'm able to get the stuff running.

kentonv:
Unfortunately it's pretty impossible to change the CSS without rebuilding the whole shell (or running it in dev mode). But it sounds like you've got the build working now. :)

JamborJan:
Yes, build is okay now. Will do the CSS stuff and tests tomorrow and let you know the results.

JamborJan:
Found a solution with:

```
#app.main-content {
  overflow: hidden;
  -webkit-overflow-scrolling:touch;
  overflow-y: scroll;
  overflow-x: scroll;
}
```

It fixes the total in-usability on mobile for now but a general overhaul of the mobile experience might be required.

PR #230, opened 2015-01-30
fix for mobile scrolling, see issue  #215

JamborJan:
This fixes the scrolling issue in my test cases with iOS. Tested with GitLab, Etherpad and WordPress where it works perfect. Ethercalc is still a mess on mobile. Didn't work before (no scrolling, no input, constant loading) and still doesn't work because of constant loading. Seems to be an issue of Ethercalc as all other tested apps are working now. It is a quick fix of the mobile experience issue but a over all mobile overhaul is required anyway.

### DIFF
--- a/shell/client/shell.css
+++ b/shell/client/shell.css
@@ -225,11 +225,8 @@ img {
   color: #428bca;
 }
 
-#app.main-content {
+#app.main-content { 
   overflow: hidden;
-  -webkit-overflow-scrolling:touch;
-  overflow-y: scroll;
-  overflow-x: scroll;
 }
 
 #grainLog {
@@ -901,6 +898,10 @@ aside #social>p {
     position: static;
     margin: 100px 150px;
   }
+  #app.main-content {
+  	-webkit-overflow-scrolling:touch;
+    overflow: auto;
+  }  
 }
 
 @media screen and (max-width: 900px) {
--- a/shell/client/shell.css
+++ b/shell/client/shell.css
@@ -227,6 +227,9 @@ img {
 
 #app.main-content {
   overflow: hidden;
+  -webkit-overflow-scrolling:touch;
+  overflow-y: scroll;
+  overflow-x: scroll;
 }
 
 #grainLog {


garply:

dwrensha:
I gave this a quick try on my server. Infinite scrolling on GitLab still doesn't work. (Try scrolling through commits on a largish repo like Sandstorm.)

However, I agree with Kenton that a partial fix is better than nothing. So I'm in favor of accepting this or something like it.

JamborJan:
Okay I tested this only on my small projects. Maybe it's an iframe issue. I will do some mor checks on that. At the moment it would help me a lot if I could use the apps more or less good on mobile which is possible with that small fix.

kentonv:
On desktop, it looks like this change creates unconditional scrollbars around every app, leading to double scrollbars if the app actually has scrolling content.

Shouldn't this change be going on `#app iframe`, not `#app.main-view`? main-view only contains one big iframe with size 100%x100%, so it seems to be that it should never overflow, whereas the iframe content can overflow and normally displays scrollbars (on the iframe) when it does.

If `overflow: scroll` is really needed (as opposed to `overflow: auto`) then we'll need to put it in the mobile-specific part of the stylesheet as this property will force the display of unnecessary scrollbars on desktop.

JamborJan:
I did two things wrong here A) I did not check the shell.css carefully enough to understand it completely and B) I did not test carefully enough. Sorry for that.

I moved the ```overflow: hidden``` part to ```.main-content``` as it was defined there as ```auto```before and overwritten anyway by ```hidden``` later. I tested it with firefox, safari and safari on iOS and it seems to be exact the same result with less confusion in the CSS.

Any way the main issue seems to be the iframe and the iOS browser and not the iframe alone. I have no Android or Windows Phone to test but only on the iOS not on the mobile tests on my desktop I have the issue that GitLab seems to have an issue on the files and commits view. These are the only two where the mobile view disappears and suddenly the width is damaged, but only on iOS.

I will do some more testing and come up with a well tested proposal.

ocdtrekkie:
I have both an Android and Windows Phone. If you can give me the specific
cases you want checked, I will check them.

-Jake Weisz

JamborJan:
@ocdtrekkie I can send you an invitation link to my sandstorm instance and a backup of the sandstorm repo for the GitLab app where it is pretty easy to test because of enough content. If this suits you let me know and I send you the invite link via email.

JJ

ocdtrekkie:
Yeah, sure.

JamborJan:
Have you received my mail 3 days ago @ocdtrekkie ?

ocdtrekkie:
I responded to it within an hour with results.

JamborJan:
Hmmm ... Can you please send it again? Maybe I deleted the spam folder without checking good enough because I have nothing so far. Sorry.

ocdtrekkie:
Check now.

JamborJan:
Have it. Thx.

JamborJan:
I close this one in favor of #247. Still no really good solution but the fewest impact on all other besides iOS and working scrolling on iOS.

PR #243, opened 2015-02-05
use on_message_complete callback to detect end of HTTP response

dwrensha:
With our current `sandstorm-http-bridge`, an app with the following node.js server will fail to respond. (Assume there is another server listening on port 10001.)

```
var http = require('http');
var request = require('request');
http.createServer(function (req, res) {
  console.log("got request");
  var proxy = request({url: "http://127.0.0.1:10001", method: req.method});
  proxy.pipe(res);
}).listen(10000, '127.0.0.1');
```

The problem is that `request().pipe(response)` does not close the response stream once its end is reached, and therefore the http bridge never calls `done()` on the response stream, because we currently wait for EOF.

We should instead use the `on_message_complete()` callback, as in this patch. The EOF case is still covered, because in that case the final call to `http_parser_execute()` will have length zero, which signifies EOF, and the parser will know to trigger the callback. (See [the documentation](https://github.com/joyent/http-parser/blob/master/README.md).)

I originally noticed this problem when I was porting ShareLaTeX back in June. This patch will mean that I can simplify my port. I've tested it on most of my other ports, and everything still works as expected.

### DIFF
--- a/src/sandstorm/sandstorm-http-bridge.c++
+++ b/src/sandstorm/sandstorm-http-bridge.c++
@@ -165,6 +165,7 @@ public:
     settings.on_header_value = &on_header_value;
     settings.on_body = &on_body;
     settings.on_headers_complete = &on_headers_complete;
+    settings.on_message_complete = &on_message_complete;
     http_parser_init(this, HTTP_RESPONSE);
   }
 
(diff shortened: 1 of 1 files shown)


dwrensha:
Hm... I think I might be mistaken in my above reasoning. I think that the parser will not trigger the `on_message_complete()` callback on EOF if the message is not well-formed. (See https://github.com/sandstorm-io/sandstorm/blob/master/src/joyent-http/http_parser.c%2B%2B#L593.)
I'll add another commit that fixes this up...

dwrensha:
Ok, fixed. It looks like the EOF check is indeed needed at least to deal with the possibility of an empty response, in which case `on_message_complete()` will not be called.

dwrensha:
On more small improvement: make sure that EOF before end-of-headers fails in a sane way.

dwrensha:
This bit of the documentation leads me to believe that we don't need a similar check in the upgrade case:

>The user is expected to check if parser->upgrade has been set to 1 after http_parser_execute() returns. >Non-HTTP data begins at the buffer supplied offset by the return value of http_parser_execute().

PR #247, opened 2015-02-08
Mitigates scrolling issue on iOS

JamborJan:
This mitigates scrolling issue on iOS but is no final solution.

I did various tests and it is kind of strange that only iOS has issues here. I still think that some combination of the iframe and iOS specific issues are the root cause here but I found no 100% solution so far.

I would like to use the apps on my mobile so I hope this small change can be included.

(closes #230)

### DIFF
--- a/shell/client/shell.css
+++ b/shell/client/shell.css
@@ -898,6 +898,10 @@ aside #social>p {
     position: static;
     margin: 100px 150px;
   }
+  #app.main-content {
+    -webkit-overflow-scrolling:touch;
+    overflow: auto;
+  }
 }
 
 @media screen and (max-width: 900px) {
--- a/shell/client/shell.css
+++ b/shell/client/shell.css
@@ -899,7 +899,8 @@ aside #social>p {
     margin: 100px 150px;
   }
   #app.main-content {
-    -webkit-overflow-scrolling:touch;
+    /* Mitigates scrolling issue on iOS; see #247 */
+    -webkit-overflow-scrolling: touch;
     overflow: auto;
   }
 }


garply:

paulproteus:
Garply, ok to test

paulproteus:
This change seems pretty enjoyably minimal! You said it's not a 100% solution; it seems like a 100% solution to me! What's missing, in your eyes?

I have a few small style suggestions; consider this a pre-review not a real review since Kenton is likely to be more opinionated.

* Leave a comment in the CSS (using `// ...` syntax not `/* ... */` syntax) indicating what this is about. That way, down the road, we'll remember. Perhaps something like `// Without this, we cannot scroll the iframe on iOS; see #247`

* Add a space between `-webkit-overflow-scrolling:` and `touch;`, to be consistent with the rest of the file.

Let me also say that I'm super excited you've been looking into this and that the patch ends up being nice and small. Thank you! I actually ran into this problem yesterday when I asked a friend to look up my grocery list on an iPhone and we couldn't scroll it. (-:

JamborJan:
Thanks for your positive feedback @paulproteus. Sorry for not meeting the style requirements, will do so the next time. Or can I change the commit? (Don't know, I'm not working with GitHub every day).

Things I don't like so far:
1) the fix may cause other small screen devices to have a scroll bar even if not required. I have this when testing on my desktop firefox in "i pretend I'm a mobile" mode. 
bildschirmfoto_2015-02-08_um_18_14_35


2) the GitLab app makes me go crazy. Only iOS seems to have an issue displaying the files and commits view. Its simply not the correct width and I spend now hours after hours figuring out why ... no clue so far. Worse on the GitLab app: when you are at the end of the commit list it doesn't load the next commits to show ...

Thats why I say it's not 100% but better then nothing as I only have iOS devices and can use nothing from sandstorm without this fix.

kpreid:
@paulproteus `//` is **not a comment syntax** in CSS. The only syntactically valid comments in CSS are `/*...*/`.

paulproteus:
re: comment syntax: Thanks @kpreid ! Oops. That's what I get for doing code review before coffee and breakfast.

re: unnecessary scrolling: interesting. Good to know. I imagine we can live with it for now; @kentonv will likely have an opinion on that.

re: change commit: On your own machine, you can do `git commit --amend` to change the contents of your commit, and then `git push --force origin HEAD:patch-1` to update the request you submitted, assuming that you have cloned your fork to your own computer using the git command line client. I see `patch-1` in the branch name, which suggests you made the changes using the github web editor. You can try the `git commit --amend` way, or you can make a new pull request. If you're willing to do that, that's best; we try to merge only exactly what people submit, rather than change it up ourselves, in part to avoid adding mistakes of our own on top of other people's work.

Also, the test runner seems to not like this, but I don't really understand why. It remarks:

```
Running:  Google: Test grain restart 

✖  Timed out while waiting for element <#publish> to be present for 30000 milliseconds.  - expected "found" but got: not found

FAILED:  1 assertions failed and 3 skipped (35.593s)
```

I don't see why grain restart should be affected by this. @jparyani might know more.

ocdtrekkie:
If you created the pull request with Github, and you update your patch-1
branch with additional edits, it will add them to the pull request.

jparyani:
Ignore the test error. We're getting intermittent failures on that test I think due to a timing issue.

jparyani:
Fixed by eb484f671387a99ada9c663587e69253a492d8a2

Thanks for everyone's help on this issue (especially @JamborJan). It turns out that always making the outer div have `overflow: auto` and `-webkit-overflow-scrolling: touch`, and changing the iframe to be `display: block` fixes the problem for mobile while not affecting desktop. Let me know if you run into any issues, but I tested on iphone5/ipad/nexus 7 (both firefox and chrome mobile) and it seems to be working fine on all of them.

JamborJan:
Thanks @jparyani i quickly tested it and it seems to solve the issue at least to the level I had too.

I still see the issue that the GitLab app is not properly displayed on iOS. But thats a different story. 

Thanks!

PR #254, opened 2015-02-12
topbar: show light grey background for menu buttons

zarvox:
The "delete", "show debug log", and other menu buttons don't change color on hover, which makes it hard to be sure that you're actually supposed to be able to click them.

This commit gives them the same background color as the grain title field on hover.

### DIFF
--- a/shell/client/shell.css
+++ b/shell/client/shell.css
@@ -157,6 +157,10 @@ body {
   cursor: pointer;
 }
 
+#menu button:hover {
+  background-color: #888;
+}
+
 #topbar #renameGrain {
   background-image: url("/edit.png");
 }


garply:

paulproteus:
Garply, ok to test

paulproteus:
Thanks @zarvox ! I think this is an improvement and that we should merge it. I'll defer to @kentonv for his opinion.

Also @kentonv I couldn't find notes about our use of `Signed-off-by` or whatever else in https://github.com/sandstorm-io/sandstorm/wiki . Does that mean I need to write such a doc? (-:

zarvox:
Heh, I looked for a page like "Contributing" on the wiki and found none, and I didn't see anyone else's commits featuring a Signed-off-by line, so I just followed suit.

If we're using the standard meaning of Signed-off-by:, I'm happy to amend my commit accordingly. :)

kentonv:
For tiny changes like this, no one is really afraid about legal issues. For larger contributions in theory we have sandstorm.io/cla (Apache-style contributor license agreement), but there are also prominent lawyers who say that none of this is necessary, so whatever. :)

The test failure is spurious (spurrrrious?), so I'll go ahead and merge. Thanks!

kentonv:
BTW at some point we will do a complete redesign of the UI, but our designer is busy on the app store right now. :)

PR #255, opened 2015-02-12
Refactor hack-session code structure a bit

jparyani:
Clean up hack-session code structure:
* Rename mail.js -> hack-session.js
* Create folder named hack-session and move ip.js there
* Split out mail specific functionality and create mail.js in that folder

Also fix up some small style issues such as using single quotes for strings.


jparyani:
Renamed hack-session folder to drivers

PR #258, opened 2015-02-13
Add dev accounts

### LINKED ISSUE #150: Implement some sort of "local developer account" login
For the purpose of a localhost Sandstorm server that is only accessible to one person anyway and is intended only for development, it would make sense to implement some way to create test users and log in as them. This way, people don't have to hook Google or Github authentication just for their silly testing.

The feature should NOT use passwords or anything else that could be mistaken for security. The developer should be able to create as many users as they want and arbitrarily log in as any of them by just choosing the account from a list.

We can implement this pretty easily using the same approach as the "demo user" code.

jparyani:
This uses a new sandstorm.conf setting named ALLOW_DEV_ACCOUNTS. If set to true, there will be a new Meteor login service named DevAccounts, and clicking it will direct you to a page where you can login as your choice of Alice, Bob, or Carol. Alice will always be an admin.

Fixes #150

### DIFF
--- a/shell/client/shell.css
+++ b/shell/client/shell.css
@@ -729,6 +729,10 @@ input.autoSelect {
   text-align: center;
 }
 
+.devLogin {
+  margin: 16px auto;
+}
+
 ul.backer-list, ul.dev-list {
   list-style: none;
   text-align: center;
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -541,6 +541,33 @@ $KEY</textarea></p>
   </div>
 </template>
 
+<template name="devAccounts">
+  <div id="topbar">{{> homeLink}} {{ pageTitle }} </div>
+
+  <div id="demo" class="main-content">
+    <div class="centered-box">
+      <h2>Developer Accounts</h2>
+        {{#if shouldShowStartDevAccounts}}
+          {{#if loggingIn}}
+            <p>Logging in...</p>
+          {{else}}
+            <button id="loginAliceDevAccount" class="devLogin">Log in as Alice (admin) »</button>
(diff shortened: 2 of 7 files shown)


paulproteus:
I am excited about this. If the target users are app developers, this seems
perfectly great as an idea.

If the target is self installers using their apps for real, I would rather
see an approach where the user enters an email address and we log them in,
via a new hypothetical flow that I nickname, Password reset as login.

I am typing on a phone so apologies if this comes off as short. I can try
to explain more in a bit.

jparyani:
Ya the target audience here is developers that don't want to bother setting up a real authentication method for a local server. It should never be used in production.

paulproteus:
Cool, though I do wonder if even they would be better off with the password
reset as login flow, if the installer asks for an email address, and prints
out a token, and then you can use that to log in.

Then again, my proposal doesn't help simulate Bob and Charlie. OK, I like
your idea now!

jparyani:
Now it will purge dev users if a server ever switches allowDevAccounts to false, and dev users also play nicely with the auto-admin logic in Sandstorm.

kentonv:
Hmm, I think if we're purging the users we'll also need to purge their grains and possibly other things. Maybe we need a purgeUser() function defined in db.js, so it can be extended whenever there's a new collection to purge. I guess it also needs to delete the grains on disk. Ugh.

Alternatively, we could change the page wording so that it doesn't promise that dev accounts will go away if you change the flag. I don't think there's much of a user case for taking a dev server and switching it to prod -- there's a lot of ways that could go badly, I suspect, so we probably shouldn't encourage it.

jparyani:
Ping @kentonv.

I went with the easy solution and removed the text that said it was safe to switch to ALLOW_DEV_ACCOUNTS=false after it's been true.

PR #257, opened 2015-02-13
add appName field to Manifest

dwrensha:
This will allow us to transition away from using the `appNameFromActionName()` function.

PR #261, opened 2015-02-16
Add appName field to UserActions collection. Unbreak things.

dwrensha:
Eek -- my change from #257 made it so that dev apps failed to appear at all if they did not have an `appName` in their manifest! This fixes the problem and actually uses the new `appName` field for installed actions.

We may also want to add an `appName` field to the `Grains` collection.

### DIFF
--- a/shell/shared/db.js
+++ b/shell/shared/db.js
@@ -60,6 +60,8 @@ UserActions = new Mongo.Collection("userActions");
 //   userId:  User who has installed this action.
 //   packageId:  Package used to run this action.
 //   appId:  Same as Packages.findOne(packageId).appId; denormalized for searchability.
+//   appName:  Same as Packages.findOne(packageId).manifest.appName; denormalized for
+//       seachability.
 //   appVersion:  Same as Packages.findOne(packageId).manifest.appVersion; denormalized for
 //       searchability.
 //   title:  Human-readable title for this action, e.g. "New Spreadsheet".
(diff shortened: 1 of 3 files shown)


shell/shared/db.js
```diff
@@ -60,6 +60,8 @@ UserActions = new Mongo.Collection("userActions");
 //   userId:  User who has installed this action.
 //   packageId:  Package used to run this action.
 //   appId:  Same as Packages.findOne(packageId).appId; denormalized for searchability.
+//   appName:  Same as Packages.findOne(packageId).manifest.appName.defaultText; denormalized for
```
kentonv: "denormalized for searchability" seems incorrect; we don't query on this field.

I guess Mongo doesn't really do joins otherwise I'd suggest we actually don't denormalize this. :)

dwrensha:
Ok, I've updated the comment.

Note that the client currently doesn't actually have access to the Packages collection. Maybe a better solution would be to get the server to publish the subset of the Packages collection that is relevant to the user. Publishing joins doesn't work natively in Meteor/Mongo, but we could probably wire something up that would manually push the relevant data, similar to what we already do for the GrainSizes pseudo-collection. Once we do that, I think we would be able to remove the UserActions.addName column.

kentonv:
I think it's probably not worth the effort to implement a manual join. This looks good as-is.

Issue #263, opened 2015-02-16
make fails on ubuntu 14.04 in Vagrant


diracdeltas:
While running `make` on the default ubuntu 14.04 vagrant box with all dependencies installed:

```
[snip]
> link: capnp/endian-test.o
> test: capnp/endian-test
> link: capnp/endian-reverse-test.o
> test: capnp/endian-reverse-test
> link: capnp/endian-fallback-test.o
> test: capnp/endian-fallback-test
> install: node-capnp/node-capnp.ekam-manifest
> terminate called after throwing an instance of 'std::logic_error'
>   what():  Already fulfilled this promise.
> Aborted (core dumped)
> make: *** [tmp/.ekam-run] Error 134
```

using Ubuntu clang version 3.5.0-4ubuntu2~trusty1 (tags/RELEASE_350/final) (based on LLVM 3.5.0)

using

paulproteus:
Perhaps we should "just" switch the base box to use Debian stable!!

```
config.vm.box = "chef/debian-7.6"
```

is one way to do that.

paulproteus:
There is also the chance that the filesystem sharing is messing it up.

paulproteus:
I notice in https://github.com/sandstorm-io/sandstorm/blob/master/Vagrantfile that we're not using NFS. If we want reasonable filesystem sharing, we should use NFS. (FWIW I'm totally to blame on that; I am the author of this Vagrantfile.)

paulproteus:
http://docs.vagrantup.com/v2/synced-folders/nfs.html discusses that.

Looks like adding this:

```
  config.vm.synced_folder ".", "/vagrant", type: "nfs"
```

would switch that. @diracdeltas can you see if that makes this problem go away? Note you might need stop & start Vagrant.

paulproteus:
And thank you for trying this. This stuff should work, and you're forging a path ahead for everyone.

paulproteus:
Turns out adding NFS support also requires changing the network configuration, so I'll see if I can make that change here. Testing it now.

paulproteus:
I found that it did help some, but that I still had other problems building Sandstorm.

Looking into NFS + 14.10 upgrade now.

I would use Debian inside the Vagrant, but Jessie isn't released yet, and Wheezy has too old a kernel to run Sandstorm.

paulproteus:
I currently consistently get this error with NFS and 14.04 and clang-3.5 (as well as clang-3.4):

[...]

PR #264, opened 2015-02-18
Adjust Vagrantfile to do file sharing via NFS

paulproteus:
The purpose of this change is to stop using the VirtualBox-native
file sharing system, which seems to be both slower and less POSIX-y
than NFS.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -35,21 +35,40 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # compatibility.
   config.vm.synced_folder ".", "/vagrant", type: "nfs"
 
-  # Set the number of CPUs within Vagrant to be the number of CPUs
-  # outside it.
+  # Calculate the number of CPUs and the amount of RAM the system has,
+  # in a platform-dependent way; further logic below.
+  cpus = nil
+  total_kB_ram = nil
+
   host = RbConfig::CONFIG['host_os']
   if host =~ /darwin/
     cpus = `sysctl -n hw.ncpu`.to_i
+    total_kB_ram =  `sysctl -n hw.memsize`.to_i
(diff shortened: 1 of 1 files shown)


paulproteus:
I believe this may fix the build problems in #263 (though this commit doesn't change the Vagrantfile to actually do the build). @diracdeltas if you want to test that, I recommend doing:

```
$ vagrant up
$ vagrant ssh
$ sudo sudo apt-get install build-essential libcap-dev xz-utils zip \
    unzip imagemagick strace curl clang-3.4 clang nodejs
$ curl https://install.meteor.com/ | sh
```

and then seeing if it builds for you.

@awans I believe this will fix the issues you were having on Mac OS where Meteor/Mongo couldn't run properly. If you can test that, I'd greatly appreciate it.

I'll probably merge this once the tests pass, rather than wait on the above, since I believe it makes things no worse.

paulproteus:
I'm adding a commit that has Vagrant use more RAM, and I'm also going to test right now that the `apt-get install` (etc.) lines above result in a successful build.

paulproteus:
I've tested locally that this setup can compile Sandstorm, so I will ship it.

@awans if you can test on Mac OS sometime, that'd be stellar.

kentonv:
BTW, if you want to reduce the parallelism of the build, you can set the `PARALLEL` make var, e.g.:

    make PARALLEL=2

The default is determined by invoking `nproc`.

awans:
Hey @paulproteus, I ran into some trouble getting the provisioning script to run—seems like my sudoers privileges weren't getting created correctly?

I figure a short screencast is the easiest way of describing the behavior I see:

https://www.dropbox.com/s/627rnjgs50vmufd/vagrantfile.mov?dl=0

I'm probably missing something obvious :/

PR #266, opened 2015-02-19
Implement ApiForwarder 

dwrensha:
... so that apps using sandstorm-http-bridge get proper access to SessionContexts.

### DIFF
--- a/src/sandstorm/sandstorm-http-bridge.c++
+++ b/src/sandstorm/sandstorm-http-bridge.c++
@@ -46,6 +46,7 @@
 #include <sandstorm/api-session.capnp.h>
 #include <sandstorm/web-session.capnp.h>
 #include <sandstorm/email.capnp.h>
+#include <sandstorm/http-bridge.capnp.h>
 #include <sandstorm/hack-session.capnp.h>
 #include <sandstorm/package.capnp.h>
 #include <joyent-http/http_parser.h>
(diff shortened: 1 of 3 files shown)


dwrensha:
I've rebased against master, resolving the conflicts with #267, and I've renamed `ApiForwarder` to `PlatformProxy`.

dwrensha:
Another rename: `PlatformProxy` is now `SandstormHttpBridge`.

src/sandstorm/sandstorm-http-bridge.c++
```diff
@@ -1358,53 +1374,26 @@ public:
...
+    explicit AcceptedConnection(SandstormHttpBridge::Client& bridge,
+                                kj::Own<kj::AsyncIoStream>&& connectionParam)
+      : connection(kj::mv(connectionParam)),
+        network(*connection, capnp::rpc::twoparty::Side::SERVER),
+        rpcSystem(capnp::makeRpcServer(network, bridge)) {}
   };
 
-  kj::Promise<void> acceptLoop(kj::ConnectionReceiver& serverPort, ApiRestorer& restorer,
+  kj::Promise<void> acceptLoop(kj::ConnectionReceiver& serverPort,
+                               SandstormHttpBridge::Client& bridge,
```
kentonv: It feels kind of weird that you're using a mutable reference to a value type here, without intending to re-assign it. I think this should probably be a non-reference instead.

Because capability clients are refcounting, you'll want to use kj::mv() when passing them anywhere where it makes sense, to skip the refcount. You'll also want KJ_MVCAP() for capturing it in a lambda.

kentonv:
I guess this change breaks any apps which currently use HackSessionContext, although of course only affects new builds of those apps. I vaguely wonder if a temporary compatibility fallback is worthwhile, though I guess you and @jparyani are probably the only people affected so if you don't care then I don't.

jparyani:
I'm fine with updating my apps. Seeing as it only affects new builds, I'll get around to it whenever we do our next app update week.

src/sandstorm/sandstorm-http-bridge.c++
dwrensha: Oh, neat -- I was unaware that I could specify capture-by-move for a single variable and still specify capture-by-ref for the default. I think my main motivation here was to avoid the verbosity of writing out each captured variable.

dwrensha:
Ok, ready for re-review.

I don't think a capability fallback would be worth the effort. Here's an example for how to use the new interface: https://github.com/dwrensha/wordpress-sandstorm/blob/master/sandstorm/getPublicId.c%2B%2B

PR #267, opened 2015-02-20
Add ApiSession

jparyani:
Now when calling newSession from a proxy that was initiated from an API
connection, we now call it with a type id of ApiSession.

As a transitional measure, we fallback to WebSession if the call asking
for an ApiSession fails. This will eventually be removed once all apps
have updated.

Added apiPath as a variable to BridgeConfig. This will enable ApiSession
functionality as well as limiting what paths an api token can access.

This is blocking on https://github.com/kentonv/node-capnp/pull/9 getting accepted first.

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -796,6 +796,8 @@ Proxy.prototype._callNewApiSession = function (request, userInfo) {
   return this.uiView.newSession(userInfo, makeHackSessionContext(this.grainId),
     ApiSession.typeId).then(function (session) {
       return session.session;
+    }, function (err) {
+      return self._callNewWebSession(request, userInfo);
     });
 };
 
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -784,7 +784,7 @@ Proxy.prototype._callNewWebSession = function (request, userInfo) {
   });
 
   return this.uiView.newSession(userInfo, makeHackSessionContext(this.grainId),
-    WebSession.typeId, params).session;
+                                WebSession.typeId, params).session;
 };
 
 Proxy.prototype._callNewApiSession = function (request, userInfo) {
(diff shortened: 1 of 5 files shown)


kentonv:
Garply, retest this please

shell/server/proxy.js
```diff
@@ -781,6 +783,23 @@ Proxy.prototype._callNewSession = function (request, viewInfo) {
...
   });
 
+  return this.uiView.newSession(userInfo, makeHackSessionContext(this.grainId),
+    WebSession.typeId, params).session;
+};
+
+Proxy.prototype._callNewApiSession = function (request, userInfo) {
+  var self = this;
+
+  // TODO(someday): We are currently falling back to WebSession if we get any kind of error upon
```
kentonv: Is this fallback implemented? I don't see it.

jparyani:
Sorry, I was testing with the fallback disabled to make sure sandstorm-http-bridge's apiPath was actually working. I've fixed it now.

shell/server/proxy.js
```diff
@@ -781,6 +783,23 @@ Proxy.prototype._callNewSession = function (request, viewInfo) {
...
+  // calling newSession with an ApiSession._id.
+  // Eventually we'll remove this logic once we're sure apps have updated.
+  return this.uiView.newSession(userInfo, makeHackSessionContext(this.grainId),
+    ApiSession.typeId).then(function (session) {
```
kentonv: Some formatting nitpicks (which I wouldn't have commented on if the PR were ready to merge, but since it's not...):
- Continuation lines (not part of a braced block) should be indented at least four spaces, not two.
- When breaking a line in a parenthesized list, I find it easiest to read if either the continuation is indented to match the open paren, like:

        someLongFunctionName(param1, param2,
                             param3, param4);

  Or the break happens immediately after the open paren and all the parameters are indented four spaces:

      someLongFunctionName(
          param1, param2, param3, param4);

- With promises, there's a lot of possible styles and none of them are great, but I've settled on indenting the closure/lambda body as a regular block -- so, the contents are indented two spaces beyond the surrounding block. Also, if `.then(...{` doesn't fit on the previous line, I like to line break before the `.`. So:

        someAsynchronousFunction(foo, bar, baz)
            .then(function (result) {
          doSomething(result);
        });

kentonv: Oh, I guess by the time I got done typing that, your change became ready to merge...

PR #271, opened 2015-02-21
Sets the window title for the grain template

### LINKED ISSUE #165: Page title doesn't identify grain
Grain URLs can be bookmarked, shared, and so on, which is great. But in a list of open tabs/bookmarks/history, the titles are all just `Sandstorm`, and it's hard to find a particular grain without going via the home page. How about including the grain title to get something like `Untitled Todo List - Sandstorm`?

jparyani:
Also adds a postMessage endpoint for "setTitle" which allows apps to
explicitly set their title. This is prepended onto the normal title.

Fixes #165

### DIFF
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -191,6 +191,7 @@ limitations under the License.
 </template>
 
 <template name="grain">
+  {{setGrainWindowTitle}}
   <div id="topbar">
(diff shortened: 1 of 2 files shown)


jparyani:
Now it will revert the title to `Sandstorm` when navigating away from a grain

coyotebush:
The second part appears to have regressed in the new UI: the applist and grainlist don't reset the window title.

PR #272, opened 2015-02-21
Changed DNS error messages to be more clear based on the error code

### LINKED ISSUE #166: DNS error message should account for type of error
The error below is written assuming that the DNS record wasn't found, but we can see in this case that's not what happened -- actually, the DNS request timed out. We should pay attention to the type of error and adjust the error text accordingly.

    Error looking up DNS TXT records for host 'www.centur.io': queryTxt ETIMEOUT

    This Sandstorm server's main interface is at: https://pandorica.centur.io

    If you were trying to connect this address to a Sandstorm app hosted at this server,
    you either have not set your DNS TXT records correctly or the DNS cache has not
    updated yet (may take a while).

    If you are the server admin and want to use this address as the main interface,
    edit /opt/sandstorm/sandstorm.conf, modify the BASE_URL setting, and restart.

    If you got here after trying to log in via OAuth (e.g. through Github or Google),
    the problem is probably that the OAuth callback URL was set wrong. You need to
    update it through the respective login provider's management console.

jparyani:
Fixes #166

### DIFF
--- a/shell/server/pre-meteor.js
+++ b/shell/server/pre-meteor.js
@@ -129,6 +129,18 @@ Meteor.startup(function () {
   });
 });
 
+var errorTxtMapping = {};
+errorTxtMapping[Dns.NOTFOUND] = "\n" +
+    "If you were trying to connect this address to a Sandstorm app hosted at this server,\n" +
+    "you either have not set your DNS TXT records correctly or the DNS cache has not\n" +
+    "updated yet (may take a while).\n";
+errorTxtMapping[Dns.NODATA] = errorTxtMapping[Dns.NOTFOUND];
+errorTxtMapping[Dns.TIMEOUT] = "\n" +
+    "The DNS query has timed out, which may be a sign of poorly configured DNS on the server.\n";
+errorTxtMapping[Dns.CONNREFUSED] = "\n" +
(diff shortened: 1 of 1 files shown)


shell/server/pre-meteor.js
```diff
@@ -129,6 +129,13 @@ Meteor.startup(function () {
   });
 });
 
+var errorTxtMapping = {};
+errorTxtMapping[Dns.NOTFOUND] = "\nIf you were trying to connect this address to a Sandstorm app hosted at this server,\n" +
+    "you either have not set your DNS TXT records correctly or the DNS cache has not\n" +
+    "updated yet (may take a while).\n";
+errorTxtMapping[Dns.TIMEOUT] = "\nThe DNS query has timed out, which may be a sign of poorly configured DNS on the server. \n";
+errorTxtMapping[Dns.CONNREFUSED] = "\nThe DNS server refused the connection, which means either your DNS server is down/unreachable, or the server has misconfigured their DNS. \n";
```
kentonv: Stick a newline somewhere in here to match the wrapping of the other text. (Also, break the string to wrap the code line.)

jparyani:
Fixed

PR #277, opened 2015-02-27
Filter out irregular files from backup

### LINKED ISSUE #100: Grain backup fails if /var contains irregular files (sockets, pipes)
Backup failed: Error: Zip process failed. [500]

jparyani:
Fixes #100

### DIFF
--- a/shell/server/backup.js
+++ b/shell/server/backup.js
@@ -41,6 +41,27 @@ function recursiveRmdirIfExists(dir) {
   }
 }
 
+function walkAndWriteFiles(dir, writeStream, replaceRoot, originalRoot) {
+  originalRoot = originalRoot || dir;
+  var files = Fs.readdirSync(dir);
+  files.forEach(function (name) {
+    name = dir + "/" + name;
+    var fileStat = Fs.lstatSync(name);
+    if (fileStat.isDirectory()) {
+      walkAndWriteFiles(name, writeStream, replaceRoot, originalRoot);
+    } else if (fileStat.isFile()) {
+      name = replaceRoot + name.slice(originalRoot.length);
+      writeStream.write(name + "\n");
+    }
+  });
+
+  // Include empty directories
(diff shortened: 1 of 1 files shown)


kentonv:
Security issue: Someone could create a file containing a '\n' character in order to inject an arbitrary filename into this list. They could use that to trick us into instructing zip to follow some symbolic link to an arbitrary place in the filesystem.

Luckily, minibox saves us here because it hides all the other user data on the filesystem, so nothing interesting could be exfiltrated. But I think we should also check for filenames containing '\n' and drop them from the list.

jparyani:
Done.

shell/server/backup.js
```diff
@@ -41,6 +41,33 @@ function recursiveRmdirIfExists(dir) {
...
+      walkAndWriteFiles(name, writeStream, replaceRoot, originalRoot);
+    } else if (fileStat.isFile()) {
+      name = replaceRoot + name.slice(originalRoot.length);
+      writeStream.write(name + "\n");
+    }
+  });
+
+  // Include empty directories
+  if (files.length === 0) {
+    dir = replaceRoot + dir.slice(originalRoot.length);
```
diracdeltas: Don't you also need to check that `dir` doesn't contain a newline?

jparyani: I think it's fine. Any call to walkAndWriteFiles except for the first can only occur within the forEach loop, after the check for the new line has already happened. This should ensure `dir` can never have a newline in it, except for on the first call. We enter walkAndWriteFiles with a known path (specifically Path.join(SANDSTORM_GRAINDIR, grainId, "sandbox")), so it should be fine.

That being said, the logic is a bit twisty and I'm not against adding another check just for clarity.

diracdeltas: Ah, makes sense. It wasn't obvious to me that this will only get called with a known entry point.

PR #278, opened 2015-02-28
Use /proc/<pid>/comm in killall() implementation.

### LINKED ISSUE #269: Grain isn't restarted when `spk dev` starts/stops
Lately I've noticed that killing `spk dev` or starting it again doesn't seem to restart the grain, so instead it ends up in a very broken state. I thought this worked before.

dwrensha:
Fixes #269.

With this change, the behavior of the `killall()` function more closely matches that of `/usr/bin/killall`. In particular, we match on the contents of `/proc/<pid>/comm`.


dwrensha:
Closing in favor of #279.

PR #279, opened 2015-03-01
Use /proc/<pid>/cmdline in killall() implementation.

### LINKED ISSUE #269: Grain isn't restarted when `spk dev` starts/stops

dwrensha:
Fixes #269.

I think this is strictly superior to #278. Here we match on the entire command string, so we don't risk accidentally killing `sandstorm-supersoaker` or whatever.

### DIFF
--- a/src/sandstorm/run-bundle.c++
+++ b/src/sandstorm/run-bundle.c++
@@ -2324,29 +2324,22 @@ private:
     }
   }
 
-  void killall(kj::StringPtr exePath) {
+  void killall(kj::StringPtr command) {
     for (auto& file: listDirectory("/proc")) {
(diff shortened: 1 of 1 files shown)


kentonv:
Ah, so it's my fault from when I merged the binaries.

We still risk "accidentally killing" in-app processes if the app happens to have a binary called `/bin/sandstorm-supervisor`, but I suppose it's irrelevant since we're killing all the supervisors anyway, which will kill all the apps. In theory we could avoid this problem by _also_ checking that readlink() resolves to the Sandstorm binary, but meh.

PR #287, opened 2015-03-11
Add basic Sandcats support to installer

paulproteus:
This commit shows how Sandcats support will be added to the installer.

If people have feedback, let me know. I intend to continue to revise this over the next few days, including making some changes to the Sandcats backend to support it, and then ship it by the end of the week.

Please *don't* merge this yet! But please feel free to review.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -749,7 +749,6 @@ function register_sandcats_name() {
     # Set these global variables to inform the installer down the
     # road.
     SS_HOSTNAME="${DESIRED_SANDCATS_NAME}.${SANDCATS_BASE_DOMAIN}"
-    DEFAULT_WILDCARD='http://*.'"${SS_HOSTNAME}"
   else
     # Show the server's output, and re-run this function.
     error "$(cat sandcats/register-log)"
--- a/install.sh
+++ b/install.sh
@@ -744,6 +744,8 @@ function register_sandcats_name() {
     # Show the server's output, which presumably is some happy
     # message.
     cat sandcats/register-log
+    # Make sure that is on a line of its own.
+    echo ''
     # Set these global variables to inform the installer down the
     # road.
     SS_HOSTNAME="${DESIRED_SANDCATS_NAME}.${SANDCATS_BASE_DOMAIN}"
--- a/install.sh
+++ b/install.sh
@@ -701,6 +701,7 @@ function register_sandcats_name() {
   # "none" or provide a name they want to register.
   if [ -z "$DESIRED_SANDCATS_NAME" ] ; then
       register_sandcats_name
+      return
   fi
 
   # If the user really wants none of our sandcats help, then bail out.
@@ -710,7 +711,10 @@ function register_sandcats_name() {
 
   # Validate the client-side, to avoid problems, against the same
   # regex that the server is using.
-  [[ $DESIRED_SANDCATS_NAME =~ ^[0-9a-zA-Z]{1,20}$ ]] || register_sandcats_name
+  if ! [[ $DESIRED_SANDCATS_NAME =~ ^[0-9a-zA-Z]{1,20}$ ]] ; then
+    register_sandcats_name
+    return
+  fi
 
   # Ask them for their email address, since we use that as part of Sandcats
   # registration.
@@ -753,6 +757,7 @@ function register_sandcats_name() {
     # Show the server's output, and re-run this function.
     error "$(cat sandcats/register-log)"
     register_sandcats_name
+    return
   fi
 }
 
--- a/install.sh
+++ b/install.sh
@@ -753,6 +753,7 @@ function register_sandcats_name() {
     # Set these global variables to inform the installer down the
     # road.
     SS_HOSTNAME="${DESIRED_SANDCATS_NAME}.${SANDCATS_BASE_DOMAIN}"
+    USE_EXTERNAL_INTERFACE="yes"
     echo "Congratulations! We have registered your ${DESIRED_SANDCATS_NAME}.${SANDCATS_BASE_DOMAIN} name."
     echo "Your credentials to use it are in $(readlink -f var/sandcats); consider making a backup."
   else
--- a/install.sh
+++ b/install.sh
@@ -332,8 +332,8 @@ if [ "yes" = "$USE_SANDCATS" ] ; then
   # possible time-of-check-time-of-use race.
   echo "As a Sandstorm user, you are invited to use a free Internet hostname as a subdomain of sandcats.io."
 
-  # The sandcats service places its authentication files in $DIR/sandcats.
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -312,6 +324,64 @@ mkdir -p "$DIR"
...
 # ========================================================================================
+# Sandcats.io - Dynamic DNS service hosted by Sandstorm.io
+if [ "yes" = "$USE_SANDCATS" ] ; then
+  # We generate the public key before prompting for a desired hostname
+  # so that when the user presses enter, we can try to register the
+  # hostname, and if that succeeds, we are totally done. This avoids a
+  # possible time-of-check-time-of-use race.
+  echo "As a Sandstorm user, you are invited to use a free Internet hostname as a subdomain of sandcats.io."
+
+  # The sandcats service places its authentication files in $DIR/sandcats.
```
kentonv: Perhaps it should be `$DIR/var/sandcats`? I like consolidating user data under `var`, while everything else is stuff that can be re-downloaded.

install.sh
```diff
@@ -312,6 +324,64 @@ mkdir -p "$DIR"
...
+    openssl \
+      req `# Invoke OpenSSL's PKCS#10 X.509 bits.` \
+      -new `# Create a new certificate/request.` \
+      -newkey rsa:4096 `# Create a new RSA key of length 4096 bits.` \
+      -days 3650 `# Make the self-signed cert valid for 10 years.` \
+      -nodes `# no DES -- that is, do not encrypt the key at rest.` \
+      -x509 `# Output a certificate, rather than a signing request.` \
+      `# Sandcats ignores the subject in the certificate; use` \
+      `# OpenSSL defaults.` \
+      -subj "/C=AU/ST=Some-State/O=Internet Widgits Pty Ltd" \
```
kentonv: Wouldn't it make more sense to leave these blank?

install.sh
```diff
@@ -312,6 +324,64 @@ mkdir -p "$DIR"
...
+    # for the `curl` command. It is just as private as id_rsa.
+    cat sandcats/id_rsa sandcats/id_rsa.pub > sandcats/id_rsa.private_combined
+
+    # Set filesystem permissions, in case the files get copied into the
+    # wrong place later.
+    chmod 0600 sandcats/id_rsa sandcats/id_rsa.pub sandcats/id_rsa.private_combined
+
+    # Delete the "..." that we left on the screen, to indicate that we
+    # are done generating cryptographic material.
+    echo -ne '\b\b\b'
```
kentonv: Not important, but FWIW this doesn't actually delete the characters, just moves the cursor back. You have to additionally print spaces to overwrite the chars. On another note, you could use `\r` to mean "go back to beginning of line".

install.sh
```diff
@@ -615,5 +685,76 @@ fi
...
+  fi
+
+  # If the user really wants none of our sandcats help, then bail out.
+  if [ "none" = "$DESIRED_SANDCATS_NAME" ] ; then
+    return
+  fi
+
+  # Validate the client-side, to avoid problems, against the same
+  # regex that the server is using.
+  [[ $DESIRED_SANDCATS_NAME =~ ^[0-9a-zA-Z]{1,20}$ ]] || register_sandcats_name
```
kentonv: Hostnames shouldn't/can't have capital letters. Better disallow them here. (An alternative would be to do the case-insensitive thing, but that's easy to mess up.)

kentonv: On another note, hostnames can have internal hyphens, but not a hyphen as the first char and I think not as the last char (unsure). Not sure if we care to allow those here...

Maybe we should require a minimum length to avoid a land-grab of the short names?

install.sh
```diff
@@ -615,5 +685,76 @@ fi
...
+
+  # Ask them for their email address, since we use that as part of Sandcats
+  # registration.
+  echo "We need your email on file so we help you recover your domain if you lose access. No spam."
```
kentonv: s/so we help/so we can help/

install.sh
kentonv: Also does the recursion on this line actually work? If the `register_sandcats_name` recursion returns won't we incorrectly go on and repeat the rest of the function? (I.e. this isn't _tail_ recursion here...)

install.sh
```diff
@@ -615,5 +685,76 @@ fi
...
 
+function register_sandcats_name() {
+  # We allow environment variables to override some details of the
+  # Sandcats service, so that during development, we can test against
+  # a non-production Sandcats service.
+  SANDCATS_BASE_DOMAIN="${OVERRIDE_SANDCATS_BASE_DOMAIN:-sandcats.io}"
+  SANDCATS_API_BASE="${OVERRIDE_SANDCATS_API_BASE:-https://sandcats.io}"
+  SANDCATS_CURL_PARAMS="${OVERRIDE_SANDCATS_CURL_PARAMS:-}"
+
+  echo "Press enter to accept a random name, or choose your own (alphanumeric, max len 20)."
```
kentonv: I worry that someone might accidentally press enter, or press enter without thinking, and get a random name, which is not what they intended. Perhaps they should type "random" for a random name. I know your argument is that simply repeatedly pressing enter should produce a reasonable result, but IMO a randomly-generated name is probably _not_ what the vast majority of people actually want, and so this will probably just lead to confusion...

kentonv:
I'm excited for this to go live!

install.sh
paulproteus: +1

install.sh
paulproteus: Intriguing. Fixed in a follow-up commit (that I haven't pushed yet, but will shortly).

install.sh
paulproteus: `openssl` doesn't seem to like creating a certificate with empty values for the `-subj`, so I figure we can blend in with the rest of the Internet and use the OpenSSL defaults.

install.sh
paulproteus: re: capital letters: I downcase them on the server-side, and I figure some people are going to type capital letters, so I figure it's OK to let them type it here, and downcase it later in the process.

re: internal hyphens: I'll deal with hyphens in a separate pull request, I think.

re: minimum length: whatever, land grabs are fun!

re: recursion: nope, it doesn't work; fixed in a follow-up commit.

install.sh
paulproteus: +1

install.sh
paulproteus: +1; I removed the random name feature.

paulproteus:
I believe these changes address all known Sandcats installer feedback. If you want, before merging, I can rebase this; I figured it's best for me not to rebase until review is done.

@kentonv ready for a fresh review.

paulproteus:
@kentonv said to me that I can merge this, so I'm going to merge this! (If I'm acting on a misunderstanding, someone please revert and accept my apologies.)

PR #290, opened 2015-03-12
First draft of database schema for sharing permissions.

dwrensha:


shell/shared/db.js
```diff
@@ -86,6 +86,46 @@ Grains = new Mongo.Collection("grains");
...
+//   recipient: The `_id` of the user who receives these permissions.
+//   permissions: An array of bools representing the received permissions.
+//   roleAssignment: A Grain.ViewSharingLink.RoleAssignment that is equivalent to `permissions`, but
+//                   may contain more semantic information. In JSON format.
+//   petname: Human-readable label chosen by and only visible to the sharer.
+//   title: Human-readable title as chosen by the recipient. Used in the same places that
+//          `grain.title` is used for the grain's owner.
+//   created: Date when this role assignment was created.
+//   expires: Optional expiration date. If undefined, the role assignment does not expire.
+//   parentKey: If present, the `_id` of the role assignment key from which this was derived.
```
kentonv: Hmm, how is `parentKey` supposed to be used? Say Alice shares to Carol who shares to Dave. Later, Alice revokes Carol's access, but now Bob separately shares the same role to Carol, restoring her access. Dave's access should be restored as well. So it seems the "parent" assignment has changed.

Or say Alice shares read-only access to Carol, and Bob shares write-only access to Carol, and Carol shares read/write access to Dave. Then there are two parents.

With that said, it may very well make sense to keep some sort of a cache of the authorization path used on a previous authorization check so that we don't have to do a full graph search each time.

shell/shared/db.js
```diff
@@ -86,6 +86,46 @@ Grains = new Mongo.Collection("grains");
...
+//
+// Any grain for which a user has received a role assignment will show up in that user's grain list.
+// Role assignments are also used to restrict access during restoration of persisted capabilities.
+//
+// Each contains:
+//   _id: random
+//   grainId: The `_id` of the grain whose permissions are being shared.
+//   sharer: The `_id` of the user who is sharing these permissions.
+//   recipient: The `_id` of the user who receives these permissions.
+//   permissions: An array of bools representing the received permissions.
```
kentonv: It might make sense to define this as being "bonus" permissions in addition to what the role assignment grants. This way an app can change the mapping of roles to permissions over time. In fact the only reason to have the `permissions` field in this case is if we want to offer an "advanced" UI in which users can select individual permissions, but I'd argue that there's no immediate need to implement, so we should just leave out `permissions` entirely and go based on `roleAssignment`.

shell/shared/db.js
```diff
@@ -86,6 +86,46 @@ Grains = new Mongo.Collection("grains");
...
+//   roleAssignment: A Grain.ViewSharingLink.RoleAssignment that is equivalent to `permissions`, but
+//                   may contain more semantic information. In JSON format.
+//   petname: Human-readable label chosen by and only visible to the sharer.
+//   title: Human-readable title as chosen by the recipient. Used in the same places that
```
kentonv: Interesting! I hadn't considered letting people modify the title in their own capability store, but I like it.

kentonv: Also I suppose it probably improves performance for the purpose of displaying the user's capability store -- no need for a "join" to get the grain metadata.

shell/shared/db.js
```diff
@@ -86,6 +86,46 @@ Grains = new Mongo.Collection("grains");
...
+// Each contains:
+//   _id: random
+//   sharer: The `_id` of the user who created this key.
+//   grainId: The `_id` of the grain whose permissions are being shared.
+//   permissions: An array of bools representing the received permissions.
+//   roleAssignment: A Grain.ViewSharingLink.RoleAssignment that is equivalent to `permissions`, but
+//                   may contain more semantic information. In JSON format.
+//   petname: Human-readable label chosen by and only visible to the sharer.
+//   created: Date when this key was created.
+//   expires: Optional date when this key and all role assignments derived from it should expire.
```
kentonv: I think we may want separate fields for when the key itself expires (and can no longer be exercised to grant access to new people) and when access granted via the key expires. In fact, the former is probably much more useful in practice than the latter: it lets you guard against long-term leakage of keys without needing to re-send new ones to the people you actually meant to share with.

shell/shared/db.js
kentonv: Oh wait, `roleAssignment` actually already contains `addPermissions` for this purpose. So, why do we separately have `permissions` here? Is this just denomalization? I think we probably should avoid that.

src/sandstorm/grain.capnp
```diff
@@ -528,7 +528,12 @@ struct SturdyRefOwner {
   # Human-readable explanation of how this capability relates to the grain. This is displayed
   # when visualizing outgoing capabilities from the calling grain in the Sandstorm UI. This string
   # will NOT be communicated to the host of the saved capability; it is intercepted and used by the
-  # Sansdtorm system.
+  # Sandstorm system.
+
+  userId @1 :Data;
+  # ID of the user whose permissions are tied to this capability. This data will NOT be communicated
```
kentonv: I don't quite understand what's going on here. Are you sure this belongs in `SturdyRefOwner`, which is meant to identify the owner of the reference itself (i.e., the grain), not the owner of the referenced object? Normally the grain fills in `SturdyRefOwner` when making an outgoing `save()` request on some arbitrary capability it has somehow received.

src/sandstorm/supervisor.capnp
```diff
@@ -64,9 +64,6 @@ interface SandstormCore {
   # When the supervisor receives a save() request for a capability hosted by the app, it first
   # calls save() on the underlying capability to get an AppSturdyRef, then calls wrapSaved() to
   # convert this to a token which it can then return.
-  #
-  # TODO(soon): How do we keep this capability associated with the user account that created it,
```
kentonv: Presumably this is what you're trying to solve with adding `userId` to `SturdyRefOwner`.

I think the right solution to this issue is actually that when a capability comes out of the Powerbox, it is automatically wrapped in a revokable membrane that is revoked if the user who performed the Powerbox interaction loses access. The later `save()` call on that powerbox capability goes through this membrane which appropriately annotates the resulting SturdyRef with this information. The membrane could possibly be implemented by the supervisor for the app which hosts the capability, in order to avoid network hops through any other intermediates.

shell/shared/db.js
dwrensha: To be clear, `parentKey` refers to an entry in the `RoleAssignmentKeys` table; it as a role assignment without a recipient.

The main purpose of storing `parentKey` is so that the sharer can know how the recipient got their permissions, and to provide a convenient way to revoke all role assignments derived from a single `RoleAssignmentKey`.

Not that while the effects of a revocation are transitive in the sharing graph, the changes to the state stored in this collection are local.

> Say Alice shares to Carol who shares to Dave. Later, Alice revokes Carol's access, but now Bob separately shares the same role to Carol, restoring her access. Dave's access should be restored as well. So it seems the "parent" assignment has changed.

The "Carol -> Dave" role assignment does not change through Alice's revocation and Bob's restoration. Its `parentKey` continues to represent the URL that Carol created the one time that she clicked the Share button.

> Or say Alice shares read-only access to Carol, and Bob shares write-only access to Carol, and Carol shares read/write access to Dave. Then there are two parents.

Similar story to the previous case.

> With that said, it may very well make sense to keep some sort of a cache of the authorization path used on a previous authorization check so that we don't have to do a full graph search each time.

Right. I'm not even attempting to do that here, but something like that may make sense.

shell/shared/db.js
dwrensha: Hmm. Yeah, I agree. Let's just keep the `roleAssignment` field.

shell/shared/db.js
dwrensha: One thing we'll need to sort out is how to merge such titles when a user receives multiple role assignments.

Maybe the oldest role assignment is the one we use for the title, and we're careful when deleting a role assignment to transfer the title if necessary.

dwrensha: Just to be explicit here: a big reason I want recipient-controlled titles is that I don't want a grain owner to be able change a title out from under the feet of a recipient.

src/sandstorm/grain.capnp
dwrensha: Good point.

shell/shared/db.js
dwrensha: Good point.

src/sandstorm/supervisor.capnp
dwrensha: Interesting! I think this addresses my concern.

src/sandstorm/grain.capnp
dwrensha: Yeah, I couldn't see where else the `userId` was going to come from. It sounds like the PowerBox is going to be able to handle this, which is cool.

shell/shared/db.js
dwrensha: It occurs to me that the problem of performing an authorization check given the `RoleAssignments` collection can be thought of as a problem in first-order logic programming, as in [Prolog](http://en.wikipedia.org/wiki/Prolog). That is to say, it's a well-studied problem that probably has some nice properties we can exploit.

dwrensha:
I've pushed a second draft which I hope clarifies exactly what is being stored in the `RoleAssignments` collection.

shell/shared/db.js
dwrensha: I think it should be feasible for the actual `(user, grain) -> permissions` map to be lazily computed and cached in-memory.

kentonv: Oops, yes, I misunderstood that this pointed into `RoleAssignmentKeys` rather than `RoleAssignments`. Maybe explicitly say `entry in RoleAssignmentKeys` instead of `role assignment key`?

shell/shared/db.js
```diff
@@ -86,6 +86,46 @@ Grains = new Mongo.Collection("grains");
...
 
+RoleAssignments = new Mongo.Collection("roleAssignments");
+// Edges in the permissions sharing graph.
+//
+// To share a permission with another user is to declare "if I have this permission, then so should
+// the recipient". A bundle of such declarations by a single sharer directed at a single recipient
+// for a single grain is called a "role assignment". A grain's owner always has every permission,
+// but permissions for other users must be computed from this collection.
+//
+// Every role assignment implicitly shares the special "can view grain" permission, which dictates
```
kentonv: This comment could be confusing in that "can view grain" could be taken to mean "has read-only access to grain content", but here I think you only mean "can tell that the grain exists".

I'm not really sure I'd describe this as a permission. I might say something like: "Receiving any role assignment causes the grain to appear in the user's file list, even if the assignment provides no actual permissions."

kentonv:
Other than comment nitpicks this looks good to me.

shell/shared/db.js
dwrensha: I want to describe it as a permission because its value will be computed that same way as for other permissions -- this is, flowing from the grain owner along RoleAssignments.

I think "can tell that the grain exists" is a weaker permission than I have in mind here. I'm thinking more like "can get a UiSession".

dwrensha: A user can get a UiSession if there's a path of RoleAssignments from the grain owner to the user.

A user can tell that a grain still exists if they have a RoleAssignment to that grain that hasn't been automatically delete by the system yet.

dwrensha: In particular, you can know that a grain still exists without being able to open a UiSession to it. And open a UiSession to a grain without having any permissions (other than this special implicit permission that I'm describing).

kentonv: OK, I see what you're getting at now -- the empty permission may actually grant the user some ability to access the grain, depending on the app. Whether this can be described as "view" depends on the app, though.

But there's another issue here: It sounds like you're saying that deleting a RoleAssignment will transitively delete downstream assignments if they aren't backed elsewhere. I actually don't think we want that -- If Alice shares to Bob who shares to Carol, and then Alice temporarily revokes Bob (perhaps by accident) and then restores his access, Carol's access needs to be restored as well. So the Bob -> Carol edge can't really be deleted. And meanwhile, we probably can't do a full access check for every capability in your account when rendering a list, so Carol will still see the grain even though she has lost access. Once she clicks on it, she'll get an error. Annoying UX but probably rare... and actually arguably less annoying then having the document disappear entirely without explanation.

dwrensha: > It sounds like you're saying that deleting a RoleAssignment will transitively delete downstream assignments if they aren't backed elsewhere.

No, that's not what I mean. I think a RoleAssignment should only be garbage collected when its grain gets deleted. All other changes to the RoleAssignments table should be local, i.e. should not propagate.

dwrensha: > If Alice shares to Bob who shares to Carol, and then Alice temporarily revokes Bob (perhaps by accident) and then restores his access, Carol's access needs to be restored as well.

Alice's state for this grain should go from "knows exists, can open" to "knows exists, can't open" and then back to "knows exists, can open".

dwrensha: An alternative design might be: if a user cannot open a grain, then they are no way of knowing whether the grain exists. But then what about Bob? After the Alice -> Bob RoleAssignment is revoked, is Bob allowed to revoke the Bob -> Carol RoleAssignment? If yes, then we can't garbage collect it when the grain gets deleted, because doing so would allow Bob to detect that the grain no longer exists.

In any case, I think these questions about who knows which grains exist are somewhat orthogonal to this pull request, and can be resolved later.

kentonv: I don't think it's a big deal to leak the fact that a grain was deleted to users who at one point had access but were later revoked, so let's not worry about that.

Separately from that, though, this comment as it stands still seems to be saying that "can open a session" and "shows up in grain list" are the same thing. But in the Alice -> Bob -> Carol case, after Bob's access is revoked, Carol cannot open a session, but will still see the capability in her grain list unless we transitively delete edges, which we don't want to do.

dwrensha:
I've clarified the comments further and I've added an 'active' field that allows a role assignment to be revoked without being removed from the collection.

shell/shared/db.js
```diff
@@ -86,6 +86,48 @@ Grains = new Mongo.Collection("grains");
...
+//                   received permissions. The sharer is allowed to modify this later.
+//   active: Flag indicating that this role assignment has not been revoked. The sharer is allowed
+//           to flip this bit, but only the recipient is allowed to delete the role assignment.
+//   petname: Human-readable label chosen by and only visible to the sharer.
+//   title: Human-readable title as chosen by the recipient. Used in the same places that
+//          `grain.title` is used for the grain's owner.
+//   created: Date when this role assignment was created.
+//   parentKey: If present, the `_id` of the entry in RoleAssignmentKeys from which this was derived.
+
+RoleAssignmentKeys = new Mongo.Collection("sharingUrls");
```
kentonv: Should the mongo collection be called `roleAssignmentKeys` for consistency?

dwrensha: Heh, good catch.

dwrensha: Fixed.

dwrensha:
It occurs to me that, with a few more tweaks, the `RoleAssignments` collection that we've designed here could in fact completely replace the `Grains` collection.  Any entry in `Grains` can be thought of as an `allAccess` role assignment where the sharer is a special omnipotent user with `userId = 0` and the recipient is the grain owner. We might even do away with the `RoleAssigments.grainId` field, as we'll need to walk back to the root assignment to check authorization anyway, so we might as well grab the ID while we're at it.

kentonv:
Hmm, I think you still need `grainId` to be able to merge role assignments coming from different people. Otherwise how would you tell that two assignments are for the same grain?

As for the `Grains` table, we do need to store `packageId` and such somewhere.

I see what you're saying about the grain initially being shared from `userId = 0`, and that sounds elegant, but note that the owner of a grain is actually very different from any other `allAccess` user in that the grain's storage is charged against the owner's quota, and the owner has the right to delete it. So, there's a wart...

All that said, the idea of having just one collection to search for both your own grains and grains shared to you sounds kind of nice.

PR #294, opened 2015-03-16
Comment out failing "Test grain restart" test for now

jparyani:
I think there's an actual transient error here, not just an artifact of
nightwatch timing issues. I'm having trouble tracking it down, so I'm
going to disable the test for now.

### DIFF
--- a/tests/tests/grain.js
+++ b/tests/tests/grain.js
@@ -134,16 +134,18 @@ module.exports = utils.testAllLogins({
       // TODO(someday): detect if error occurred, since there's no way for selenium to verify downloads
   },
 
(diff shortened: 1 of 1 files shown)


paulproteus:
Sad to disable the test, but maybe this is for the best? I'm not sure, but wanted to express general sadness without actively blocking the pull request.

kentonv:
As discussed I'm inclined not to merge this because we should actually fix this bug, and until we do we should be reminded that it's not fixed. :)

jparyani:
Feel free to close this. We haven't seen this issue in a while, and as discussed, this is never going to be merged anyways.

PR #295, opened 2015-03-17
Add interface for background processing and expand SturdyRef infrasturcture

kentonv:
This started as an attempt to add interfaces for background processing -- that is, letting the app request that it not be shut down even if no sessions are open.

This expanded into defining some basic interfaces for notifications. Background processing requires displaying a notification to the user, and I didn't want to write a hack that would change as soon as notifications are truly implemented.

Things got complicated when I realized that the front-end is able to restart independently of any particular grain. This implies that capabilities related to keeping the notification open need to be persistent. (We could perhaps add more hacks that define some new kind of ID, but I'd rather start writing things the way we're supposed to be writing them long-term.)

But thinking about that made me realize that the SturdyRef stuff is missing some things. For instance, sometimes the supervisor or the front-end need to host persistent capabilities, therefore ApiToken needs a way to represent those.

Moreover, I realized that we really do need a way to delete a SturdyRef if the "handle" pattern is to extend to persistent handles. So, I went through and added a `drop()` method to go with every `restore()` method, along with a discussion of all of the issues that come along with `drop()` (basically, if you want to use `drop()`, then you must implement `save()` such that it returns a new unique ref every time, because reference counting is not really possible here.)

So, this is what I came up with.

@dwrensha does this affect what you're working on right now? Or does basic sharing turn out not to use SturdyRefs?

### DIFF
--- a/shell/shared/db.js
+++ b/shell/shared/db.js
@@ -186,8 +186,9 @@ ApiTokens = new Mongo.Collection("apiTokens");
 //
 // Originally API tokens were only used by external users through the HTTP API endpoint. However,
 // now they are also used to implement SturdyRefs, not just held by external users, but also when
-// an app holds a SturdyRef to another app within the same server. See `SturdyRef` in
-// `grain.capnp` -- the type of `external` is an API token.
+// an app holds a SturdyRef to another app within the same server. See the various `save()`,
+// `restore()`, and `drop()` methods in `grain.capnp` (on `SansdtormApi`, `AppPersistent`, and
+// `MainView`) -- the fields of type `Data` are API tokens.
 //
 // Each contains:
 //   _id:       A SHA-256 hash of the token.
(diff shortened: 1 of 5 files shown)


kentonv:
I've made an attempt at simplifying these interfaces, particularly from the app's perspective: To save a capability, the app should no longer call save() on the cap itself but rather on SandstormApi, passing in the capability. This allows all saved capabilities to be represented as plain API tokens, which is probably easier for developers to get their head around. If the app passes one of its own local capabilities to save(), the supervisor can still create an API token for it, making it look the same as everything else. This also means that drop() can now be meaningful, as the supervisor will call it only when all known refs to the object are dropped.

Apps also no longer implement Cap'n Proto's `Persistent` to make their own objects persistent, but rather implement the new `AppPersistent` interface. Since this interface is only intended to be called by the supervisor to handle incoming save requests -- _not_ for use by the app to make outgoing save requests, since `SandstormApi.save()` now does that -- there is no need for this type to return a union of a "loopback" app ID vs. an "external" API token. This should reduce cognitive load for people implementing this. (One possible problem is that the app can no longer forward an incoming `save()` call to an external capability, but it's unclear that that would ever be desirable -- in fact, it would more likely be a security bug.)

It may actually make sense to eliminate the Cap'n Proto common `Persistent` interface and instead define a generalized notion of "realm gateways" that lets you intercept arbitrary interfaces and translate them to arbitrary other interfaces. Then each realm defines its own equivalent of `Persistent` as appropriate for it, and we can still bridge between them at gateways. E.g. the supervisor would bridge incoming save requests from SystemPersistent (supervisor.capnp) to AppPersistent (grain.capnp).

dwrensha:
> It may actually make sense to eliminate the Cap'n Proto common Persistent interface and instead define a generalized notion of "realm gateways" that lets you intercept arbitrary interfaces and translate them to arbitrary other interfaces.

I like this idea a lot.

kentonv:
I've addressed David's comments, other than the one I replied to.

I've decided that modeling the wake lock as an RPC call probably doesn't make any particular sense after all, so I'm going to merge this as-is. (Though this is just an interface so it can still be tweaked before/during implementation.)

kentonv:
(That is to say, modelling it as a call would not simplify anything, and would perhaps in fact be more complex.)

PR #296, opened 2015-03-19
spk: make --pkg-def switch work correctly

zarvox:
The variable sawPkgDef was initialized to false, and never set to true when the
pkgdef was loaded successfully.

As a result, specifying --pkg-def would load the user-specified package def, and then ensurePackageDefParsed would additionally attempt to load the default one, which would fail.


garply:

kentonv:
Thanks! Embarrassing, apparently we never tested that flag.

PR #303, opened 2015-03-26
Refactor login and add settings

jparyani:
This had a lot of parts to it:
* Add sandstorm-accounts-oauth package, forked from accounts-oauth to
  add a `deregisterService` method, and to check at runtime what oauth
  services are allowed to login
* Add sandstorm-accounts-packages. It includes github/google meteor
  packages, but doesn't register the accounts service at startup like
  accounts-{github,google} do
* Add sandstorm-accounts-ui to fix some display bugs with non OAuth login
  services like demo and devAccounts
* Move shared/db.js to lib/db.js to ensure it gets loaded first
* Had to fix up demo/devAccounts to follow the new accounts registration
  system
* Made the adminSettings page


ocdtrekkie:
Is this going to make it possible to add/remove login providers?

jparyani:
Yep, this makes login providers configurable. Hopefully this will allow us to add all kinds of new login providers and let the server administrator pick which they want to support.

Unfortunately, we don't have a story for runtime plugins yet. That means to add a custom login provider, you'll have to add it as a meteor package and re-build Sandstorm. That being said, we're happy to accept PRs that add login services, but ask on the mailing list first since there's some trickiness in how to make the IDs for users global. I'll also probably add twitter and/or facebook login at some point to show how adding services should go.

shell/lib/db.js
```diff
@@ -250,6 +250,16 @@ Misc = new Mongo.Collection("misc");
 //   _id:       The name of the setting. ie. "BASE_URL"
 //   value:     The value of the setting.
 
+Settings = new Mongo.Collection("settings");
+// Settings for this Sandstorm instance go here. They are configured through the adminSettings
+// route. This collection differs from misc in that this collection is completely user controlled.
+//
+// Each contains:
+//   _id:       The name of the setting. ie. "MAIL_URL"
```
kentonv: s/ie/eg/

shell/lib/db.js
```diff
@@ -321,6 +331,18 @@ isAdmin = function() {
   }
 }
 
+isAdminById = function(id) {
+  check(id, String);
+  // Returns true if the user's id is the administrator.
+
+  var user = Meteor.users.findOne({_id: id})
```
kentonv: use `{fields: {isAdmin: 1}}` to select just the `isAdmin` field. (See examples in `stats.js` -- you could probably change them to use this function, in fact.)

shell/lib/db.js
```diff
@@ -321,6 +331,18 @@ isAdmin = function() {
   }
 }
 
+isAdminById = function(id) {
+  check(id, String);
+  // Returns true if the user's id is the administrator.
```
kentonv: doc comment should go before check

shell/lib/db.js
```diff
@@ -321,6 +331,18 @@ isAdmin = function() {
   }
 }
 
+isAdminById = function(id) {
+  check(id, String);
```
kentonv: Maybe just return false if id is undefined? Otherwise I think all the places where you used this will throw if used by a non-logged-in user.

shell/shared/admin.js
```diff
@@ -0,0 +1,204 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+var NUM_SETTINGS = 4;
+var accountServiceNames = ["google", "github", "devAccounts"];
+
+Router.map(function () {
+  this.route("adminSettings", {
+    path: "/adminSettings",
```
kentonv: How about just `/admin`?

shell/client/shell.html
```diff
@@ -441,6 +442,33 @@
   </div>
 </template>
 
+<template name="adminSettings">
+  <div id="topbar">{{> homeLink}} Admin settings</div>
+
+  <div id="admin-settings" class="main-content">
+    <div class="centered-box">
+      <form id="admin-settings-form">
+        <label title="This setting enables Google OAuth accounts for this server">Enable Google Login <input type="checkbox" name="googleLogin" value="value" checked={{googleEnabled}} /></label>
```
kentonv: These tooltips seem not very discoverable. Could we just put the description of each setting directly under it in smaller text?

shell/client/shell.html
```diff
@@ -441,6 +442,33 @@
...
+        <br />
```
kentonv: `/>` is not valid HTML. Just use `>`. (With the `input` tags as well.)

shell/shared/devAccounts.js
```diff
@@ -14,112 +14,91 @@
...
-          userId = Accounts.insertUserDoc({ profile: { name: displayName } },
-                                          { signupKey: "devAccounts", devName: displayName, isAdmin: isAdmin });
-        }
-        // Log them in on this connection.
-        return Accounts._loginMethod(this, "createDevAccount", arguments,
-            "devAccounts", function () { return { userId: userId }; });
+if (Meteor.isServer) {
+  Meteor.methods({
+    createDevAccount: function (displayName, isAdmin) {
+      if (!allowDevAccounts()) {
```
kentonv: I hate to say this after you've done all this refactoring in this file, but I wonder if we really want dev accounts to be toggle-able at runtime. Probably, we should force the user to choose between dev and prod at install time and not allow them to easily switch since it would be easy to get wrong. E.g. someone might idly enable dev accounts just to see what they are, not reading the fine print and not realizing they just opened a security hole on their server.

Perhaps we should keep this strictly in the config file...

shell/packages/sandstorm-accounts-ui/LICENSE
```diff
@@ -0,0 +1,20 @@
+forked from Meteor's accounts-ui from https://github.com/meteor/meteor/tree/ab9ef40d258bbb9b9de453600495c8337577c565/packages/accounts-ui-unstyled
```
kentonv: Sorry, I guess when you asked if I wanted this in multiple changes I wasn't really thinking about the fact that you were bringing in Meteor code. It would be nice to be able to see a diff between the original meteor code and your code. Could we submit the code verbatim first, then update it as a second change?

kentonv: (I'm less worried about the fork of accounts-oauth as that's only temporary and I saw the changes that went upstream.)

shell/packages/sandstorm-accounts-packages/accounts.js
```diff
@@ -0,0 +1,50 @@
+if (Meteor.isClient) {
+  Meteor.loginWithGoogle = function(options, callback) {
```
kentonv: I notice that the place that calls `loginWith*` is actually in `accounts-ui`, which we've forked, so we can change this. Considering that the Google and Github versions of this method are identical except for the package they call into, maybe it makes sense to have just one method that either takes the package as a parameter or takes a name and looks up the package in a map?

jparyani:
Addressed all your changes, except for the last one:

>I notice that the place that calls loginWith* is actually in accounts-ui, which we've forked, so we can change this. 

Moving this functionality to accounts-ui doesn't make sense, and would actually hurt us whenever we move back to using upstream accounts-oauth/accounts-github/accounts-google/etc.

kentonv:
Doh, apparently this is no longer mergeable. Can you update it?

kentonv:
Hmm, moving db.js arguably should have happened as a separate PR since it conflicts with probably everything everyone is doing. I'm about to review some other PRs with changes there...

jparyani:
Ping me when you're done with the rest of the PRs and I'll rebase.

kentonv:
OK Asheesh's change affecting db.js is in.

jparyani:
Done

PR #305, opened 2015-03-27
Change install to redirect to appdemo for non logged in users

### LINKED ISSUE #292: Sandstorm Apps page with demo requires confusing login
If I go to https://sandstorm.io/apps , choose Demo then try to install an app, I get 

Failed to install app!
You must sign in to install packages.

I think this "Configure Demo Login" should be automatic for now, it's an unnecessary step and probably confusing because trying the other forms of login gives another error:

Failed to install app!
Sorry, Sandstorm is in closed alpha.  You must receive an alpha key before you can install packages.

jparyani:
Along with https://github.com/sandstorm-io/sandstorm-website/pull/53,
this fixes #292

### DIFF
--- a/shell/shared/demo.js
+++ b/shell/shared/demo.js
@@ -14,7 +14,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-var allowDemo = Meteor.settings && Meteor.settings.public &&
+allowDemo = Meteor.settings && Meteor.settings.public &&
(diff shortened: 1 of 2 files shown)


paulproteus:
In theory, this means it'll be easy to cause demo-enabled Sandstorms to show a "Try No such package >>" button, but I guess that's already doable, so it doesn't make things any worse.

paulproteus:
Generally: yippee for this change. I'm +1 on it.

PR #306, opened 2015-03-27
Change DNS error messages to be html

### LINKED ISSUE #283: If the user visits Sandstorm at a host that shares an IP with the IP of the BASE_URL, redirect to BASE_URL
Right now, when I surf to localhost:6080, I get:

```
Error looking up DNS TXT records for host 'localhost': queryTxt ENOTFOUND

This Sandstorm server's main interface is at: http://local.sandstorm.io:6080
```

I would prefer that this do a HTTP redirect to local.sandstorm.io:6080.

I'm open to figuring under what circumstances it should do this. Answers include:

* Always, if it's not a host that's part of the wildcard host spec.

* Only if the host where Sandstorm is being seems to DNS-resolve to the same IP address as the host of the BASE_URL

* Your idea here

jparyani:
This allows us to make the link back to the server an actual <a> tag.

Fixes #283

### DIFF
--- a/shell/server/pre-meteor.js
+++ b/shell/server/pre-meteor.js
@@ -123,23 +123,23 @@ Meteor.startup(function () {
         }
       });
     }).catch(function (err) {
-      res.writeHead(500, { "Content-Type": "text/plain" });
+      res.writeHead(500, { "Content-Type": "text/html" });
       res.end(err.message);
     });
   });
 });
 
 var errorTxtMapping = {};
-errorTxtMapping[Dns.NOTFOUND] = "\n" +
-    "If you were trying to connect this address to a Sandstorm app hosted at this server,\n" +
-    "you either have not set your DNS TXT records correctly or the DNS cache has not\n" +
-    "updated yet (may take a while).\n";
+errorTxtMapping[Dns.NOTFOUND] = "<br>\n" +
+    "If you were trying to connect this address to a Sandstorm app hosted at this server,<br>\n" +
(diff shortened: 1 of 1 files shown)


shell/server/pre-meteor.js
```diff
@@ -123,23 +123,23 @@ Meteor.startup(function () {
...
     });
   });
 });
 
 var errorTxtMapping = {};
-errorTxtMapping[Dns.NOTFOUND] = "\n" +
-    "If you were trying to connect this address to a Sandstorm app hosted at this server,\n" +
-    "you either have not set your DNS TXT records correctly or the DNS cache has not\n" +
-    "updated yet (may take a while).\n";
+errorTxtMapping[Dns.NOTFOUND] = "<br />\n" +
```
kentonv: `<br />` (and, in general, ending a tag with `/>`) is actually not valid HTML. It is valid XHTML, but contrary to popular belief XHTML is not a subset of valid HTML, and serving XHTML with `text/html` as the content type is incorrect. See: http://hixie.ch/advocacy/xhtml (<- this is the document that basically killed XHTML ~10 years ago. As it turns out, almost no one has ever actually used XHTML in production because almost everyone has always used `text/html` as the content type, but a lot of people _thought_ they were using XHTML.)

So, these should just be `<br>`.

jparyani:
Fixed

ocdtrekkie:
I think I probably learn more reading Kenton's code review notes than anything else I read on the Internet. I swear we were "supposed" to use <br /> to be all correct and such, though I don't think I really ever do write it that way.

PR #307, opened 2015-03-27
Add appMarketingVersion to the right of user actions

### LINKED ISSUE #276: Show app "marketing version" in UI
The app manifest contains a field `appMarketingVersion` which is an arbitrary string that should indicate the app version (e.g. "1.23.9-alpha1~asheesh2" or whatever else you want). We should actually display this in the UI, and make sure our ports all fill this in, so that users can tell what version they have installed. Note that since old grains could be pinned to older versions, we'll need to have a way to inspect the version per-grain. Probably, we need to merge a bunch of the top bar functionality into a drop-down menu that could also include an "about" option, or something.

jparyani:
Fixes #276

### DIFF
--- a/shell/shared/shell.js
+++ b/shell/shared/shell.js
@@ -257,6 +257,12 @@ if (Meteor.isClient) {
       return Session.get("selectedApp");
     },
 
+    selectedAppMarketingVersion: function () {
+      var appMap = this.appMap;
+      var app = appMap && appMap[Session.get("selectedApp")];
+      return app && app.appMarketingVersion && app.appMarketingVersion.defaultText;
+    },
+
     selectedAppIsDev: function () {
       var app = Session.get("selectedApp");
       return app && DevApps.findOne(app) ? true : false;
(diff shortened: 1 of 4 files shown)


shell/shared/shell.js
```diff
@@ -257,6 +257,14 @@ if (Meteor.isClient) {
       return Session.get("selectedApp");
     },
 
+    selectedAppMarketingVersion: function () {
+      var controller = Iron.controller();
+      var appMap = controller.state.get("appMap");
```
kentonv: Hmm, it would be more consistent with the other helpers to do UserActions.find() here.

Otherwise, put appMap into the data object (returned by data()) and access it there, since that's what we do everywhere else. I haven't seen this `controller` approach before and it seems a little confusing to have multiple ways of filling in the template...

jparyani:
Fixed. Went with adding `appMap` to data since it better masks the difference between devApps and real userActions.

PR #310, opened 2015-03-28
Use find -exec syntax for chown/chgrp for long names

canadaduane:
When executing the xargs version on my machine, the number and length of files exceeded the allowed size and failed. The -exec worked, however.

### DIFF
--- a/shell/run-dev.sh
+++ b/shell/run-dev.sh
@@ -20,13 +20,11 @@ fi
 . $SANDSTORM_HOME/sandstorm.conf
 
 if [ "$SERVER_USER" != "$USER" ]; then
-  echo "Please change your Sandstorm installation to be own by your own user" >&2
+  echo "Please change your Sandstorm installation to be owned by your own user" >&2
   echo "account. E.g. run as root:" >&2
   echo "  $SANDSTORM_HOME/sandstorm stop" >&2
-  echo "  find $SANDSTORM_HOME/var -user $SERVER_USER -print0 | \\" >&2
-  echo "      xargs -0 chown -h $USER" >&2
-  echo "  find $SANDSTORM_HOME/var -group $(id -gn $SERVER_USER) -print0 | \\" >&2
-  echo "      xargs -0 chgrp -h $(id -gn)" >&2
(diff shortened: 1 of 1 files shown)


garply:

paulproteus:
Garply, ok to test

kentonv:
Hmm, `xargs` is supposed to automatically split the argument list into multiple invocations if it's too long for one. Was that not happening here? What was the error message?

Would it make sense to use the `-exec command +` variant rather than `-exec command ;` in order to reduce the number of times the command is executed, or does that lead to the same problem?

canadaduane:
Ahh, I see what I did wrong. I put 'sudo' in the wrong place, which gives a deceptive message:

```
$ find /opt/sandstorm/var -user sandstorm -print0 | xargs -0 sudo chown -h duane
find: `/opt/sandstorm/var/sandstorm/grains': Permission denied
sudo: unable to execute /bin/chown: Argument list too long
sudo: unable to execute /bin/chown: Argument list too long
sudo: unable to execute /bin/chown: Argument list too long
sudo: unable to execute /bin/chown: Argument list too long
sudo: unable to execute /bin/chown: Argument list too long
sudo: unable to execute /bin/chown: Argument list too long
sudo: unable to execute /bin/chown: Argument list too long
sudo: unable to execute /bin/chown: Argument list too long
sudo: unable to execute /bin/chown: Argument list too long
sudo: unable to execute /bin/chown: Argument list too long
sudo: unable to execute /bin/chown: Argument list too long
sudo: unable to execute /bin/chown: Argument list too long
sudo: unable to execute /bin/chown: Argument list too long
sudo: unable to execute /bin/chown: Argument list too long
```

kentonv:
Ah, yeah, that'd do it. In fact, if you don't run `find` under `sudo` you may miss files that are not visible to your user account.

I do like the `-exec` approach. It's simpler and easier to combine with `sudo`. I think it would be a good idea to change the `;` to `+` but then I'm happy to accept this.

canadaduane:
I'm at a different computer right now, so I don't have easy access. If
you'd like to switch it, by all means do :)

kentonv:
Will do, thanks.

PR #313, opened 2015-03-30
Suggest make update-deps if ekam fails

paulproteus:
This changeset also makes the color function not silence its own
output -- that is, it removes the @ from the printf -- because otherwise,
invoking $(call color,foo) as part of x || $(call color,foo) would result
in the shell complaining that @printf is not found.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,7 +139,8 @@ tmp/.ekam-run: tmp/ekam-bin src/sandstorm/* tmp/.deps
 
 continuous:
 	@CC="$(CC)" CXX="$(CXX)" CFLAGS="$(CFLAGS)" CXXFLAGS="$(CXXFLAGS2)" \
-	    LIBS="$(LIBS)" NODEJS=$(NODEJS) ekam -j$(PARALLEL) -c -n :41315
+	    LIBS="$(LIBS)" NODEJS=$(NODEJS) ekam -j$(PARALLEL) -c -n :41315 || \
+	    $(call color,You probably need to install ekam and put it on your path; see github.com/sandstorm-io/ekam) && false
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -130,9 +131,10 @@ tmp/ekam-bin: tmp/.deps
 	     cd ../.. && ln -s ../deps/ekam/bin/ekam-bootstrap tmp/ekam-bin)
 
 tmp/.ekam-run: tmp/ekam-bin src/sandstorm/* tmp/.deps
-	$(call color,building sandstorm with ekam)
+	@$(call color,building sandstorm with ekam)
 	@CC="$(CC)" CXX="$(CXX)" CFLAGS="$(CFLAGS)" CXXFLAGS="$(CXXFLAGS2)" \
-	    LIBS="$(LIBS)" NODEJS=$(NODEJS) tmp/ekam-bin -j$(PARALLEL)
+	    LIBS="$(LIBS)" NODEJS=$(NODEJS) tmp/ekam-bin -j$(PARALLEL) || \
+	$(call color,ekam failed. You might need to: make update-deps) && false
```
kentonv: Indent to make clear that this is a continuation of the previous line.

Maybe say "build failed" rather than "ekam failed" to avoid confusing people who don't know what ekam is?

paulproteus:
Good points; addressed in the above.

paulproteus:
I also added a patch that prints an informative error if ekam fails to start; see above.

(Also unrelated, @shaunagm and @jvns might appreciate this pull request; see thread at https://twitter.com/b0rk/status/582356512489734144 )

paulproteus:
Yay, thanks @kentonv !

PR #315, opened 2015-03-31
Export stats on appDemoUsers separately

paulproteus:
This change permits us to export appDemo user stats separately when viewing the JSON stats export.

@jparyani promised me that if I pushed this, he'd make the corresponding change on the Sandstorm internal stats dashboard.

For non-internal people interested in Sandstorm stats, you might like to see how to get JSON stats data from your own Sandstorm install:

* http://local.sandstorm.io:6080/stats (or whatever URL reaches your own Sandstorm)
* Click on the API token at the bottom of that page
* Ta-da, stats!

I have tested this in the following way:

* Create two demo users, each of which use an app
* Create an appDemo user by visiting an appdemo URL and logging in, which also creates an instance of that app
* View the mongo data on the User objects; note that there are 2 'demo' users and 1 'appDemo' user
* Remove the time constraint from `cleanupExpiredUsers()` and then look at the `DeleteStats` Mongo collection and notice that there are 2 `users` and 1 `appDemoUser`s

I re-did the manual test steps above, and will leave my laptop on overnight in order to trigger the real time-based `cleanupExpiredUsers()` call in the hopes of having an even more accurate test, but I already believe this is ready to merge.

### DIFF
--- a/shell/shared/demo.js
+++ b/shell/shared/demo.js
@@ -86,12 +86,13 @@ if (Meteor.isServer) {
         // isAppDemoUser is important for stats; see
         // cleanupExpiredUsers().
         check(displayName, String);
+        check(isAppDemoUser, Boolean);
 
         // Create the new user.
         var expires = new Date(Date.now() + DEMO_EXPIRATION_MS);
         var userId = Accounts.insertUserDoc({ profile: { name: displayName } },
                                             { expires: expires,
-                                              isAppDemoUser: !!isAppDemoUser
+                                              isAppDemoUser: isAppDemoUser
                                             });
 
         // Log them in on this connection.
--- a/shell/shared/demo.js
+++ b/shell/shared/demo.js
@@ -48,7 +48,7 @@ if (Meteor.isServer) {
     // Delete expired demo accounts and all their grains.
 
     var now = new Date();
-    Meteor.users.find({expires: {$lt: now}}, {fields: {_id: 1, lastActive: 1}})
+    Meteor.users.find({expires: {$lt: now}}, {fields: {_id: 1, lastActive: 1, isAppDemoUser: 1}})
                 .forEach(function (user) {
       Grains.find({userId: user._id}, {fields: {_id: 1, lastUsed: 1}})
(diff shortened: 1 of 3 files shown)


shell/shared/demo.js
```diff
@@ -71,15 +79,20 @@ if (Meteor.isServer) {
...
-      createDemoUser: function (displayName) {
-        // This is a login method that creates a new temporary user every time it is used.
-
+      createDemoUser: function (displayName, isAppDemoUser) {
+        // This is a login method that creates a new temporary user
+        // every time it is used.
+        //
+        // isAppDemoUser is important for stats; see
+        // cleanupExpiredUsers().
         check(displayName, String);
```
kentonv: Please add a check() for `isAppDemoUser`'s type. Granted, your code will deal reasonably with incorrect types due to your `!!`, but as a rule we should always add check()s because it's easy to have Mongo injection bugs otherwise.

shell/shared/stats.js
```diff
@@ -20,7 +20,11 @@ if (Meteor.isServer) {
   function computeStats(since) {
     return {
       activeUsers: Meteor.users.find({lastActive: {$gt: since}}).count() +
-          DeleteStats.find({type: "user", lastActive: {$gt: since}}).count(),
+          DeleteStats.find({$or: [{type: "user"},
```
kentonv: Instead of doing a possibly-expensive "or" query, maybe just add `appDemoUsers` into `activeUsers` after the fact?

paulproteus: +1

shell/shared/demo.js
paulproteus: +1

paulproteus:
I've pushed this so Jenkins can sanity-check it; I'm also testing it locally, which will take a few minutes.

paulproteus:
I've re-tested this, and I believe it's ready to merge, @kentonv (assuming that Jenkins likes it).

paulproteus:
Yippee!

PR #318, opened 2015-04-03
Refactor supervisor to allow some limited reuse as a library

kentonv:
(I'm making this a pull request so that it goes through Jenkins, since otherwise I'm not completely confident that this didn't break something.)


paulproteus:
< 3​

PR #320, opened 2015-04-04
Simplify session management.

dwrensha:
This gets rid of the ugly bit where we have a separate subscription for each session.

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -154,10 +154,9 @@ Meteor.methods({
       grainId: grainId,
       hostId: proxy.hostId,
       timestamp: new Date().getTime(),
-      userId: userId
+      userId: userId,
+      connectionId: this.connection.id
     });
-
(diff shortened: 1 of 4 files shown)


kentonv:
Hmm, what happens if the WebSocket gets disconnected and reconnected, or if the server gets updated and restarts? Currently app sessions may survive these things...

dwrensha:
Good point. In the future, Meteor might automatically handle reconnects, according to [the docs](http://docs.meteor.com/#/full/meteor_onconnection):

> Currently when a client reconnects to the server (such as after temporarily losing its Internet connection), it will get a new connection each time. The onConnection callbacks will be called again, and the new connection will have a new connection id.

> In the future, when client reconnection is fully implemented, reconnecting from the client will reconnect to the same connection on the server: the onConnection callback won't be called for that connection again, and the connection will still have the same connection id.

Maybe once that's implemented we'll be able to simplify along the lines of this pull request. Until then, closing.

PR #328, opened 2015-04-10
implement sharing of grains with permissions (was: "Work in progress on sharing")

dwrensha:
I'm making this a pull request to encourage discussion. It's not yet ready to merge.

Two important things that I'm not completely happy with yet are the URL scheme and the database schema.


Makefile
```diff
@@ -48,6 +48,7 @@ IMAGES= \
     shell/public/wrench.png \
     shell/public/download.png \
     shell/public/key.png \
+	shell/public/share.png \
```
kentonv: seems your editor inserted a tab here where the surrounding lines use spaces (below for share-m.png, too)

kentonv:
UI comments:

* The "You have these permissions" thing feels overly technical. I think this information should show up somewhere, but maybe not top-and-center. Maybe remove it for now.
* I'd like to avoid "grain" in UI text because users might not know what that is. It would be nice to have another noun, though. Maybe each app could declare such a noun? "Document", "Album", etc. This would be part of ViewInfo, I suppose, so different UiViews could use different nouns.
* "Key label" -> "Label"? Really what we want is to ask "Who are you sharing this with?", but in a way that doesn't confuse the user into thinking that they need to be precise.
* "Your existing keys for sharing this grain:" -> "Links you've created:"; if there aren't any, "You have not shared this grain"
* "delete" -> "revoke"

ocdtrekkie:
As grains are supposed to be granular units of data, I would like to submit that each grain is a "file". In properly granular Sandstorm apps, a grain is relatively equivalent to a file on a desktop application. I'd kinda suggest avoiding having different apps refer to grains as different things.

kentonv:
@ocdtrekkie I've thought about using "file" but I worry that it is widely understood to mean a blob of bytes, not a thing that is potentially running and communicating with the world. Could confuse people?

ocdtrekkie:
@kentonv I'm not sure it's much different than your/Sandstorm's definition of a "driver", which many admins may confuse with their servers' hardware drivers. My notion is that a "file" in the cloud may potentially do more things than a "file" in other contexts. Obviously grains are pretty varied in what they can do in Sandstorm though, and it's unlikely a single existing term will cover them all perfectly, as a grain can be my five line Etherpad document or an entire website. My notion is you have to do the imperfect solution of using "file", which is at least relatively accurate for adequately granular Sandstorm apps (and quite a few ideal use cases), or sell people on "grain" as a completely newly defined word in the tech space.

dwrensha:
I'm declaring this ready for serious review now.

shell/lib/db.js
```diff
@@ -201,6 +188,11 @@ ApiTokens = new Mongo.Collection("apiTokens");
 //   userInfo:  For API tokens created by the app through HackSessionContext, the UserInfo struct
 //              that should be passed to `newSession()` when exercising this token, in decoded (JS
 //              object) format. This is a temporary hack.
+//   roleAssignment: If this API token represents a UiView, this field contains a JSON-encoded
+//              Grain.ViewSharingLink.RoleAssignment representing the received permissions.
```
kentonv: Add that this assignment strictly narrows permissions held by `userId`.

shell/server/permissions.js
```diff
@@ -0,0 +1,212 @@
...
+  // userId -> PermissionSet
+  var permissionsMap = {}
+  permissionsMap[owner] = roleAssignmentPermissions({allAccess: null}, viewInfo);
+
+  var userStack = [owner];
+  var user;
+
+  while (userStack.length > 0) {
+    var user = userStack.pop();
+    RoleAssignments.find({active: true, sharer: user,
```
kentonv: It seems like you'll end up querying all of the role assignments for the grain, so it's probably best to do that as one query upfront and then operate on the set in-memory.

shell/server/permissions.js
```diff
@@ -0,0 +1,212 @@
...
+  // information from the`mayOpenGrain()` computation to use here.
+
+  var grain = Grains.findOne(grainId);
+  var owner = grain.userId;
+
+  // userId -> PermissionSet
+  var permissionsMap = {}
+  permissionsMap[owner] = roleAssignmentPermissions({allAccess: null}, viewInfo);
+
+  var userStack = [owner];
```
kentonv: Instead of searching "forwards" from the owner (ultimately to all leaf nodes), what about searching "backwards" from the requesting user up to the owner? I think in most cases this would cut a large portion of the graph out of consideration.

shell/server/permissions.js
```diff
@@ -0,0 +1,212 @@
...
+      }
+    }
+  }
+  return false;
+}
+
+Meteor.startup(function () {
+  RoleAssignments.find().observe({
+    changed : function (newRoleAssignment, oldRoleAssignment) {
+      if (newRoleAssignment.active != oldRoleAssignment.active) {
```
kentonv: What if assigned role is modified but the assignment remains active? Or do you think that should be implemented by removing the old assignment and creating a new one? IMO we should be able to change them in-place.

Also we probably only want to reset the user's session if their overall permissions changed (and maybe only if their permissions changed negatively?) to avoid disruptions when someone shares to you a permission you already had.

shell/server/permissions.js
dwrensha: That's what I do for the `mayOpenGrain()` function below, but I think it only works for a single permission at a time.

shell/server/proxy.js
```diff
@@ -129,38 +130,68 @@ Meteor.methods({
...
-    // Start the grain if it is not running.
-    var runningGrain = runningGrains[grainId];
-    var grainInfo;
-    if (runningGrain) {
-      grainInfo = waitPromise(runningGrain);
-    } else {
-      grainInfo = continueGrain(grainId);
+    var grain = Grains.findOne(grainId);
+    if (!grain) {
+      throw new Meteor.Error(403, "Unauthorized", "No grain found.");
```
kentonv: Are we intentionally trying to hide from the user whether the grain exists? I'm not sure if that's really necessary, so I'd probably make this 404.

shell/server/proxy.js
```diff
@@ -129,38 +130,68 @@ Meteor.methods({
...
-    });
+    check(token, String);
+    var hashedToken = Crypto.createHash("sha256").update(token).digest("base64");
+    var apiToken = ApiTokens.findOne(hashedToken);
+    if (!apiToken) {
+      throw new Meteor.Error(403, "Unauthorized", "Invalid token.");
+    }
+    var grain = Grains.findOne({_id: apiToken.grainId});
+    if (!grain) {
+      throw new Meteor.error(403, "No such grain.");
```
kentonv: This case should definitely either be 404 or the user-visible part of the error should be "Unauthorized". I favor 404, as stated above.

shell/shared/shell.js
```diff
@@ -223,10 +224,19 @@ if (Meteor.isClient) {
...
+      var result = [];
       if (selectedApp) {
         return Grains.find({userId: userId, appId: selectedApp}, {sort: {lastUsed: -1}}).fetch();
       } else {
-        return Grains.find({userId: userId}, {sort: {lastUsed: -1}}).fetch();
+        var result = Grains.find({userId: userId}, {sort: {lastUsed: -1}}).fetch()
+        var uniqueGrains = {};
+        RoleAssignments.find({}, {sort:{created:1}}).forEach(function(roleAssignment) {
+          if (!(roleAssignment.grainId in uniqueGrains)) {
+            result.push({_id : roleAssignment.grainId, title: roleAssignment.title});
```
kentonv: I think we should keep the "shared with me" files in a separate list from the user's own files. So the sidebar should have "My Files" (renamed from "All Files") followed by "Shared With Me" showing RoleAssignments. (And as discussed previously (I think), the per-app tabs should only show the user's own files.)

shell/server/permissions.js
kentonv: Hmm, I don't see why that would be. I'm pretty sure the algorithm is the same whether you're calculating owner -> recipient vs. recipient -> owner. In the owner -> recipient case your computing the set of permissions that flow to each node from the owner, with the ultimate answer being the set that flows to the recipient. In the recipient -> owner case you're computing the set of permissions that flow to the recipient _from_ each node, with the ultimate answer being the set of permissions that flow from the owner.

As a heuristic, if you keep track of the set of permissions that the recipient has already been proven to possess, then you can stop following edges that don't add any more permissions beyond that. In most cases, once you find any path from the owner to the recipient, you'll be able to quickly eliminate all other edges from consideration.

A further heuristic: Whenever inserting a new RoleAssignment, store its "distance" from the owner -- that is, the minimum number of edges that must be traversed from the owner to get to the user who is making the assignment. Then when you are later computing permissions, you can prefer to search RoleAssignments with lower distances first, to more quickly fill out the permissions set allowing you to eliminate other paths from consideration. (This optimization probably isn't necessary for now, though.)

There's probably a lot of other heuristics we could think of along these lines, too...

(FWIW, I spent several months working on a slightly different version of this problem at Google. Sadly, that code never made it into a product, because politics.)

dwrensha:
I just pushed some new commits that address the comments so far.

kentonv:
I had one more thought: When you open the "share" dialog for an old "non-private" grain, it should simply say something like "This grain uses the old sharing model. You may share it by copy/pasting the URL from the location bar. You can upgrade this grain to the new model, but anyone with whom you shared this grain previously will lose access; you will have to share it with them again. This cannot be undone. [button: Upgrade to new model]"

Of course, for non-owners, only the first two sentences should display.

dwrensha:
Done.

dwrensha:
Our method of clearing the error whenever DevApps changed was broken. The intent was that if DevApps changed then we should retry opening a session. The problem is that if DevApps is nonempty, then `clearError()` will *always* get executed, because the `added` callback gets triggered even for existing entries in DevApps. I'm confused about how this ever seemed to do the right thing. For now I think it's best to delete this logic, as I do in https://github.com/dwrensha/sandstorm/commit/f5b22e4fcf4af9a76d6e7ff58999f5f8fe73f342.

This means that if you do `spk dev` on a an app that is not installed, it won't automatically clear out any "package not installed" errors.

I think it should be possible to regain the error-clearing behavior, but it should be a separate pull request.

dwrensha:
Sure enough, I just verified that on the current master branch it's possible to get into an infinite loop in certain situations if you do `spk dev` with two non-installed apps at the same time.

PR #329, opened 2015-04-10
Add sandcats updater

jparyani:


src/sandstorm/run-bundle.c++
```diff
@@ -1822,6 +1825,7 @@ private:
           ", \"allowDevAccounts\":", config.allowDevAccounts ? "true" : "false",
           ", \"isTesting\":", config.isTesting ? "true" : "false",
           ", \"wildcardHost\":\"", config.wildcardHost, "\"",
+          ", \"sandcatsHostname\":\"", config.sandcatsHostname, "\"",
```
kentonv: I guess you're relying on Javascript treating empty strings as falsy, but it seems safer to not include this field or explicitly set it null if sandcatsHostname isn't set...

shell/server/sandcats.js
```diff
@@ -0,0 +1,100 @@
...
+var SANDCATS_HOSTNAME = Meteor.settings && Meteor.settings.public &&
+                          Meteor.settings.public.sandcatsHostname;
+var SANDSTORM_ALTHOME = Meteor.settings && Meteor.settings.home;
+SANDCATS_VARDIR = (SANDSTORM_ALTHOME || "") + "/var/sandcats";
+
+var ROOT_URL = Url.parse(process.env.ROOT_URL);
+var HOSTNAME = ROOT_URL.hostname;
+var SANDCATS_NAME; // Look at `startup` below to see where this is set
+
+updateSandcats = function () {
```
kentonv: Seems like these functions don't need to be global?

jparyani:
Done

shell/server/sandcats.js
```diff
@@ -0,0 +1,100 @@
...
+var querystring = Npm.require("querystring");
+var https = Npm.require("https");
+var fs = Npm.require("fs");
+var dgram = Npm.require("dgram");
+var Url = Npm.require("url");
+
+var SANDCATS_HOSTNAME = Meteor.settings && Meteor.settings.public &&
+                          Meteor.settings.public.sandcatsHostname;
+var SANDSTORM_ALTHOME = Meteor.settings && Meteor.settings.home;
+var SANDCATS_VARDIR = (SANDSTORM_ALTHOME || "") + "/var/sandcats";
```
kentonv: Actually, doesn't proxy.js already expose these as globals? I don't think you need to copy them here.

shell/server/sandcats.js
```diff
@@ -0,0 +1,100 @@
...
+  });
+  req.write(post_data);
+  req.end();
+
+  req.on("error", function(err) {
+    console.error("Couldn't send update sandcats hostname", err);
+  });
+};
+
+var pingUdp = function () {
```
kentonv: should probably be `function pingUdp() {`. :)

(also for `updateSandcats`)

jparyani:
Done

PR #330, opened 2015-04-10
Add admin configurable dialogs for the splash page and signup

### LINKED ISSUE #213: Splash page link on self-hosted should be configurable
Currently when you visit a Sandstorm server and are logged out or otherwise don't have permission to install apps, you see the splash page with text saying that you need to join the Sandstorm mailing list for an invite.

For self-hosted instances, this is totally wrong. The server's admin is the one that can send invites. It somehow hadn't occurred to me before, but this text actually looks pretty bad because it implies that you need permission from us for self-hosted servers.

We should make this configurable. A simple config file option specifying HTML to inject should be good, with a default along the lines of: "Contact the server admin for an invite."

jparyani:
Fixes #213

### DIFF
--- a/shell/shared/shell.js
+++ b/shell/shared/shell.js
@@ -276,7 +276,12 @@ if (Meteor.isClient) {
       }
     },
 
-    dateString: function (date) { return makeDateString(date); }
+    dateString: function (date) { return makeDateString(date); },
+
+    splashDialog: function() {
+      var setting = Settings.findOne("splashDialog");
+      return (setting && setting.value) || DEFAULT_SPLASH_DIALOG;
+    }
   });
 
   Template.root.events({
--- a/shell/shared/signup.js
+++ b/shell/shared/signup.js
@@ -161,6 +161,13 @@ if (Meteor.isClient) {
       Session.set("inviteMessage", undefined);
     },
   });
+
+  Template.signup.helpers({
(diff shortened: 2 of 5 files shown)


shell/shared/shell.js
```diff
@@ -276,7 +276,12 @@ if (Meteor.isClient) {
       }
     },
 
-    dateString: function (date) { return makeDateString(date); }
+    dateString: function (date) { return makeDateString(date); },
+
+    splashDialog: function() {
+      var setting = Settings.findOne("splashDialog");
+      return (setting && setting.value) || "Contact the server admin for an invite.";
```
kentonv: Can you share a string constant between here and admin.js?

shell/shared/signup.js
```diff
@@ -161,6 +161,13 @@ if (Meteor.isClient) {
       Session.set("inviteMessage", undefined);
     },
   });
+
+  Template.signup.helpers({
+    signupDialog: function () {
+      var setting = Settings.findOne("signupDialog");
+      return (setting && setting.value) || "You've been invited to join this Sandstorm server!";
```
kentonv: Also here.

kentonv:
Oops, merge conflict since I just pulled the email token login change.

jparyani:
Done.

PR #332, opened 2015-04-15
Begin implementing a separate 'back-end' process responsible for managin...

kentonv:
...g non-mongo storage and app execution.

The front-end talks to the back-end process via RPC.

So far this change implements starting and deleting grains via the API. This has the pleasant side effect of simplifying the means by which a run-dev.sh frontend starts grains.

Future changes will make backup/restore and package installation go through the API as well.


kentonv:
LEEROOOOOOYYYYYYYYYYYYYYYYYYYYYYYYY

PR #333, opened 2015-04-15
Update selenium version and instructions on how to start it. Line wrap the README.

dwrensha:
The Selenium version we're pinned at is broken for newer versions of Firefox: https://code.google.com/p/selenium/issues/detail?id=8399

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -2,7 +2,8 @@
 
 ## Setup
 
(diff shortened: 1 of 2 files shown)


dwrensha:
@jparyani: note that the new `selenium-standalone start` command means that you may need to update the init scripts on the Jenkins servers.

jparyani:
Will do

PR #335, opened 2015-04-16
Clear mock user data before running test suite. Use 'upsert' for creating mock users.

dwrensha:
This fixes the problem where running the tests multiple times on the same database would fail.


kentonv:
@jparyani can you review this?

jparyani:
Looks good to me

PR #336, opened 2015-04-16
Set isTesting to true in dev mode.

dwrensha:


paulproteus:
I just read through `shell/shared/testing.js` and I'm not sure I understand what this is really for yet. I grok the implementation; it's the purpose that's not super clear to me. I'm fine with this particular pull request, but I figured I'd remark on my uncertainty somewhere.

dwrensha:
Yeah, now that we have dev accounts, the mock users defined in testing.js seem somewhat redundant. There are, however, a bunch a places where we branch on the login service, e.g. [here](https://github.com/sandstorm-io/sandstorm/blob/dec99284770ddea214477289c4cc79bf6c0796a0/shell/server/proxy.js#L695). If we used dev accounts instead of these mock users, we would lose coverage of such code.

dwrensha:
I think that in the future we should have a generic way to get an ID, display name, email, etc. from a user, and then we should be able to eliminate testing.js and the isTesting option.

PR #338, opened 2015-04-20
For appDemo feature, use appTitle if it exists

paulproteus:
This addresses an issue where right now, if you want to Let's Chat, you
end up being offered to "Try Let's >>" because the word "Chat" is eaten
by the appNameFromActionName function.

### DIFF
--- a/shell/shared/demo.js
+++ b/shell/shared/demo.js
@@ -264,8 +264,9 @@ Router.map(function () {
       var appName = 'missing package';
 
       if (thisPackage) {
-        var actionTitle = thisPackage.manifest.actions[0].title.defaultText;
-        appName = appNameFromActionName(actionTitle);
+        var manifest = thisPackage.manifest;
+        var action = manifest.actions[0];
+        appName = (manifest.appTitle && manifest.appTitle.defaultText) || appNameFromActionName(action.title.defaultText);
       }
 
       return {


paulproteus:
Please don't merge yet; still testing this.

paulproteus:
OK, I've tested this and I'm now requesting review+merge.

Note that until this is fixed, I can't ask the nice Let's Chat people to use an appdemo link!

shell/shared/demo.js
```diff
@@ -264,8 +264,9 @@ Router.map(function () {
       var appName = 'missing package';
 
       if (thisPackage) {
-        var actionTitle = thisPackage.manifest.actions[0].title.defaultText;
-        appName = appNameFromActionName(actionTitle);
+        var manifest = thisPackage.manifest;
+        var action = manifest.actions[0];
+        appName = (manifest.appTitle && manifest.appTitle.defaultText) || appNameFromActionName(action.title.defaultText);
```
kentonv: I'll fix this, but note that the style guide calls for lines to be wrapped at 100 chars.

https://github.com/sandstorm-io/capnproto/blob/master/style-guide.md#spacing-and-bracing

paulproteus: Thanks @kentonv for the merge and the feedback. I'll look into http://www.emacswiki.org/emacs/ColumnMarker to avoid the >100-char lines.

Issue #340, opened 2015-04-21
Grain isn't restarted when `spk dev` starts/stops


dwrensha:
Looks like https://github.com/sandstorm-io/sandstorm/issues/269 has resurfaced due to the changes in https://github.com/sandstorm-io/sandstorm/pull/332.

PR #341, opened 2015-04-21
Shut down dev app grains from the frontend.

dwrensha:
Fixes https://github.com/sandstorm-io/sandstorm/issues/340.
Bonus: now `spk dev` will only restart grains whose appId matches the app being developed.


dwrensha:
Note that we need the `delete runningGrains[grainId];` line because otherwise we get "Capability has been closed" errors. An alternative might be to get rid of the `grain.close()` lines, but that seems uglier.

PR #344, opened 2015-04-22
Add token for admin settings page

jparyani:
This will automatically generate an admin token in the installer and direct users there to configure a login system. Also, added a command `admin-token` to the sandstorm binary to generate said token.

### DIFF
--- a/src/sandstorm/run-bundle.c++
+++ b/src/sandstorm/run-bundle.c++
@@ -652,6 +652,17 @@ public:
                   .build();
             },
             "For internal use only.")
+        .addSubCommand("admin-token",
+            [this]() {
+              return kj::MainBuilder(context, VERSION,
+                      "Generates a new admin token that you can use to access the admin settings "
+                      "page. This is meant for initial setup, or if an admin account is locked out.")
+                  .addOption({"short"}, [this]() { shortOutput = true; return true; },
+                      "Pass this flag if you wish to output only the token.")
+                  .callAfterParsing(KJ_BIND_METHOD(*this, adminToken))
+                  .build();
+            },
+            "Resets OAuth configuration.")
         .build();
   }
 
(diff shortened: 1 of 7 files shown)


src/sandstorm/util.c++
```diff
@@ -601,6 +601,18 @@ kj::Array<byte> base64Decode(kj::StringPtr input) {
 
 // =======================================================================================
 
+kj::String bytesToHex(kj::ArrayPtr<const byte> input) {
```
kentonv: We actually already have hexEncode() in sandstorm-http-bridge.c++. Perhaps move that out to util.{h,c++}. (It's shorter.)

src/sandstorm/run-bundle.c++
```diff
@@ -992,6 +1003,30 @@ public:
     context.exitInfo(kj::str("reset OAuth configuration"));
   }
 
+  kj::MainBuilder::Validity adminToken() {
+    changeToInstallDir();
+
+    // Get 20 random bytes for token.
+    kj::byte bytes[20];
+    kj::FdInputStream random(raiiOpen("/dev/urandom", O_RDONLY));
```
kentonv: Use libsodium randombytes_buf(). (We should be switching over to that everywhere.)

src/sandstorm/run-bundle.c++
```diff
@@ -652,6 +652,17 @@ public:
                   .build();
             },
             "For internal use only.")
+        .addSubCommand("admin-token",
+            [this]() {
+              return kj::MainBuilder(context, VERSION,
+                      "Generates a new admin token that you can use to access the admin settings "
+                      "page. This is meant for initial setup, or if an admin account is locked out.")
+                  .addOption({"short"}, [this]() { shortOutput = true; return true; },
+                      "Pass this flag if you wish to output only the token.")
```
kentonv: style nitpick: "Pass this flag if you wish to" seems redundant. Just "Output only the token." seems sufficient.

src/sandstorm/run-bundle.c++
```diff
@@ -652,6 +652,17 @@ public:
                   .build();
             },
             "For internal use only.")
+        .addSubCommand("admin-token",
+            [this]() {
+              return kj::MainBuilder(context, VERSION,
+                      "Generates a new admin token that you can use to access the admin settings "
+                      "page. This is meant for initial setup, or if an admin account is locked out.")
+                  .addOption({"short"}, [this]() { shortOutput = true; return true; },
```
kentonv: Typically this would be called "quiet", with 'q' as a short option: {'q', "quiet"}

kentonv:
The install script just got rewritten so you'll have to update for that. Sorry.

jparyani:
Done

PR #346, opened 2015-04-23
Use git tags to record commit <=> version info

paulproteus:
With this commit, we will start publishing git tags, which would have made my life a little easier earlier today.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -58,15 +58,6 @@ echo "**** Building build $BUILD ****"
 
 make BUILD=$BUILD
 
-echo "**** Pushing build $BUILD ****"
-
-echo $BUILD > tmp/$CHANNEL
-gcutil push fe $TARBALL /var/www/dl.sandstorm.io
-gcutil push fe tmp/$CHANNEL /var/www/install.sandstorm.io
-
-gcutil ssh smalldemo sudo service sandstorm update
-gcutil ssh alpha sudo service sandstorm update dev
-
 echo "**** Tagging this commit ****"
 
 # The git tag stores the version number as a normal-looking version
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -62,3 +66,16 @@ gcutil push fe tmp/$CHANNEL /var/www/install.sandstorm.io
...
+
+echo "**** Tagging this commit ****"
+
+# The git tag stores the version number as a normal-looking version
+# number, like 0.75 for build 75 within branch 0, or 2.121 for build
+# 121 within branch 2.
+
+BUILD_MINOR="$(( $BUILD % 1000 ))"
+DISPLAY_VERSION="${BRANCH_NUMBER}.${BUILD_MINOR}"
+TAG_NAME="sandstorm-${DISPLAY_VERSION}"
```
kentonv: "sandstorm" seems redundant since the tag is already in the Sandstorm repository. Maybe just "v$DISPLAY_VERSION"?

release.sh
```diff
@@ -62,3 +66,16 @@ gcutil push fe tmp/$CHANNEL /var/www/install.sandstorm.io
 
 gcutil ssh smalldemo sudo service sandstorm update
 gcutil ssh alpha sudo service sandstorm update dev
+
+echo "**** Tagging this commit ****"
```
kentonv: Nitpick: IMO this section should execute before "Pushing build", so that a random problem with production machines doesn't mean we lose the tag.

release.sh
paulproteus: +0

release.sh
paulproteus: +1

paulproteus:
OK, please re-review! This addresses the feedback above, I believe.

paulproteus:
@kentonv merge this, and then I will push all my other git tags that I made locally!!!

PR #354, opened 2015-04-23
Move oauth config to admin settings page

jparyani:


jparyani:
Merge conflicts have been fixed. Please re-review

shell/client/shell.html
```diff
@@ -547,15 +547,61 @@
...
+              <td>
+                <input id="configure-login-service-dialog-{{property}}" type="text" value="{{value}}">
+              </td>
+            </tr>
+          {{/each}}
+        </table>
+      </p>
+      <div class="new-section">
+        <div class="login-button configure-login-service-dismiss-button">
+          I'll do this later
```
kentonv: Perhaps change this to "Cancel"?

shell/shared/admin.js
```diff
@@ -214,10 +313,22 @@ if (Meteor.isServer) {
     }
   };
 
+  var updateLoginStyleToRedirect = function (serviceName) {
+    var configurations = Package["service-configuration"].ServiceConfiguration.configurations;
+    var config = configurations.findOne({service: serviceName});
+
+    if (config) {
```
kentonv: Maybe also check if loginStyle is already "redirect", in which case you can skip setting it?

kentonv:
It looks like if I enable a login service but fail to configure it, then any random user can subsequently configure it by calling the `configureLoginService` method.

Perhaps we should either:
a. Don't let the admin enable a service that doesn't have a configuration.
b. Auto-add an empty configuration if none already exists.
c. Somehow disable the `configureLoginService` method from `accounts-base`.

If (c) is possible without being ugly, it would be nice, but otherwise I think (a) is probably best.

A snag: what now happens if `BASE_URL` changes and we clear the OAuth config? I guess we have to disable the no-longer-configured services as well? But how does the admin then figure out what to do, when they browse to their site and find they can't log in, and there is no "configure login" button to click on to get more info? (I think if we can managed to use solution (c), this isn't a problem, since then they'd see the "configure login" buttons which take them to the admin page.)

jparyani:
Done.

jparyani:
Sorry, I just realized in #376, I said the error.details/reason change was going to get fixed in this PR, but I forgot to mention that it's actually working as intended. Do you want me to proceed with the change in this PR?

PR #356, opened 2015-04-24
Release install.sh along with the rest of Sandstorm

paulproteus:
This avoids a condition where we do a release but forget to update
the installer.

@ocdtrekkie just ran into this -- the install.sh fix from last week seemingly didn't get pushed to `https://install.sandstorm.io`.

@kentonv if you prefer to handle this a different way, I'm all ears, but it seems to me that we'll be happiest having exactly one release process, rather than one for `install.sh` and another for the rest of Sandstorm.


kentonv:
Ostensibly the reason why they are not released together is because I usually push installer updates immediately rather than wait for the next release. But I guess doing it in the release script too is a nice backup for when I fail like this.

PR #357, opened 2015-04-24
Enable DNS configuration (sandcats) by default

paulproteus:
This change adjusts the install script so that if you choose that you
want Sandstorm to listen on external network interfaces, then we ask
you about configuring sandcats.io for dynamic DNS.

This tiny change is desireable, rather than rewriting the installer
immediately, because @ocdtrekkie is giving a talk about Sandstorm
on Sat Apr 25 at the Chicago Linux Users Group, and I would prefer
to ensure that his talk goes smoothly than to land a big install.sh
rewrite right now.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -75,6 +75,9 @@ while getopts ":cdeu" opt; do
       ;;
     e)
       USE_EXTERNAL_INTERFACE="yes"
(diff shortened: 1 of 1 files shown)


paulproteus:
I tested this and it doesn't actually work. Re-working pull request.

PR #359, opened 2015-04-24
Rewrite install.sh to clarify server vs. dev mode

paulproteus:
Server mode emphasizes (but does not require) integration with the
Sandcats.io dynamic DNS service.

Dev mode emphasizes dev accounts and requires root by default, for
spk dev to work.

MAIL_URL configuration is gone.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -122,6 +122,7 @@ USE_LOCAL_INSTALL_SH_FILE="no"
 
 # Defaults for some config options, so that if the user requests no
 # prompting, they get these values.
+DEFAULT_PORT=6080
 DEFAULT_DIR_FOR_ROOT="/opt/sandstorm"
 DEFAULT_DIR_FOR_NON_ROOT="$HOME/sandstorm"
 DEFAULT_UPDATE_CHANNEL="dev"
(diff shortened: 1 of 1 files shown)


paulproteus:
Testing done: none.

(kidding!)

paulproteus:
Testing done:

* Install with "./install.sh -l" and ensure that it uses the local install.sh file.

* Install with ./install.sh" without root, choose option 1, and happily find myself asked to switch to root and end up with working Sandcats enablement. (Note: I actually did: `OVERRIDE_SANDCATS_BASE_DOMAIN=sandcats-dev.sandstorm.io OVERRIDE_SANDCATS_API_BASE=https://sandcats-dev-machine.sandstorm.io OVERRIDE_SANDCATS_CURL_PARAMS=-k bash ./install.sh` to use the Sandcats dev server instead.)

* Install as root with -d and happily be asked no questions and end up with a working Sandstorm.

* Attempt to install not as root with -d and be told that it can't work.

* Choose option `2` for dev mode, try saying no to getting root, be told it can't work.

* Choose option `2` for dev mode, say yes to getting root, have a working server. (Note that this one still asks a few questions. Fixing this now to ask fewer questions.)

paulproteus:
@kentonv (once the tests pass) this is ready for review+merge.

install.sh
```diff
@@ -296,176 +293,309 @@ int main() {
...
+  # as root, at the moment.
+
+  if [ "yes" = "$PREFER_ROOT" ] && [ "no" = "$CURRENTLY_UID_ZERO" ] ; then
+    # We are not root, but we would like to be root.
+    echo "If you want app developer mode for a Sandstorm install, you need root"
+    echo "due to limitations in the Linux kernel."
+    echo ""
+
+    echo "To set up Sandstorm, we will need to use sudo."
+    echo "Rest assured that Sandstorm itself won't run as root."
```
kentonv: This is debatable, in this case. The sandstorm bundle runner actually does run as root, but runs Mongo and Node as non-root.

install.sh
```diff
@@ -296,176 +293,309 @@ int main() {
...
+    fi
   fi
-  if [ "${UPDATE_CHANNEL:-none}" != none ]; then
-    CHANNEL=$UPDATE_CHANNEL
+
+  # If they did not pass -d, then let them opt into that.
+  if [ "yes" != "$USE_DEFAULTS" ] ; then
+    echo "We're going to:"
+    echo ""
+    echo "* Install Sandstorm in ${DEFAULT_DIR_FOR_ROOT}."
```
kentonv: If the script executed with `-u`, then you need DEFAULT_DIR_FOR_NON_ROOT here. (Also for the live server path.)

paulproteus: It never occurred to me that people would run the script with `-u` and not also `-d`. But sure, that makes sense.

I think I'll just skip the nice helpful summaries if you go with `-u`. At that point, I'm going to presume you know what you're doing.

paulproteus: I pushed code changes along these lines.

install.sh
```diff
@@ -780,10 +1171,87 @@ function register_sandcats_name() {
...
+save_config
+download_latest_bundle_and_extract_if_needed
+extract_bundle_if_provided
+make_runtime_directories
+set_permissions
+install_sandstorm_symlinks
+ask_about_starting_at_boot
+configure_start_at_boot_if_desired
+print_success
+}
```
phildini: why is there a second closing brace here?

phildini: to put another way, this brace may be correct, but in code this long, I'm a big fan of labelling (with comments, maybe) what opening brace a closing brace is referring to.

paulproteus: Right-o. This close-brace matches the `_` function definition at the top of the file.

PR #369, opened 2015-04-28
Resize Share Icon

ocdtrekkie:
I realize the likely impending futility of this with a shell redesign coming. However, as a note most icons on the toolbar have about a 15% margin on them. The Share icon was both very small with a lot of margin and fairly heavily off center.


garply:

paulproteus:
Hey @jparyani -- when I follow this link, it tells me:

```
Error: remote exception: remote exception: remote exception: connect(): Connection refused
C++ location:(remote):??
type: disconnected
```

I was hoping for something a little more wiki-like : D

paulproteus:
Garply, ok to test

paulproteus:
Dear @ocdtrekkie ,

You rule. Thanks for submitting this! I'll let the build sanity-check it before reviewing it (although someone else might end up beating me to the review but ha-ha! they won't beat me to complimenting you).

jparyani:
@paulproteus Hmm, that's weird, it looks like a weird issue with nginx. The grain log had a bunch of lines like:

nginx: [emerg] bind() to 0.0.0.0:33411 failed (98: Address already in use)

kentonv:
FWIW, I think the "share" button should be more prominent than the others, probably by giving it an actual text label "share". We want people to know that it's there.

I also think the icon should be a [Granovetter Diagram](https://www.google.com/search?q=granovetter+diagram). (Actually, [Google's preferred share icon](https://www.google.com/search?q=google%20share%20icon) is almost a Granovetter Diagram already, so this won't even be confusing.)

ocdtrekkie:
I wasn't going to propose major UI changes before the shell redesign, as far as text buttons. All of the existing options are icons, so I just fixed the one we had.

Ideally, noting how much empty space is in that black bar on my 1080p screen, I would kinda suggest that on desktops you have an icon and text combined, on mobile width, it condenses to just the icon. I would say this should apply for heavily used features (or features that should be heavily used) like Delete, Get Webkey, Share, and Download. And then for features that are more of a troubleshooting nature that (in a perfect world) aren't needed heavily, like Restart and Debug Log, those should go in a dropdown menu at the end of the list of top-level menu items.

I do think the rough share icon that Google uses seems pretty close to standard these days, whereas this icon could possibly be misconstrued as a pop-out or fullscreen command.

ocdtrekkie:
Worthwhile to tag @neynah here?

kentonv:
Yeah I'm not asking you to make the change, just commenting that we should do it.

I think that we should probably shove all the icons other than share under a menu of some sort (possibly looking a lot like what you get in the mobile view today). The share button should remain prominent, though -- it's more important than any of the other icons.

ocdtrekkie:
bar

I made an awful mock-up. Which icons are in a menu and which aren't and what order they're in all not really all that important, just showing the format of navigation that strikes me. I suppose I could see everything but share in the dropdown menu though.

Also, with the login/profile thing being on the right, and presumably the possibility of a few platform-wide features being in that bar in the future, here's a crazy thought: Should the Sandstorm button be on the far right? Like, could the left side of the bar be app (grain) details and options, and the right side be platform (Sandstorm) details and options? Logo in the far right is perhaps weird, but if it functions as a "return to home" button, it shares commonality with the function of the "X"/Close button in other platforms and applications, which is indeed often found on the top right.

ocdtrekkie:
Arrow out of box and the Granovetter diagram share icons definitely seem to be fierce competitors on the Noun Project: https://thenounproject.com/search/?q=share

neynah:
@ocdtrekkie A logo (as a home icon) on the top left is more common place and perhaps intuitive but I see what you mean about the "X"/Close button and that's an interesting point. We'll explore our options and see what works best. I also think the Granovetter diagram share icons are a good fit.

ocdtrekkie:
@neynah As I said, just a crazy thought. ;)

PR #373, opened 2015-05-01
Switch to a more granovetter-diagram-like share icon and make the share button more prominent.

dwrensha:
This involved telling ImageMagick to not fill in a background color for the generated pngs.

The share button now has an icon and the word "Share".


ocdtrekkie:
Eyeballing this I am guessing it'll need to be resized.

PR #374, opened 2015-05-01
Resize share.svg

ocdtrekkie:
This icon is "squared off" with a slight horizontal stretch, as well as scaled up from it's old position to the rough padding seen on other icons of about 15%.

I am not really fond of the stretching, but it's relatively imperceptible at the size the icon appears, and since it's an SVG, it stretches well. The overall look with similarly sized icons is worth it.

There are some other issues I noticed with the current build and the current code of the Share Icon, @dwrensha. Namely, that on mobile on the Alpha server (a.k.a., code prior to your latest PR), the Share icon is the wrong color. On mobile the background of the menu is white and the icons are black, which is relatively inverse to the desktop experience. However, the Share icon is forced to appear that color.

As this behavior is prior to your current PR, I am not sure what explains this. However, I noticed your PR to change the Share icon specifies a lot of CSS details for color and padding and such that the other icons do not use. It seems weird that you have that code, and you probably shouldn't for reasons such as the above.


garply:

dwrensha:
> on mobile on the Alpha server (a.k.a., code prior to your latest PR), the Share icon is the wrong color.

This was because I had overlooked the need to override the appearance inside the media query [here](https://github.com/sandstorm-io/sandstorm/blob/1c5ec4cfac64f704b5cdeac339cddbfeb450d15b/shell/client/shell.css#L1074). There should have been a statement like this:
```
#topbar #show-share-grain { background-image: url("/share-m.png"); }
```


> I noticed your PR to change the Share icon specifies a lot of CSS details for color and padding and such that the other icons do not use.

The new share button is a blue with rounded corners and includes both an icon and the text "Share". It needs extra CSS to specify all of that and to get the alignments reasonable. That said, I wouldn't be surprised if there's a more elegant way than what I came up with.

PR #375, opened 2015-05-01
spk pack: require nonempty appTitle and appMarketingVersion

dwrensha:


src/sandstorm/spk.c++
```diff
@@ -944,6 +948,15 @@ private:
   kj::MainBuilder::Validity doPack() {
     ensurePackageDefParsed();
 
+    if (!packageDef.getManifest().getAppTitle().hasDefaultText()) {
+      context.exitError(kj::str("Expected manifest.appTitle.defaultText to be nonempty."));
```
kentonv: This error message feels like it may take more cognitive power than necessary to understand. Maybe a better one would be:

    Your package manifest is missing `appTitle`. In `sandstorm-pkgdef.capnp`, under `pkgdef.manifest`, add something like:
        appTitle = (defaultText = "My App"),

kentonv: The `kj::str()` is redundant here.

dwrensha:
Updated. I moved the validation into `setPackageDef()`, where it can make use of the value of the `-p` flag if it was set. This additionally performs the validation for `spk dev`.

dwrensha:
(Note that the error message now automatically includes the name of the pkgdef file.)

kentonv:
Nice!

PR #376, opened 2015-05-01
Disallow users from disabling their login service

### LINKED ISSUE #352: Don't allow admin to disable the login mechanism used by their own account.
Otherwise this could brick your server.

jparyani:
This prevents your from bricking your Sandstorm install by having no
admin users that are able to login.

Fixes #352

### DIFF
--- a/shell/shared/admin.js
+++ b/shell/shared/admin.js
@@ -226,6 +226,9 @@ if (Meteor.isServer) {
         throw new Meteor.Error(403, "Unauthorized", "User must be admin or provide a valid token");
       }
 
+      // TODO(someday): currently this relies on the fact that an account is tied to a single
+      // identity, and thus has only that entry in "services". This will need to be looked at when
+      // multiple login methods/identities are allowed for a single account.
       if (!value && !tokenIsValid(token) && (serviceName in Meteor.user().services)) {
         throw new Meteor.Error(403, "Unauthorized",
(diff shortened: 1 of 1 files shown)


shell/shared/admin.js
```diff
@@ -226,6 +226,14 @@ if (Meteor.isServer) {
         throw new Meteor.Error(403, "Unauthorized", "User must be admin or provide a valid token");
       }
 
+      // TODO(someday): currently this relies on the fact that an account is tied to a single
+      // identity, and thus has only that entry in "services". This will need to be looked at when
+      // multiple login methods/identities are allowed for a single account.
+      if (!value && !tokenIsValid(token) && (serviceName in Meteor.user().services)) {
+        throw new Meteor.Error(403, "Unauthorized",
+          "You can not disable the login service that your account uses.");
```
kentonv: The third argument to Meteor.Error() is hidden from the user in production mode, so users will only see "Unauthorized" here, which will be confusing. We actually make this mistake in a number of places. At one point I thought that the second argument was supposed to be an HTTP error string, but it's really supposed to be an arbitrary human-readable error description. Anyway, in this case just replace the second argument with the third and omit a third argument.

jparyani:
The fix to error messages is blocked on #354, because I'd have to change them all at once and then modify the display code to use `error.reason` instead of `error.details`.

PR #379, opened 2015-05-01
Fix typo in install.sh when `sysctl -w` fails

jparyani:


install.sh
```diff
@@ -269,7 +269,7 @@ assert_userns_clone() {
 kernel.unprivileged_userns_clone = 1
 __EOF__
         sysctl -w kernel.unprivileged_userns_clone=1 || fail "'sysctl -w" \
-          "kernel.unprivileged_userns_clone=1' failed. If you are inside docker, please run the"
+          "kernel.unprivileged_userns_clone=1' failed. If you are inside docker, please run the" \
```
paulproteus: @jparyani massive props to you for finding this and fixing it!

PR #387, opened 2015-05-06
Integrate Sandcats domain recovery into installer

paulproteus:
@kentonv This is now ready for review.

Tested manually against the sandcats-dev.sandstorm.io domain via:

`sudo OVERRIDE_SANDCATS_BASE_DOMAIN=sandcats-dev.sandstorm.io OVERRIDE_SANDCATS_API_BASE=https://sandcats-dev-machine.sandstorm.io OVERRIDE_SANDCATS_CURL_PARAMS=-k bash ./install.sh`

Sandcats prod as well as dev have the code needed to use this feature.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -63,6 +63,29 @@ fail() {
   exit 1
 }
 
+dotdotdot_curl() {
+  # This function calls curl, but first prints "..." to the screen, in
+  # an attempt to indicate to the user that the script is waiting on
+  # something.
+  #
+  # It then moves the cursor to the start of the line, so that future
+  # echo-ing will overwrite those dots.
+  #
+  # Since the script is -e, and in general we don't have a reliable
+  # thing that we do in the case that curl exits with a non-zero
+  # status code, we don't capture the status code; we allow the script
+  # to abort if curl exits with a non-zero status.
+
+  # Functions calling dotdotdot_curl expect to capture curl's own
+  # stdout. Therefore we do our echo-ing to stderr.
+
+  echo -n '...' >&2
+
+  curl "$@"
+
+  echo -ne '\r' >&2
+}
+
 # writeConfig takes a list of shell variable names and saves them, and
 # their contents, to stdout. Therefore, the caller should redirect its
 # output to a config file.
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -1137,6 +1284,12 @@ sandcats_register_name() {
     return
   fi
 
+  # If the wants help, offer help.
```
paulproteus: s/the wants/the user wants/

install.sh
```diff
@@ -1122,7 +1269,7 @@ sandcats_register_name() {
   SANDCATS_CURL_PARAMS="${OVERRIDE_SANDCATS_CURL_PARAMS:-}"
 
   echo "Choose your desired Sandcats subdomain (alphanumeric, max 20 characters)."
-  echo "Type the word none to skip this step."
+  echo "Type the word none to skip this step, or help for help."
```
paulproteus: FWIW I played with the text here for a while, unsure exactly how to phrase this. I am reasonably happy with this, having tried a number of options.

install.sh
```diff
@@ -1114,6 +1114,153 @@ print_success() {
...
+
+  # Print something indicating we're working
+  echo -n '...'
+
+  # Now we can do a call to /update, which we will do silently on the
+  # user's behalf. This uses the new key we registered (via /recover)
+  # and sets the IP address for this host in the sandcats.io DNS
+  # service.
+  LOG_PATH="var/sandcats/update-log"
+  HTTP_STATUS=$(
```
paulproteus: It would be vaguely nice to move this stuff to a shell function, but I'm not totally sure at the moment how to do that cleanly, so I think it's better to leave the copy-pasta in.

paulproteus:
I think the TODO before merging this is:

* Review the comments carefully.

* Make sure that before every call to `curl` we print `...` so that the user knows something is going on.

* Deploy the current Sandcats code into production, so that when merged, this flow will actually work.

Remarks to self about deploying the current Sandcats code:

* Adjust the `sandcats.service` to provide a MAIL_URL to enable outbound SMTP.

* Adjust the `/etc/sandcats-meteor-settings.json` file to include an `EMAIL_FROM_ADDRESS`. It's approximately safe to add this to the conf file while the service is running, but note that the next service start has to be the new version, since the codebase aggressively checks which settings are present.

* Make sure firewall rules etc. are happy with outbound SMTP from the Sandcats server. (Arguably I should add a Nagios check for this, or something, but that machine doesn't have Nagios at the moment.)

* Use the script to do a domain recovery of one of my own domains, to make absolutely sure that this really works.

* Re-submit.

Other notes from testing this:

* In manual testing on the Sandcats dev service, outbound mail would sometimes time out, resulting in `curl` in the `install.sh` script aborting because it has a 20 second timeout without printing any information. I think/hope this is not going to be very likely in production, but probably we should instrument the Sandcats backend so that any request that takes longer than (e.g.) 2 seconds gets logged somewhere. Perhaps the place to do that is the nginx configuration. References on that: http://wiki.nginx.org/HttpLogRequestSpeed https://lincolnloop.com/blog/tracking-application-response-time-nginx/

paulproteus:
Other Sandcats issue discovered during testing:

* https://github.com/sandstorm-io/sandcats/issues/57

Probably a show-stopper before attempting to merge this.

paulproteus:
Issue found during testing:

* The Sandstorm `install.sh` doesn't allow a hyphen, even though the backend does.

paulproteus:
Issues found in testing:

* The end of the flow, where we call `update`, prints all sorts of weird garbage. That should be cleaned up.

install.sh
kentonv: Maybe "none" and "help" should be quoted?

kentonv:
I think most users who are trying to recover a domain will start by typing in that name again rather than going to "help", so I think the error message when the domain is already token should say "If this is your name and you lost the key, type 'help' to recover it."

install.sh
paulproteus: I like not quoting this, fwiw, since otherwise people might type literally `"help"`.

install.sh
```diff
@@ -1137,9 +1302,15 @@ sandcats_register_name() {
...
-  if ! [[ $DESIRED_SANDCATS_NAME =~ ^[0-9a-zA-Z]{1,20}$ ]] ; then
+  # If the user wants help, offer help.
+  if [ "help" = "$DESIRED_SANDCATS_NAME" ] ; then
+    sandcats_provide_help
+    return
+  fi
+
+  # Validate the client-side, to avoid problems, against a slightly
+  # less rigorous regex than the server is using.
+  if ! [[ $DESIRED_SANDCATS_NAME =~ ^[0-9a-zA-Z-]{1,20}$ ]] ; then
```
paulproteus: Just a note for reviewers: the change here is that I updated this code to allow hyphens anywhere; the server checks our hyphen use against a more rigorous regexp.

paulproteus:
@kentonv please merge this if you're OK with it!

PR #390, opened 2015-05-08
Set BIND_IP to 0.0.0.0 for Vagrant, so port forwarding works

zarvox:
At some point, we changed the default BIND_IP to 127.0.0.1,
but with Vagrant, this is a no-go because we need the guest to
listen to port-forwarded requests, which will come in on eth0,
not lo.

Tested with @phildini at the meetup.


phildini:
I can certify this worked on my local machine, so :ship: :stuck_out_tongue_winking_eye:

paulproteus:
I'm thrilled to merge this. Thanks for testing, @phildini .

PR #394, opened 2015-05-08
Compute transitive shares.

dwrensha:
This adds a `transitiveShares()` function to permissions.js and uses it in two places:
 1. To compute a closer approximation to which sessions need to be killed when permissions change.
 2. To implement a new "show transitive shares" button in the share menu. Looks like this:

transitive-shares

### DIFF
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -333,6 +333,27 @@ limitations under the License.
             </li>
           {{/each}}
           </ul>
+          {{#if transitiveShares}}
+            <button class="hide-transitive-shares">hide transitive shares</button>
+            {{#if transitiveShares.empty}}
+              <p> You currently grant access to nobody. </p>
+            {{else}}
+              <ul>
+                {{#each transitiveShares}}
+                  <li>
+                    <div>{{displayName recipient}} was granted access by:</div>
+                    <ul>
+                      {{#each sharers}}
+                        <li>{{displayName sharer}} ({{dateString created}})</li>
+                      {{/each}}
+                    </ul>
+                  </li>
+                {{/each}}
+              </ul>
+            {{/if}}
+          {{else}}
+            <button class="show-transitive-shares">show transitive shares</button>
+          {{/if}}
         {{/if}}
       {{/if}}
     {{/if}}
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -334,9 +334,10 @@ limitations under the License.
           {{/each}}
           </ul>
           {{#if transitiveShares}}
-            <button class="hide-transitive-shares">hide transitive shares</button>
+            <button class="hide-transitive-shares">hide reshares</button>
             {{#if transitiveShares.empty}}
-              <p> You currently grant access to nobody. </p>
+              <p></p>
+              <p> There are no reshares. </p>
             {{else}}
               <ul>
                 {{#each transitiveShares}}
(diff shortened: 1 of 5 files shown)


ocdtrekkie:
Question: Is only allowing you to revoke your direct shares a "by design" notion of delegation? Or just a not implemented feature?

There's a valid use case that someone may want to revoke a transitive share without having to revoke the original person you shared to. For instance, this example, maybe I want to remove those Unknown users from here, because Bob wasn't supposed to take it out of the company. Maybe I even want to disable Bob for having shared outside the company, but Alice didn't do anything wrong sharing to Bob. But with this chain, I'd have to disable Alice to disable Bob to disable the Unknowns.

shell/client/shell.html
```diff
@@ -333,6 +333,27 @@
...
+                      {{#each sharers}}
+                        <li>{{displayName sharer}} ({{dateString created}})</li>
+                      {{/each}}
+                    </ul>
+                  </li>
+                {{/each}}
+              </ul>
+            {{/if}}
+          {{else}}
+            <button class="show-transitive-shares">show transitive shares</button>
```
zarvox: I might suggest using a less-technical term than "transitive" for the user-facing copy.

Perhaps "reshares" or something less scary that doesn't require the user to have a mental model of graphs?

kentonv: +1 for "reshares"

kentonv:
@ocdtrekkie Directly removing reshares might be OK if you are the owner, but so far we've implemented everything such that the owner is not special-cased in any way. If you aren't the owner, the problem is that Bob may have received his own permissions from multiple sources (of which you are only one), in which case obviously you should not be allowed to control how he reshares those permissions.

Theoretically, the ideal solution would be that everyone can set policies on the permissions they personally share, and those policies flow with the permissions whenever they go. So if you set a "no resharing" policy on the permissions you shared to Bob, then he cannot reshare the permission he got from _you_, but he can possibly reshare permissions he got from other people, if they don't have similar policy restrictions.

If all of that were implemented, then removing transitive shares would simply be a matter of setting a policy that counteracts it. So if you shared to Bob who shared to Carol, and you wanted to revoke Carol without revoking Bob, you'd modify the permission you shared to Bob with the policy "cannot share to Carol". The Bob->Carol edge remains in the graph, but no longer carries any permissions.

We haven't implemented policies just yet, but we're planning on it. It might make sense to start out by saying that only the owner can set policies, so that we don't have to track their flow through the graph. OTOH, tracking their flow through the graph may not be all that hard, as long as we can define "intersect" and "union" operations across policies, just as we have for permissions.

(We will, of course, want to design a UX around all this that doesn't require users to actually understand the details of how policies work. I think we can do that -- for example, we can present the user with the ability to revoke a transitive share, but actually implement the feature as a policy.)

ocdtrekkie:
Alright, all of that makes sense. I just wanted to point out that your blog example of why delegation was important, was so if the owner was on vacation, the other person with permissions could share them themselves. But similarly, it'd be a shame if the owner needed to wait for the say, Alice, who reshared with Bob, to come in to disable Bob's access.

kentonv:
@dwrensha Let's replace "transitive shares" with "reshares" in the UI. Longer-term we'll want to do a bunch more UX work of course, but it's better to have something than nothing.

I think that long-term we will want the initial sharing window to focus entirely on sharing to new people, with some sort of a "see who has access" link that can be used to open a new window to see the current list. That window should probably be structured as a table, with columns "name", "role", "received access from". (Also the graph visualization... someday.)

dwrensha:
I've switched over to "reshares" and changed the logic a bit so that direct shares do not show up in the reshares list.

kentonv:
I suppose technically I might have given write access to Alice and read access to Bob, and then Alice may have given write access to Bob, in which case it is actually relevant to know that Alice shared to Bob even though I also shared directly to Bob. That said, this isn't supposed to be a final version of the UI anyway, so let's not fret over it for now.

PR #397, opened 2015-05-11
Fix copy/pasto in sandstorm --help message and adjust style for consistency.

dwrensha:


kentonv:
I actually prefer to write doc comments in descriptive rather than imperative style (i.e. the opposite of the style change you're making here), though I commonly forget to do so. The Google C++ style guide actually specified this and I feel like it does in fact improve readability, albeit mostly in longer, more-complex descriptions. (With imperative style, in docs with multiple paragraphs, it can become unclear which sentences are describing the behavior of the documented command vs. instructing the user on things they need to do before/after the command, so longer doc comments tend to switch over to descriptive style.)

dwrensha:
I agree that descriptive style makes more sense than imperative style in longer-form help text. I've added a commit that adjusts the `briefDescription`s of all `sandstorm` subcommands to be non-imperative.

I think that the `briefHelpText` field should remain imperative, if only to match prevailing style used by other commandline tools that have subcommands, e.g. `capnp`, `cargo`, `git`.

PR #400, opened 2015-05-12
Work In Progress: sandstorm install script tests

paulproteus:
This is some code to automatically test `install.sh`.

It doesn't test all cases yet, not by a long shot, but I wanted to get @jparyani and/or @zarvox to give it a quick look-over and suggest ways it could be better.

If you can't think of any, that's fine, but that probably means you're not being creative enough. (-;

(I plan to make some changes before shipping this.)

Open questions:

* If I'm driving the install script's tests via Vagrant, then how can this run on our Jenkins server? (One answer: Adjust Vagrant to use the libvirt backend. Another answer: Run this stuff somewhere we can run VirtualBox, like @kentonv's closet.)

* Is this syntax (see `installer-tests/*.t`) basically reasonable for people?

* Should I document this syntax somewhere?

* Should I go find an actual parsing library and learn how to use it?

* I'd like to capture the full conversation on the PTYs, so that in the case of test failures, one doesn't have to eyeball pexpect's error output (which is not too bad, so this isn't a showstopper).

* How do we avoid making our install stats look inflated due to running the tests? (Answer for now: (a) pass a `User-Agent` header through to `curl` when it's being run through the installer, and (b) adjust the stats system to ignore all requests from the "office" as well as all requests that come in with that `User-Agent` header.)

Things I know I'll have to do and don't totally know how, but intend to figure out:

* Figure out an enhancement to this syntax to let us specify which VM to run it on. This way, we can e.g. test on Fedora, and Docker inside Ubuntu 14.04, and so on.

Things I already know how to do, and plan to:

* Adjust the handling of `Vagrant-Precondition-bash` so that if the precondition fails, the test destroys the VM and then recreates it and then brings the VM back up again.

Things I'm considering doing:

* Creating a library of helper bash functions, and having Vagrant inject those into the VM, so I can simplify the various `Vagrant-Precondition-bash` lines to call those functions rather than contain lots of copy-pasta.

Please don't merge this yet, but feedback welcome.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,7 @@ bundle/
 .vagrant/
 phantomjsdriver.log
 selenium-debug.log
+
+*~
+*#
+.#*
--- a/installer-tests/no-root.t
+++ b/installer-tests/no-root.t
@@ -1,0 +1,26 @@
+Title: Can't install Sandstorm without root, by default
+Cwd: SANDSTORM_SOURCE_TREE
+Vagrant-Precondition-bash: ! -d $HOME/sandstorm
+Vagrant-Precondition-bash: ! -d /opt/sandstorm
+Vagrant-Postcondition-bash: ! -d $HOME/sandstorm
+Vagrant-Postcondition-bash: ! -d /opt/sandstorm
+
+$[run]/vagrant/install.sh
+$[slow]Sandstorm makes it easy to run web apps on your own server. You can have:
+
+1. A full server with automatic setup (press enter to accept this default)
+2. A development server, for writing apps.
+
+How are you going to use this Sandstorm install? [1] $[type]2
+If you want app developer mode for a Sandstorm install, you need root
+due to limitations in the Linux kernel.
+
+To set up Sandstorm, we will need to use sudo.
+Sandstorm's database and web interface won't run as root.
+OK to continue? [yes] $[type]no
+If you are OK with a local Sandstorm install for testing
+but not app development, re-run install.sh with -u to bypass this message.
+For developer mode to work, the script needs root, or read above to bypass.
+*** INSTALLATION FAILED ***
+Report bugs at: http://github.com/sandstorm-io/sandstorm
+$[exitcode]1
--- a/installer-tests/run_tests.py
+++ b/installer-tests/run_tests.py
@@ -1,5 +1,154 @@
+import glob
+import os
+import pexpect
+import subprocess
+import re
+
+def _expect(line, current_cmd, do_re_escape=True, do_detect_slow=True, strip_comments=True,
+            verbose=True):
+    timeout = 1
+    if do_detect_slow:
+        if line.startswith('$[slow]'):
+            print 'Slow line...'
+            timeout = 30
+            line = line.replace('$[slow]', '', 1)
+
+    if verbose:
+        print 'expecting', line
+
+    if do_re_escape:
+        line = re.escape(line)
+
+    current_cmd.expect(line, timeout=timeout)
+
+
+TEST_ROOT=os.path.dirname(os.path.abspath(__file__))
+SANDSTORM_SOURCE_TREE=os.path.abspath(os.path.join(os.path.abspath(__file__), '..', '..'))
+SANDSTORM_IN_HOME=os.path.join(os.environ['HOME'], 'sandstorm')
+
+def vagrant_destroy():
+    subprocess.check_output(['vagrant', 'destroy', '-f'], cwd=TEST_ROOT)
+
+def handle_test_script(lines):
+    current_cmd = None
+    next_timeout = 1
+
+    for line in lines:
+        # Figure out what we want to do, given this line.
+        if line.startswith('$[run]'):
+            arg = line.replace('$[run]', '')
+            arg = 'vagrant ssh -c "' + arg + '"'
+            print 'starting', arg, '...'
+            current_cmd = pexpect.spawn(arg, cwd=TEST_ROOT)
+        elif '$[exitcode]' in line:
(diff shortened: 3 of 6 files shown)


.gitignore
```diff
@@ -25,3 +25,7 @@ bundle/
 .vagrant/
 phantomjsdriver.log
 selenium-debug.log
+
+*~
```
kentonv: Can you put this in a per-user ignore rather than the project-wide ignore? I personally really like git to tell me when random tools or editors have decided to leave trash lying around my directory tree.

kentonv:
Let's say you own `installer-tests`, so you don't need my review, but if you'd like review then I think Jason would be a good reviewer. :)

installer-tests/run_tests.py
```diff
@@ -0,0 +1,155 @@
...
+
+    if do_re_escape:
+        line = re.escape(line)
+
+    current_cmd.expect(line, timeout=timeout)
+
+
+TEST_ROOT=os.path.dirname(os.path.abspath(__file__))
+SANDSTORM_SOURCE_TREE=os.path.abspath(os.path.join(os.path.abspath(__file__), '..', '..'))
+SANDSTORM_IN_HOME=os.path.join(os.environ['HOME'], 'sandstorm')
```
zarvox: This appears to be unused?

installer-tests/run_tests.py
```diff
@@ -0,0 +1,155 @@
...
+            full_bash_cmd = 'if [ %s ] ; then exit 0 ; else exit 1 ; fi' % (value,)
+            exitcode = subprocess.call(['vagrant', 'ssh', '-c', full_bash_cmd], cwd=TEST_ROOT)
+            if exitcode == 0:
+                print 'Destroying...'
+                vagrant_destroy()
+                print 'Recreating...'
+                vagrant_up()
+
+        if key == 'vagrant-precondition-bash':
+            full_bash_cmd = 'if [ %s ] ; then exit 0 ; else exit 1 ; fi' % (value,)
```
zarvox: You can simply exit [ %s ] here, no?

installer-tests/no-root.t
```diff
@@ -0,0 +1,26 @@
+Title: Can't install Sandstorm without root, by default
+Cwd: SANDSTORM_SOURCE_TREE
+Vagrant-Precondition-bash: ! -d $HOME/sandstorm
+Vagrant-Precondition-bash: ! -d /opt/sandstorm
+Vagrant-Postcondition-bash: ! -d $HOME/sandstorm
```
zarvox: these assertions don't seem to be implemented.

installer-tests/run_tests.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+TEST_ROOT=os.path.dirname(os.path.abspath(__file__))
+SANDSTORM_SOURCE_TREE=os.path.abspath(os.path.join(os.path.abspath(__file__), '..', '..'))
+SANDSTORM_IN_HOME=os.path.join(os.environ['HOME'], 'sandstorm')
+
+def vagrant_destroy():
+    subprocess.check_output(['vagrant', 'destroy', '-f'], cwd=TEST_ROOT)
+
+def handle_test_script(lines):
```
zarvox: If you plan to do more with this tooling, consider refactoring these into a TestScript class?  Something where the phases of (parse, read headers, interpret script) are more neatly structured?

I could see something like:

    class TestScript(object):
        def __init__(self, file_like_object):
            self.headers = []
            self.command_sequence = []
            self.preconditions = []
            self.postconditions = []
            self.parse(file_like_object) # does the stuff that iterating over lines currently does

        @staticmethod
        def load_from_file(path):
        ...
        def run(self):
            for condition in self.preconditions:
                self.verify_condition(condition)
            for command in self.command_sequence:
                self.run_command(command)
            for condition in self.post_conditions:
                self.verify_condition(condition)
        ....
        def verify_condition(self, condition):
             # dispatch on condition type
        ...

Or maybe I'm bikeshedding and this is totally fine for now. :P

.gitignore
paulproteus: +1

installer-tests/run_tests.py
paulproteus: +1, ran `flake8` on it and fixed many warnings

installer-tests/run_tests.py
paulproteus: Is it `exit [ %s ]` or `exit %s` ?

To avoid having to think about that, I wrote it the above way.

FWIW:

```
asheesh@laptop:~ $ if [ -f /tmp ] ; then echo yes ; else echo no ; fi
no
asheesh@laptop:~ $ exit [ -f /tmp ]
exit
bash: exit: [: numeric argument required
```

So I don't think literally `exit [ %s ]` works, but if I'm mistaken by all means let me know.

installer-tests/no-root.t
paulproteus: Fixed in current version.

installer-tests/run_tests.py
paulproteus: I like the idea of there being a `TestScript`. I will probably adjust that in the next revision.

paulproteus:
Closing in favor of #405 but thanks for the review!

installer-tests/run_tests.py
zarvox: Sorry, you don't need the `exit`.  It's just `[ %s ]`.  `man [` should elucidate.

Perhaps a more-readable alternative is to have bash simply execute the entire test-provided-line verbatim, and then users can also have slightly more interesting assertions that don't depend on being expressed as flags to `[`.

PR #401, opened 2015-05-14
Allow signed-in but non-invited users to see a Shared With Me tab.

dwrensha:


kentonv:
Can we set this up so that if the user is not invited and doesn't have any grains shared with them, we continue to show them the old splash screen?

dwrensha:
Done.

PR #402, opened 2015-05-14
Add ongoing notifications and persistence

jparyani:
Sorry, the bulk of the code is in one big commit since I was developing both persistence and the ongoing notification code in tandem. Let me know if it's hard to review, and I'll try to split it into smaller chunks.

As a bonus, I've written tests :)


shell/shared/shell.js
```diff
@@ -445,8 +450,78 @@ if (Meteor.isClient) {
   Template.homeLink.helpers({
     origin: getOrigin
   });
-}
 
+  Template.notifications.helpers({
+    notifications: function () {
+      return Notifications.find({userId: Meteor.userId()}, {sort: {timestamp: -1}}).map(function (row) {
+        // TODO(someday): make this work with grains that are not owned by the current user.
+        // Currently we only subscribe to grains owned by this user, and the following query will
```
kentonv: It's a bit worse than that: We only subscribe to all of the user's grain titles when they are looking at the grain list. If they are looking at a specific grain, then we'll only have subscribed to that one grain, and if they're on some other page, we won't have any titles. Currently if a grain is running in the background and you open another grain, the notifications view looks wrong because the grain title is missing.

shell/client/shell.html
```diff
@@ -41,11 +41,41 @@
...
+  {{#if notificationsIsEnabled}}
+    <div id="notification-dropdown">
+      <h4>Notifications</h4>
+      <ul id="notification-list">
+        {{#each notifications}}
+          <li>
+            <div class="notification-item">
+              <div class="notification-header">
+                  <span>{{grainTitle}}</span>
+                  <span title="Dismiss" class="cancel-notification" data-notificationid="{{_id}}">x</span>
```
kentonv: We'll need to make it clear that the user is canceling the task, not simply dismissing the notification. Designing this is probably a task for Nena but for the moment maybe we should at least use a button labeled "cancel" rather than a little red x.

shell/server/core.js
```diff
@@ -0,0 +1,220 @@
...
+      dropToken(notification.grainId, new Buffer(notification.ongoing));
+    }
+  }
+}
+
+function tryDropNotification(notificationId) {
+  // TODO(soon): This has serious timing issues with Mongo, since we're relying on it for refcounting
+  // Typically a notification handle will be saved and then immediately closed by the caller.
+  // This means that if Mongo doesn't actually find the ApiToken that was just stored, this method
+  // will dismiss the notification even though it was meant to be persisted.
```
kentonv: I think this timing issue can be solved by having `NotificationHandle` contain a boolean field indicating whether `save()` was called. Then, in `close()`, check for that boolean rather than searching the `ApiTokens` table.

It looks like the other place `tryDropNotification()` is called (inside `dropToken()`), there's no serious timing issue.

Or am I misunderstanding?

shell/server/core.js
```diff
@@ -0,0 +1,220 @@
...
+  var token = ApiTokens.findOne({"frontendRef.notificationHandle": notificationId});
+  if (!token) {
+    dismissNotification(notificationId);
+  }
+}
+
+Meteor.methods({
+  dismissNotification: function (notificationId) {
```
kentonv: I guess this will both be for "dismissing" one-time notifications and "canceling" ongoing notifications? Maybe add a comment to that effect.

shell/server/core.js
```diff
@@ -0,0 +1,220 @@
...
+}
+
+function dropWakelock(grainId, wakeLockNotificationId) {
+  // For some reason, Mongo returns an object that looks buffer-like, but isn't a buffer.
+  // We have to explicitly copy it before we can pass it on to node-capnp.
+  var copiedId = new Buffer(wakeLockNotificationId);
+  waitPromise(openGrain(grainId).supervisor.drop({wakeLockNotification: copiedId}).catch(function (err) {
+    if (shouldRestartGrain(err, 0)) {
+      return openGrain(grainId, true).supervisor.drop({wakeLockNotification: copiedId});
+    }
```
kentonv: You should re-throw the error if shouldRestartGrain() returns false (in all the places where you're calling shouldRestartGrain()).

Maybe we should have a helper method for this common pattern.

    useGrain(grainId, function (supervisor) {
      // do stuff with supervisor; whole function may be retried on disconnect
    });

shell/server/core.js
```diff
@@ -0,0 +1,220 @@
...
+      }
+    });
+    return {sturdyRef: sturdyRef};
+  });
+};
+
+SandstormCoreImpl.prototype.restore = function (sturdyRef) {
+  var self = this;
+  return inMeteor(function () {
+    var hashedSturdyRef = Crypto.createHash("sha256").update(sturdyRef).digest("base64");
```
kentonv: Maybe factor the hash-creation into a function since you call it a lot?

shell/server/core.js
```diff
@@ -0,0 +1,220 @@
...
+  });
+};
+
+var dropToken = function (grainId, sturdyRef) {
+  var hashedSturdyRef = Crypto.createHash("sha256").update(sturdyRef).digest("base64");
+  var token = ApiTokens.findOne({_id: hashedSturdyRef});
+  if (!token) {
+    return;
+  }
+  if (token.frontendRef) {
```
kentonv: In this if/else, you're assuming that all frontendRefs are notificationHandles and that all objectIds are wakeLockNotifications, but these are both unions that could contain other things, so please check and throw an error on unsupported types.

shell/server/core.js
```diff
@@ -0,0 +1,220 @@
...
+    dropToken(self.grainId, sturdyRef);
+  });
+};
+
+SandstormCoreImpl.prototype.makeToken = function (ref, owner) {
+  var self = this;
+  return inMeteor(function () {
+    var sturdyRef = new Buffer(Random.id(20));
+    var hashedSturdyRef = Crypto.createHash("sha256").update(sturdyRef).digest("base64");
+    // TODO(soon): should userId be filled?
```
kentonv: This relates to the TODO in the doc comment for the RPC method. What `userId` is really meant to track is the edge in the sharing graph leading to this cap. In the case of background processing, there is none. Eventually `makeToken()` should take another parameter which can be filled in to indicate a parent user or a parent ApiToken.

src/sandstorm/run-bundle.c++
```diff
@@ -1796,13 +1796,18 @@ private:
       dropPrivs(config.uids);
       clearSignalMask();
 
-      capnp::TwoPartyServer server(kj::heap<BackendImpl>(*io.lowLevelProvider, network));
+      auto redirector = kj::refcounted<CapRedirector>();
+      capnp::Capability::Client coreFactoryCap = kj::addRef(*redirector);
+      TwoPartyServerWithBootstrap server(
+        kj::heap<BackendImpl>(*io.lowLevelProvider, network,
+          coreFactoryCap.castAs<SandstormCoreFactory>()),
+        kj::addRef(*redirector));
```
kentonv: It looks like this is the last use of `redirector` in this scope so shouldn't it be kj::mv() rather than kj::addRef()?

src/sandstorm/supervisor.capnp
```diff
@@ -107,6 +107,10 @@ interface SystemPersistent extends(Persistent(Data, ApiTokenOwner)) {
   # created for the purpose of HTTP APIs).
 }
 
+interface HandlePersistent extends(SystemPersistent, Util.Handle) {}
```
kentonv: I've generally used `Persistent` as a prefix rather than a suffix for these cases. (`SystemPersistent` is different because it's not "a persistent System", it's "the system's Persistent interface".)

src/sandstorm/supervisor.capnp
```diff
@@ -154,7 +158,7 @@ struct SupervisorObjectId(AppObjectId) {
     appRef @0 :AppObjectId;
     # A reference restorable by the app.
 
-    wakeLockNotification @1 :UInt32;
+    wakeLockNotification @1 :Data;
```
kentonv: This was a UInt32 before because I figured you'd just use a counter. SupervisorObjectId is allowed to contain guessable content, since it will be suitably obscured behind an ApiToken.

src/sandstorm/util.h
```diff
@@ -334,6 +335,86 @@ class SubprocessSet {
...
+    return context.tailCall(kj::mv(req));
+  }
+
+private:
+  uint iteration = 0;
+  capnp::Capability::Client target;
+  kj::Own<kj::PromiseFulfiller<capnp::Capability::Client>> fulfiller;
+};
+
+class TwoPartyServerWithBootstrap: private kj::TaskSet::ErrorHandler {
```
kentonv: It took me a bit to figure out how this was different from TwoPartyServer, because I was thinking "but all servers have a bootstrap". But the difference is that it accepts a client bootstrap.

So maybe call this `TwoPartyServerWithClientBootstrap`, and update the comment to explain this.

src/sandstorm/run-bundle.c++
```diff
@@ -1796,13 +1796,18 @@ private:
       dropPrivs(config.uids);
       clearSignalMask();
 
-      capnp::TwoPartyServer server(kj::heap<BackendImpl>(*io.lowLevelProvider, network));
+      auto redirector = kj::refcounted<CapRedirector>();
```
kentonv: It seems the reason you need to create the redirector yourself and then pass it in to `TwoPartyServerWithBootstrap` is because of a cyclic problem where TPSWB requires the server's bootstrap as a constructor input, which requires the redirector as an input. But it would be kind of nice if TPSWB could fully encapsulate the concept of the CapRedirector, and only have the one-argument constructor.

A way to solve this would be to use promises. Create a PromiseAndFulfiller<Backend::Client>, pass that as the server's bootstrap, and then immediately fulfill the promise by constructing BackendImpl based on the client's bootstrap from TPSWB::bootstrap().

[Here's an example of something like this from Blackrock.](https://github.com/sandstorm-io/blackrock/blob/16a82a53080c424caa2a743b83f9d9ff8703b143/src/blackrock/blackrock.c++#L507-L521)

src/sandstorm/supervisor.c++
```diff
@@ -1526,8 +1531,11 @@ void SupervisorMain::DefaultSystemConnector::checkIfAlreadyRunning() const {
   KJ_UNREACHABLE;
 }
 
-class SupervisorMain::SandstormApiImpl final: public SandstormApi<>::Server {
+class SupervisorMain::SandstormApiImpl final:
+  public SandstormApi<>::Server, public kj::Refcounted, private kj::TaskSet::ErrorHandler  {
```
kentonv: It's dangerous to make capability servers refcounted because then `thisCap()` (which is like `this` except returning a capability) doesn't work (due to implementation complications).

I think what you really want here is a separate class called `WakeLockSet` which is itself refcounted with `SandstormApiImpl` holding one reference. `WakeLockSet` would implement `restoreWakeLockNotification()` and such.

src/sandstorm/supervisor.c++
```diff
@@ -1586,14 +1594,195 @@ public:
...
+      return;
+    }
+    auto req = sandstormCore.dropRequest();
+    req.setToken(iter->second.sturdyRef);
+    tasks.add(req.send().then([](auto args) {}));
+    iter->second.isHandleActive = false;
+    stopWakelock(iter);
+  }
+
+  kj::Promise<void> cancelOngoingCapPromise(kj::StringPtr wakelockId, OngoingNotification::Client ongoingNotification) {
```
kentonv: Please wrap at 100 chars (various places in this file).

src/sandstorm/supervisor.c++
```diff
@@ -1586,14 +1594,195 @@ public:
...
+      stopWakelock(iter);
+    }
+  }
+private:
+  class LockInfo {
+  public:
+    kj::String id;
+    kj::Array<byte> sturdyRef;
+    OngoingNotification::Client ongoingNotification;
+    bool isHandleActive;
```
kentonv: Could use some comments describing some of these fields.

src/sandstorm/supervisor.c++
```diff
@@ -1586,14 +1594,195 @@ public:
...
+    uint32_t notificationRefCount;
+
+    LockInfo(kj::String&& id) :
+      id(kj::mv(id)), ongoingNotification(nullptr), isHandleActive(true), notificationRefCount(0) {
```
kentonv: I find it's a little easier to tell the initializers apart from the function body by writing like:

    LockInfo(kj::String&& id)
        : id(kj::mv(id)), ... {
      SANDSTORM_LOG(...);
      ...
    }

jparyani:
All comments addressed. The bulk of the refactor occurred in the last commit (7ced8de)

shell/server/core.js
```diff
@@ -0,0 +1,241 @@
...
+      return;
+    }
+    if (token.frontendRef) {
+      if (token.frontendRef.notificationHandle) {
+        var notificationId = token.frontendRef.notificationHandle;
+        ApiTokens.remove({_id: hashedSturdyRef});
+        var anyToken = ApiTokens.findOne({"frontendRef.notificationHandle": notificationId});
+        if (!anyToken) {
+          // No other tokens referencing this notification exist, so dismiss the notification
+          dismissNotification(notificationId);
```
kentonv: In the case that the handle is dropped, we don't actually want to call cancel() on the underlying notification, but simply drop it as well. `cancel()` should be called when the user is actively asking to interrupt the task, not when the task merely declares itself done.

src/sandstorm/supervisor.c++
```diff
@@ -1535,10 +1539,123 @@ void SupervisorMain::DefaultSystemConnector::checkIfAlreadyRunning() const {
...
+  public:
+    WrappedOngoingNotification(OngoingNotification::Client ongoingNotification,
+                               WakelockSet& wakelockSet)
+      : ongoingNotification(ongoingNotification), wakelockSet(wakelockSet), isCancelled(false) {
+      ++sandstorm::wakelockCount;
+    }
+    WrappedOngoingNotification(WrappedOngoingNotification&&) = delete;
+    KJ_DISALLOW_COPY(WrappedOngoingNotification);
+
+    ~WrappedOngoingNotification() {
```
kentonv: Make sure your destructors are always declared `noexcept(false)`, because C++11 rules unfortunately made them `noexcept` by default.

(Of course, you still need to make sure your destructor can't throw during unwind.)

kentonv:
Looks like there are more merge conflicts to resolve.

shell/server/proxy.js
```diff
@@ -307,6 +308,21 @@ shouldRestartGrain = function (error, retryCount) {
   return error.type === "disconnected" && retryCount < 1;
 }
 
+useGrain = function (grainId, cb) {
+  // This will open a grain for you, handlind restarts if needed, and call the passed function with
```
dwrensha: s/handlind/handling/

shell/shared/shell.js
```diff
@@ -169,6 +178,8 @@ if (Meteor.isClient) {
   };
   Template.registerHelper("dateString", makeDateString);
 
+  Template.registerHelper("dateString", makeDateString);
```
dwrensha: Duplicate code?

shell/shared/shell.js
```diff
@@ -450,8 +461,75 @@ if (Meteor.isClient) {
...
+        document.querySelector("body").addEventListener("click", globalClickHandler);
+      } else {
+        document.querySelector("body").removeEventListener("click", globalClickHandler);
+      }
+    });
+  });
+
+  Meteor.subscribe("notifications");
+  // TODO(someday): don't leave naked subscribes at the top level.
+  // The following works in Meteor 1.1+. Remove the above line and use the following when we update.
```
dwrensha: Should you switch this over now that we're upgraded Meteor?

jparyani:
Done.

PR #404, opened 2015-05-15
Add install script tests, and fix sysctl/userns detection (WIP (do not merge yet))

### LINKED ISSUE #396: curl https://install.sandstorm.io | bash fails when not run as root
Related to issue #377: when install.sh is run as non-root, it fails while trying to set  kernel.unprivileged_userns_clone because it runs "cat >> /etc/sysctl.conf ..." without non-root permissions.

Easy to reproduce: run "docker build -t sandstorm ." or "curl https://install.sandstorm.io | bash" as non-root.

```
$ docker build -t sandstorm .
[...]
Step 8 : RUN /install.sh -d -u
 ---> Running in b72a5c17c5ef
Sandstorm requires sysctl kernel.unprivileged_userns_clone to be enabled.
Currently, it is not enabled on your system.
/install.sh: line 266: /etc/sysctl.conf: Permission denied
INFO[0052] The command [/bin/sh -c /install.sh -d -u] returned a non-zero code: 1 
```

paulproteus:
This commit:

* Introduces the installer-tests directory, which contains

* A Vagrantfile, indicating how to run a variety of operating systems
  for testing the Sandstorm installer, and

* run_tests.py, a quick sketch of how to read sample command line
  interaction scripts and verify that they still work, and

* Three tests for run_tests.py, to avoid regressions in the install
  script.

It also fixes install.sh to properly handle enabling the sysctl for
unprivileged_userns_clone.

Close #396

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -144,6 +144,9 @@ USE_SANDCATS="no"
 SANDCATS_SUCCESSFUL="no"
 CURRENTLY_UID_ZERO="no"
 PREFER_ROOT="yes"
+USERNS_CLONE_AT_ALL=""
+USERNS_CLONE_UNPRIVILEGED_NEEDS_SYSCTL_SET=""
+SYSCTL_PROBABLY_WORKS="yes"
 
 # Defaults for some config options, so that if the user requests no
 # prompting, they get these values.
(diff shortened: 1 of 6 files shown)


paulproteus:
@jparyani I believe I need to make sure the `User-Agent` is set to something you can detect, to avoid ruining stats, is that right? I'll work on that next.

Other than that, any grave objections to this?

jparyani:
Yep, `User-Agent` would be a good way to filter out these installs. Feel free to pick anything that's reasonably unique.

paulproteus:
I'll use `testing`!

install.sh
```diff
@@ -276,77 +279,171 @@ assert_usable_kernel() {
...
+  fi
 
-  return 0;
+  # OK, so they need the sysctl set.
+  #
+  # If, however, they _can't_ set the sysctl, make the install fail
+  # now.
+  if [ "$SYSCTL_PROBABLY_WORKS" = "no" ] ; then
+    echo "# sysctl -w kernel.unprivileged_userns_clone=1"
+    echo "# cat >> /etc/sysctl.conf << __EOF__"
```
kentonv: Using `#` as a prompt character strikes me as slightly confusing here because it's also used as a comment just a couple lines later.

Maybe we don't need to append the comment in the case where the user is adding it manually, so you could say:

    echo 'kernel.unprivileged_userns_clone = 1' >> /etc/sysctl.conf

And now since it's just two single-liners, you could merely indent them rather than putting a fake-prompt in front of them, which the user might otherwise confusedly try to include in the command. This strikes me as consistent with your preference about putting quotes around "help". :)

paulproteus: +1

install.sh
```diff
@@ -276,77 +279,171 @@ assert_usable_kernel() {
...
+  # work, and the user doesn't need a sysctl set, then we are happy.
+  if [ "$USERNS_CLONE_UNPRIVILEGED_NEEDS_SYSCTL_SET" = "no" ] ; then
+    return
+  fi
+
+  # At this point, we're going to need this sysctl set.
+  #
+  # But that won't work under one of a few circumstances. Let's identify
+  # those so we know to bail out.
+  local SYSCTL_PROBABLY_WORKS="yes"
```
kentonv: Should this also be set to "no" if the user is not root?

install.sh
paulproteus: Just pushed this change; will re-test it in a moment.

install.sh
```diff
@@ -276,77 +279,171 @@ assert_usable_kernel() {
...
+  if [ "${PRINT_USERNS_PROMPT}" = "yes" ] ; then
+    echo "Sandstorm requires sysctl kernel.unprivileged_userns_clone to be enabled."
+    echo "Currently, it is not enabled on your system."
+    if prompt-yesno "Shall I enable it for you?" yes; then
+      ACCEPTED_SYSCTL_SWITCH="yes"
     fi
+  fi
+
+  if [ "$ACCEPTED_SYSCTL_SWITCH" = "yes" ] ; then
+    if [ ! -e /etc/sysctl.conf ]; then
```
kentonv: Didn't we already check this in assert_userns_clone_works_or_can_be_made_to_work?

install.sh
```diff
@@ -276,77 +279,171 @@ assert_usable_kernel() {
...
+
+  if [ "yes" = "${ACCEPTED_FULL_SERVER_INSTALL:-}" ] ; then
+    PRINT_USERNS_PROMPT="no"
+    ACCEPTED_SYSCTL_SWITCH="yes"
+  fi
+
+  if [ "${PRINT_USERNS_PROMPT}" = "yes" ] ; then
+    echo "Sandstorm requires sysctl kernel.unprivileged_userns_clone to be enabled."
+    echo "Currently, it is not enabled on your system."
+    if prompt-yesno "Shall I enable it for you?" yes; then
```
kentonv: Should we give some indication that we plan to enable it permanently? (Mainly so that the rare user who has any idea what a sysctl is doesn't get the impression that they'll have to go do that themselves later.)

paulproteus:
I'm going to close this for now to avoid confusion but still process your comments.

install.sh
paulproteus: +0, will have to figure out how to phrase this

install.sh
paulproteus: Tested.

install.sh
paulproteus: If the user gives us consent to re-run as root later, then we don't want to bail out now.

I haven't fully tested the sysctl stuff with `-u` (for not using root). I plan to do that in a follow-up patch set, for what it's worth.

install.sh
paulproteus: Yeah. I can remove this `! -e` check here.

install.sh
paulproteus: I went with:

```
-    if prompt-yesno "Shall I enable it for you?" yes; then
+    if prompt-yesno "Shall I enable it for you (including in sysctl.conf)?" yes; then
```

PR #408, opened 2015-05-16
Add "Self-install" link for demo users

paulproteus:
**Goal**

People seem to use the "appDemo" links we have in the wild, e.g. etherpad.org => "Or try a demo" => https://demo.sandstorm.io/appdemo/h37dm17aa89yrd8zuqpdn36p6zntumtv08fjpu8a8zrte7q1cn60 , and for those people, I don't think they actually understand they can install Sandstorm then+there and get the app.

**Status**

I've gotten color feedback from @neynah and I think this is ready to ship.

The `/install/` page supports `#appname` as a way to provide extra info. For now, this code doesn't use that; it merely uses the generic `#demo`, which does everything except add the specific app name.

(I am hoping to ship this as-is, and then add the app name stuff later.)

The `/install/` page will go to this thing: https://github.com/sandstorm-io/sandstorm-website/pull/69

**Technical details**

This re-uses the styles from the Indiegogo campagin, and partially
reverts commit 0eb64bfb2b2d10a4194a62d02ce576cacdcf308f.

In the future, for demo users, they will get a link to
https://sandstorm.io/demo/#appname so that they see a customized page
that appeals to their love of appname.

### DIFF
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -765,7 +765,7 @@ $KEY</textarea></p>
             secure sandbox.</li>
           <li>We only have one machine. It may get slow or crashy under excessive load.</li>
           <li>We know the UI design needs work. This is a tech demo. ;)</li>
-          <li><a href="http://igg.me/at/sandstorm">Read more and fund us on Indiegogo.</a></li>
+          <li><a href="https://sandstorm.io">Read more at Sandstorm.io.</a></li>
         </ul>
       {{else}}
         <p>Sorry, this server does not allow demo accounts.</p>
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -41,6 +41,9 @@ limitations under the License.
   </div>
   {{else}}
   <div id="topbar-right">
+    {{#if isDemoUser}}
+    <a id="alert" href="https://sandstorm.io/install/">Run your own »</a>
+    {{/if}}
     {{> loginButtons align="right"}}
   </div>
   {{/if}}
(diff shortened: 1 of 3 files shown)


paulproteus:
What it looks like for demo users in the shell:

run-your-own


What it looks like when you hover over that:

run-your-own-hover

paulproteus:
@neynah I realize you're probably in the middle of totally re-imagining the shell. I'm hopeful I can land this very soon since I'm going on vacation soon and want to start capitalizing on all that appdemo traffic, if possible!

shell/client/shell.html
```diff
@@ -41,6 +41,9 @@
   </div>
   {{else}}
   <div id="topbar-right">
+    {{#if isDemoUser}}
+    <a id="alert" href="https://sandstorm.io/install/">Run your own »</a>
```
zarvox: I assume the target of this link will point to some new page with helpful instructions?

paulproteus: Thanks for asking! I just added a link to
https://github.com/sandstorm-io/sandstorm-website/pull/69 in the pull
request description, which contains proposed contents for /install/

neynah:
The 2nd option underneath has a lighter purple that looks less "bleached." I didn't want to suggest too many changes (there isn't much time) but I hope this is enough for now.

sandstorm-shell1

paulproteus:
I've adjusted this pull request to use the 9D3AA5 color when :hover-ing!

paulproteus:
Closing in favor of https://github.com/sandstorm-io/sandstorm/pull/411

PR #411, opened 2015-05-18
Self install now

paulproteus:
**Goal**

People seem to use the "appDemo" links we have in the wild, e.g. etherpad.org => "Or try a demo" => https://demo.sandstorm.io/appdemo/h37dm17aa89yrd8zuqpdn36p6zntumtv08fjpu8a8zrte7q1cn60 , and for those people, I don't think they actually understand they can install Sandstorm then+there and get the app.

**Status**

* This supports linking to `/install/#appName` in the case that you are a demo user, looking at a particular app.

* This defaults to `/install/#demo` if it can't find an `appName` (e.g. if you're in the shell).

* The `/install/` page supports `#appname` as a way to provide extra info. To test that, visit e.g. https://sandstorm.io/install/#EtherCalc

* I had the colors sanity-checked by @neynah.

* This pull request supercedes https://github.com/sandstorm-io/sandstorm/pull/408 . Please review and/or merge.

**Technical details**

This re-uses the styles from the Indiegogo campagin, and partially
reverts commit 0eb64bfb2b2d10a4194a62d02ce576cacdcf308f.

When the demo is enabled, users get Package information about the grains they own.

### DIFF
--- a/shell/client/shell.css
+++ b/shell/client/shell.css
@@ -97,7 +97,7 @@ body {
 }
 
 #topbar-right #alert:hover {
-  background-color: #ba97c3
+  background-color: #9d3aa5;
 }
 
 #topbar a {
--- a/shell/client/shell.css
+++ b/shell/client/shell.css
@@ -86,9 +86,9 @@ body {
   z-index: 3;
 }
 
-#topbar-right #crowdfunding {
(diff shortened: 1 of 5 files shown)


paulproteus:
: D

PR #415, opened 2015-05-19
Meteor dev scripts: work from any folder

zarvox:
Some assumptions about PWD or meteor install location were made.  This cleans
those up.

### DIFF
--- a/shell/run-dev.sh
+++ b/shell/run-dev.sh
@@ -7,6 +7,7 @@ set -euo pipefail
 SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 
 (cd "$SCRIPT_DIR/.." && make shell-env)
+cd "$SCRIPT_DIR"
 
 if [ $# -gt 0 ]; then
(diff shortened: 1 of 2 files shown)


shell/run-dev.sh
```diff
@@ -69,6 +71,6 @@ __EOF__
 
 # Work-around for problem where Meteor's bundled npm prefers the system gyp
 # over its own bundled version, and the system gyp doesn't work.
-export PYTHONPATH=$HOME/.meteor/tools/latest/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib
+export PYTHONPATH=$("$SCRIPT_DIR/../find-meteor-dev-bundle.sh")/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib
 
 exec meteor run -p $PORT --settings $SETTINGS
```
kentonv: Meteor itself needs to run inside the `shell` directory. Maybe just `cd` there from the beginning, rather than for each command?

zarvox: Done.

PR #416, opened 2015-05-20
Invoke sub-make with the same make

emaste:
On some systems GNU make is installed as gmake.  $(MAKE) will pass that
through.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -129,7 +129,7 @@ tmp/ekam-bin: tmp/.deps
 	@mkdir -p tmp
 	@rm -f tmp/ekam-bin
 	@which ekam >/dev/null && ln -s "`which ekam`" tmp/ekam-bin || \
-	    (cd deps/ekam && make bin/ekam-bootstrap && \
+	    (cd deps/ekam && $(MAKE) bin/ekam-bootstrap && \
 	     cd ../.. && ln -s ../deps/ekam/bin/ekam-bootstrap tmp/ekam-bin)
 
 tmp/.ekam-run: tmp/ekam-bin src/sandstorm/* tmp/.deps


garply:

kentonv:
Hi Ed,

Are you trying to build Sandstorm on FreeBSD? Unfortunately Sandstorm is very much Linux-specific, using kernel features like namespaces, seccomp, etc. which I don't think FreeBSD has. It would be interesting to build something like Sandstorm on jails and capsicum instead, but that would be a pretty big project. Also all the app packages assume a Linux ABI, though perhaps FreeBSD's ABI compatibility layer could deal with that.

Since portability seems infeasible on a macro level, I tend to ignore it on a micro level as well, such as assuming gmake or assuming pipe2() is available (which I just noticed _is_ now available on FreeBSD, which is awesome, but you get the idea).

I'm not against changing `make` to `$(MAKE)` but it seems like this is only useful in the context of a broader porting effort that would be difficult and invasive, so I'd like to know what your plans are here.

emaste:
Hi Kenton,

While I think it would be very interesting to build Sandstorm on FreeBSD, I don't think it's feasible today and have no plans to really try to do so.  I do want to better understand what features are important to contemporary, leading-edge Linux software so that we can make sure FreeBSD has them (or equivalents) as well, and sometimes just trying to build that software provides insight.  In this case I was looking to understand the state of the bits under deps/ on FreeBSD.

In general I wouldn't bother suggesting micro-level changes of this scale, but I think $(MAKE) vs make is a Makefile best practice so sent it on.

kentonv:
Garply, ok to test

kentonv:
Hi Ed,

Fair enough! I've sent this to the CI to verify.

Note that the fist problem you'll probably run into is that Ekam (the build system) doesn't work on FreeBSD. Ironically, Ekam was actually originally written on FreeBSD, but support for it has since bitrotted. :/

jparyani:
The CI failed, but this change looks fine. It looks like we just saw a new transient error where the email driver never started up or was for some other reason not connectable. Opening an issue.

Garply, retest this please.

jparyani:
Garply, retest this please

emaste:
> Note that the fist problem you'll probably run into is that Ekam (the build system) doesn't work on FreeBSD.

Ekam's interesting - as far as I can tell the main issue is the lack of epoll. Interestingly, we do already have support in FreeBSD for epoll - in the Linux ABI compat layer. We've been pondering moving some of this into the native ABI as well.

kentonv:
@emaste Yes, that's the main problem currently; there used to be kqueue() code but it bitrotted so I deleted it. IIRC I also struggled a bit with filesystem change notifications (Ekam wants to watch the source directory for changes and trigger rebuilds). It seems that kqueue()s method for this requires opening a file descriptor for every file you want to watch, whereas Linux's inotify seems more scalable (watching a directory at a time, and using "watch descriptors" rather than file descriptors). On OSX I ran into the file descriptor limit quickly, though FreeBSD admittedly had much higher limits so maybe it's not a problem?

Meanwhile my LD_PRELOAD approach to intercepting syscalls actually worked better on FreeBSD because it seemed I was able to catch intra-libc calls whereas on Linux such calls cannot be intercepted. So e.g. if I intercepted `open()` it also caught `fopen()`, whereas on Linux I have to intercept both.

PR #422, opened 2015-05-21
Add "make test" and extend tests/run-local.sh

zarvox:
run-local.sh now spawns an Xvfb and selenium-standalone and cleans up after
itself.  This behavior can be disabled (if you wanted to run a long-lived
selenium server) with the --no-selenium flag.

@jparyani Take a look at this when you have a chance, and adjust the test machine accordingly/show me the appropriate knobs to turn?  Also, should I be calling run-tests.sh as well as run-local.sh in make test?

### DIFF
--- a/tests/run-local.sh
+++ b/tests/run-local.sh
@@ -18,6 +18,10 @@
 
 set -euo pipefail
 
+XVFB_PID=""
+RUN_SELENIUM="true"
+BUNDLE_PATH=""
+
 cleanExit () {
   rc=$1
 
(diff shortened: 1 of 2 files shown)


tests/run-local.sh
```diff
@@ -29,14 +33,63 @@ cleanExit () {
   "$SANDSTORM_DIR/sandstorm" stop
   sleep 1
   rm -rf "$SANDSTORM_DIR"
+  if [ -n "$XVFB_PID" ] ; then
+    # Send SIGINT to the selenium-server child of the backgrounded xvfb-run, so
+    # it will exit cleanly and the Xvfb process will also be cleaned up.
+    # We don't actually know that PID, so we find it with pgrep.
+    kill -SIGINT $(pgrep --parent $XVFB_PID node)
```
jparyani: This doesn't work on debian wheezy (specifically the -SIGINT), which is what our test box runs. Perhaps we should update...

tests/run-local.sh
```diff
@@ -29,14 +33,63 @@ cleanExit () {
...
+checkInstalled npm npm
+checkInstalled firefox firefox
+
+npm install
+
+if [ "$RUN_SELENIUM" != "false" ] ; then
+  checkInstalled java default-jre-headless
+  checkInstalled xvfb-run Xvfb
+  checkInstalled pgrep procps
+  xvfb-run ./node_modules/selenium-standalone/bin/selenium-standalone start &
```
jparyani: It's interesting that I expected your PR to fail since selenium is already running on the CI server. What actually happens is this command starts a server that exits immediately, and everything goes on fine.

This is probably fine as is, just wanted to comment on it.

tests/run-local.sh
jparyani: Woops sorry, I misread the error message. SIGINT works fine, it's just that pgrep would fail to find the process as per my other comment.

jparyani:
Looks good to me. I just got a bit confused over an error message that `kill` was putting out, but I think it's fine as-is.

I've given you admin rights on https://build.sandstorm.io/ (login with your github account), and I'll show you how to change the config tomorrow.

zarvox:
Garply, retest this please

PR #424, opened 2015-05-21
Add admin user management

jparyani:
Added it as a tab in the admin settings page. Also moved invite users into a tab.


jparyani:
Garply, retest this please

shell/client/shell.html
```diff
@@ -527,14 +523,13 @@
       <p>Body:
       <textarea id="invite-message">Click on the following link to get access to my Sandstorm.io server.
 
-$KEY</textarea></p>
+  $KEY</textarea></p>
```
kentonv: Why the added leading spaces?

PR #428, opened 2015-05-23
Add token auth for invite

jparyani:
Allow invite functionality for users authenticating with admin token


jparyani:
Garply, retest this please

PR #433, opened 2015-05-25
Support default role in the shell.

mquandalle:


garply:

kentonv:
Hmm I think line 254 should be updated as well (the API tokens dialog).

kentonv:
Actually, no, scratch that. API tokens should probably have all the creator's permissions by default, since they're not for sharing but for connecting apps.

kentonv:
Garply, ok to test.

kentonv:
Apparently Meteor doesn't like this.

https://build.sandstorm.io/job/Sandstorm%20PRs/330/console

    Errors prevented bundling:
    While building the application:
    client/shell.html:312: Reactive HTML attributes must either have a constant
    name or consist of a single {{helper}} providing a dictionary of names and
    values.  A template tag of type BLOCKOPEN is not allowed here.
    ...                     {{#if obsolete}}styl...

mquandalle:
Thanks Garply, this is fixed (this syntax used to work in Meteor <0.8 :-))

PR #438, opened 2015-05-28
Schema for contacts.

dwrensha:
Submitting this by itself to make sure that it meshes with other powerbox plans, before I build anything on top of it.


kentonv:
I think we want "name" in there, so that we don't have to do a "join" to generate a selection dialog. This name should also be editable by the contact "owner", i.e. it is a petname.

This would be a good time to review petnames and Zooko's Triangle:

http://www.skyhunter.com/marcs/petnames/IntroPetNames.html

dwrensha:
"petname" field added.

PR #439, opened 2015-05-28
fix ownership and permissions for var/sandstorm

dwrensha:
Currently, the install script makes <install_dir>/var/sandstorm owned by root:root. I might be missing something (cc @paulproteus), but I think this is a mistake; it should have the same ownership and permissions as all of the other directories.

Note that admin tokens live in this directory. After this patch you will be able to do `sandstorm admin-token` without `sudo`.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1084,10 +1084,6 @@ set_permissions() {
   chown -R $SERVER_USER:$GROUP var/{log,pid,mongo} var/sandstorm/{apps,grains,downloads,adminToken}
   chown root:$GROUP var/{log,pid,mongo,sandstorm} var/sandstorm/{apps,grains,downloads,adminToken}
   chmod -R g=rwX,o= var/{log,pid,mongo,sandstorm} var/sandstorm/{apps,grains,downloads,adminToken}
-
-  # Don't allow listing grain IDs directly.  (At the moment, this is faux security since
-  # an attacker could just read the database, but maybe that will change someday...)
-  chmod g-r var/sandstorm/grains
 }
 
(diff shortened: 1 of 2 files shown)


kentonv:
Hmm, I can't remember what my exact reason for doing it that way might have been. (@paulproteus only copied the code from my original installer.)

It seems like anyone in the sandstorm group should indeed be allowed to create admin tokens? Anyone in the group can, after all, arbitrarily edit the database and make themselves admin, so I guess that's not a reduction in security.

Should we add code to run-bundle.c++ which fixes up permissions on existing installs? Otherwise the differing permissions could easily lead to bugs later on.

kentonv:
BTW perhaps we should remove the line that makes the grains directory unlistable, since grain IDs are no longer secrets.

paulproteus:
Are grain IDs still secret for grains using the old permission model?

kentonv:
@paulproteus Yes, but making the grain directory unlistable has never actually truly hidden them since anyone in the group can just list the database. The original idea was that some day we would do some magic encryption that prevents someone from trivially pulling grain IDs out of the database.

As it turns out, we now do magic encryption that prevents people from pulling sharing tokens out of the database. But that means we now have no reason to pursue hiding grain IDs from mongo dumps, so therefore making the grain directory unlistable will never actually be useful.

paulproteus:
+1​

dwrensha:
> Should we add code to run-bundle.c++ which fixes up permissions on existing installs?
> ...
> perhaps we should remove the line that makes the grains directory unlistable

Done and done.

PR #441, opened 2015-05-29
Compress tarballs with more cores

zarvox:
For release tarballs:
  - increases compressed tarball size by 8 bytes
  - reduces compression time from 139.3 to 135.5 seconds

For -fast tarballs:
  - increases compressed tarball size by ~400k
  - reduces compression time from 9.7 seconds to 2.4 seconds


zarvox:
Garply, retest this please

PR #443, opened 2015-05-29
Autoselect default role in webkey popup too.

dwrensha:
Like https://github.com/sandstorm-io/sandstorm/pull/433, but for webkeys.

### DIFF
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -251,7 +251,8 @@ limitations under the License.
           <select id="api-token-role">
             {{#each roles}}
               <option title={{verbPhrase.defaultText}}
-                      style="{{#if obsolete}}display:none{{else}}{{/if}}">
+                      style="{{#if obsolete}}display:none{{/if}}"
+                      selected={{default}}>
                 {{title.defaultText}}</option>
             {{/each}}
           </select>


kentonv:
As discussed on #433, I actually don't think we want this. Since webkeys are for connecting apps, the intuitive (and backwards-compatible) default is "all access". We don't actually have a selection for that currently, but perhaps we should.

Separately, it occurs to me now that #433 is actually overloading the meaning of "default". Previously it was meant specifically for enabling backwards-compatibility when adding sharing to an app. The backwards-compatible access level is not necessarily the same as the access level which we want to encourage users to use for new shares, so perhaps these need to be two different concepts.

dwrensha:
> the intuitive (and backwards-compatible) default is "all access".

Ah, I see. If an app does not define any permissions, any tokens created through the API token popup should get an "allAccess" role. If the app does define permissions, an "allAccess" role should be offered in the drop-down, and maybe it should be selected by default. I think this makes a lot of sense.

> The backwards-compatible access level is not necessarily the same as the access level which we want to encourage users to use for new shares, so perhaps these need to be two different concepts.

I'm not sure that these need to be different concepts. In the apps I've dealt with so far, it has always made sense for the two roles to coincide. I think we should wait to add this distinction until a compelling use case comes up.

kentonv:
So I think we decided this PR isn't the right thing to do? Re-open if not.

Meanwhile, we should add support for "all access" for API tokens.

PR #444, opened 2015-05-30
Add changelog and display it on Sandstorm about page

### LINKED ISSUE #440: Show changelog on about page
We should show a change log on the "about" page, summarizing what's new in each build. It would perhaps be nice to pull the text from a changelog file in the repository home directory for easy maintenance, and so that people can read it there too. This probably requires a Makefile rule which copies the changelog from the source root into the shell's `public` directory, then code for the about page which downloads that file dynamically (much like is currently done for the contributor list), but I'm open to other ideas.

jparyani:
Fixes #440

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ IMAGES= \
 all: sandstorm-$(BUILD).tar.xz
 
 clean:
-	rm -rf bin tmp node_modules bundle shell-build sandstorm-*.tar.xz shell/.meteor/local $(IMAGES) shell/packages/*/.build* shell/packages/*/.npm/package/node_modules
+	rm -rf bin tmp node_modules bundle shell-build sandstorm-*.tar.xz shell/.meteor/local $(IMAGES) shell/client/changelog.html shell/packages/*/.build* shell/packages/*/.npm/package/node_modules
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -162,11 +162,14 @@ shell-env: tmp/.shell-env
 
 # Note that we need Ekam to build node_modules before we can run Meteor, hence
 # the dependency on tmp/.ekam-run.
-tmp/.shell-env: tmp/.ekam-run $(IMAGES)
+tmp/.shell-env: tmp/.ekam-run $(IMAGES) CHANGELOG.md
 	@mkdir -p tmp
 	@touch tmp/.shell-env
 	@mkdir -p node_modules/capnp
 	@bash -O extglob -c 'cp src/capnp/!(*test*).capnp node_modules/capnp'
+	@echo '<template name="changelog">' > shell/client/changelog.html
```
kentonv: Can you write a separate make rule for shell/client/changelog.html and have tmp/.shell-env depend on it, as we do for images?

jparyani:
Done

kentonv:
Oh, also make sure it's deleted by `make clean`.

jparyani:
Done

PR #470, opened 2015-06-05
Installer tests can live in jenkins

paulproteus:
This adjusts the tests in `installer-tests/` so that they:

* Run using `libvirt` and `qemu`, so they can run on Jenkins, and

* Have a `README.md` that explains how the tests work.

I'm running all the tests locally to make sure they all pass this way. If so, I'd like to merge this. For now, still waiting on verification that they all pass.

### DIFF
--- a/installer-tests/Vagrantfile
+++ b/installer-tests/Vagrantfile
@@ -5,7 +5,8 @@
 VAGRANTFILE_API_VERSION = "2"
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
-  # We have a few box types.
+  # We have a few box types. The boxes, such as trust64 or
+  # thoughtbot_jessie, are created by prepare-for-tests.sh.
 
   # In general, we use a shell script to "provision" the box. just
   # sets up a different hostname than vagrant, because I find
@@ -37,16 +38,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   end
 
   config.vm.define "default", primary: true do |default|
-    # The default is the Ubuntu 14.04 ("trusty tahr") cloud image.
     default.vm.box = "trusty64"
-
-    # The url from which to fetch that base box.
-    default.vm.box_url = "https://cloud-images.ubuntu.com/vagrant/trusty/current/trusty-server-cloudimg-amd64-vagrant-disk1.box"
   end
 
   config.vm.define "jessie" do |jessie|
-    # This is the thoughtbot jessie image, converted to libvirt; see
-    # prepare-for-tests.sh.
     jessie.vm.box = "thoughtbot_jessie"
 
     jessie.vm.provision "shell",
--- a/installer-tests/Vagrantfile
+++ b/installer-tests/Vagrantfile
@@ -12,7 +12,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # "vagrant" to be a weird hostname.
   config.ssh.shell = "bash"
   config.vm.provision "shell",
-                       inline: "cd /vagrant && echo localhost > /etc/hostname && hostname localhost"
+                       inline: "echo localhost > /etc/hostname && hostname localhost"
 
 
   # Make sure to share the dir one *above* the current working directory. That way, we get
--- a/installer-tests/Vagrantfile
+++ b/installer-tests/Vagrantfile
@@ -65,6 +65,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     end
   end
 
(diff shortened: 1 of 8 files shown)


installer-tests/Vagrantfile
```diff
@@ -70,8 +79,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   end
 
   config.vm.define "jessie" do |jessie|
-    # The default is the Ubuntu 14.04 ("trusty tahr") cloud image.
-    jessie.vm.box = "thoughtbot/debian-jessie-64"
+    # This is the thoughtbot jessie image, converted to libvirt; see
+    # prepare-for-tests.sh.
+    jessie.vm.box = "thoughtbot_jessie"
```
zarvox: Consider using debian/jessie64, the more-official Debian base image.

installer-tests/Vagrantfile
```diff
@@ -60,6 +65,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     end
   end
 
+  # Use qemu driver for libvirt. See README.md for rationale.
+  config.vm.provider :libvirt do |libvirt|
+    libvirt.driver = 'qemu'
```
zarvox: Here would be a good place to adjust memory/CPU offers, since we seem to only do that for virtualbox above.

Feel free to steal the block from https://github.com/zarvox/vagrant-spk/blob/master/vagrant-spk#L60-98 , which computes cpus/ram to offer once, and then applies it to each provider.

installer-tests/prepare-for-tests.sh
```diff
@@ -0,0 +1,40 @@
...
+done
+
+# Check if Vagrant has the mutate plugin; if not, we install it, since
+# we use it during this script to convert a few Vagrant base boxes
+# into libvirt format.
+(vagrant plugin list | grep -q mutate) || vagrant plugin install vagrant-mutate
+
+# Download this particular random Debian Jessie VM and then convert it to
+# libvirt format.
+(vagrant box list | grep -q thoughtbot_jessie) || vagrant box add thoughtbot_jessie https://vagrantcloud.com/thoughtbot/boxes/debian-jessie-64/versions/0.1.0/providers/virtualbox.box
```
zarvox: Consider not pinning the version, so future runs can capture updates to the base image.

(also consider using debian/jessie64)

installer-tests/Vagrantfile
paulproteus: On the other hand, maybe this thing can live with a static amount of RAM and number of CPUs, since all we're testing is the install script!

installer-tests/Vagrantfile
paulproteus: +1, will look into it, but probably in a follow-up pull request.

installer-tests/prepare-for-tests.sh
paulproteus: There's a fascinating thing where I think `vagrant mutate` refuses to operate on any box with a `/` in the name. Which means I can't use `vagrant box add`'s integration with `vagrant cloud`, which means I have to specify the URL.

I'm open to other ideas and/or being shown some of the above statements are false!

paulproteus:
This should now actually be good enough! I'm briefly afk but will then see about installing the requirements on Jenkins.

paulproteus:
@jparyani let me know if you think the `Makefile` stuff is reasonable.

installer-tests/prepare-for-tests.sh
zarvox: I think the above statement is false!  Just two days ago I ran:

```
vagrant box add debian/jessie64
vagrant mutate debian/jessie64 libvirt
```
and then `vagrant box list` listed debian/jessie for both libvirt and virtualbox at 8.0.4

So I think that works just fine!

paulproteus:
```
➜  installer-tests git:(installer-tests-can-live-in-jenkins) ✗ vagrant box add debian/jessie64
==> box: Loading metadata for box 'debian/jessie64'
    box: URL: https://vagrantcloud.com/debian/jessie64
==> box: Adding box 'debian/jessie64' (v8.0.4) for provider: virtualbox
    box: Downloading: https://atlas.hashicorp.com/debian/boxes/jessie64/versions/8.0.4/providers/virtualbox.box
==> box: Successfully added box 'debian/jessie64' (v8.0.4) for 'virtualbox'!
➜  installer-tests git:(installer-tests-can-live-in-jenkins) ✗ vagrant mutate debian/jessie64 libvirt
Vagrant-mutate can not download from Vagrant Cloud for you.
```

for what it's worth, on my system!

PR #471, opened 2015-06-05
Add token templating

zarvox:
Grains can now make a request to the shell via postMessage() with a template
which can substitute a privileged API token for $API_TOKEN and the API host for
$API_HOST.  The grain will receive a postMessage reply from the shell with a
single-use URL in the shell's origin that will display the rendered template.
The app should iframe this URL in a place in the document inline where it wants
the template to appear.  Apps should *not* persist this URL anywhere.

Implementation detail: the privileged token will expire 5 minutes after
creation, if it is not used in that timeframe.  This helps avoid cluttering the
DB (and UI) with a bunch of tokens that were generated for this templating magic
that the user didn't end up using.

### DIFF
--- a/shell/public/tokentemplate.html
+++ b/shell/public/tokentemplate.html
@@ -1,0 +1,48 @@
+<!doctype html>
+<html>
+  <head>
+    <title>Sandstorm template</title>
+    <style>
+body {
+  margin: 0;
+  padding: 0;
+  border-collapse: collapse;
+  border-spacing: 0;
+  border: 0;
+}
+pre {
+  margin: 0;
+  padding: 0;
+}
+    </style>
+  </head>
+  <body>
+    <pre id="text"></pre>
+<script>
+// Templates are namespaced in localStorage by the prefix "template"
+var templateToken = "template" + window.location.hash.substring(1);
+var record = JSON.parse(localStorage.getItem(templateToken));
+console.log(record);
+document.getElementById("text").textContent = record.renderedTemplate;
+
+// Clean up localStorage templates that have expired.
+// First, collect expired template keys
+var i;
+var toDelete = [];
+for (i = 0; i < localStorage.length ; i++) {
+  var key = localStorage.key(i);
+  // we only care about templates
+  if (key.startsWith("template")) {
+    var data = localStorage.getItem(key);
+    if (data.expires < Date.now()) {
+        toDelete.push(key);
+    }
+  }
+}
+// Then, wipe expired keys.
+for (i = 0; i < toDelete.length; i++) {
+  localStorage.deleteItem(toDelete[i]);
+}
+</script>
+  </body>
+</html>
--- a/shell/server/hack-session.js
+++ b/shell/server/hack-session.js
@@ -251,8 +251,13 @@ Meteor.methods({
       allAccess: Match.Optional(null),
       roleId: Match.Optional(Match.Integer),
     });
-    check(destroyIfNotUsedByTime, Match.Optional(Number));
-    var selfDestructAt = destroyIfNotUsedByTime > 0 ? new Date(destroyIfNotUsedByTime) : null;
+    // Meteor bug #3877: we get null here instead of undefined when we
+    // explicitly pass in undefined.
+    if (destroyIfNotUsedByTime) {
+      check(destroyIfNotUsedByTime, Number);
+    }
+    var selfDestructAt = (destroyIfNotUsedByTime && (destroyIfNotUsedByTime > 0)) ?
+        new Date(destroyIfNotUsedByTime) : null;
 
     var grain = Grains.findOne(grainId);
     if (!grain) {
--- a/shell/server/hack-session.js
+++ b/shell/server/hack-session.js
@@ -237,9 +237,8 @@ HackSessionContextImpl.prototype.generateApiToken = function (petname, userInfo,
 };
 
 Meteor.methods({
-  newApiToken: function (grainId, petname, roleAssignment, forSharing) {
+  newApiToken: function (grainId, petname, roleAssignment, forSharing, destroyIfNotUsedByTime) {
     // Create a new user-oriented API token.
-
     if (!this.userId) {
       throw new Meteor.Error(403, "Must be logged in to create an API token.");
     }
@@ -252,6 +251,8 @@ Meteor.methods({
       allAccess: Match.Optional(null),
       roleId: Match.Optional(Match.Integer),
     });
+    check(destroyIfNotUsedByTime, Match.Optional(Number));
+    var selfDestructAt = destroyIfNotUsedByTime > 0 ? new Date(destroyIfNotUsedByTime) : null;
 
     var grain = Grains.findOne(grainId);
     if (!grain) {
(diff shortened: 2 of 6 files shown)


zarvox:
Failed due to [a bug in meteor](https://github.com/meteor/meteor/issues/3877).  Workaround is in testing.

paulproteus:
Good spotting.​

paulproteus:
Garply, retest this please

shell/public/tokentemplate.html
```diff
@@ -0,0 +1,48 @@
...
+  padding: 0;
+}
+    </style>
+  </head>
+  <body>
+    <pre id="text"></pre>
+<script>
+// Templates are namespaced in localStorage by the prefix "template"
+var templateToken = "template" + window.location.hash.substring(1);
+var record = JSON.parse(localStorage.getItem(templateToken));
```
kentonv: Use `sessionStorage` rather than `localStorage`? Then you don't really need to clean up expired templates.

shell/public/tokentemplate.html
```diff
@@ -0,0 +1,48 @@
...
+  margin: 0;
+  padding: 0;
+}
+    </style>
+  </head>
+  <body>
+    <pre id="text"></pre>
+<script>
+// Templates are namespaced in localStorage by the prefix "template"
+var templateToken = "template" + window.location.hash.substring(1);
```
kentonv: How about "offerTemplate", to be somewhat more specific? Filename could be `offer-template.html`.

shell/server/hack-session.js
```diff
@@ -270,12 +276,22 @@ Meteor.methods({
...
     return {token: token, endpointUrl: endpointUrl};
   }
 });
 
+var cleanupSelfDestructing = function() {
+  var now = new Date();
+  ApiTokens.remove({expiresIfUnused: {$lt: now}});
+};
+
+// Make self-destructing tokens actually self-destruct, so they don't
```
kentonv: Could we also not display the token in the webkey UI until it has had its first use?

shell/shared/grain.js
```diff
@@ -511,6 +511,60 @@ if (Meteor.isClient) {
             currentGrainId + event.data.setPath);
       } else if (event.data.setTitle) {
         Session.set("grainFrameTitle", event.data.setTitle);
+      } else if (event.data.renderTemplate) {
+        // Request creation of a single-use template with a privileged API token.
+        // Why?  Apps should not be able to obtain capabilities to themselves
```
kentonv: Specifically, capabilities-as-keys to themselves, that is capabilities that can be leaked through an arbitrary bit stream.

shell/shared/grain.js
```diff
@@ -511,6 +511,60 @@ if (Meteor.isClient) {
...
+          assignment = {roleId: call.roleId};
+        } else {
+          assignment = {none: null};
+        }
+        var petname = undefined;
+        if (call.petname) {
+          check(call.petname, String);
+          petname = call.petname;
+        } else {
+          petname = "template token";
```
kentonv: Users aren't likely to know what this means. Is there anything more understandable we could use? Maybe just "connected external app"?

shell/public/tokentemplate.html
```diff
@@ -0,0 +1,48 @@
...
+var record = JSON.parse(localStorage.getItem(templateToken));
+console.log(record);
```
kentonv: Remove debug logging?

shell/shared/grain.js
```diff
@@ -511,6 +511,60 @@ if (Meteor.isClient) {
...
+        }
+        // Tokens expire by default in 5 minutes from generation date
```
kentonv: Let's make sure to implement some sort of periodic ping from the template HTML that prevents the token from expiring while it's still open.

zarvox: Hmm, is that possible without making the template load all of Meteor's JS to get an authenticated connection?  At that point, does it make more sense to make the currently-static-HTML page become just another route on the meteor app?

zarvox:
Addressed all comments, except for the "unused offer token iframe sends keepalives while still visible" request, which I'm not sure offhand how to actually integrate in both a lightweight and safe fashion.  At present, the template is short and entirely self-contained; sending keepalives would require Meteor and a session and so on.

It should be possible to add the keepalives at a later date without breaking any of the current APIs or behaviors, though, so perhaps we could punt on that and revisit later?

kentonv:
I'm fine with not adding keepalives in this PR. But I think it will be easy. You don't need a Meteor context. Instead, we just need to expose a way to send an XHR to the API endpoint which refreshes the timeout to 5 minutes. There's no security concerns, since a similar XHR could just as easily count as the token's first use, keeping it alive permanently.

shell/shared/grain.js
```diff
@@ -511,6 +511,62 @@ if (Meteor.isClient) {
...
+            // Generate random key id2.
+            var id2 = Random.secret();
+            // Store apitoken id1 and template in local storage in the offer
+            // template namespace under key id2.
+            var key = "offerTemplate" + id2;
+            var renderedTemplate = template.replace("$API_TOKEN", tokenId)
+                                           .replace("$API_HOST", makeWildcardHost("api"));
+            sessionStorage.setItem(key, JSON.stringify({
+                "renderedTemplate": renderedTemplate,
+                "expiry": selfDestructTime
```
dwrensha: It looks like the code in offer-template.html expects an "expires" field, not an "expiry" field.

shell/shared/grain.js
```diff
@@ -511,6 +511,62 @@ if (Meteor.isClient) {
...
+                    selfDestructTime, function (error, result) {
+          if (error) {
+            //Session.set("api-token-" + currentGrainId, undefined);
+            window.alert("Failed to create token for offer template.\n" + error);
+            console.error(error.stack);
+          } else {
+            var tokenId = result.token;
+            // Generate random key id2.
+            var id2 = Random.secret();
+            // Store apitoken id1 and template in local storage in the offer
```
dwrensha: s/local storage/session storage/

shell/shared/grain.js
```diff
@@ -511,6 +511,62 @@ if (Meteor.isClient) {
...
+        // the platform, we can ensure that the token is only visible to the
+        // shell's origin.
+        var call = event.data.renderTemplate;
+        check(call, Object);
+        var rpcId = call.rpcId;
+        check(rpcId, String);
+        var template = call.template;
+        check(template, String);
+        var assignment = undefined;
+        if (call.roleId) {
```
dwrensha: Does this mean that these templates can't be used for `allAccess` tokens? Is there any reason to allow an arbitrary Grain.ViewSharingLink.RoleAssignment here?

shell/shared/grain.js
zarvox: Whoops, you're right.  Good thing this is strictly advisory for now!

shell/shared/grain.js
zarvox: I'm not sure I know the meaning of allAccess tokens yet - is that shorthand for "all permissions defined in this app's pkgdef.capnp"?

As for having a particular role rather than allAccess: in the case of Piwik, we'll want to use the token to allow the pages with tracking code to submit pageview events, but probably don't want to grant that token access to view config data, or actual site statistics.  So I think supporting some limits to the permissions granted is necessary.

shell/shared/grain.js
zarvox: another good catch

shell/shared/grain.js
dwrensha: Oops, sorry. I meant to ask "Is there any reason **not** to allow an arbitrary Grain.ViewSharingLink.RoleAssignment here?"

The allAccess role carries all permissions. This means that whoever uses the token will get exactly the permissions of the user who created the token. As discussed in https://github.com/sandstorm-io/sandstorm/pull/443, this is commonly the right thing.

PR #473, opened 2015-06-05
Add migrations, simplify login provider login in admin.js

### LINKED ISSUE #466: Google and Github login should not be enabled by default (since they're not configured)
Currently, awkwardly, when Sandstorm first runs, Google and Github login are enabled but not configured. If the admin tries to change anything in the admin settings without either disabling or configuring them, they get an error. (I think it might also be possible for a non-admin to configure the services at this time, though the right method calls?)

These should both default to disabled.

When no login services are enabled, the sign-in drop-down should display some sort of "configure server" button or link that goes to `/admin` (where the user will be instructed to generate an admin token).

zarvox:
Fixes #466.


shell/lib/server/migrations.js
```diff
@@ -0,0 +1,79 @@
...
+
+// This must come after all the functions named within are defined.
+// Only append to this list!  Do not modify or remove list entries;
+// doing so is likely change the meaning and semantics of user databases.
+var MIGRATIONS = [
+  updateLoginStyleToRedirect,
+  enableLegacyOAuthProvidersIfNotInSettings
+];
+
+migrateToLatest = function () {
```
kentonv: Currently you're declaring this function globally, but I don't think it needs to be. `function migrateToLatest() {` will make it local.

zarvox:
Done.

PR #480, opened 2015-06-06
Check for kernel version before user namespaces

### LINKED ISSUE #478: Installer: check kernel version before checking userns
We received a report that the installer complained about being unable to create user namespaces, but it turned out the user in question was using kernel version 3.10. Our installer is supposed to verify that the user is on 3.13 or newer, but I guess that check is occurring after the userns check? We should probably do the version check first, since it's simpler and more informative.

paulproteus:
This way, if someone has a Linux distribution that is too old to
run Sandstorm, we give them relevant advice.

Close #478

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1542,13 +1542,13 @@ sandcats_generate_keys() {
 
 # Now that the steps exist as functions, run them in an order that
 # would result in a working install.
+assert_usable_kernel
 detect_current_uid
 detect_userns_clone
 assert_userns_clone_works_or_can_be_made_to_work
 handle_args "$@"
 assert_on_terminal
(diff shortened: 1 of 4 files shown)


paulproteus:
I'm running the installer tests locally (on my laptop) now.

paulproteus:
I'm going to merge this as-is, since supposedly I own `install.sh` and `installer-tests/`.

paulproteus:
@kentonv can you update `install.sandstorm.io` with the latest `install.sh`?

I should have tagged you on this when I merged it.

kentonv:
Done

paulproteus:
Yay thanks!

PR #484, opened 2015-06-09
Fix up ownership of notification sturdyrefs.

dwrensha:


kentonv:
When the frontend restores a token for itself, I think it should check that it is in fact the owner of the token. The purpose would be to detect confused-deputy bugs, where someone convinces the frontend to open a token it wasn't supposed to. It looks like that isn't currently implemented in this change?

dwrensha:
Ok, I've pushed a commit that checks the owner in the `frontend` case as well. Additionally, it checks the owner for the `drop()` method.

dwrensha:
Good catch, garply. Fixed now.

PR #486, opened 2015-06-10
Add documentation illustrating the depedencies among the fields of ApiTokens.

dwrensha:


shell/lib/db.js
```diff
@@ -236,6 +236,30 @@ ApiTokens = new Mongo.Collection("apiTokens");
...
+//      union {
+//        uiView: group {
+//          userId: Text;
+//          roleAssignment: RoleAssignment;
+//          forSharing: Bool;
+//        }
+//        objectId: SupervisorObjectId;
+//      }
+//    }
+//    frontendRef: union {
```
kentonv: nitpick: It looks like liens from `owner` to here are missing a space of intent? Also the positioning of the colons is inconsistent.

dwrensha:
Done.

PR #494, opened 2015-06-11
spk pack: only use mmap for files >=128k in size

zarvox:
Too many mmaps breaks on VirtualBox shared folders, where VirtualBox has a
hardcoded 4096-file limit.  Keeping every file mapped while building the
Archive means we eventually get an ENFILE when opening the next file.

While most apps probably don't actually need >4096 files, this limitation makes
life harder for vagrant-spk.

Since most files are small, and most data is in large files, we can get most of
the RAM-saving benefit of mmap()ing files by doing so for files above some
threshold size, and reading the rest greedily into RAM, so we do that.

### DIFF
--- a/src/sandstorm/spk.c++
+++ b/src/sandstorm/spk.c++
@@ -1037,37 +1037,54 @@ private:
       if (S_ISREG(stats.st_mode)) {
         KJ_ASSERT(children.empty(), "got file, expected directory", target);
 
-        mapping = MemoryMapping(raiiOpen(target, O_RDONLY), target);
+        kj::AutoCloseFd fd = raiiOpen(target, O_RDONLY);
+        size_t size = getFileSize(fd, target);
 
-        if (mapping.size() >= (1ull << 29)) {
+        if (size >= (1ull << 29)) {
           context.exitError(kj::str(target, ": file too large. The spk format currently only "
             "supports files up to 512MB in size. Please let the Sandstorm developers know "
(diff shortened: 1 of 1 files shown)


zarvox:
@JamborJan This patchset, once merged and released, should remove the need for NFS mounts when using vagrant-spk.

JamborJan:
FYI: having fixed this https://github.com/JamborJan/paperwork/issues/12 allows me to undo this https://github.com/JamborJan/paperwork/issues/10 without the improvements made here. 

I'm waiting now for this issue to be solved / committed then I can test again the new ```vagrant-spk```.

src/sandstorm/spk.c++
```diff
@@ -1037,37 +1037,54 @@ private:
...
+          // Small file; direct read preferable.
+          ::capnp::Data::Builder buf = nullptr;
+          if (stats.st_mode & S_IXUSR) {
+            warnIfMongod(context);
+            buf = builder.initExecutable(size);
+          } else {
+            buf = builder.initRegular(size);
+          }
+          kj::FdInputStream stream(kj::mv(fd));
+          size_t readSize = stream.tryRead(buf.begin(), size, size);
```
kentonv: Use `read(buf.begin(), size)`, which throws an exception if it fails to read the full size. Then you don't need the subsequent `KJ_REQUIRE`.

zarvox: Done, thanks.

PR #496, opened 2015-06-12
allow users to open a shared grain in incognito mode

dwrensha:


shell/shared/grain.js
```diff
@@ -71,6 +73,20 @@ if (Meteor.isServer) {
            ];
   });
 
+  Meteor.publish("tokenInfo", function (token) {
+    // Allows the client side to map a raw token to its entry in ApiTokens.
+
+    var hashedToken = Crypto.createHash("sha256").update(token).digest("base64");
+    var apiToken = ApiTokens.findOne({_id: hashedToken});
```
kentonv: We need to restrict what fields are revealed here. For example, `objectId` may contain secrets private to a grain which the token bearer is not intended to be able to see.

jparyani:
Garply, retest this please

jparyani:
Garply, retest this please

PR #502, opened 2015-06-15
Deps -> Tracker

dwrensha:
Deps has been deprecated since https://github.com/meteor/meteor/commit/8e7cd297bf5addd96ffba9b5309e471ffb8fe79a.


kentonv:
I guess I was confused because docs.meteor.com does not document `Tracker` as having any members except `autorun`, whereas http://manual.meteor.com/#tracker has code samples that use `Deps` and not `Tracker`...

PR #503, opened 2015-06-15
Add Powerbox copy/paste functionality

jparyani:


shell/server/proxy.js
```diff
@@ -249,6 +249,10 @@ function validateApiToken (apiToken) {
       ApiTokens.update(apiToken._id, {$set: {expiresIfUnused: null}});
     }
   }
+
+  if (apiToken.objectId) {
```
dwrensha: `"frontendRef" in apiToken` is another case when it's not a UiView.

tests/apps/powerbox.js
```diff
@@ -0,0 +1,60 @@
...
+    medium_wait = utils.medium_wait,
+    long_wait = utils.long_wait,
+    very_long_wait = utils.very_long_wait;
+
+module.exports = {};
+
+module.exports["Install Powerbox"] = function (browser) {
+  browser
+    .init()
+    .installApp("http://sandstorm.io/apps/jparyani/powerbox-1.spk", "b0704d58e11f67494b6b203dff5cc8b6")
```
dwrensha: Do you post the source code of your test apps anywhere? When I was working with the backgrounding stuff there were a few times when it would have been helpful to see the source code.

paulproteus: ​Along those lines, I would be very +1 on a section of the app list,
perhaps grayed-out by default, that is labeled, "Developer testing apps". I
think we should list these publicly.

jparyani: By convention, I've been pointing to the source code of the package a test relies on in a comment at the top of the file (in this case it's https://github.com/jparyani/sandstorm-test-app/tree/powerbox).

We obviously need a better solution here, but I'm not sure what.

shell/server/core.js
```diff
@@ -80,6 +80,10 @@ function hashSturdyRef(sturdyRef) {
   return Crypto.createHash("sha256").update(sturdyRef).digest("base64");
 }
 
+function generateSturdyRef() {
+  return Random.secret();
```
kentonv: Note that `Random.secret()` uses base64 which means it contains characters that are problematic for some contexts. I think Random.id(n) with a sufficiently-large n is what we want to use here.

shell/server/hack-session.js
```diff
@@ -34,12 +36,78 @@ var Url = Npm.require("url");
...
+    try {
+      cap = waitPromise(restoreInternal(token,
+                                        Match.Optional({webkey: Match.Optional(Match.Any)}))).cap;
+    } catch (err) {
+      throw new Meteor.Error(500, err.toString());
+    }
+    var castedCap = cap.castAs(SystemPersistent);
+    var save = castedCap.save({grain: {
+      grainId: grainId,
+      saveLabel: {
```
kentonv: This label should be specified by the app, which I guess means it needs to be a parameter to finishPowerboxRequest().

shell/shared/grain.js
```diff
@@ -357,7 +363,37 @@ if (Meteor.isClient) {
...
+      Session.set("show-powerbox-request", false);
+      powerboxRequestInfo.source.postMessage(
+        {
+          rpcId: powerboxRequestInfo.rpcId,
+          error: "User cancelled request"
+        }, powerboxRequestInfo.origin);
+      powerboxRequestInfo = null;
+    },
+    "click #powerbox-offer-popup-closer": function (event) {
+      // TODO(soon): handle errors and what not
```
kentonv: What errors?

jparyani: Errors from Meteor.call("finishPowerboxOffer"). I'll fix it.

src/sandstorm/grain.capnp
```diff
@@ -421,9 +421,19 @@ interface SessionContext {
   #
   # The capability is implicitly tied to the user as if via `tieToUser()`.
 
-  request @3 (query :PowerboxQuery) -> (cap :PowerboxCapability);
-  # Searches for capabilities in the user's powerbox matching the given query and displays a
-  # selection UI to the user.  Returns the selected capability.
+  # request is no longer an rpc call on the session. Instead you use the postMessage api to get a
```
kentonv: Let's leave the method defined (since it could make sense at some point), but extend the comment to specify the postMessage approach.

src/sandstorm/supervisor.c++
```diff
@@ -1541,6 +1541,35 @@ void SupervisorMain::DefaultSystemConnector::checkIfAlreadyRunning() const {
...
+      req.getRef().setAppRef(response.getObjectId());
+      req.setOwner(owner);
+      return req.send().then([context](auto response) mutable {
+        context.getResults().setSturdyRef(response.getToken());
+      });
+     });
+  }
+
+  kj::Promise<void> export_(ExportContext context) override {
+    KJ_FAIL_REQUIRE("Cannot call save() on capabilities across realms. Use SandstormApi.save().");
```
kentonv: Text: "Cannot directly call save() on capabilities outside the grain. Use SandstormApi.save() instead."

src/sandstorm/supervisor.c++
```diff
@@ -1541,6 +1541,35 @@ void SupervisorMain::DefaultSystemConnector::checkIfAlreadyRunning() const {
...
+typedef capnp::RealmGateway<capnp::Data, capnp::AnyPointer, ApiTokenOwner,
+                            capnp::AnyPointer> SupervisorRealmGateway;
+
+class SupervisorRealmGatewayImpl final: public SupervisorRealmGateway::Server {
+public:
+  explicit SupervisorRealmGatewayImpl(SandstormCore::Client&& sandstormCore)
+    : sandstormCore(kj::mv(sandstormCore)) {}
+
+  kj::Promise<void> import(ImportContext context) override {
+    auto cap = context.getParams().getCap().castAs<AppPersistent<> >();
```
kentonv: FWIW, the space between '>'s is not necessary in C++11.

src/sandstorm/supervisor.c++
```diff
@@ -1541,6 +1541,35 @@ void SupervisorMain::DefaultSystemConnector::checkIfAlreadyRunning() const {
...
+    auto owner = context.getParams().getParams().getSealFor();
+    return cap.saveRequest().send()
+        .then([this, owner, context](auto response) mutable {
+      auto req = sandstormCore.makeTokenRequest();
+      req.getRef().setAppRef(response.getObjectId());
+      req.setOwner(owner);
+      return req.send().then([context](auto response) mutable {
+        context.getResults().setSturdyRef(response.getToken());
+      });
+     });
```
kentonv: nitpick: indentation

src/sandstorm/supervisor.c++
```diff
@@ -1824,11 +1859,20 @@ public:
...
-      return kj::READY_NOW;
-    } else {
-      KJ_FAIL_REQUIRE("Supervisor can only restore wakelocks for now.");
+    switch (objectId.which()) {
+      case SupervisorObjectId<>::WAKE_LOCK_NOTIFICATION: {
+        context.getResults().setCap(wakelockSet.restore(objectId.getWakeLockNotification()));
+        return kj::READY_NOW;
+      }
+      case SupervisorObjectId<>::APP_REF: {
+        auto req = mainView.castAs<MainView<>>().restoreRequest();
```
kentonv: I think we actually want `mainView` to be of type `MainView<>::Client` all along rather than `castAs` here.

This brings up another issue, which is that technically the supervisor is supposed to filter requests to this capability such that `MainView`'s methods are not visible to outsiders (but `UiView`'s methods are).

jparyani:
Done

.gitignore
```diff
@@ -27,3 +27,8 @@ bundle/
 .vagrant/
 phantomjsdriver.log
 selenium-debug.log
+src/sandstorm/*.capnp.c++
+src/sandstorm/*.capnp.h
```
dwrensha: Why are these needed? Don't the capnp build artifacts get put in tmp/sandstorm?

shell/server/core.js
```diff
@@ -112,33 +116,35 @@ NotificationHandle.prototype.close = function () {
...
+restoreInternal = function (sturdyRef, ownerPattern, requiredPermissions) {
+  // Restores `sturdyRef`, checking first that its owner matches `ownerPattern`.
   var hashedSturdyRef = hashSturdyRef(sturdyRef);
   var token = ApiTokens.findOne(hashedSturdyRef);
   if (!token) {
     throw new Error("No token found to restore");
   }
   check(token.owner, ownerPattern);
-
+  if (requiredPermissions && !_.isEqual(token.requiredPermissions, requiredPermissions)) {
```
dwrensha: There is no `requiredPermissions` field in `ApiTokens`.

shell/server/core.js
```diff
@@ -112,33 +116,35 @@ NotificationHandle.prototype.close = function () {
...
   });
 };
 
-function restoreInternal(sturdyRef, ownerPattern) {
-  // Restores `sturdyRef`, checking first that its owner matches `ownerPattern`.
+NotificationHandle.prototype.save = function (params) {
+  return saveFrontendRef({notificationHandle: this.notificationId}, params.sealFor);
+};
 
+restoreInternal = function (sturdyRef, ownerPattern, requiredPermissions) {
```
dwrensha: I think you can get away without the `requiredPermissions` parameter here. To check the permissions for a `grain.introducerUser` owner, you can use [`Match.Where`](http://docs.meteor.com/#/full/matchpatterns). The other permissions you'll need to check will be in the `token.requirements` list; they won't depend on any parameters to this function.

shell/server/hack-session.js
```diff
@@ -34,12 +36,84 @@ var Url = Npm.require("url");
...
+SessionContextImpl.prototype.offer = function (cap, requiredPermissions) {
+  var self = this;
+  return inMeteor((function () {
+    var castedCap = cap.castAs(SystemPersistent);
+    var save = castedCap.save();
+    var sturdyRef = waitPromise(save).sturdyRef;
+
+    // TODO(someday): This actually needs to be implemented with the help of the supervisor, and we
+    // should be storing a reference to the parent as well as the requiredPermissions
+    ApiTokens.update({_id: hashSturdyRef(sturdyRef)}, {$set: {requiredPermissions: requiredPermissions}});
```
dwrensha: The required permissions should go in `token.requirements`.

shell/server/hack-session.js
```diff
@@ -34,12 +36,84 @@ var Url = Npm.require("url");
...
+
+Meteor.methods({
+  finishPowerboxRequest: function (webkeyUrl, saveLabel, grainId) {
+    check(webkeyUrl, String);
+    check(grainId, String);
+
+    var parsedWebkey = Url.parse(webkeyUrl.trim());
+    if (parsedWebkey.host !== makeWildcardHost("api")) {
+      console.log(parsedWebkey.hostname, makeWildcardHost("api"));
+      throw new Meteor.Error(500, "Hostname does not match this server. External webkeys are not " +
```
dwrensha: I think a 400 would be more appropriate. The user can paste anything in, right?

shell/server/hack-session.js
```diff
@@ -34,12 +36,84 @@ var Url = Npm.require("url");
...
+        "supported (yet)");
+    }
+
+    var token = parsedWebkey.hash;
+    if (!token) {
+      throw new Meteor.Error(500, "Invalid webkey. You must pass in the full webkey, " +
```
dwrensha: Again, I think a 400 would be more appropriate.

shell/server/hack-session.js
```diff
@@ -34,12 +36,84 @@ var Url = Npm.require("url");
...
+        "including domain name and hash fragment");
+    }
+    token = token.slice(1);
+
+    var cap;
+    try {
+      cap = waitPromise(restoreInternal(token,
+                                        Match.Optional({webkey: Match.Optional(Match.Any)}))).cap;
+    } catch (err) {
```
dwrensha: Hm. Maybe `restoreInternal` should use `Match.test` instead of `check`. Would that let you omit this `catch`?

shell/server/hack-session.js
```diff
@@ -34,12 +36,84 @@ var Url = Npm.require("url");
...
+    var castedCap = cap.castAs(SystemPersistent);
+    var save = castedCap.save({grain: {
+      grainId: grainId,
+      saveLabel: {
+        defaultText: saveLabel
+      }
+    }});
+    var sturdyRef = waitPromise(save).sturdyRef;
+    // TODO(someday): Same as with `offer`, this should eventually be part of the supervisor
+    ApiTokens.update({_id: hashSturdyRef(sturdyRef)}, {$set: {parentToken: hashSturdyRef(token)}});
```
dwrensha: This is an abuse of the `parentToken` field. Since you've already saved the capability, the entry in `ApiTokens` already has a provider. You should either manually save the token with `parentToken` as the provider, or you should not set `parentToken` and wait until membranes are implemented (and `save()` will appropriately set the provider).

dwrensha: Hm. I suppose that getting `parentToken` to be set automatically upon `save` doesn't need membranes per se. I think a non-membrane wrapper would do.

In any case, I don't think we actually want a `parentToken` cap here. `parentToken` is useful in the case where the sharer might modify something about the shared token, e.g. the granted permissions of a UiView. Here, it suffices to add the original token as a `tokenValid` requirement to the `requirements` list.

.gitignore
jparyani: Hmm that's weird, they must've ended up there for me from some other `capnp compile` gone rogue.

shell/server/core.js
jparyani: Can you take a look at e7660d3 and see if that answers your question? requiredPermissions is being passed in to set a constraint, and seems necessary based on how we designed `restore`

dwrensha: I don't think that the MembraneRequirement represented by `requiredPermissions` should be added to the existing token. I think that `SandstormCore.restore(token, requiredPermissions)` should call `restoreInternal()` and then call `SystemPersistent.addRequirements()` on the result. If the restored capability eventually gets re-saved, then that's when the new MembraneRequirement should get persisted.

shell/server/core.js
```diff
@@ -205,13 +247,20 @@ SandstormCoreImpl.prototype.drop = function (sturdyRef) {
 SandstormCoreImpl.prototype.makeToken = function (ref, owner) {
   var self = this;
   return inMeteor(function () {
-    var sturdyRef = new Buffer(Random.id(20));
+    if (owner.grain && !owner.grain.introducerUser) {
+      // In cases where the caller doesn't know the introducerUser to save under (eg. in the case of
```
dwrensha: Would there be anything wrong with leaving `introducerUser` empty here?

jparyani: Actually after just discussing with Kenton, we do want to allow `introducerUser` to be empty. In the case where it is empty, `requiredPermissions` will be ignored.

jparyani:
This is ready for review.

shell/server/core.js
```diff
@@ -112,53 +116,121 @@ NotificationHandle.prototype.close = function () {
...
+    var requirement = requirements[i];
+    if (requirement.tokenValid) {
+      var token = ApiTokens.findOne({_id: requirement.tokenValid}, {fields: {permissions: 1}});
+      if (!checkRequirements(token.permissions)) {
+        return false;
+      }
+    } else if (requirement.permissionsHeld) {
+      var p = requirement.permissionsHeld;
+      var viewInfo = Grains.findOne(p.grainId, {fields: {cachedViewInfo: 1}}).cachedViewInfo;
+      var set = new PermissionSet(grainPermissions(p.grainId, p.userId, viewInfo));
```
dwrensha: This will throw an exception if `viewInfo` is `null` or `undefined`. You should either pass in `{}` in such cases or change `grainPermissions()` so that it is more forgiving.

.gitignore
kentonv: Yes, can you please remove these? They shouldn't be needed and could cause confusion in the future.

jparyani:
Alright, it's ready for review now. All the commits since last review have been squashed to 5bccc71

jparyani:
Done

paulproteus:
: D

PR #506, opened 2015-06-16
Get rid of extraneous junk in icons.

dwrensha:
When I build Sandstorm on Ubuntu 15.04, I get icons that look like this:

icons

This patch fixes it.

Amusingly, the 'display="none"' elements that I'm deleting are entire other icons, including a boat and a knife & fork.


ocdtrekkie:
This is a cool bug. I think it's fun. I guess it explains why some of these SVGs were so large for what they are.

zarvox:
so much +1

kentonv:
... wat.

PR #514, opened 2015-06-18
Add shell/lib to the dependencies of the shell-build target.

dwrensha:
Right now, if you edit migrations.js and then do `make update`, your changes won't get applied because the Makefile thinks that nothing has changed. This fixes the problem.


zarvox:
+1

If we need to rebuild for anything changing under `shell` recursively, perhaps we should should shell out to `find`?

I guess we still have the usual `make` bug of not rebuilding if a file is deleted, since it no longer is included in the glob.

PR #516, opened 2015-06-18
Merge RoleAssignments into ApiTokens.

dwrensha:
We're also going to need to update a bunch of code for this to work (particularly in permissions.js). I'm submitting this as a request for comments about the big picture.

The migration translates each RoleAssignment into a direct edge from the attenuated UiView to the user. They essentially become "shares by identity". In contrast, new ApiTokens created by redeeming webkeys will always have the `parentToken` field set.

I think we could go even farther and merge Grains into ApiTokens. That would entail adding a new "raw UiView" provider in addition to the existing "attenuated UiView" provider.

### DIFF
--- a/shell/lib/db.js
+++ b/shell/lib/db.js
@@ -91,32 +91,9 @@ Grains = new Mongo.Collection("grains");
 //       web publishing. This field is initialized when first requested by the app.
 
 RoleAssignments = new Mongo.Collection("roleAssignments");
-// Edges in the permissions sharing graph.
-//
-// To share a permission with another user is to declare "if I have this permission, then so should
-// the recipient". A bundle of such declarations by a single sharer directed at a single recipient
-// for a single grain is called a "role assignment". A grain's owner always has every permission,
-// but permissions for other users must be computed from this collection.
-//
-// Any grain for which a user has received a role assignment should show up in that user's grain
-// list. However, the user may only access a grain if there is a path of *active* role assignments
-// leading from the grain owner to that user, precisely as if every role assignment carried a
-// special "can open grain" permission.
-//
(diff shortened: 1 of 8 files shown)


kentonv:
I like this. I would also like to see ApiTokens reworked into a new, cleaner structure, but I can live with that being a separate migration I suppose.

I may be warming to the idea of merging grains in as well. Basically the table becomes "the capability graph"... Arguably backup tokens could join, even.

Long-term, particularly when we think about moving away from Mongo, it would be nice to think about how this database can be broken up into small, manageable collections rather than one world-spanning collection. You could almost do it by owner -- logically, any code that is trying to restore a capability knows its owner in advance, and it would actually be more secure if it first looked up the owner's collection and then looked up the capability within. The one problem is auditing: we need some way to find all incoming references in order to revoke them. But maybe that just means each edge has entries in two collections, one forward and one backwards.

(Probably I should read up on graph databases...)

dwrensha:
This is ready for review.

shell/lib/db.js
```diff
@@ -280,7 +259,13 @@ ApiTokens = new Mongo.Collection("apiTokens");
...
 //     child :group {
 //       parentToken :Text;
-//       roleAssignment :Maybe(RoleAssignment);
+//       union {
+//         uiView :group {
+//           grainId :Text;
+//           userId :Text;
+//           roleAssignment :RoleAssignment = (allAccess = ());
+//         }
+//         other :Void;
```
kentonv: missing a '}' here

kentonv: Eh I'll just merge and fix this myself. :)

PR #521, opened 2015-06-19
Allow extra paths to passthrough to apps for `shared` links

jparyani:


shell/shared/grain.js
```diff
@@ -753,7 +753,7 @@ Router.map(function () {
     }
   });
 
-  this.route("/shared/:key", {
+  this.route("/shared/:key/:path(.*)?", {
```
zarvox: Does this make the trailing slash after `:key` required?

jparyani: Nope, it works the same as the grain route (look back ~20 lines)

jparyani:
Added a fix for a bug that was causing the page to refresh upon data() recalc (which happens every time we send a keep-alive).

PR #522, opened 2015-06-20
Add two nudge arrows for demo users

paulproteus:
For demo users with no grains, point an arrow at "Install apps" and
"New App Instance" if they are looking at an app.

This visually conflicts with the information about the managed hosting
and mailing list signup, so we hide that information until the demo user
has created a grain.

Requesting @kentonv 's review + merge.


paulproteus:
OK, now with the indentation fixes, this is _actually_ ready for review + merge.

kentonv:
I think I'm going to modify this to apply to *all* users, because why not?

paulproteus:
Interesting. I think that's a good idea.

Issue #524, opened 2015-06-20
'make' from scratch fails


yuvipanda:
/bin/sh: 1: cannot create tmp/changelog.html: Directory nonexistent  
  Makefile:185: recipe for target 'shell/client/changelog.html' failed

Creating the 'tmp' directory fixes this

paulproteus:
Closed by #526 being merged in.

paulproteus:
Thanks to you for reporting this bug, @yuvipanda !

PR #526, opened 2015-06-20
Ensure tmp/ exists before generating changelog in it

paulproteus:
This is the smallest possible fix for #524.

Other imaginable fixes include:

* Depending on a dir called `tmp/` which has its own `Makefile` target.
* Adding `tmp/.gitkeep` to the repository so there is always a `tmp/` directory.

I think this is an OK fix. If others have opinions, I'm interested in them, although I'm much more interested in them if the opinions come with their own pull request that supercede this one. (But if people think I should adjust this one, OK.)

Also perhaps @jparyani we should be `git clean -d -x -f -f`-ing from the tests to catch problems like this?


kentonv:
I think this is the correct fix.

PR #535, opened 2015-06-22
Fix link to packaging tutorial

paulproteus:
Thanks to @sunu for reporting this on IRC.


kentonv:
FWIW in the future for things like this where something is broken and the fix is obvious, feel free to submit directly rather than wait for my review. :)

paulproteus:
Heh, okay (-:​

PR #538, opened 2015-06-24
Pull MembraneRequirements out into its own collection.

dwrensha:
When computing permissions and token validity, I foresee often wanting to look up which tokens are impacted by a given modification. (E.g. Alice's permissions on grain X have just changed -- which tokens does that cause to become invalid?) The current schema would not make this easy.

With this patch, such lookups become easy, and we can later add a [mongo index](http://docs.mongodb.org/v2.6/core/index-single/#indexes-on-embedded-fields) that will make them fast.

### DIFF
--- a/shell/lib/db.js
+++ b/shell/lib/db.js
@@ -275,7 +275,7 @@ ApiTokens = new Mongo.Collection("apiTokens");
 //   ...
 // }
 
-MembraneRequirements = new Mongo.collection("membraneRequirements");
+MembraneRequirements = new Mongo.Collection("membraneRequirements");
 // For each sturdy ref, this collection lists conditions that must hold for it to be considered
 // valid. Semantically, a sturdy ref's requirements specify the powers which were used to originally
 // create the capability. If any condition in the requirements becomes untrue, then the capability
--- a/shell/lib/db.js
+++ b/shell/lib/db.js
@@ -246,12 +246,6 @@ ApiTokens = new Mongo.Collection("apiTokens");
 //   expiresIfUnused:
 //              Optional Date after which the token, if it has not been used yet, expires.
 //              This field should be cleared on a token's first use.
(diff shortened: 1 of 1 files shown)


dwrensha:
Terminological note: I'm preferring "sturdy ref" over "token" because I think that in the nearish future we are going to replace the `ApiTokens` collection with a `SturdyRefs` collection.

shell/lib/db.js
```diff
@@ -278,10 +272,25 @@ ApiTokens = new Mongo.Collection("apiTokens");
 //       roleAssignment :Maybe(RoleAssignment);
 //     }
 //   }
-//   requirements: List(Supervisor.MembraneRequirement);
 //   ...
 // }
 
+MembraneRequirements = new Mongo.collection("membraneRequirements");
```
jparyani: This should Mongo.Collection with a capital 'C'

jparyani:
I'm not sure this is the right thing to do. Normalizing data (ie. moving fields out of monolithic collections into their own collection), is a thing that doesn't make a lot of sense in Mongo. It comes with the large downside of having to do a double query if we want to access the ApiToken at the same times as its MembraneRequirement.

The upside that it's easier to query MembraneRequirements is actually not entirely true. You can query into arrays just as easily as normal fields. For example, in your case of trying to find all Requirements with a given `tokenValid` field:

```
ApiTokens.find({requirements: {tokenValid: "MY_TOKEN"}})
```
Will return all ApiTokens that have a requirement that matches (see http://docs.mongodb.org/manual/reference/operator/query/elemMatch, specifically the note where it says you can leave out $elemMatch if there's only 1 query). Similarly, we can add indices to fields inside an array.

dwrensha:
Indeed, and the indexing would work with a [multikey index](http://docs.mongodb.org/v2.6/core/index-multikey/).

Closing because it looks like this change doesn't actually buy us anything.

PR #540, opened 2015-06-25
Add opt-in passthrough of remote address to ApiSession as X-Real-IP

zarvox:
While in most cases, we want to avoid exposing a client's IP address to
sandboxed apps, having the option of sharing it is acutely valuable for
analytics apps like Piwik, which benefit substantially from having external IP
addresses for geolocation.

If an authenticated client sends the API endpoint a header:

    X-Sandstorm-Passthrough: address

then we will pass the client's observed IP address as a parameter in the
ApiSession, and sandstorm-http-bridge will pass that to the app's backend as
an X-Real-IP header.

Additionally, if the request to the API endpoint is made from an RFC1918
address, we will substitute the value of an X-Real-IP header, if present, for
the client's address.  This should make things work out-of-the-box for most
simple SSL termination setups, while preventing spoofing of the address seen by
apps by anyone on a separate network from the sandstorm server.

We only allow this for API sessions because doing so for web sessions would
allow a malicious app to serve JS that makes requests back to the app with the
opt-in header, leaking the user's real IP without first requiring user action.
That would break our sandboxing guarantees.

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -23,6 +23,7 @@ var Http = Npm.require("http");
 var Url = Npm.require("url");
 var Promise = Npm.require("es6-promise").Promise;
 var Capnp = Npm.require("capnp");
+var Net = Npm.require("net");
 
 var ByteStream = Capnp.importSystem("sandstorm/util.capnp").ByteStream;
 var ApiSession = Capnp.importSystem("sandstorm/api-session.capnp").ApiSession;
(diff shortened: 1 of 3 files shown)


src/sandstorm/api-session.capnp
```diff
@@ -19,12 +19,18 @@
...
 
 interface ApiSession @0xc879e379c625cdc7 extends(WebSession.WebSession) {
   # A special case of WebSession but for APIs. It doesn't provide much other
   # than a unique type id that we can identify ApiSessions with
 
   struct Params {
-    # Currently there are no params for api sessions
+    # Normally, we strip the remote address from requests, since most applications shouldn't need
+    # it.  However, for those that benefit from it (like analytics), clients can opt into passing
+    # their IP on to the backend by adding an "X-Sandstorm-Pass-IP: yes" header to their request.
```
kentonv: Comment is outdated?

shell/server/proxy.js
```diff
@@ -936,13 +937,69 @@ Proxy.prototype._callNewWebSession = function (request, userInfo) {
...
 
+var isRfc1918OrLocal = function(address) {
+  if (Net.isIPv4(address)) {
+    quad = address.split(".").map(function(x) { return parseInt(x, 10); });
+    return (quad[0] === 127 || quad[0] === 10 ||
+            (quad[0] === 192 && quad[1] === 168) ||
+            (quad[0] === 172 && quad[1] >= 16 && quad[1] < 32));
+  } else if (Net.isIPv6(address)) {
+    // IPv6 specifies ::1 as localhost and fd:: as reserved for private networks
+    return (address === "::1" || address.indexOf("fd") === 0);
```
kentonv: You can avoid redundantly scanning through the whole string looking for "fd" by writing it like:

    address.lastIndexOf("fd", 0) === 0

This is a common idiom, actually.

shell/server/proxy.js
```diff
@@ -936,13 +937,69 @@ Proxy.prototype._callNewWebSession = function (request, userInfo) {
...
+      var addressToPass = request.socket.remoteAddress;
+      if (isRfc1918OrLocal(addressToPass) && "x-real-ip" in request.headers) {
+        // Allow overriding the socket's remote address with X-Real-IP header if the request comes
+        // from either localhost or an RFC1918 address.  These are not useful for geolocation
+        // anyway.
+        addressToPass = request.headers["x-real-ip"];
+      }
+      if (Net.isIPv4(addressToPass)) {
+        // Map IPv4 addresses in IPv6
+        var v4Int = 0xFFFF00000000 + addressToPass.split(".")
```
kentonv: Maybe comment on how this comes out to 48 bits which manages to be precisely-representable in a double, hence you don't need to use bignum here?

shell/server/proxy.js
zarvox: Ah, searching backwards starting at 0 (and ending at 0) means testing precisely what we want.

Cool!  Today I learned.

shell/server/proxy.js
zarvox: Done

src/sandstorm/api-session.capnp
zarvox: Thanks, fixed.

src/sandstorm/sandstorm-http-bridge.c++
```diff
@@ -1324,6 +1337,47 @@ private:
     }
     return kj::strArray(permissionVec, ",");
   }
+  inline kj::String addressToString(::sandstorm::IpAddress::Reader&& address) {
```
zarvox: Done.

zarvox:
Comments addressed.

PR #546, opened 2015-06-30
IpInterface/IpNetwork implementation

jparyani:
This adds a page to the admin UI where you can offer these capabilities.


kentonv:
Hmm, so, the one problem I have with this is the approach to accountability. It looks like you're tracking the originating user ID in order to record in the table which admin created a capability. When a token is restored and then re-saved, the new token is not associated with the original except for getting the same user ID.

I think instead restore/re-save should create a child token, to be consistent with our model elsewhere.

It feels weird to indicate the creating user under `frontendRef`, which is meant to identify the object. Every type of frontendRef that is created through the UI will want to track the creating user, so putting it in a different place for each seems wrong. I can think of a few options here:

1. Use `ApiToken.userId`. I think, though, that we decided to move away from such overloading.
2. Have a new field `ApiToken.creatorUserId`, which is explicitly documented as only being for informational purposes -- that is, you might display this user in a table to show who created the capability, but there is no permission enforcement done based on this value.
3. If `ApiToken.requirements` contained a requirement that the user is still admin (or something), we should use that as a hint that this is the user who created the capability.

I think (3) is no good: Arguably it would also be "consistent" to say that if an admin is demoted, all the admin capabilities they created are revoked. However, I suspect that in this case, that is almost never what anyone wants in practice. So there is no real "requirement" that we can write here.

If we did (2), I think we want to apply it to all SturdyRefs. This will be useful for the audit UI, which should display which user created each incoming SturdyRef -- technically it could dig into `requirements` to find this, but it might be weird in cases where additional requirements were picked up by passing through membranes. The one question is whether it really makes sense to have `creatorUserId` and `userId` be separate. I do not think there is ever a case where they would be different, other than when `creatorUserId` is present but `userId` is not. If that's only ever the case for non-UiView capabilities, we could very well document this as: "For non-UiView capabilities, `userId` is only informational." So the big question is: Is there a case where a token might represent a raw (unattenuated) `UiView` capability (therefore have no `userId`) but have a `creatorUserId`?

I think I have a use case: say a non-main UiView is minted during a powerbox request made by Alice in order to be embedded in another grain (e.g. within an iframe in that other grain). Alice doesn't really want this UiView appearing directly in her grain list, since that's not the point. She created this UiView just for embedding purposes. In this case, we have a capability representing a UiView that was created by Alice but is not attenuated by Alice's permissions.

So it seems like option (2) is best?

kentonv:
On a separate note, are we going to add the ability to mint capabilities for listening on specific ports or talking to specific remote hosts? I think it makes sense to do this in a subsequent PR but just wondering what the plan is.

jparyani:
> On a separate note, are we going to add the ability to mint capabilities for listening on specific ports or talking to specific remote hosts? I think it makes sense to do this in a subsequent PR but just wondering what the plan is.
Yep that was the plan. They will be persistent similar to IpNetwork/IpInterface.

shell/lib/db.js
```diff
@@ -236,6 +238,11 @@ ApiTokens = new Mongo.Collection("apiTokens");
 //       notificationHandle: A `Handle` for an ongoing notification, as returned by
 //                           `NotificationTarget.addOngoing`. The value is an `_id` from the
 //                           `Notifications` collection.
+//       ipNetwork: An IpNetwork capability that is implemented by the frontend. Eventually, this
+//                  will be moved out of the frontend and into the backend, but we'll migrate the
+//                  database when that happens. This field contains the userId who originally
+//                  minted this capability.
+//       ipInterface: Ditto IpNetwork, except it's an IpInterface.
```
dwrensha: You should update the pseudo-capnproto schema below with these new variants.

shell/shared/admin.js
```diff
@@ -759,6 +855,51 @@ if (Meteor.isServer) {
...
+          throw new Meteor.Error(403, "Offering IpNetwork is only allowed for logged in users " +
+            "(a token is not sufficient). Please sign in with an admin account")
+        } else {
+          throw new Meteor.Error(403, "User must be admin.");
+        }
+      }
+
+      var ipNetwork = new IpNetworkImpl(Meteor.userId());
+      var sturdyRef = waitPromise(ipNetwork.save({
+        sealFor: {
```
dwrensha: Does this parameter actually make it through the RPC? I think that to deal with the quirks of node-capnp you need to do `ipNetwork.save({webkey: null})` here.

dwrensha: cf https://github.com/kentonv/node-capnp/pull/22

jparyani: You're right, it doesn't get passed through. The weird thing is that it actually works in this case, since it is ignored, and the union for ApiTokenOwner defaults to being a webkey anyways.

dwrensha:
@kentonv wrote:
> I think I have a use case: say a non-main UiView is minted during a powerbox request made by Alice in order to be embedded in another grain (e.g. within an iframe in that other grain). Alice doesn't really want this UiView appearing directly in her grain list, since that's not the point.

The view will appear in Alice's grain list if its `owner` field points to Alice. I think that is an orthogonal concern.

dwrensha:
@kentonv wrote:

>I think (3) is no good: Arguably it would also be "consistent" to say that if an admin is demoted, all the admin capabilities they created are revoked. However, I suspect that in this case, that is almost never what anyone wants in practice. So there is no real "requirement" that we can write here.

What makes these admin capabilities special? Or, to come at it from the other direction: if nobody would want such automatic-transitive-revocation for these admin capabilities, why would they want such automatic-transitive-revocation for ordinary capabilities hosted by grains? Will ordinary capabilities have a way to opt out of such semantics?

kentonv:
> The view will appear in Alice's grain list if its owner field points to Alice. I think that is an orthogonal concern.

Right, but, the point is, we need `creatorUserId` here because if we use `userId` then we're saying the capability should not work if the grain is not in Alice's grain list.

That said, I'm becoming more conflicted about whether we need a new field vs. just digging into `requirements`. Arguably, if there are multiple user requirements, we actually want to attribute the capability to all of them... hmm...

dwrensha:
@kentonv wrote:
> The one question is whether it really makes sense to have creatorUserId and userId be separate.

If an anonymous user "creates" a UiView capability, then I guess `creatorUiView` should not be set, but `userId` should still be set to the user who is providing the capability. So I don't think these can be merged into one field.

kentonv:
> What makes these admin capabilities special?

Only that it vaguely feels like "admin" is more commonly a rotating role, because admins set up system things, and those system things are not really tied to the individual.

However, this is a _very_ fuzzy argument and there are plenty of cases where you don't want transitive revocation with "regular" capabilities.

What we really want here, I suppose, is that when you demote an admin to non-admin, the UI should prompt you about whether to transitively revoke things they set up.

kentonv:
> If an anonymous user "creates" a UiView capability, then I guess creatorUserId should not be set, but userId should still be set to the user who is providing the capability.

Who is "providing" the capability if not the creator? The user who shared access to the anonymous user? That ought to be expressed as a requirement that the sharing token is still valid.

dwrensha:
> Right, but, the point is, we need creatorUserId here because if we use userId then we're saying the capability should not work if the grain is not in Alice's grain list.

Ah, I think I understand now what you're getting at. And I don't think that this argument needs to invoke non-main UiViews to make sense. If Alice wants to forward a UiView from one grain to another without needing to save a copy for herself, then we can't use `userId` and `creatorUserId` to mean the same thing here.

dwrensha:
> Who is "providing" the capability if not the creator?

Part of my problem here is that I'm unsure of exactly what it means to be the "creator" of a sturdyref.

In this case, the provider of the second token is `parentToken`, pointing back to the first token, which has a (grain, user) pair that can be used to compute permissions.

My impression is that the intended meaning of "creator" is something like "the user who was offered this strudyref". So if the second token's creator is the anonymous user.

dwrensha:
> That said, I'm becoming more conflicted about whether we need a new field vs. just digging into requirements. Arguably, if there are multiple user requirements, we actually want to attribute the capability to all of them... hmm...

+1

jparyani:
Ready to merge

shell/server/core.js
```diff
@@ -116,8 +116,16 @@ NotificationHandle.prototype.close = function () {
...
 
-saveFrontendRef = function (frontendRef, owner, userId) {
+saveFrontendRef = function (frontendRef, owner, creatorUserId, parentToken, grainId) {
   return inMeteor(function () {
+    if (parentToken) {
+      var ret = makeChildTokenInternal(parentToken, owner, [], grainId);
+      return {sturdyRef: ret.token};
+    }
+    var requirement = {
+      userIsAdmin: creatorUserId
```
kentonv: This seems pretty confusing: `saveFrontendRef()` takes a parameter called `creatorUserId`, but we interpret it as "this user is required to be admin". Do we even want such a requirement in all cases? It seems like the parameter should be a requirement list, and the caller should decide what requirements to pass.

shell/server/drivers/ip.js
```diff
@@ -37,7 +37,24 @@ ByteStreamConnection.prototype.write = function (data) {
...
+function IpInterfaceImpl (userId, parentToken, grainId) {
+  // TODO(someday): maybe check if userId is admin? This could break things if we don't allow
+  // handing off capabilties from a deleted/deactivated user, so for now we're just allowing all
+  // admins to revoke any ipInterface/ipNetwork caps from the admin settings page.
+  this.userId = userId;
+  this.parentToken = parentToken;
+  this.grainId = grainId;
+}
+
+makeIpInterface = function (userId, parentToken, grainId) {
```
kentonv: I think it's the case that when `parentToken` is provided, `userId` is irrelevant. Moreover, I think in the case where `userId` is provided, the object is used only to handle a subsequent save() call.

Maybe you can simplify by removing the `userId` parameter entirely. When creating a fresh top-level IpInterface token, create the token directly in the table without creating the object at all.

kentonv: Tracing through how `grainId` is used, I think it's not being used correctly.

It looks like the `grainId` passed in is the grain that restored the token. It looks like it eventually gets used in `save()` to fill in `grainId` of the `ApiToken`. An ApiToken for a frontend ref should not have a `grainId` at all.

I think you can remove all use of `grainId` in this file and you can also remove it as a parameter to `restoreInternal()`.

jparyani:
Done

shell/shared/admin.js
```diff
@@ -759,6 +856,49 @@ if (Meteor.isServer) {
...
+      }];
+      var sturdyRef = waitPromise(saveFrontendRef({ipInterface: true}, {webkey: null},
+                                  requirements)).sturdyRef;
+      return ROOT_URL.protocol + "//" + makeWildcardHost("api") + "#" + sturdyRef;
+    },
+    adminToggleDisableCap: function (token, capId, value) {
+      checkAuth(token);
+
+      if (value) {
+        ApiTokens.update({_id: capId}, {$set: {expires: new Date(0)}});
```
dwrensha: Looks like your intent here is "soft" deletion. However, expired tokens get collected every hour:
https://github.com/sandstorm-io/sandstorm/blob/45c2ba13f56b6976ffb32a1a9eb38eab983a8516/shell/shared/grain.js#L130-L136

dwrensha: My proposal for soft deletion: #570.

jparyani:
I've changed this to use new `revoked` field (thanks @dwrensha). Ready to merge.

shell/client/shell.html
```diff
@@ -899,6 +902,41 @@
...
+    <h3>Powerbox Offer</h3>
+    <div>
+      <input type="text" class="autoSelect" value="{{powerboxOfferUrl}}" readonly>
+    </div>
+  </div>
+  <div id="powerbox-offer-popup-closer" class="popup-closer"></div>
+  {{/if}}
+  <p><button id="offer-ipnetwork">Offer IpNetwork Capability</button>
+     <button id="offer-ipinterface">Offer IpInterface Capability</button><br>
+    <span class="details">These buttons create IpNetwork and IpInterface capabilities respectively, and will return webkeys that you can restore in other apps. Typically these apps will be "drivers" that offer more granular protocols such as IRC or SMTP.</span></p>
```
kentonv: When writing text for consumption by the user, we need to make sure we don't use a bunch of implementation jargon than only we understand.

How about: "These buttons create webkeys (special secret addresses) representing permission to directly access the network. Some apps -- especially drivers -- will ask for these permissions. To grant them, you'll need to copy/paste the webkey from here into the app's request box. (Eventually, this UI will be superseded by the Powerbox picker.)"

kentonv: Maybe also: "IpNetwork represents the capability to make outbound requests to the network. IpInterface represents the capability to receive inbound connections on the machine's main network interface."

kentonv: Wait let me try editing this through github!

kentonv:
@jparyani Please review my changes.

jparyani:
LGTM

kentonv:
Garply, retest this please.

kentonv:
@jparyani Any idea why the IpNetwork test seems to have started failing as of my HTML-text-only change?

jparyani:
Ready to merge

PR #547, opened 2015-06-30
Allow viewing the log of a grain without starting the UIView

jparyani:
This is useful for the case where a grain crashes too fast to be able to
open the debug log. It works like `getWwwFileHack` in that the
supervisor is started, but the main UIView is never started.


paulproteus:
I'm super-duper excited about this.​

PR #551, opened 2015-07-01
Clean up offer template argument validation.

dwrensha:
This puts `roleAssignment` into its own field. Additionally, it validates `roleAssignment` on the client side, resulting in a better error message if the match fails.

(Unfortunately there's a related issue where Firefox does not report the full exception information when a `Meteor.Error` is thrown from within the event listener callback. (Chrome does just fine.) I think that problem should be dealt with in a separate pull request.)

I couldn't think of a good way to precisely express an exclusive union in a Meteor [pattern](http://docs.meteor.com/#/full/matchpatterns).

Something like the following would work, but it starts to get really verbose:

```
roleAssignmentPattern = Match.OneOf([
 {none : null,
  addPermissions: Match.Optional([Boolean]),
  removePermissions: Match.Optional([Boolean])},
 {allAccess : null,
  addPermissions: Match.Optional([Boolean]),
  removePermissions: Match.Optional([Boolean])},
 {roleId : Match.Integer,
  addPermissions: Match.Optional([Boolean]),
  removePermissions: Match.Optional([Boolean])}]);
```

### DIFF
--- a/shell/lib/db.js
+++ b/shell/lib/db.js
@@ -473,3 +473,11 @@ allowDevAccounts = function () {
            Meteor.settings.public.allowDevAccounts;
   }
 };
+
+roleAssignmentPattern = {
+  none : Match.Optional(null),
+  allAccess: Match.Optional(null),
+  roleId: Match.Optional(Match.Integer),
+  addPermissions: Match.Optional([Boolean]),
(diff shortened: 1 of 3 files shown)


dwrensha:
Okay, I've updated this to return an error when possible, instead of throwing an exception.

PR #552, opened 2015-07-01
Detect architecture before running userns test program

paulproteus:
* Add an installer test, and
* Fix it, with @zarvox 's help

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -275,7 +275,7 @@ assert_linux_x86_64() {
   fi
 
   if [ "$(uname -m)" != x86_64 ]; then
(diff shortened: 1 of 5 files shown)


paulproteus:
Once this goes green, I will merge it.

Thanks to @zarvox for his brief review on this; that's how I discovered the need for the second-to-last commit in the series, and then thanks to me for actually trying the "improved" pull request and discovering that the final commit is also needed.

zarvox:
:+1:

paulproteus:
@kentonv , can you update install.sandstorm.io with this improved script? I've checked that it passes `installer-tests/` and it addresses an issue that a few people have run into, so I'd love to get the fix out soon.

Thanks!

PR #559, opened 2015-07-06
Fix copy/paste bug that prevents ApiSession proxies from being properly killed.

dwrensha:


jparyani:
Garply, retest this please

PR #560, opened 2015-07-07
Update powerbox query and related interfaces to match current thinking

kentonv:
Namely that the responding app is rendered embedded inside the powerbox and can display an additional UI there to choose a capability or even mint a new capability.

Thoughts? @zarvox @jparyani @dwrensha

### DIFF
--- a/src/sandstorm/grain.capnp
+++ b/src/sandstorm/grain.capnp
@@ -71,7 +71,7 @@ struct PowerboxDescriptor {
   #
   # Note that, in some use cases, where the "object" being granted is in fact just static data,
   # that data may be entirely encoded in tags, and the object itself may be a null capability.
-  # For exmaple, a powerbox request for a "contact" may result in a null capability with a tag
+  # For example, a powerbox request for a "contact" may result in a null capability with a tag
   # containing the contact details. Apps are free to define such conventions as they see fit; it
   # makes no difference to the system.
 
@@ -600,7 +600,7 @@ interface SessionContext {
   #
   # The capability is implicitly tied to the user as if via `tieToUser()`.
 
-  request @3 (query :List(PowerboxDescriptor)) -> (cap :Capability);
+  request @3 (query :List(PowerboxDescriptor)) -> (cap :Capability, descriptor :PowerboxDescriptor);
   # Although this method exists, it is unimplemented and currently you are meant to use the
   # postMessage api to get a token, and then restore that token with SandstormApi.restore().
   #
--- a/src/sandstorm/grain.capnp
+++ b/src/sandstorm/grain.capnp
@@ -53,16 +53,27 @@ struct PowerboxDescriptor {
   # Describes properties of capabilities exported by the powerbox, or capabilities requested
   # through the powerbox.
   #
-  # Usually, descriptors come in a list, where the list indicates one of two things:
+  # A PowerboxDescriptor specified individually describes the properties of a single object or
+  # capability. It is a conjunction of "tags" describing different aspects of the object, such as
+  # which interfaces it implements.
+  #
+  # Often, descriptors come in a list, i.e. List(PowerboxDescriptor). Such a list is usually a
+  # disjunction describing one of two things:
   # - A powerbox "query" is a list of descriptors used in a request to indicate what kinds of
   #   objects the requesting app is looking for. (In a powerbox "offer" interaction, the "query"
   #   is the list of descriptors that the accepting app indicated it accepts in its `ViewInfo`.)
-  # - A powerbox "provision" is a list of descriptors used to describe what kinds of objecst an
+  # - A powerbox "provision" is a list of descriptors used to describe what kinds of objects an
   #   app provides, which can be requested by other apps. (In a powerbox "offer" interaction, the
-  #   "provision" is the list of descriptors that the offering app passed to `offer()`.)
+  #   "provision" consists of the single descriptor that the offering app passed to `offer()`.)
   #
   # For a query to match a provision, at least one descriptor in the query must match at least one
   # descriptor in the provision (with an acceptable `matchQuality`; see below).
+  #
+  # Note that, in some use cases, where the "object" being granted is in fact just static data,
+  # that data may be entirely encoded in tags, and the object itself may be a null capability.
+  # For exmaple, a powerbox request for a "contact" may result in a null capability with a tag
+  # containing the contact details. Apps are free to define such conventions as they see fit; it
+  # makes no difference to the system.
 
   tags @0 :List(Tag);
   # List of tags. For a query descriptor to match a provision descriptor, every tag in the query
@@ -74,17 +85,27 @@ struct PowerboxDescriptor {
     id @0 :UInt64;
     # A unique ID naming the tag. All such IDs should be created using `capnp id`.
     #
-    # If `id` is the Cap'n Proto type ID of an interface, it indicates that the described powerbox
-    # capability will implement this interface. The interface's documentation may define what
-    # `value` should be in this case; otherwise, it should be null.
+    # It is up to the developer who creates a new ID to decide what type the tag's `value` should
+    # have (if any). This should be documented where the ID is defined, e.g.:
     #
-    # If `id` is the type ID of a struct type, then `value` is an instance of that struct type.
-    # The struct type's documentation describes how the tag is to be interpreted.
+    #     const preferredFrobberTag :UInt64 = 0xa170f46ec4b17829;
+    #     # The value should be of type `Text` naming the object's preferred frobber.
     #
-    # Apps may also define `typeId` values which do not correspond to any Cap'n Proto type. In
-    # this case, the ID should be declared as a constant with documentation defining what the
-    # `value` should be in this case. For example, you might declare a tag type ID which is for
-    # strings.
+    # By convention, however, a tag ID is *usually* a Cap'n Proto type ID, with the following
+    # meanings:
+    #
+    # * If `id` is the Cap'n Proto type ID of an interface, it indicates that the described
+    #   powerbox capability will implement this interface. The interface's documentation may define
+    #   what `value` should be in this case; otherwise, it should be null. (For example, a "file"
+    #   interface might define that the `value` should be some sort of type descriptor, such as a
+    #   MIME type. Most interfaces, however, will not define any `value`; the mere fact that the
+    #   object implements the interface is the important part.)
+    #
+    # * If `id` is the type ID of a struct type, then `value` is an instance of that struct type.
+    #   The struct type's documentation describes how the tag is to be interpreted.
+    #
+    # Note that these are merely conventions; nothing in the system actually expects tag IDs to
+    # match Cap'n Proto type IDs, except possibly debugging tools.
 
     value @1 :AnyPointer;
     # An arbitrary value expressing additional metadata related to the tag.
@@ -94,7 +115,7 @@ struct PowerboxDescriptor {
     # this field null.
     #
     # When "matching" two descriptors (one of which is a "query", and the other of which describes
-    # a "provider"), the following algorithm is used to decide if they match:
+    # a "provision"), the following algorithm is used to decide if they match:
     #
     # * A null pointer matches any value (essentially, null = wildcard).
     # * Pointers pointing to different object types (e.g. struct vs. list) do not match.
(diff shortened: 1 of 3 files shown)


src/sandstorm/grain.capnp
```diff
@@ -23,80 +23,165 @@ using Util = import "util.capnp";
...
+    # different interfaces for different file types. So, you will want to use tags here. A
+    # "mime type"
   }
-}
 
-interface PowerboxAction {
-  apply @0 (cap: PowerboxCapability) -> (view :UiView);
-  # Invoke the action on the given capability, producing a view which is displayed to the user.
+  quality @1 :MatchQuality = acceptable;
+  # Use to indicate a preferenc or anti-preference for this descriptor compared to others in the
```
dwrensha: s/preferenc/preference/

src/sandstorm/grain.capnp
```diff
@@ -23,80 +23,165 @@ using Util = import "util.capnp";
...
+
+struct PowerboxDescriptor {
+  # Describes properties of capabilities exported by the powerbox, or capabilities requested
+  # through the powerbox.
+  #
+  # Usually, descriptors come in a list, where the list indicates one of two things:
+  # - A powerbox "query" is a list of descriptors used in a request to indicate what kinds of
+  #   objects the requesting app is looking for. (In a powerbox "offer" interaction, the "query"
+  #   is the list of descriptors that the accepting app indicated it accepts in its `ViewInfo`.)
+  # - A powerbox "provision" is a list of descriptors used to describe what kinds of objecst an
```
dwrensha: s/objecst/objects/

src/sandstorm/grain.capnp
```diff
@@ -23,80 +23,165 @@ using Util = import "util.capnp";
...
+    #   recursively.
+    # * Two lists of non-struct elements match if their contents match exactly.
+    # * Lists of structs are treated as *sets*. They match if every element in the query list
+    #   matches at least one element in the provider list. Order of elements is irrelevant.
+    #
+    # The above algorithm may appear quirky, but is designed to cover common use cases while being
+    # relatively simple to implement. Consider, for example, a powerbox query seeking to match
+    # "video files". All "files" are just byte blobs; file managers probably don't implement
+    # different interfaces for different file types. So, you will want to use tags here. A
+    # "mime type"
```
dwrensha: Looks like you intended to finish this sentence.

src/sandstorm/grain.capnp
```diff
@@ -23,80 +23,165 @@ using Util = import "util.capnp";
...
-    # TODO(someday):  Other match criteria.
+    value @1 :AnyPointer;
+    # An arbitrary value expressing additional metadata related to the tag.
+    #
+    # This is optional. "Boolean" tags (where all that matters is that they are present or
+    # absent) -- including tags that merely indicate that an interface is implemented -- may leave
+    # this field null.
+    #
+    # When "matching" two descriptors (one of which is a "query", and the other of which describes
+    # a "provider"), the following algorithm is used to decide if they match:
```
dwrensha: "provision" is the term you use above.

src/sandstorm/grain.capnp
```diff
@@ -23,80 +23,165 @@ using Util = import "util.capnp";
...
+  # Describes properties of capabilities exported by the powerbox, or capabilities requested
+  # through the powerbox.
+  #
+  # Usually, descriptors come in a list, where the list indicates one of two things:
+  # - A powerbox "query" is a list of descriptors used in a request to indicate what kinds of
+  #   objects the requesting app is looking for. (In a powerbox "offer" interaction, the "query"
+  #   is the list of descriptors that the accepting app indicated it accepts in its `ViewInfo`.)
+  # - A powerbox "provision" is a list of descriptors used to describe what kinds of objecst an
+  #   app provides, which can be requested by other apps. (In a powerbox "offer" interaction, the
+  #   "provision" is the list of descriptors that the offering app passed to `offer()`.)
```
dwrensha: s/list of descriptors/list of tags/ ?

src/sandstorm/grain.capnp
```diff
@@ -23,80 +23,165 @@ using Util = import "util.capnp";
...
+# accept this offer. Grains can register interest in receiving offers by filling in the powerbox
+# metadata in `UiView.ViewInfo`. Apps can also indicate in their manifest that it makes sense for a
+# user to create a whole new grain to accept a powerbox offer. In either case, a session is created
+# using `UiView.newOfferSession()`.
+
+struct PowerboxDescriptor {
+  # Describes properties of capabilities exported by the powerbox, or capabilities requested
+  # through the powerbox.
+  #
+  # Usually, descriptors come in a list, where the list indicates one of two things:
```
dwrensha: "descriptors come in a list" is inaccurate. Maybe "a descriptor is a list of tags" would be better?

src/sandstorm/grain.capnp
```diff
@@ -23,80 +23,165 @@ using Util = import "util.capnp";
...
-    # could be displayed e.g. in a tooltip.
-
-    interfaces @3 :List(UInt64);
-    # Type IDs of Cap'n Proto interfaces implemented by this capability.  Only interfaces which
-    # should be used for powerbox searching/matching purposes need to be listed.  Superclasses
-    # must be listed explicitly if they should be considered for matching.  The powerbox does not
-    # know anything about the actual interface schemas; it just matches the IDs.
+    # If `id` is the Cap'n Proto type ID of an interface, it indicates that the described powerbox
+    # capability will implement this interface. The interface's documentation may define what
+    # `value` should be in this case; otherwise, it should be null.
```
dwrensha: What if I want to match on a specific instance of a generic interface, i.e. something like `Assignable(MyStruct)`?

src/sandstorm/grain.capnp
```diff
@@ -467,6 +588,40 @@ interface SessionContext {
...
+  provide @4 (cap :Capability, requiredPermissions :PermissionSet,
+              descriptor :PowerboxDescriptor, displayInfo :PowerboxDisplayInfo);
+  # For sessions started with `newRequestSession()`, fulfills the original request. If only one
+  # capability was requested, the powerbox will close upon `provide()` being called. If multiple
+  # capabilities were requested, then the powerbox remains open and `provide()` may be called
+  # repeatedly.
+  #
+  # If the session was not started with `newRequestSession()`, this method is equivalent to
+  # `offer()`. This can be helpful when building a UI that can be used both embedded in the
+  # powerbox and stand-alone.
```
dwrensha: Is there any situation where we would want `offer()` and `provide()` to act differently? Do they need to be distinct methods?

src/sandstorm/grain.capnp
```diff
@@ -23,80 +23,165 @@ using Util = import "util.capnp";
...
-    # Type IDs of Cap'n Proto interfaces implemented by this capability.  Only interfaces which
-    # should be used for powerbox searching/matching purposes need to be listed.  Superclasses
-    # must be listed explicitly if they should be considered for matching.  The powerbox does not
-    # know anything about the actual interface schemas; it just matches the IDs.
+    # If `id` is the Cap'n Proto type ID of an interface, it indicates that the described powerbox
+    # capability will implement this interface. The interface's documentation may define what
+    # `value` should be in this case; otherwise, it should be null.
+    #
+    # If `id` is the type ID of a struct type, then `value` is an instance of that struct type.
+    # The struct type's documentation describes how the tag is to be interpreted.
```
dwrensha: Oh, interesting. So I can offer plain old data by passing in a null capability and putting the data in the value of a tag?

src/sandstorm/grain.capnp
dwrensha: If we want full generality, I think that we need to be able to specify a `schema.Type` here rather than just a `UInt64`. I'm somewhat in favor of such a change, though I'm not entirely certain that it would be worth the complexity.

src/sandstorm/grain.capnp
```diff
@@ -23,80 +23,165 @@ using Util = import "util.capnp";
...
+    #
+    # Apps may also define `typeId` values which do not correspond to any Cap'n Proto type. In
+    # this case, the ID should be declared as a constant with documentation defining what the
+    # `value` should be in this case. For example, you might declare a tag type ID which is for
+    # strings.
```
dwrensha: > you might declare a tag type ID which is for strings.

If we had a `tag.type` field rather than a `tag.id` field (as I suggest above), there would be no need for such ad-hoc declarations.

src/sandstorm/grain.capnp
```diff
@@ -23,80 +23,165 @@ using Util = import "util.capnp";
...
-    # must be listed explicitly if they should be considered for matching.  The powerbox does not
-    # know anything about the actual interface schemas; it just matches the IDs.
+    # If `id` is the Cap'n Proto type ID of an interface, it indicates that the described powerbox
+    # capability will implement this interface. The interface's documentation may define what
+    # `value` should be in this case; otherwise, it should be null.
+    #
+    # If `id` is the type ID of a struct type, then `value` is an instance of that struct type.
+    # The struct type's documentation describes how the tag is to be interpreted.
+    #
+    # Apps may also define `typeId` values which do not correspond to any Cap'n Proto type. In
```
dwrensha: I assume that `typeId` refers to the `id` field?

src/sandstorm/grain.capnp
dwrensha: Or are these data tags intended to be used solely for matching purposes? If so, then why do they even need a type id to be specified?

src/sandstorm/grain.capnp
dwrensha: Oops, I was mistaken here. `List(PowerboxDesciptor)` is indeed typical, and specifies a disjunction of descriptors, where each descriptor is a conjunction.

src/sandstorm/grain.capnp
dwrensha: > If so, then why do they even need a type id to be specified?

Oh, right, we need the IDs because the lists of tags are not ordered, so there needs to be a way to match the corresponding elements of the query and the provision.

src/sandstorm/grain.capnp
```diff
@@ -467,6 +588,40 @@ interface SessionContext {
   #     // pass event.data.token to your app's server and call SandstormApi.restore() with it
   #   }
   # }, false)
+
+  provide @4 (cap :Capability, requiredPermissions :PermissionSet,
+              descriptor :PowerboxDescriptor, displayInfo :PowerboxDisplayInfo);
+  # For sessions started with `newRequestSession()`, fulfills the original request. If only one
+  # capability was requested, the powerbox will close upon `provide()` being called. If multiple
+  # capabilities were requested, then the powerbox remains open and `provide()` may be called
```
dwrensha: What would it look like for multiple capabilities to be requested? The `requestInfo` parameter of `newRequestSession()` is a disjunctive list, right?

src/sandstorm/grain.capnp
kentonv: You can always declare a tag ID as a constant and document it as having a value of a particular brand. I've re-arranged the comment to make this clearer (not pushed yet).

src/sandstorm/grain.capnp
kentonv: I vaguely worry that there might be cases where, in the course of a powerbox request, the embedded app may want to `offer()` another capability that isn't meant to be a response to the request but is rather meant to start a new powerbox. Admittedly this seems like it would be crappy UX but I'm not sure it's safe to say it will never happen.

But maybe the answer here is that such an app should call `goToView()` (which I think I'm going to rename to `openView()`, BTW) to open a new tab and then make the new offer from there? Thoughts?

src/sandstorm/grain.capnp
kentonv: I hadn't thought about plain old data, but indeed you could pass plain data as a tag and have the capability itself be null. The receiving app receives. That may actually be pretty useful. I'm adding a note about this.

src/sandstorm/grain.capnp
kentonv: > > you might declare a tag type ID which is for strings.
>
> If we had a `tag.type` field rather than a `tag.id` field (as I suggest above), there would be no need for such ad-hoc declarations.

I disagree. The problem is, the tag ID doesn't just name a type, it identifies the tag _purpose_. If you simply say that the value is Text, what does it mean? When would you ever match on `Text`? In fact there may need to be many different tag IDs which have the type `Text`, each with some specific meaning.

src/sandstorm/grain.capnp
kentonv: > What would it look like for multiple capabilities to be requested?

This is something I'm still thinking about. Basically I think `request()` should probably have an additional boolean parameter called `multiple`. When set, then the request is for a list of capabilities rather than a single one. In this case the powerbox presents the user with a UI for building the list by visiting one or more grains and adding items one by one.

The `requestInfo` would apply to all elements in the list equally. E.g. for requesting a list of contacts, `requestInfo` describes the different contact formats that are accepted. All the items in the list might be of just one of those formats or could be a variety.

kentonv:
Updated!

src/sandstorm/grain.capnp
```diff
@@ -23,80 +23,203 @@ using Util = import "util.capnp";
...
+  # - A powerbox "provision" is a list of descriptors used to describe what kinds of objects an
+  #   app provides, which can be requested by other apps. (In a powerbox "offer" interaction, the
+  #   "provision" consists of the single descriptor that the offering app passed to `offer()`.)
+  #
+  # For a query to match a provision, at least one descriptor in the query must match at least one
+  # descriptor in the provision (with an acceptable `matchQuality`; see below).
+  #
+  # Note that, in some use cases, where the "object" being granted is in fact just static data,
+  # that data may be entirely encoded in tags, and the object itself may be a null capability.
+  # For exmaple, a powerbox request for a "contact" may result in a null capability with a tag
```
dwrensha: s/exmaple/example/

src/sandstorm/grain.capnp
```diff
@@ -431,18 +587,20 @@ interface SessionContext {
...
+  # capability/grain store.  Other type-specific actions may be offered by the platform or by other
+  # applications.
   #
   # For example, offering a UiView will give the user options like "open in new tab", "save to
   # grain list", and "share with another user".
   #
   # The capability is implicitly tied to the user as if via `tieToUser()`.
 
-  request @3 (query :PowerboxQuery) -> (cap :PowerboxCapability);
+  request @3 (query :List(PowerboxDescriptor)) -> (cap :Capability);
```
dwrensha: If my query is a disjunction of descriptors, I'm going to want to which of them is satisfied by the result that I get back. Is there going to be a way for me to find that out without try/catch?

Maybe it makes more sense to change this signature so that `query` is not a list?

src/sandstorm/grain.capnp
dwrensha: OK. I don't mind having these as separate methods, at least until we have a better understanding about how they will be used.

src/sandstorm/grain.capnp
dwrensha: I now understand better what's going on here, and I think this is OK.

src/sandstorm/grain.capnp
kentonv: Oops, the return value should have included a `PowerboxDescriptor` describing the returned capability. (This is not one of the descriptors from the query, but rather the complete descriptor for the specific object as passed to provide().)

PR #561, opened 2015-07-07
Get offer templates working for anonymous users.

### LINKED ISSUE #555: Offer templates do not work for anonymous users.
The `renderTemplate` handler calls the `newApiToken()` Meteor method, which throws an exception if there is no current user. It used to be that throwing an exception was the only reasonable option here, because we did not have a good way to create a token with appropriately attenuated permissions. Now that we have the notions of parent tokens and child tokens, we have a potentially more natural way to handle this case. Once #516 lands, we should be able to revise `newApiToken()` such that it can create a child token when an anonymous user requests a token.

dwrensha:
This fixes #555.

### DIFF
--- a/shell/server/hack-session.js
+++ b/shell/server/hack-session.js
@@ -323,10 +323,26 @@ HackSessionContextImpl.prototype.generateApiToken = function (petname, userInfo,
 };
 
 Meteor.methods({
-  newApiToken: function (grainId, petname, roleAssignment, forSharing, destroyIfNotUsedByTime) {
-    // Create a new user-oriented API token.
-    if (!this.userId) {
-      throw new Meteor.Error(403, "Must be logged in to create an API token.");
+  newApiToken: function (grainId, petname, roleAssignment, forSharing, destroyIfNotUsedByTime,
+                         rawParentToken) {
+    // Creates a new UiView API token. If `rawParentToken` is set, creates a child token.
+    var userId = this.userId;
+    var parentToken;
+    if (!userId) {
+      if (!rawParentToken) {
+        throw new Meteor.Error(403, "If you are not logged in, you must specify a parent token to " +
+                                    "create a new API token.");
+      } else {
+        check(rawParentToken, String);
+        parentToken = Crypto.createHash("sha256").update(rawParentToken).digest("base64");
+        var parentApiToken = ApiTokens.findOne({_id: parentToken, grainId: grainId,
+                                                objectId: {$exists: false}});
+        if (!parentApiToken) {
+          throw new Meteor.Error(403, "No such parent token found.");
+        }
+        userId = parentApiToken.userId;
+        forSharing = parentApiToken.forSharing;
+      }
     }
 
     check(grainId, String);
@@ -349,9 +365,9 @@ Meteor.methods({
     var token = Random.secret();
     var endpointUrl = ROOT_URL.protocol + "//" + makeWildcardHost("api");
 
-    ApiTokens.insert({
+    var apiToken = {
       _id: Crypto.createHash("sha256").update(token).digest("base64"),
-      userId: this.userId,
+      userId: userId,
       grainId: grainId,
       roleAssignment: roleAssignment,
       petname: petname,
@@ -359,7 +375,13 @@ Meteor.methods({
       expires: null,
       forSharing: forSharing,
       expiresIfUnused: selfDestructAt
-    });
+    };
+
+    if (parentToken) {
+      apiToken.parentToken = parentToken;
+    }
+
+    ApiTokens.insert(apiToken);
 
     return {token: token, endpointUrl: endpointUrl};
   }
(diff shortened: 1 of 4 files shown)


shell/server/hack-session.js
```diff
@@ -323,10 +323,26 @@ HackSessionContextImpl.prototype.generateApiToken = function (petname, userInfo,
...
+    var parentToken;
+    if (!userId) {
+      if (!rawParentToken) {
+        throw new Meteor.Error(403, "If you are not logged in, you must specify a parent token to " +
+                                    "create a new API token.");
+      } else {
+        check(rawParentToken, String);
+        parentToken = Crypto.createHash("sha256").update(rawParentToken).digest("base64");
+        var parentApiToken = ApiTokens.findOne({_id: parentToken, grainId: grainId,
+                                                objectId: {$exists: false}});
```
jparyani: I think you forgot about `frontendRef` in addition to `objectId`. I do kind of wish there was a field that identifies a token as a UiView...

dwrensha: If `frontendRef` is set then `grainId` is not set, so I think we're covered.

jparyani: Yep, you're right, we should be good.

dwrensha:
I've pushed a new commit to this branch, adding some significant changes.

The main improvement is that offer template tokens get kept alive as long as their corresponding templates are open.

Another change is that the fifth argument to `newApiToken()` is now a duration rather than an absolute time. I think this is better for situations when the user's browser might be out of sync.

dwrensha:
Oops, I just realized that the expiration stuff doesn't quite make sense in what I just pushed. Please stand by...

dwrensha:
OK, this should work now.

shell/shared/grain.js
```diff
@@ -590,9 +590,14 @@ if (Meteor.isClient) {
           assignment = call.roleAssignment;
         }
         // Tokens expire by default in 5 minutes from generation date
-        var selfDestructTime = Date.now() + (5 * 60 * 1000);
+        var selfDestructDuration = 5 * 60 * 1000;
+
+        var rawParentToken;
+        if (Router.current().route.getName() === "shared") {
```
zarvox: When the user navigates away from the grain, do we dispose of the `#grain-frame` iframe, or do we simply hide it?  I can see a TOCTOU bug hiding here if the grain frame is hidden, the user navigates elsewhere (changing `Router.current().route`), then the app `postMessage()`s a renderTemplate request.

shell/server/hack-session.js
```diff
@@ -323,10 +323,26 @@ HackSessionContextImpl.prototype.generateApiToken = function (petname, userInfo,
...
+        if (!parentApiToken) {
+          throw new Meteor.Error(403, "No such parent token found.");
+        }
+        userId = parentApiToken.userId;
+        forSharing = parentApiToken.forSharing;
```
kentonv: Arguably it makes sense for a child to set `forSharing = true` when the parent had it `false`. This would happen e.g. when Alice creates an API token for use by an app acting on her behalf (forSharing = false), and then the app shares it to another user (forSharing = true).

OTOH, a child token should NOT be able to set forSharing = false when the parent has it true. (Or rather, if it does, this should be ignored.)

Abstractly, the `forSharing` flag on any link disassociates the consumer from any upstream identity.

shell/server/hack-session.js
```diff
@@ -323,10 +323,26 @@ HackSessionContextImpl.prototype.generateApiToken = function (petname, userInfo,
...
+    var userId = this.userId;
+    var parentToken;
+    if (!userId) {
+      if (!rawParentToken) {
+        throw new Meteor.Error(403, "If you are not logged in, you must specify a parent token to " +
+                                    "create a new API token.");
+      } else {
+        check(rawParentToken, String);
+        parentToken = Crypto.createHash("sha256").update(rawParentToken).digest("base64");
+        var parentApiToken = ApiTokens.findOne({_id: parentToken, grainId: grainId,
```
kentonv: Mongo injection vulnerability: `grainId` hasn't been validated.

There are a bunch of `check()` calls after this whole new block you added; please move them back to the start of the function.

shell/shared/grain.js
kentonv: We dispose the iframe.

Also, we have this check at the top of the function:

    if (event.origin !== currentAppOrigin) {

So I think that should prevent any app's postMessage() from being interpreted as from some other app.

shell/server/hack-session.js
dwrensha: If the parent has (forSharing = false) and a user creates a new token using the hypothetical "Asheesh button" (i.e. the "remember token" button) in the webkey menu, then it would make sense for the child to have (forSharing = false). However, maybe that's not an important case now that we have offer templates.

kentonv: What I'm saying is that if the parent has (forSharing = false) then the child should have the option of setting it true or false. But if it's true for the parent, then it is necessarily true for the child.

dwrensha:
Comments addressed.

PR #563, opened 2015-07-07
Stop crashing when Meteor.user() is null

paulproteus:
This affects a purely-cosmetic stacktrace that arises when visiting
https://demo.sandstorm.io/ -- the admin alert wants to show information
about when this user expires, but there is no user yet, so it would
crash with:

```
Exception in template helper: TypeError: Cannot read property expires of null
```

in `formatAccountExpiresIn()`.

It's "purely cosmetic" in that it didn't affect functionality.

Regardless, this fixes that.


paulproteus:
Thanks!

PR #567, opened 2015-07-09
Implement storage quotas.

kentonv:
The basic idea here is:
- Admin assigns storage quota to users e.g. at invite time.
- The back-end can determine how much space the user's grains are currently using. (Only on Blackrock, though; regular Sandstorm doesn't have any efficient way to do this.)
- We'll use this to display to the user how much space they are using.
- When the user goes over their quota, we'll stop letting them create new grains.
- When the user goes over 120% of their quota, we'll stop letting them start their existing grains.

@jparyani can you review this?

### DIFF
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -152,7 +152,8 @@ limitations under the License.
                   data-devid="{{_id}}" data-index="{{index}}">{{title}}</button>
             {{else}}
               {{#each actions}}
-                <button class="new-grain-button" data-actionid="{{_id}}">{{title}}</button>
+                <button class="new-grain-button {{#if overQuota}}over-quota{{/if}}"
+                    data-actionid="{{_id}}">{{title}}</button>
               {{/each}}
               <button class="uninstall-app-button" data-appid="{{selectedTab.appId}}">
                 Uninstall</button>
@@ -166,12 +167,26 @@ limitations under the License.
         {{else}}
           {{#if selectedTab.myFiles}}
             <div class="button-box">
-              <button id="install-apps-button">Install apps</button>
-              <button id="upload-app-button">Upload an app</button>
-              <button id="restore-backup-button">Restore a backup</button>
+              <button id="install-apps-button" class="{{#if overQuota}}over-quota{{/if}}">
+                  Install apps</button>
+              <button id="upload-app-button" class="{{#if overQuota}}over-quota{{/if}}">
+                  Upload an app</button>
+              <button id="restore-backup-button" class="{{#if overQuota}}over-quota{{/if}}">
+                  Restore a backup</button>
             </div>
           {{/if}}
         {{/if}}
+
+        {{#unless selectedTab.sharedWithMe}}
+          {{#if quotaEnabled}}
+            <div id="storage-usage" class="{{#if overQuota}}over-quota{{/if}}">{{storageUsage}}
+              {{#with storageQuota}}
+                of {{.}}
+              {{/with}}
+              used
+            </div>
+          {{/if}}
+        {{/unless}}
 
         {{#if filteredGrains}}
           <table>
@@ -576,7 +591,10 @@ limitations under the License.
         if you just want to share your files with them.</p>
       <h3>Create a link:</h3>
       <p>This will create a magic link which you can send to someone.</p>
-      <p style="text-align: center">
+      <p>
+        {{#if quotaEnabled}}
+          Quota: <input id="key-quota" type="text" placeholder="(bytes; blank for none)"><br>
+        {{/if}}
         <input id="key-note" type="text" placeholder="notes, e.g. name of the recipient">
         <button id="create">Create</button></p>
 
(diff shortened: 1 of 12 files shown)


jparyani:
Consider moving the "quotaEnabled" template helper to a globally registered one. It's defined in at least 3 different places already.

shell/client/shell.html
```diff
@@ -166,13 +167,27 @@
...
+                  Upload an app</button>
+              <button id="restore-backup-button" class="{{#if overQuota}}over-quota{{/if}}">
+                  Restore a backup</button>
             </div>
           {{/if}}
         {{/if}}
 
+        {{#unless selectedTab.sharedWithMe}}
+          {{#if quotaEnabled}}
+            <div id="storage-usage" class="{{#if overQuota}}over-quota{{/if}}">{{storageUsage}}
```
jparyani: This shows global storage beneath each app, which is a bit unintuitive. I definitely understand that we want users to be constantly aware of how close they are to the limit, but this text appears in a seemingly app specific context, and could be easily confused to read the space usage of only this app's grains.

This may have to wait on the shell re-design to fix, since I can't think of a better place to put it for now.

jparyani:
I don't think `updateLastActive` is ever called for api access through webkeys or for grains that are started because a powerbox capability is restored from it. This would lead to storage quota leaks.

shell/shared/admin.js
```diff
@@ -409,7 +409,16 @@ if (Meteor.isClient) {
     },
     userIsGuest: function () {
       return !this.isAdmin && !this.signupKey;
-    }
+    },
+    userQuota: function () {
+      return (typeof this.quota === "number") ? prettySize(this.quota) : "";
```
jparyani: Consider returning "unlimited" or something equivalent instead of displaying an empty string.

jparyani:
Interestingly, we only check quotas on installing new apps and on starting up grains. Theoretically, a grain that's already been opened (or backgrounded) can continue to use up storage space endlessly. An easy way to fix this may be to reactively kill all grains for a user that fails `isUserExcessivelyOverQuota`.

shell/lib/db.js
```diff
@@ -326,7 +326,7 @@ if (Meteor.isServer) {
 
     if (this.userId) {
       return Meteor.users.find({_id: this.userId},
-          {fields: {signupKey: 1, isAdmin: 1, expires: 1}});
+          {fields: {signupKey: 1, isAdmin: 1, expires: 1, quota: 1, storageUsage: 1}});
```
dwrensha: Are the fields in the Users collection documented anywhere?

zarvox:
UX question: once a user is excessively over quota, are they able to delete grains they can't open to get back down under quota?  My impression is that the only place in the UI users can delete grains is the trashcan button in the shell after opening that grain.  (Does the shell let you activate a grain without starting it, so you can delete it?)

shell/client/shell.html
kentonv: Yeah I consider this provisional until the UI redesign kicks in. FWIW, gmail shows storage usage at the bottom of every screen, even when you're under a label or something that doesn't include all your mail. (In fact, I think the storage quota it shows is shared with Google Drive, even.)

shell/shared/admin.js
kentonv: Eh, I feel it's more readable to have people with no quota show up blank in the table. Easier to skim through and find the people who do have quota (or vice versa).

shell/lib/db.js
kentonv: not currently. :/

kentonv:
> Consider moving the "quotaEnabled" template helper to a globally registered one. It's defined in at least 3 different places already.

Good call.

> I don't think updateLastActive is ever called for api access through webkeys or for grains that are started because a powerbox capability is restored from it. This would lead to storage quota leaks.

Ehh, I don't think that's a huge deal since the quota *will* sync as soon as the user looks at their grain list, or opens any grain at all in the UI.

> Interestingly, we only check quotas on installing new apps and on starting up grains. Theoretically, a grain that's already been opened (or backgrounded) can continue to use up storage space endlessly.

Yes, this is mentioned in a TODO in proxy.js. I'm not super worried about it for a first pass. This "exploit" is not sustainable long-term as eventually the grain will die for some reason (e.g. worker reboot) and then won't be able to come back up until the user fixes their quota. We can fix it if we observe people actually doing this, though.

> UX question: once a user is excessively over quota, are they able to delete grains they can't open to get back down under quota?

Yes. You can actually open the grain view, it's just that the iframe will display an exception. The delete button is still functional. You can even save a backup of the grain before deleting it.

shell/lib/db.js
dwrensha: We've discussed that we eventually plan to allow user accounts to encompass multiple identities. I imagine that entries in `Meteor.users()` will continue to correspond to identities, and we'll have a new `Accounts` table to collect identities. That `Accounts` table is where the quota data should ultimately end up, I think. So I guess that putting the quota data here is just a temporary solution?

kentonv: Yes, I think we'll have to do a migration at that point.

PR #570, opened 2015-07-10
Add a 'revoked' field to ApiTokens.

dwrensha:


dwrensha:
Garply, retest this please

shell/client/shell.html
```diff
@@ -325,12 +325,14 @@
         <p>Your existing webkeys:</p>
         <ul>
         {{#each existingTokens}}
-          <li>
-            <span class="token-petname" data-token-id="{{_id}}">
-              {{petname}} ({{dateString created}})
-            </span>
-            <button class="revoke-token" title="revoke" data-token-id="{{_id}}">revoke</button>
-          </li>
+          {{#if revoked}}{{else}}
```
kentonv: `#unless revoked`?

shell/shared/grain.js
```diff
@@ -148,8 +139,6 @@ Meteor.methods({
       var grain = Grains.findOne({_id: grainId, userId: this.userId});
       if (grain) {
         Grains.remove(grainId);
-        ApiTokens.remove({grainId : grainId, $or: [{owner: {$exists: false}},
```
kentonv: Hmm, do we really want API tokens for deleted grains to stick around?

shell/shared/grain.js
```diff
@@ -126,13 +124,6 @@ if (Meteor.isServer) {
       promise.cancel();
     });
   });
-
-  function cleanupExpiredTokens() {
```
kentonv: For offer templates, the whole point of having these expire was to clear them out of the database, since a new token is created every time the user looks at the page with the template. So I think we still want to clear expired tokens...

dwrensha: The transient offer-template tokens are deleted in a separate check: https://github.com/sandstorm-io/sandstorm/blob/c47f4ab7e304933ba1b968f89fb4625a8af931b3/shell/server/hack-session.js#L368-L375

shell/shared/grain.js
dwrensha: My thought was that we could give a better error message if we keep them around. But I suppose that the only way that'd be useful would be to distinguish "typo in token" from "resource no longer exists", and I guess that "typo in token" is probably not going to be a common problem.

shell/shared/grain.js
dwrensha: The "expired" case is different because then the user might actually want to request access.

dwrensha: Oops, I meant the previous comment as a continuation of the discussion below about deleting tokens when grains get deleted.

dwrensha:
Updated.

PR #575, opened 2015-07-10
Show spinner animation before a grain has responded to its first request

### LINKED ISSUE #564: wishlist: show a loading throbber for the time between grain launch and first HTTP reply
Right now, some apps take a while to start up because they do migrations, or create databases, or scan thousands of directories, or do some otherwise expensive work on startup.  For instance, for the Piwik package I'm working on, grain launch to content in frame is over 6 seconds.  I intend to reduce this for Piwik, but it'd be nice if the experience were decent even if a third-party app is not the most efficient.

It would be nice if we showed some sort of "Loading..." spinner while the grain spins up and hid it as soon as the first request is completed.  I'm not sure if doing so is compatible with iframe sandboxing, though, since I think we then can't spy the iframe.onload.  Another approach would be to have the proxy emit some signal as it starts/completes proxying the first response to the frontend to hide the spinner, which could go over the Meteor socket.

(sidenote: the fact that we have the disk usage number updating in realtime as the grain launches is actually quite helpful, since it shows the user that something is happening.  But the empty grain frame still feels empty.)

jparyani:
This has the slight downside of sometimes showing two seperate loading
screens (1 from this and 1 internal to the app).

Fixes #564


kentonv:
This change makes it impossible to click the "reveal your identity" / "open in incognito mode" buttons. Reverting.

kentonv:
Let's make sure to add a test case for this UI before we try again. Since it only triggers for new users it's really easy for us to miss yet really embarrassing.

When you tell the test runner to click on something, does it do so by delivering the click directly to the DOM or does it actually render the page and then click on a pixel? Because in this case I think only the latter would have caught the problem (since the problem was that the buttons were covered by a div).

jparyani:
Unfortunately since this was a CSS issue, I know of no good way to get Selenium to test for it. The element was still visibile, and clicking it by ID would work just fine. If we could get Selenium to simulate a mouse click instead it might've caught it, but I don't know think that's possible. I've added a test anyways, since it's good to have some coverage.

I've tweaked this and submitted a new PR https://github.com/sandstorm-io/sandstorm/pull/584

PR #583, opened 2015-07-13
Offer templates can set forSharing: true to dissociate identity

zarvox:
This is relevant for Piwik, where you're trying to create a token that is not
bound to a user, but is intended for public consumption.


paulproteus:
Yay!​

PR #585, opened 2015-07-14
Fix various things pertaining to sharing.

dwrensha:
The biggest change here is that now when you open a child token UiView anonymously, the permissions are correctly computed. I ended up pushing a lot of shared logic in the `collectEdges()` function, which I think makes things nicer and opens the door to further simplifications later. (Eventually I think we'll want to call `collectEdges()` only once per grain-startup request; right now we call it once under `mayOpenGrain()` and once under `grainPermissions()`.)

### DIFF
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -327,14 +327,14 @@ limitations under the License.
         <p>Your existing webkeys:</p>
         <ul>
         {{#each existingTokens}}
-          {{#unless revoked}}
+          {{#if displayToken}}
             <li>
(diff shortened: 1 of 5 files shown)


dwrensha:
Ok, updated to pass the tests.

I also updated `checkRequirements()` so that `permissionsHeld` requirements imply unwritten "may access grain" requirements. So now even if the `permissionsHeld.permissions` is an empty list, the requirement enforces that `permissionsHeld.userId` must still have access to `permissionsHeld.grainId`.

PR #588, opened 2015-07-15
[WIP] Add dynamic topbar package, refactor UI to use it.

kentonv:
(Not ready for merge yet.)

The topbar package allows topbar controlls to be dynamically registered and unregistered, and provides a way to actually embed those registrations syntactically within other templates.

This refactoring is incomplete:
- The mobile view is broken.
- Tests are probably wildly broken due to element IDs changing.
- A lot of dead code is still lying around.

With that said, the desktop view *should* look and behave as it did before this change.

### DIFF
--- a/shell/.meteor/packages
+++ b/shell/.meteor/packages
@@ -21,3 +21,4 @@ browser-policy
 
 reactive-var
 sandstorm-db
+sandstorm-ui-topbar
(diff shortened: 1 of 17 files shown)


kentonv:
Misc broke again. @jparyani after 5a01652 was pushed to the build server, it showed high CPU usage for 5 minutes, decreasing CPU usage for the next 5, then spiked up to 100% and stayed there, without the test apparently completing. In lots of these cases I've seen the graph have a 5-minute hump, then a valley, then spike up to 100 and stay there. It seems like maybe something towards the end of a test run is triggering the problem?

kentonv:
test pass locally. Merging.

@jparyani here is the test run that died when I rebooted the machine, if you want to investigate: https://build.sandstorm.io/job/Sandstorm%20PRs/567/

PR #593, opened 2015-07-16
Add sharing test

jparyani:
@paulproteus reported an issue with sharing, and I tried to debug it (and failed). I wrote a failing test that shows off what's wrong.

@dwrensha, can you take a look when you get the chance?

PR #594, opened 2015-07-16
Fix bug in resharing.

dwrensha:
The bug was reported in #593.


dwrensha:
One fix more! I think this should pass @jparyani's test case now.

PR #595, opened 2015-07-16
Fix frontend disconnects causing problems with SandstormCore

jparyani:


src/sandstorm/backend.c++
```diff
@@ -66,8 +65,14 @@ kj::Promise<Supervisor::Client> BackendImpl::bootGrain(
...
       // We should send a keepAlive() to make sure the supervisor is still up.
       auto promise = client.keepAliveRequest().send();
-      return promise.then([KJ_MVCAP(client)](auto) mutable -> kj::Promise<Supervisor::Client> {
+      return promise.then([=,KJ_MVCAP(client)](auto) mutable -> kj::Promise<Supervisor::Client> {
         // Success.
+        // Make new SandstormCore, since the last one points to a frontend that disconnected
+        auto coreRequest = coreFactory.getSandstormCoreRequest();
+        coreRequest.setGrainId(grainId);
+        auto core = coreRequest.send().getCore();
+        iter->second.coreRedirector->setTarget(core);
```
kentonv: `iter` could be invalidated by the time we get here. You'll need to grab a reference to `coreRedirector` early on and pass it in.

Incidentally, I think I may introduce a new style rule that lambda capture lists must never use `=` and can only use `&` when the callback is to be called synchronously. Otherwise it's hard to tell what's being captured, and lambda capture lists need careful review for these kinds of issues...

kentonv:
Hmm, so you're only restoring the `SandstormCore` when the front-end attempts to reconnect? But that only happens when a user visits the UI, right? So this could be problematic in cases of background processing. It seems like instead, whenever a new frontend connects to the backend, the backend should iterate over all running grains and redirect them.

src/sandstorm/backend.c++
dwrensha: > Incidentally, I think I may introduce a new style rule that lambda capture lists must never use = and can only use & when the callback is to be called synchronously.

+1

jparyani:
Closing, since this was fixed in #1788

kentonv:
Wow, totally forgot that there was a PR open for this.

Man, 9 months ago.

PR #597, opened 2015-07-17
Extend spk manifest format with market metadata.

kentonv:
The goal here is that the app market should not need authors to fill in any form; they just upload an spk.

Note that I plan to extend Cap'n Proto with an `embed` keyword, which is like `import` except that it turns the raw contents of the specified file into `Text` or `Data`. Thus things like image files and license text would not actually be specified inside `sandstorm-pkgdef.capnp` but would be embedded from separate files. When the manifest is compiled to binary, those separate files are pulled into it, forming one big Cap'n Proto blob.

### DIFF
--- a/src/sandstorm/package.capnp
+++ b/src/sandstorm/package.capnp
@@ -310,7 +310,7 @@ struct Metadata {
     # valid address with someone paying attention to it.
 
     pgpSignature @11 :Data;
-    # PGP signature attesting responsibility for the app ID. This is a binary-format non-detached
+    # PGP signature attesting responsibility for the app ID. This is a binary-format detached
     # signature of the following ASCII message (not including the quotes, and replacing <app-id>
     # with the standard base-32 text format of the app's ID):
     #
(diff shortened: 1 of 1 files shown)


src/sandstorm/package.capnp
```diff
@@ -201,6 +205,163 @@ struct BridgeConfig {
...
 }
 
+struct Metadata {
+  # Data which is not needed specifically to execute the app, but is useful for purposes like
+  # marketing and display.
+  #
+  # Technically, appMarketingVersion and appTitle belong in this category, but they were defined
+  # before MarketData became a thing.
+
+  icon :group {
```
dwrensha: Wouldn't it be better to define a struct `IconSet` and have `icon @0 :IconSet`?

kentonv: Well, originally I had `mainIcon`, `grainIcon`, `bannerIcon`. Then I defined a group basically for documentation purposes.

An `IconSet` type makes sense if there's some reason we'd expect it to be reused elsewhere or passed around independently of the surrounding Metadata object. I couldn't think of any reason to expect this, but arguably there's no good reason not to?

src/sandstorm/package.capnp
```diff
@@ -201,6 +205,168 @@ struct BridgeConfig {
...
+
+  codeUrl @4 :Text;
+  # URL of the app's source code repository, e.g. a Github URL.
+
+  license :group {
+    text @5 :Util.LocalizedText;
+    # Text of the app's license.
+
+    notices @6 :Util.LocalizedText;
+    # Contains any third-party copyright notices that the app is required to display, for exmaple
```
dwrensha: s/exmaple/example/

PR #599, opened 2015-07-17
Delete text from inbox icon.

dwrensha:


dwrensha:
This is what I currently see with the mobile view in Firefox:

icon-screenshot

This patch gets rid of the blurry line representing the text.

PR #606, opened 2015-07-21
Parse and pass on IPv6 addresses for X-Real-IP

zarvox:


kentonv:
`::ffff:192.168.0.1` is a legal ipv6 address but I don't think you cover this. Maybe we should use a library? Or maybe we don't really care about covering all the corner cases since we're expecting to parse strings coming from computers, not humans? Thoughts?

zarvox:
Oooh, good catch.  Theoretically we could use a library if we found a suitable one; I didn't really look.  For future reference, if we took that approach, would we vendor it in, and then update our acknowledgements?  Or specify it as a meteor dependency, if there were a meteor package for it?

I checked the nginx source to see whether it'll pass v4-mapped-v6 addresses, and the answer is that it will under the right conditions.

Thus, I've updated this code to also handle IPv6-mapped-IPv4 addresses.  I don't know of any other legal v6 address formats, so I think this patchset will do for now.

kentonv:
@zarvox For libraries, ideally use a Meteor or npm dependency. Copying code is also possible if the code is in one file and unlikely to ever change -- in that case it should be listed in LICENSE.

As for this PR, could you add a `// TODO(test): Unit test this`?

zarvox:
Done.

PR #613, opened 2015-07-22
Clear dev user accounts before sharing tests.

dwrensha:
Currently, the test suite fails if you run it twice, because the dev account data is never cleared.


dwrensha:
I've learned that with `make test` it's very easy to run tests locally on an isolated fresh install. (Thanks @zarvox!) So I no longer think that this patch is needed.

PR #619, opened 2015-07-22
If hostname -f exits with failure, do not crash

### LINKED ISSUE #614: Installer fails if hostname is unrecognized
Don't have the exact error at hand, but when the installer reaches the point where it asks for the instance URL, if your hostname is unresolvable you'll get a "Name or service not known" error and the script fails.

I have a default VM template whose hostname is "debian". This, naturally, doesn't resolve. I had to temporarily set the hostname to something that resolved in order to complete the installer.

It would be nice if, instead of attempting to look up the hostname to place in the URL automatically, the option could be blanked out so those of us who know what we're doing can enter it from scratch. :) I don't have DNS set up for the instance yet and was just trying to see if the installer would complete before going further.

Thanks.

paulproteus:
Close #614 

**Status**: Running tests locally; ETA 5 min or so.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -749,7 +749,7 @@ configure_hostnames() {
   # configure the hostname and IP address accordingly.
   if [ "yes" = "$USE_EXTERNAL_INTERFACE" ]; then
     BIND_IP=0.0.0.0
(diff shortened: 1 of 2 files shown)


paulproteus:
**Status**: installer tests pass. @kentonv requesting your merge + update to `install.sh`.

PR #630, opened 2015-07-26
Initial work at improving accessibility

ndarilek:


garply:

paulproteus:
Garply, ok to test

paulproteus:
@ndarilek thanks for writing these patches!

paulproteus:
Hi @ndarilek , there appears to be a problem with our automated test running. We'll look into it during the week. Hang tight, and sorry about the error message that appears here.

ndarilek:
No hurries, just wanted to get the PR up. Thanks!

paulproteus:
One other thing - if you can "rebase" it against current origin/master,
that'll make it cleaner to merge.​ If you're familiar with that, great. If
not, and you want to learn, great; ping me if you need anything. If that's
more than you want to deal with right now, I don't think it matters at this
stage.

ndarilek:
AFAIK it should already merge with current master. I'll try a rebase,
hopefully the fact that I've already made a few merge commits won't
break things too badly.

ndarilek:
Rebased.

ndarilek:
Any updates on this?

paulproteus:
I think @zarvox is looking at making some changes that overlap. Drew, any
chance you'd be willing to review this pull request?

kentonv:
Hi @ndarilek,

Thanks a lot for doing this! Sorry we're a little slow to review things.

@zarvox is currently reworking the whole sidebar UI, so merging right now might be disruptive -- but I think he should work these changes into his rewrite. Having this available as a guide should help.

Regarding the specifics of the change, I worry that the `href="#"` trick often has weird side-effects, like changing the URL (to add a '#') and sometimes scrolling the page to the top. I think we should use `<a>` tags when there is a logically-correct `href` value, and use `<button>` tags otherwise. For all the specific cases in this PR, there actually either are or should be paths to use for the `href`.

Could you describe semantically what the tab index stuff is doing? I think we'll need to document this carefully, otherwise we're likely to break it.

ndarilek:
The tabindex code makes it possible to quickly tab through some of the
larger UI elements to reach areas of interest.

Right now, very few of your UI elements are keyboard-focusable. What I
do is set tabindex on the first element of a menu or toolbar, then I
rely on arrow navigation to move within the element. So, for instance,
if you press tab twice then you'll hit the menubar, then the toolbar. If
you want to navigate within the menubar you tab once, then use up/down
arrows to move between the items (I assume that menu is vertical.) On a
toolbar, you'd use left/right arrows to move between the various buttons.

I'm trying to make these changes automatic if you use the correct ARIA
roles. For instance, if you use _role="toolbar"_ on a list of buttons,
my screen reader reports them as a toolbar. This hints to me as a
keyboard user that the buttons are likely horizontal, and that I can
arrow left/right between them. And, in fact, that's what works. Same
with _role="menu"_, _role="menuitem"_ and up/down arrows. Using those
roles will automatically get you the expected keyboard behavior,
(hopefully) no other work needed.

Unfortunately I need to incorporate those roles throughout, but I
thought I'd test the waters with a simpler PR and then go from there.

Oh, and _role="presentation"_ is needed on intermediate items to hint to
the screen reader that they're only for visual presentation. So if you
have something like:

`<ul><li><a href="#">Menu item</a></li></ul>`

You'd mark it up like so initially:

`<ul role="menu"><li><a role="menuitem" href="#">Menu item</a></li></ul>`

But the problem there is that you'd hear something like:

"list item Menu item: menu item

The first "Menu item" being the title, the second item being what it is
in the UI. But I don't care that it's a list item. That's only there to
keep your HTML clean. So what I should do is:

`<ul><li role="presentation"><a href="#">Menu item</a></li></ul>`

This tells the screen reader that the `<li>` is only there for visual
presentation, and when the speech representation of the element is
prepared, the `<li>` isn't even considered. So I'd just hear:

Menu item: menu item

Hope that helps a bit.

kentonv:
Re: tab tops and arrow keys: Makes sense! And seems useful for non-screenreader-users as well.

Re: Roles. What happens if you put `role="menuitem"` on the `<li>`? It seems a bit sad to put `role="presentation"` on the `<li>` to "mute" it, only to put `role="menuitem"` on the `<a>` to introduce a new spoken delimiter. Feels like pushing against the grain.

ndarilek:
On 8/7/2015 8:32 PM, Kenton Varda wrote:
> Re: Roles. What happens if you put `role="menuitem"` on the `<li>`? It seems a bit sad to put `role="presentation"` on the `<li>` to "mute" it, only to put `role="menuitem"` on the `<a>` to introduce a new spoken delimiter. Feels like pushing against the grain.
>
>
You'll then hear something like "Link Home: Menu item" instead of "Home:
Menu item". At the point where I'm being told it's a menu item, I don't
really care what it is, so the additional information is redundant.

I also want to say that it messes up something regarding how screen
readers present the items themselves, but I don't recall specifics. I
thought much as you did, tried what you did and something odd happened.
I just don't recall what.

When you mark the menu up as described, I get very nice behavior in my
screen reader (NVDA). It presents the menu as if it were a menu in a
desktop app, and I can arrow up and down through it with no additional
code needed.

See http://heydonworks.com/practical_aria_examples/ for lots of
practical examples.

paulproteus:
Hey @zarvox and @ndarilek -- do the changes that landed recently obviate these changes? Or are there changes in this pull request whose spirit did not make it into the last round?

If all the goodness from this pull request has landed (albeit perhaps in different form), then maybe it's sensible to close this and declare victory.

ndarilek:
Probably makes sense to just close this one. The new interface is so 
different that it probably isn't applicable anymore.

That said, the new interface uses links rather than clickable text, and 
does seem substantially easier to use. Haven't seen all the changes but 
the trend is positive.

paulproteus:
Thanks @ndarilek ! I'll close this for now, and I hope you keep telling us about accessibility issues that pop up.

PR #637, opened 2015-07-28
README specifies just clang, rather than (outdated) clang-3.4

zarvox:
On Debian stable, clang-3.4 does not provide /usr/bin/clang, only
/usr/bin/clang-3.4.  The "clang" package (which provides /usr/bin/clang) pulls
in clang-3.5, which is even better.

On Ubuntu 14.04 LTS, the "clang" package provides clang-3.4 as /usr/bin/clang,
which is acceptable.


zarvox:
Rebased.

PR #645, opened 2015-07-30
Move pre-meteor.js to its own package

jparyani:


dwrensha:
I was confused about how the dependencies here (e.g. `useGrain()`, which is defined in shell/server/proxy.js) work at all.

I found this, in http://docs.meteor.com/#/full/namespacing :

> Technically speaking, globals in an app (as opposed to in a package) are actually true globals. They can't be captured in a scope that is private to the app code, because that would mean that they wouldn't be visible in the console during debugging! This means that app globals actually end up being visible in packages. That should never be a problem for properly written package code (since the app globals will still be properly shadowed by declarations in the packages). You certainly shouldn't depend on this quirk, and in the future Meteor may check for it and throw an error if you do.

jparyani:
Alright, although it makes me sad, I'm closing this PR.

kentonv:
Recap: We decided this code is essentially high-level "stitching" code which ties together various other modules, and as such it is probably one of the last things we should be pushing into a package. If we were to move it into a package, we would want to redesign it to be a general-purpose library: provide an interface which high-lever stitching code can use to register arbitrary handlers for arbitrary hostnames. It's not clear there's much value in doing that refactoring right now, though.

PR #646, opened 2015-07-30
admin.js: rename checkAuthForPublish

zarvox:
Most functions named check() throw an exception, rather than returning a
boolean.


kentonv:
agreed

Issue #605, opened 2015-07-20
signing in while visiting a "/shared/" link


dwrensha:
If a user signs in while visiting a sharing link, they end up in a somewhat broken state in which the grain does not get added to their "shared with me" tab.

PR #647, opened 2015-07-31
new sharing UI

dwrensha:
This implements most of the "Share with others..." functionality of the mockup:
sharing ux


kentonv:
For the email, let's put the share link at the bottom, after the personal message. It should look kind of like an attachment.

It would be nice to send both a text and an HTML version. The HTML version could have a big blue "open shared grain" button rather than an obscure link. The "Follow this link" text could also be omitted from that version. The text version should keep the text, but should have a blank space before and after the link line.

Note that in the HTML version, the user's personal message should be treated as plain text -- you'll need to HTML-escape it and also insert `<br>`s at newlines.

For the "be careful" text after the link, let's make it: "Note: If you forward this email to other people, they will be able to access the share as well. To prevent this, remove the link/button before forwarding." In the HTML version, this text should be de-emphasized, e.g. slightly smaller, italic, and gray.

kentonv:
I tried inviting myself to a grain and observed several bugs, some of which may be preexisting:

* After sending an email, the form didn't clear itself, which felt odd. It did say "success" but it feels like it should actually clear and close the form, and maybe show a separate dialog saying "invite sent! [ok]". Playing the part of a confused user, I clicked "send" again, thus sending multiple emails.
* Of the two emails, one of them initially reported "invalid token" when I first tried to open it, but later on worked fine.
* I was not logged in. When I tried to log in, and then chose "reveal my identity" at the interstitial, etherpad then opened, but still saw me as "anonymous user", and I did not yet show up in the "who has access" list. I had to refresh manually before it did the right thing. Is there possibly a timing issue here?
* The "table view" does not seem to be reactive.
* Each time I click a new sharing link, a new bullet appears under "Kenton Varda was grated access by", all with the same content other then the timestamp (since the links were created by the same person). Maybe we should collapse those, even though they're technically different links.
* All my sharing links show "[has custom permissions]" in the drop-down but I definitely chose "can edit" (and in fact there's no way to choose anything other than the regular roles).

shell/client/shell.html
```diff
@@ -367,89 +367,157 @@
...
-                    <li>{{displayName userId}} ({{dateString created}})</li>
-                  {{/each}}
-                </ul>
-              </li>
-            {{/each}}
-          </ul>
+        <ul>
+          {{#each transitiveShares}}
+            <li>
+              <div>{{displayName recipient}} was granted access by:</div>
```
kentonv: Is the `<div>` tag actually needed here?

shell/client/shell.html
```diff
@@ -367,89 +367,157 @@
...
-                <div>{{displayName recipient}} was granted access by:</div>
-                <ul>
-                  {{#each shares}}
-                    <li>{{displayName userId}} ({{dateString created}})</li>
-                  {{/each}}
-                </ul>
-              </li>
-            {{/each}}
-          </ul>
+        <ul>
```
kentonv: I suppose this will become a `<table>` later?

shell/client/shell.html
```diff
@@ -367,89 +367,157 @@
...
+    <p>This grain uses the old sharing model. You can share it by copy/pasting the URL from the
+       location bar.</p>
+    {{#if isOwner}}
+      <p>You can upgrade this grain to the new model, but anyone with whom you shared this grain
+         previously will lose access; you will have to share it with them again.</p>
+      <p>Upgrading cannot be undone.</p>
+      <p><button id="privatize-grain">upgrade</button></p>
+    {{/if}}
+   {{else}}
+     {{#basicTabs name="" tabs=tabs}}
```
kentonv: A lot of the indentation is inconsistent around here.

kentonv:
I'm not super excited about the tabs library, but I'll probably work on fixing it myself at some point. The approach I prefer there is to have the tabs actually be attached to their bodies, like in https://oasis.sandstorm.io/terms. So you'd have HTML like:

    <section>
      <h3>Send an invite</h3>
      ... send invite form HTML ...
    </section>
    <section>
      <h3>Get sharable link</h3>
      ... create link form HTML ...
    </section>

Then you'd use CSS to stack these, have the h3's stick out above the section contents (e.g. `position: absolute; top: -16px;`), and make sure the h3's don't overlap. Clicking the h3 would manipulate the `z-index` on the whole section to put one or the other on top.

Anyway, just commenting, not an action item.

dwrensha:
> I was not logged in. When I tried to log in, and then chose "reveal my identity" at the interstitial, etherpad then opened, but still saw me as "anonymous user", and I did not yet show up in the "who has access" list. I had to refresh manually before it did the right thing. Is there possibly a timing issue here?

This is https://github.com/sandstorm-io/sandstorm/issues/605

dwrensha:
> For the email, let's put the share link at the bottom, after the personal message. It should look kind of like an attachment.
It would be nice to send both a text and an HTML version. 

Done and done.

>After sending an email, the form didn't clear itself, which felt odd. It did say "success" but it feels like it should actually clear and close the form, and maybe show a separate dialog saying "invite sent! [ok]". Playing the part of a confused user, I clicked "send" again, thus sending multiple emails.

I added a "send more invites" button that appears upon completion and clears the form when pushed.

>Of the two emails, one of them initially reported "invalid token" when I first tried to open it, but later on worked fine.

It sound like this might be related to https://github.com/dwrensha/groovebasin/issues/1. I have never observed such behavior, but clearly there's some bug lurking. I'd be grateful for a reliable reproduction.

>I was not logged in. When I tried to log in, and then chose "reveal my identity" at the interstitial, etherpad then opened, but still saw me as "anonymous user", and I did not yet show up in the "who has access" list. I had to refresh manually before it did the right thing. Is there possibly a timing issue here?

#605

> The "table view" does not seem to be reactive.

Correct, it's not reactive yet. However, the table does get refreshed when the user revokes a share link.

>Each time I click a new sharing link, a new bullet appears under "Kenton Varda was grated access by", all with the same content other then the timestamp (since the links were created by the same person). Maybe we should collapse those, even though they're technically different links.

OK. I've merged these in the table view.

> All my sharing links show "[has custom permissions]" in the drop-down but I definitely chose "can edit" (and in fact there's no way to choose anything other than the regular roles).

Fixed.

shell/client/shell.html
dwrensha: I made this a table.

shell/client/shell.html
dwrensha: Fixed.

shell/client/shell.html
dwrensha: Nope.

dwrensha:
Garply, retest this please.

kentonv:
The table styling needs work but I think this is good to merge now and we'll improve the styling before the next release!

Issue #586, opened 2015-07-14
Add some help text for email login


kentonv:
Some people are surprised when they don't get prompted for a password. We should add an info thingy (letter 'i' in a circle, click for more info) explaining how it works.

kentonv:
> Passwordless E-mail Login allows you to log in using an e-mail address. No password is required; instead, every time you log in, we send you a new email confirmation code. You will only need to log in once per device.

Maybe we could have some sort of a link to a wiki page, which could say things like:

> Sandstorm does not use passwords because passwords are extremely insecure. Most people use the same password for many sites, which means the operator of any of those sites could hijack the user's account on any of the others. Most people also choose easily-guessable passwords.
>
> If you would rather use a password, we recommend logging in using a third-party service like Google or Github, which implement their own password-based login. These services have large security teams well-positioned to deal with account hijacking when it happens. We also recommend enabling two-step authentication when using these services.

paulproteus:
@neynah says this is closed, so I am going to mark it as such!

PR #648, opened 2015-07-31
Add an info bubble to email login.

jparyani:
Addresses #586. Ideally this would be something better than title text, and allow having a link to the docs, but this is a decent first step.

### DIFF
--- a/shell/packages/sandstorm-accounts-ui/login_buttons.html
+++ b/shell/packages/sandstorm-accounts-ui/login_buttons.html
@@ -67,7 +67,7 @@
       {{/if}}
       <form class="email">
         <label>Email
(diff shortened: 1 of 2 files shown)


shell/packages/sandstorm-accounts-ui/login_buttons.html
```diff
@@ -66,7 +66,10 @@
         {{> _loginButtonsSeparator}}
       {{/if}}
       <form class="email">
-        <label>Email <input name="email" type="email"> </label>
+        <label>Email
+         <label title="Passwordless E-mail Login allows you to log in using an e-mail address. No password is required; instead, every time you log in, we send you a new email confirmation code. You will only need to log in once per device." class="info">i</label>
```
kentonv: `<label>` has a specific meaning in HTML: it's supposed to be attached to an input element, either by having the input be a child of the label or by specifying it by ID. So I think you'll need to make this a `<span>`.

jparyani:
Done

PR #650, opened 2015-07-31
New Icons

neynah:
To eventually replace existing ones in shell


kentonv:
These look great! But let's break out the little downward triangles into a separate SVG so we can control it independently. (An inverted version would be useful too, in case we want it to toggle.)

kentonv:
I think these are ready to use so I've merged them!

One random thing: In capability-based security theory, we often draw what Mark Miller calls a "granovetter diagram":



You'll notice this looks pretty close to the sharing icon. I would love to adjust our sharing icon slightly so that it actually is a granovetter diagram. That is, it should show not just three dots (Alice, Bob, and Carol) with two edges (Alice<->Bob, Alice<->Carol), but also show that Alice is sending a message to Bob, and that message contains a new edge to Carol. (After Bob receives the message, all three dots would be connected in a triangle, but we want to capture it "in transit".)

Most users won't notice the difference but I'm sure Mark will love it. :)

neynah:
@kentonv Not sure if I'm conveying it clearly enough...
granovetter-1

kentonv:
Close. The curved line should merge with the lower straight line as it hits the lower dot ("Carol"). However, it shouldn't be exactly symmetrical: the top end of the curved line is actually travelling along the edge between Alice and Bob. You might even draw in a little arrow representing the message ("foo"), although it will usually be too small to see.

If this isn't making sense, I can show you on a whiteboard later. :)

neynah:
Closer?
granovetter-2

kentonv:
Yes, this is closer.

In the image I posted, you'll notice that the arrow is not actually the tip of the curved edge, but rather an independent thing, and the curved edge is itself coming from a point inside the arrow. The arrow is meant to represent a message (called "foo") sent from Alice to Bob, and the curved edge represents that within that message is a reference to Carol. Whereas, in your version the arrow appears to be the head of the curved edge.

Note that the top end of the curved edge isn't really supposed to look like it is merging with the straight edge between Alice and Bob. It's supposed to look like it originates from the message arrow, but could e.g. be perpendicular. Whereas the lower end is supposed to merge, as you have it.

PR #658, opened 2015-08-03
Pass along the user's profile info to apps as new headers.

kentonv:


docs/developing/auth.md
```diff
@@ -42,6 +42,43 @@ related to user identity and permissions:
...
 
+* `X-Sandstorm-Preferred-Handle`: The user's preferred "handle". A
+  handle is like a Unix username. It contains only lower-case ASCII
+  letters, digits, and underscores, and it never starts with a digit.
+  The user can set their preferred handle in their account settings.
+  This handle is NOT UNIQUE; it is only a hint from the user. Apps
+  that use handles must decide for themselves whether they need
+  unique handles and, if so, implement some mechanism to deal with
+  duplicates (such as prompting the user to choose a different one,
+  or just appending some digits). Apps should storngly consider
```
paulproteus: storngly => strongly

shell/.meteor/versions
```diff
@@ -39,7 +39,7 @@ iron:middleware-stack@1.0.9
 iron:router@1.0.9
 iron:url@1.0.9
 jparker:crypto-core@0.1.0
-jparker:crypto-md5@0.1.1
+jparker:crypto-sha256@0.1.1
```
paulproteus: Fascinating change!

docs/developing/auth.md
```diff
@@ -42,6 +42,43 @@ related to user identity and permissions:
...
+  distorting a non-square picture. If this header is missing, the
+  user has no profile picture. In this case, it is recommended that
+  apps use [identicon.js](https://github.com/stewartlord/identicon.js),
+  with the user's ID (from `X-Sandstorm-User-Id`) as the input, to
+  produce identicons consistent with those that Sandstorm's own UI
+  would produce. Note that you should NOT hash the ID; just pass the
+  hex ID directly to the `Identicon` constructor as the `hash`
+  argument.
+
+* `X-Sandstorm-User-Pronouns`: Indicates by which pronouns the user
```
paulproteus: I think current thinking is we're supposed to let people choose their own pronouns rather than from our list. If that's eventually on the TO-DO list, we could indicate as such.

Requesting sanity-check from @brittag if willing.

docs/developing/auth.md
```diff
@@ -42,6 +42,43 @@ related to user identity and permissions:
...
+  as primary keys or for security; use `X-Sandstorm-User-Id`
+  instead.**
+
+* `X-Sandstorm-User-Picture`: The URL of the user's profile picture.
+  The exact resolution of the picture is not specified, but assume
+  it is optimized for a 64x64 viewport. Although profile pictures
+  are normally square, it is recommended to use CSS `max-width` and
+  `max-height` instead of `width` and `height` in order to avoid
+  distorting a non-square picture. If this header is missing, the
+  user has no profile picture. In this case, it is recommended that
```
paulproteus: Not essential, but I suggest:

Avoid passive voice. s/it is recommended that apps use/use/

docs/developing/auth.md
paulproteus: Proposed addition:

"In the future, users might be able to set their own pronouns. Those values would presumably flow into apps via a separate header, and this header might contain the word `user-chosen`."

shell/.meteor/versions
kentonv: I decided the identicon algorithm doesn't need to be perfectly consistent with github because we'll always fetch github users' identicons from actual github. So it now is based on the user ID that we pass into apps, which is sha-256 based. This allows apps to use identicon.js themselves to generate consistent identicons.

Eliminating MD5 was a nice side-effect. (Previously I needed it because that's what Github uses.)

docs/developing/auth.md
paulproteus: It will be nice if `sandstorm-http-bridge` can make for per-grain uniqueness. That's still the plan in the near term, as I understand it?

We don't have to tell people that yet.

docs/developing/auth.md
kentonv: I don't think so -- that's not i18n-friendly.

docs/developing/auth.md
kentonv: Right, we can implement this at some point in the future, but I don't think it's at the top of the priority queue currently.

docs/developing/auth.md
```diff
@@ -42,6 +42,43 @@ related to user identity and permissions:
...
+  apps use [identicon.js](https://github.com/stewartlord/identicon.js),
+  with the user's ID (from `X-Sandstorm-User-Id`) as the input, to
+  produce identicons consistent with those that Sandstorm's own UI
+  would produce. Note that you should NOT hash the ID; just pass the
+  hex ID directly to the `Identicon` constructor as the `hash`
+  argument.
```
paulproteus: It'd be nice if we did this by default, for what it's worth, and sent it along in the headers.

paulproteus: To help apps know if it's a real user icon vs. the identicon, we could always send a X-User-Identicon-URL header which apps could choose to use or not.

kentonv: The nice thing about identicon.js is that it can run client-side and avoid sending anything over the network. As it stands, it actually produces poorly-compressed PNGs so I'm not sure I want them sent over the network as-is. Perhaps with some processing, we could in fact do this server-side and then start filling in `X-Sandstorm-User-Picture` on every request.

paulproteus: +1 to "punting for now"

paulproteus:
Wa​-hoo!

https://docs.sandstorm.io/en/latest/developing/auth/

PR #661, opened 2015-08-05
Always fchown the admin token to the sandstorm group.

dwrensha:
Today, if you do a default development install, SERVER_USER gets configured to a new user "sandstorm". If a non-root user gets added to the "sandstorm" group, they should be allowed to do `sandstorm admin-token` without sudo. Currently, the command succeeds, but it leaves an adminToken file that causes an error when the frontend tries to open it; the user ends up seeing the admin page stuck on "loading...".

This patch fixes that case and preserves the current functionality for the other cases we care about.


paulproteus:
Garply, retest this please

PR #662, opened 2015-08-05
Add `onlyPopup` parameter to sandstormTopbar

jparyani:
This will create a popup in the center of the screen that extends a
caret into the topbar. There will be no icon in the topbar associated
with these items.

### DIFF
--- a/shell/client/shell.scss
+++ b/shell/client/shell.scss
@@ -73,7 +73,6 @@ body {
 
 // Styles used by the user account settings page
 @import "_account.scss";
-
 // =======================================================================================
 
 a.copy-me {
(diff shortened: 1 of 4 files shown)


shell/client/shell.scss
```diff
@@ -616,6 +616,12 @@ body>.popup {
     >*:last-child {
       margin-bottom: 16px;
     }
+
+    &.centered {
+      margin: 0px auto;
+      width: 600px;
+      margin-left: calc(50% - 300px);
```
kentonv: You set margin left/right to `auto` above, and now you're setting it to calc. Since the element is `position: fixed` (on desktop), I think you actually don't want to do anything with margin here, but instead want: `left: calc(50% - 300px);`

It also seems like this styling is not what you want on mobile, so perhaps it should be moved into the desktop-specific section above?

kentonv:
I directly edited some things. Please review.

jparyani:
Fixed a syntax error with the `#unless` commit, and rebased with the new shell.scss changes. Ready to merge.

PR #663, opened 2015-08-05
Instantiate and display blackrock payments if it exists

jparyani:


ocdtrekkie:
I'm curious... is this just here to maintain as close of code parity with Blackrock as possible?

kentonv:
@ocdtrekkie More or less. Blackrock uses all the Sandstorm code, plus adds some additional code on top. We need to put some hooks into Sandstorm so that the additional functionality (like billing for Oasis) can be inserted without maintaining a fork. So that's what this PR does.

PR #668, opened 2015-08-07
Split shell.scss into several files

zarvox:
`shell.scss` is about two and a half thousand lines long, which is a lot to sift through.  Let's split things up into smaller chunks where possible.

This PR is primarily structural and should produce equivalent effective styles as the monolith.

Note that SCSS files starting with an underscore are considered "partials" and do not produce their own .css file as output.  The is ideal when you want to include the whole content of one SCSS file in another, which is precisely what we want.


zarvox:
FYI, I'm happy to rebase this atop other folks' style changes, since @kentonv mentioned there are some conflicts with other PRs.  Just let me know when is a good time to go through the rebasing effort. :)

kentonv:
Ready to rebase.

zarvox:
Rebased.

PR #674, opened 2015-08-10
Add docs on user accounts

paulproteus:


kentonv:
I think this kind of database tweaking is likely to break, or might even be broken already in certain sharing situations, so I'd rather not recommend it in the official docs. But I'm OK with suggesting this to people as a short-term solution on the mailing list, etc., perhaps by linking them to this pull request.

paulproteus:
If we think this is too brittle to put in the docs, then I'll close this pull request!

PR #676, opened 2015-08-12
Permit null as name

### LINKED ISSUE #675: Github auth broken in dev?
I'm using `vagrant-spk` and have and up-to-date dev install of sandstorm (at the time of writing).  I tried enabling Github auth in the admin console.  Setup works fine, but when I log out and try to log in using Github, login fails and I see this in `/opt/sandstorm/var/log/sandstorm.log`: 

```
Exception while invoking method 'login' Error: Match error: Expected string, got object in field name
    at checkSubtree (packages/check/match.js:159:1)
    at packages/check/match.js:296:1
    at Function._.each._.forEach (packages/underscore/underscore.js:113:1)
    at checkSubtree (packages/check/match.js:290:1)
    at check (packages/check/match.js:32:1)
    at configurations (packages/sandstorm-db/user.js:81:1)
    at Object.Accounts.insertUserDoc (packages/accounts-base/accounts_server.js:999:1)
    at Object.Accounts.updateOrCreateUserFromExternalService (packages/accounts-base/accounts_server.js:1164:1)
    at [object Object].Package (packages/sandstorm-accounts-oauth/oauth_server.js:55:1)
    at packages/accounts-base/accounts_server.js:358:1
Sanitized and reported to the client as: Match failed [400]
```

Looks like Github is returning something with a schema that Sandstorm(/Meteor?) doesn't expect.

paulproteus:
This occurs when a GitHub user attempts to sign in to Sandstorm where
they have no name configured in GitHub, just a username (aka handle).

Close #675

### DIFF
--- a/shell/packages/sandstorm-accounts-ui/accounts-ui-methods.js
+++ b/shell/packages/sandstorm-accounts-ui/accounts-ui-methods.js
@@ -26,7 +26,7 @@ Meteor.methods({
   updateProfile: function (profile) {
     // TODO(cleanup): This check also appears in sandstorm-db/users.js.
     check(profile, {
-      name: Match.Optional(Match.OneOf(null, Match.Optional(String))),
+      name: Match.OneOf(null, Match.Optional(String)),
       handle: Match.Optional(ValidHandle),
       pronoun: Match.Optional(Match.OneOf("male", "female", "neutral", "robot")),
     });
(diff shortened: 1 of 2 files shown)


paulproteus:
Garply, retest this please

paulproteus:
For what it's worth, https://github.com/sandstorm-io/sandstorm/issues/679 shows that having `profile.name` be null is not something we expect very often in the codebase.

I think that it's OK to merge this as-is, and fix those issues separately, but wanted to mention that. Assuming I can get the tests to pass, anyway.

PR #682, opened 2015-08-13
Extract identicons into sandstorm-identicons package

zarvox:


kentonv:
Please update the LICENSE file in the root directory which lists copyrights on these.

zarvox:
Good catch.

PR #685, opened 2015-08-13
WIP app index with `spk publish` command.

kentonv:
The app index is itself a Sandstorm app. It presents an API through which apps can be submitted (with the new `spk publish` being the client to this API) and presents a review interface to the admin. Once apps are approved, they are published via web publishing such that the app market client can download and display the index, and Sandstorm servers can install packages from the index.

Currently `spk publish` only works with a webkey pointing to an app index instance, but once we have a standard one up we'll embed the webkey directly into the spk source code.

This code is generally not done and is kind of messy at present, but it basically works.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ IMAGES= \
 # Meta rules
 
 .SUFFIXES:
-.PHONY: all install clean continuous shell-env fast deps bootstrap-ekam deps update-deps test installer-test
+.PHONY: all install clean continuous shell-env fast deps bootstrap-ekam deps update-deps test installer-test app-index-dev
 
 all: sandstorm-$(BUILD).tar.xz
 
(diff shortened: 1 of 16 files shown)


paulproteus:
non-showstopper-comment:

I'd personally prefer to have an app for Sandstorm not live in the
Sandstorm repo, if possible.

If it's best to have it live in the Sandstorm repo, I'd prefer to see a
directory called bundled-apps/ and see the app in there. I've watched a lot
of people look at new codebases for the first time, so this suggestion
comes from my opinions on how to minimally confuse new prospective
contributors to Sandstorm, as well as just minimizing "ls" noise for me the
next time we add a bundled app.

Other prospective bundled apps probably include powerbox backend apps, so I
think it's possible we'd see a few more apps in that directory.

PR #689, opened 2015-08-14
improvements to the sharing UI

### LINKED ISSUE #678: "Send an invite" should de-emphasize "See who has access"
In #677 I filed an issue about a UI element.

However, the first thing I tried when trying to send the email was clicking the **See who has access** button. I realize that's sort of insane of me, since it doesn't say **Send** but I'm tired, I woke up early to get on a plane, so I clicked the wrong button. In particular I clicked it because I expect the button in the bottom-right of a dialog to mean "Commit the action I have just configured through typing", like how there is a big green **Submit new issue** button in this GitHub issues submit form.

dwrensha:
Fixes #678. Removes the Sandstorm-Tabs package. Uses the new sharing icons. Handles the case when there are too many shares to fit on the screen in the Who Has Access view. Makes the Who Has Access view its own separate popup.


kentonv:
Merged. See tweaks in: 51347e50f49b93c1d67a7d29c3c645a93aa9f322

PR #688, opened 2015-08-14
Add new "new grain" UI (and supporting refactorings)

zarvox:
This patchset implements the new "new grain" UI.

PR #691, opened 2015-08-16
Add new "open grain" UI

zarvox:
Like the "new grain" UI, this is exposed at a route, but not linked by default.  The path to try out is `/grain`.

Depends on #688 for the crazy server-side reactive publish, icon helpers, and other fun helpers.

### DIFF
--- a/shell/.meteor/packages
+++ b/shell/.meteor/packages
@@ -24,5 +24,6 @@ sandstorm-db
 sandstorm-identicons
 sandstorm-ui-topbar
 sandstorm-ui-applist
+sandstorm-ui-grainlist
 fourseven:scss
 sandstorm-tabs
--- a/shell/.meteor/versions
+++ b/shell/.meteor/versions
@@ -72,6 +72,7 @@ sandstorm-email@0.1.0
 sandstorm-identicons@0.1.0
 sandstorm-tabs@0.1.0
 sandstorm-ui-applist@0.1.0
+sandstorm-ui-grainlist@0.1.0
 sandstorm-ui-topbar@0.1.0
 service-configuration@1.0.4
 session@1.1.0
--- a/shell/client/_grainlist.scss
+++ b/shell/client/_grainlist.scss
@@ -1,0 +1,47 @@
+.grain-list {
+  overflow: auto;
+  padding: 0px 32px;
+  >input.search-bar {
+    font-size: 16pt;
+    width: 100%;
+  }
+  >table {
+    width: 100%;
+    padding: 0;
+    margin: 8px 0 0 0;
+    border-collapse: collapse;
+    background-color: rgba(255, 255, 255, 0.5);
+    >thead>tr {
+      font-weight: 600;
+      border-bottom: 1px solid #aaa;
+      >td.td-app-icon {
+        padding-left: 16px;
+      }
+    }
+    >tbody>tr {
+      width: 100%;
+      height: 32px;
+      border-bottom: 1px solid #ddd;
+      >td {
+        cursor: pointer;
+        &.td-app-icon {
+          >img.app-icon {
+            width: 24px;
+            height: 24px;
+          }
+          width: 48px;
+          padding-left: 16px;
+        }
+        &.grain-name {
+          width: 400px;
+        }
+      }
+      &.grain {
+        cursor: pointer;
+        &:hover {
+          background-color: #f0f0f0;
+        }
+      }
+    }
+  }
+}
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -617,6 +617,12 @@ limitations under the License.
   </div>
 </template>
 
+<template name="selectGrain">
+  <div class="main-content">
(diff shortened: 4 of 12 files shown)


kentonv:
Added fixes for your future review, but merging now.

There are some big problems with the code handling "shared with me" grains which we'll need to deal with before launch:
- It leaks information to people whose access to the grains has been revoked.
- It leaks package IDs of private packages, inadvertently allowing sharing recipients to install said packages.
- Reporting that a grain has been used might be seen as a data leak in itself, depending on the app.
- (Less serious) The existing "shared with me" list intentionally still displays items that have been deleted upstream, because it's confusing to the user if they cannot find a document they knew existed before. Instead we let them click and then find out the grain is gone. We arguably should visually distinguish it in the list as well, but having the grain disappear entirely is problematic. (The recipient can explicitly remove things from their "shared with me" list at their leisure.)

I think it may be the case that we simply _cannot_ report up-to-date information to sharing recipients, because there's no efficient way to do it. We certainly don't want to compute whether or not you still have access to every single grain when you view your grain list; these access checks are relatively expensive.

Instead I think it makes sense to fall back to showing the last time the viewer themselves opened that particular grain, rather than the last time the grain executed. We can store this on the ApiToken. We may also need to store the app icon asset ID on the ApiToken, possibly updating it every time the token is exercised (and therefore verified to still have access).

In the future, I think we should develop an explicit API for apps to notify users of activity, being sure to specify what permission level is required to receive the notification, and without notifying anyone whose access has been revoked. But we'll have to punt on that for now.

BTW, I believe we were still planning to have separate lists for "my grains" and "shared with me", but it looks like currently you're merging them.

zarvox:
Thanks for all the helpful feedback!

Regarding leaking information to users with revoked API tokens: this leak is limited to the `lastUsed` field, right?  So this is fixed if we track and show the token owner's usage instead of the grain's usage?

Any thoughts on what should we do in cases where a user has multiple API tokens for the same grain?  Which token should this information be stored on?  (Ditto for grain petnames - right now I'm picking a token's title to show arbitrarily, but this seems wrong.)

I'm surprised to learn that the packageId of a grain that you have an API token is still intended to be a secret - I guess I hadn't realized that having a packageId was sufficient for a user to install the package, even if the .spk itself is not available to that user.  I would have expected such packages to have required the user upload the .spk again to get the appropriate userActions.  Are app IDs for grains you don't own (but do have access to) also considered secret, or just the package IDs?  If the former, then identicons break down, since those hash on app ID.

I guess the proposed alternative for needing access to package IDs to extract icons is denormalizing the asset IDs into the ApiTokens collection?

The rationale for keeping deleted shared grains in the sharee's list makes sense; I'll update this code accordingly.

+1 on storing the last time the viewer opened the grain on the APIToken itself, probably under owner.user with the other owner-specific data.

-0 on storing the asset ID on ApiToken and only updating it when the token is exercised - you'll wind up with stale icons if shared grains are updated, and it'll be inconsistent across even a user's own grains.  It also means that assets can't really be garbage-collected, even if no longer referenced by any extant packages or grains.  These downsides would be mitigated if we walk the ApiTokens table when a grain is updated and update any assetId replicas. where appropriate, which both can be a background task and should be relatively rare compared to reads.

As for separate lists for "my grains" and "shared with me": I've been under the impression we were merging them, since:

1. That's what I see in https://oasis.sandstorm.io/grain/CrZtrwsdLRXesfLfvscnsx/#3/14395029154948 and also the paper mockups we drew
2. Filter-search is less cluttered when matched with a single time-series
3. If you are use shared grains a lot, you will want quick access to them, and if you're not, then they will not be cluttering your grain list anyway.

If your concern is that it'll be hard for users to distinguish their own grains from those shared with them, I do expect to add something to help distinguish the two; likely in the "collaborators" column or something similar.  There's a couple of options in the mockups album for this.

paulproteus:
Just to reply to one thing here:

On Monday, August 17, 2015, Drew Fisher <notifications@github.com> wrote:

> As for separate lists for "my grains" and "shared with me": I've been
> under the impression we were merging them, since:
>
>    1. That's what I see in
>    https://oasis.sandstorm.io/grain/CrZtrwsdLRXesfLfvscnsx/#3/14395029154948
>    and also the paper mockups we drew
>    2. Filter-search is less cluttered when matched with a single
>    time-series
>    3. If you are use shared grains a lot, you will want quick access to
>    them, and if you're not, then they will not be cluttering your grain list
>    anyway.
>
>
I am massively +1 on merging "My grains" and "Shared with me". Almost all
the time, I want them merged. I frequently click on "Hacker CMS" and then
look at the list and go, "Where is the Sandstorm blog?"


I could also live with the default being called "All grains", and there
being some way to sub-categorize on "Mine" vs "Shared with me". I think I'd
never use it, though.

>

ocdtrekkie:
I think you may want some sort of icon modifier on "Shared With Me" grains, like a shortcut icon on Windows. Something like a little Share Icon in the corner of things that aren't yours.

PR #695, opened 2015-08-17
Prevent the role select element from getting too big.

dwrensha:
[The change here](https://github.com/sandstorm-io/sandstorm/commit/51347e50f49b93c1d67a7d29c3c645a93aa9f322#diff-f41c8458d337dd4e150b2ed41c120dc1R687) allows the following to happen:

too-big

This patch fixes it.

### DIFF
--- a/shell/client/_topbar.scss
+++ b/shell/client/_topbar.scss
@@ -639,7 +639,7 @@ body>.popup {
       text-transform: uppercase;
     }
     .share-token-role {
-      max-width: 40%;
+      width: 40%;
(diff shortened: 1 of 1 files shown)


kentonv:
Hmm. Maybe we should give it a fixed width. The reason I increased the other input size is because there was a lot of empty space. Ideally I'd like the text box to expand to fill all space available but I'm not sure if there's a way to do that.

dwrensha:
OK, updated to be a fixed width.

PR #697, opened 2015-08-18
Applist/grainlist cleanup

zarvox:
Addresses the infoleak issues from #691, and the infoleak and perf issues mentioned in #688.


src/sandstorm/supervisor.capnp
```diff
@@ -266,6 +266,24 @@ struct ApiTokenOwner {
...
       # Title as chosen by the user.
+
+      appTitle @8 :Util.LocalizedText;
+      # A copy of the app name for the corresponding UIView for presentation in the grain list.
+
+      lastUsed @9 :Int64;
+      # The last time the user used this API token with the associated grain, in milliseconds
+      # since the epoch (equivalent to javascript's new Date().getTime())
+
+      icon :group {
```
kentonv: It probably makes sense to define this as a separate struct rather than a group since it seems there's a decent chance we'll need to use it elsewhere. In fact it might even make sense to wrap all the new fields here in a struct called `DenormalizedGrainMetadata` or something like that.

kentonv: Also is appId missing? Should be in a union with icon, right?

src/sandstorm/supervisor.capnp
```diff
@@ -266,6 +266,24 @@ struct ApiTokenOwner {
...
+        assetFormat @10 :Text;
```
kentonv: Isn't this actually called just `format`, not `assetFormat`?

src/sandstorm/supervisor.capnp
zarvox: Seems like `lastUsed` and `title` are fields that owner.user.userId owns, whereas `appTitle`, `icon`, and `appId` are all denormalized grain information (which are owned by the grain owner).  I think I'd want to divide them on those lines.

And yes, appId should be in a union with icon.

kentonv: I agree.

zarvox:
Adjustments made - any further comments?

PR #698, opened 2015-08-18
Pass mongo command over stdin to work around MAX_ARG_STRLEN.

dwrensha:
After adding metadata to GitWeb, I was getting the following error:

```
sandstorm/run-bundle.c++:2335: failed: execv(args[0], const_cast<char**>(args.begin())): Argument list too long
```

This patch should fix the problem.

### DIFF
--- a/src/sandstorm/run-bundle.c++
+++ b/src/sandstorm/run-bundle.c++
@@ -2282,25 +2282,23 @@ private:
   }
 
   void mongoCommand(const Config& config, kj::StringPtr command, kj::StringPtr db = "meteor") {
-    pid_t pid = fork();
-    if (pid == 0) {
-      // We don't want to unwind the stack in this subprocess.
-      KJ_IF_MAYBE(exception, kj::runCatchingExceptions([&]() {
-        // Don't run as root.
-        dropPrivs(config.uids);
+    int pipeFds[2];
+    KJ_SYSCALL(pipe2(pipeFds, O_CLOEXEC));
+    kj::AutoCloseFd inPipe(pipeFds[0]);
+    kj::AutoCloseFd outPipe(pipeFds[1]);
 
-        execMongoClient(config, {"--quiet", "--eval", command }, db);
-      })) {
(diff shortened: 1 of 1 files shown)


dwrensha:
Hm. Apparently Mongo swallows errors when invoked this way, and always returns an exit code of 0. I think we may have to use temp files instead...

dwrensha:
OK, updated to use temporary files. I've verified that this works for large commands, and that malformed commands produce an error that gets propagated correctly.

dwrensha:
Also, after this change, `spk dev` no longer prints out "WriteResult(...)".

src/sandstorm/run-bundle.c++
```diff
@@ -2282,29 +2282,27 @@ private:
...
-      })) {
-        context.exitError(kj::str(*exception));
-      }
+    char commandFile[] = "/tmp/mongo-command.XXXXXX";
+    int commandFd;
+    KJ_SYSCALL(commandFd = mkstemp(commandFile));
+    KJ_DEFER(unlink(commandFile));
+    KJ_SYSCALL(fchown(commandFd, -1, config.uids.gid));
+    KJ_SYSCALL(fchmod(commandFd, 0660));
+    kj::FdOutputStream(kj::AutoCloseFd(commandFd)).write(command.begin(), command.size());
```
kentonv: Exception safety: It's important that `commandFd` be claimed by an `AutoCloseFd` immediately after the fd is created, before anything else that could possibly throw is executed. Otherwise the FD will leak on exceptions.

src/sandstorm/run-bundle.c++
```diff
@@ -2282,29 +2282,27 @@ private:
...
+
+    Subprocess process([&config, db, commandFile, this]() -> int {
```
kentonv: Might as well just use `[&]` since the callback is effectively called with the whole stack frame still in-scope.

dwrensha:
Fixed now.

paulproteus:
Yay!

PR #710, opened 2015-08-23
Docs: Reverse Proxy

vezquex:
Documentation for an HTTPS reverse-proxy with NGINX.

Figured it may be useful for others.

### DIFF
--- a/docs/administering/reverse-proxy.md
+++ b/docs/administering/reverse-proxy.md
@@ -1,3 +1,8 @@
+This document helps you serve Sandstorm with HTTPS and take the port (6080) out the URL.
+
+[nginx](http://nginx.org/en/) will listen on port 80 to redirect HTTP to HTTPS on port 443.
+On port 443, it will reverse proxy Sandstorm on port 6080.
+
 ## Prerequisites
 Create DNS entries for `example.com` and `*.example.com`.
 
@@ -7,7 +12,7 @@ Install `nginx` with your package manager.
 
 ## Configuration
 
-### NGINX
+### nginx
 Copy [nginx-example.conf](https://github.com/sandstorm-io/sandstorm/blob/master/nginx-example.conf) to `/etc/nginx/sites-enabled/`.
 
 `nginx-example.conf` may be renamed to anything, such as `example.com.conf`.
@@ -28,7 +33,7 @@ ssl_certificate /etc/keys/example.com.crt;
 ssl_certificate_key /etc/keys/example.com.key;
 ```
 
-Test your NGINX configuration:
+Test your nginx configuration:
 `sudo nginx -t`
 
 ### Sandstorm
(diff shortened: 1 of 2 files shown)


garply:

paulproteus:
Garply, ok to test

paulproteus:
Hi @vezquex ,

First of all, please ignore the build failure. It seems to be caused by flaky tests, not by you.

Thanks for submitting this pull request! It looks really good. This is a question that many people ask, and I'm grateful that you were bold enough to submit a fix to the documentation so that people can fix this.

I have some content feedback, none of which are show-stoppers:

* Please add this to `mkdocs.yml` at the top level of the repository so that it shows up in the docs navigation.

* Please add a note to the top of the document explaining the purpose of this. Perhaps this sentence:

> This document helps you set up Sandstorm listening on port 443 and port 80, rather than 6080 (the default). Those are the default ports for HTTPS and HTTP, so users will be able to type `example.com` into their address bars and be redirected to a HTTPS Sandstorm install.

* For the HTTPS certificate authority I personally recently used, I can use the same certificate for `example.com` and `*.example.com`. I'd prefer that this documentation follow that design, since it's slightly easier for users.

* Consider setting some constraints on which ciphers and HTTPS versions (TLS versions) are allowed. We have some suggestions in this file: https://github.com/sandstorm-io/sandstorm/blob/master/nginx-example.conf - I hope you'll integrate them into the docs.

One other thing -- for significant changes, we need you to fill out a form so that we can use your work in the main Sandstorm project: https://sandstorm.io/cla/ . The contributor license agreement (CLA) includes a requirement that you're OK with us re-licensing this code under a new license, which we want to be able to do in case there is a problem with the Apache License 2.0 and we need to e.g. upgrade to Apache License 3.0, if the time comes for that. This is the same CLA that is used by the Apache Software Foundation, so it's hopefully something that you'll be OK with.

Thank you so much, @vezquex , and I'm looking forward to hearing back from you!

vezquex:
Thanks for the feedback.

Generating a multi-domain certificate was tricky for me. I think that warrants a guide as well.

Thanks for telling me about `nginx-example.conf`. I didn't think to look for it.

I've signed the contributor license agreement.

docs/administering/reverse-proxy.md
```diff
@@ -0,0 +1,54 @@
...
+```
+BASE_URL=https://example.com
+WILDCARD_HOST=*.example.com
+```
+
+## Run
+Finally, start NGINX, and restart Sandstorm to use the new config.
+
+```bash
+sudo nginx
```
paulproteus: On my system, I think this has to be:

```
sudo service nginx restart
```

I can make this change, if you're OK with it, or you can, or if it's not the right change, then we can figure something else out.

vezquex: OK

paulproteus:
Hi @vezquex ,

Your changes are great. Thank you so much for writing this and contributing to the project! I'll merge this right away, and update the docs.

paulproteus:
You can see your changes here! https://docs.sandstorm.io/en/latest/administering/reverse-proxy/

PR #714, opened 2015-08-24
Shell redesign

zarvox:
Not ready for merge yet, but a central place to track feedback, issues, and patches.


shell/packages/sandstorm-ui-topbar/topbar.js
```diff
@@ -45,6 +45,7 @@ Template.sandstormTopbarBlockReload.onDestroyed(function () {
 
 Template.sandstormTopbar.onCreated(function () {
   Template.instance().popupPosition = new ReactiveVar(undefined, _.isEqual);
+  // TODO: subscribe to the list of open grains for this session
```
kentonv: If this is still relevant, add parenthetical (e.g. `TODO(soon)`) and add more details.

shell/packages/sandstorm-ui-topbar/topbar.html
```diff
@@ -24,24 +24,43 @@
...
-              {{> template data.get}}
-            {{/if}}
+              {{/if}}
+            </li>
+          {{/unless}}
+        {{/each}}
+      </ul>
+      <ul class="navbar {{#unless _showSidebar.get}}hide-desktop{{/unless}}">
+        <li class="navitem-open-grain {{#if currentRoute 'selectGrain'}}current{{/if}}"><a href="{{pathFor route='selectGrain'}}"><span>Open...</span></a></li>
+        <li class="navitem-create-grain {{#if currentRoute 'newGrain'}}current{{/if}}"><a href="{{pathFor route='newGrain'}}">New...</a></li>
```
dwrensha: These items also end up appearing under a hamburger menu inside the Debug Log window!

zarvox: Good catch.

Do we actually need a topbar at all for the debug log?  I'm inclined to remove the topbar from lightLayout entirely, or maybe replace it with a single line with the grain title.

(my usual experience with the debug log is that I only look at it in dev mode, and then I wind up opening a new one every time I make a new grain, and then I have like 15 indistinguishable dev mode debug log windows)

paulproteus:
You might think I'm joking, but:

I :heart: commits named `WIP`.

shell/lib/db-deprecated.js
```diff
@@ -65,15 +65,43 @@ if (Meteor.isServer) {
...
-    set: function (value) {
-      Session.set("topbar-expanded", value);
+    this.set = function(value) {
+      this._val = value;
+      this._dep.changed();
     }
-  });
+    return this;
+  };
+  globalGrains = new FancyReactiveVar([]);
```
dwrensha: It'd be nice to have some documentation about the type of this variable. An array of ... what?

zarvox: Of GrainView objects - see new file `client/grainview.js`.  I'll add a comment.

(FYI, it turns out a plain ReactiveVar is adequate here.  I was confused when I was trying to put things in Session and they were coming out stripped of their prototype functions.  I think Kenton has a commit to simplify that queued.)

shell/packages/sandstorm-ui-topbar/topbar.js
```diff
@@ -159,8 +183,53 @@ Template.sandstormTopbar.events({
     Template.instance().data.closePopup();
   },
 
+  "click .toggle-sidebar": function (event) {
+    var topbar = Template.instance().data;
```
dwrensha: Would it be possible to animate the sidebar-toggling, so that it doesn't happen instantaneously? I'm getting some whiplash when I click on it.

zarvox: Maybe, though it's a little trickier than "just add a CSS animation to the show/hide of the sidebar" because you also have to resize the rest of the main-content.  Also, the main-content resizes horizontally, whereas we'd probably want to show the sidebar dropping from the top (or raising up into it), since the control for toggling it lives in the topbar.  I wasn't sure what a coordinated animation should look like, so I punted. :)

Also since this winds up reflowing all the grain frames, animating them is pretty expensive here.

I can try a couple options and see if anything feels better - I'd probably try "just animate sidebar and let content reflow immediately) and "try animating them both at the same time in perpendicular directions."  I'm sensitive to the performance implications when there's a bunch of grains, though.

shell/packages/sandstorm-ui-topbar/topbar.html
dwrensha: I do think that it's important for the debug window to show the grain title. I don't have strong opinions about how that title is displayed.

shell/shared/grain.js
```diff
@@ -559,11 +666,32 @@ if (Meteor.isClient) {
...
+    },
+    generatedApiToken: function () {
+      var current = getActiveGrain(globalGrains.get());
+      return current && current.generatedApiToken();
+    },
+    generatedApiTokenPending: function() {
+      var current = getActiveGrain(globalGrains.get());
+      return (current && current.generatedApiToken()) == "pending";
+    },
+    viewInfo: function () {
```
dwrensha: The shareWithOthers template needs a helper like this too, so that the role selection element shows up there.

ocdtrekkie:
+1 on wariness of animation performance. And as far as whiplash, the multi-directional animating of the G+ top bar was making people motion sick for a bit.

shell/shared/grain.js
zarvox: Added some helpers in b115a0208a406ed8d46ceeb8c4828ffb7d209420

shell/packages/sandstorm-ui-topbar/topbar.js
zarvox: removed in 257ebc9b9fc7d01c0d64bb2897de7450a0493cc3

shell/packages/sandstorm-ui-topbar/topbar.html
zarvox: All right, hopefully 63936876328226ad2ea638af2e8548cfab5fcd24 floats everyone's :boat:

kentonv:
Merging for testing purposes. Keep reviewing and iterating, though.

shell/client/shell.html
```diff
@@ -94,15 +95,46 @@
...
   {{/if}}
   {{/with}}
 </template>
 
-<template name="lightLayout">
-  {{>sandstormTopbar globalTopbar}}
+<template name="grainView">
+  {{!-- This template is rendered manually using Blaze.renderWithData() for each currently-open
+        grain and then injected into .main-content. We do things manually so that we can be
+        sure that adding or removing grains from teh open grain list does not cause other grains'
```
dwrensha: s/teh/the/

dwrensha:
Is "navbar" a synonym for "sidebar"? If so, can we stick to using just one of the terms?

dwrensha:
In the mobile view, the grain topbar/menubar buttons are only readable upon mouseover.

mobileview

shell/shared/grain.js
```diff
@@ -1148,52 +1312,68 @@ Router.map(function () {
...
+            };
+            openView();
           } else {
-            Session.set("visit-token-" + this.params.token, "redeem");
+            makeGrainIdActive(grainId);
           }
+
+          var currentGrain = getActiveGrain(grains);
+          if (!currentGrain.shouldShowInterstitial() && currentGrain.sessionStatus() == "closed") {
+            currentGrain.openSession();
```
dwrensha: I'm having trouble understanding what's going on here. I thought that the openView() thing above took care of opening the session. Why do we need this also?

zarvox: Perhaps `openSession()` should be renamed "tryOpenSession" - it will not actually open the session if `shouldShowInterstitial()` is true (because we need the user's input to decide whether incognito should be true or false in the newApiTokenSession).  To make sure that we attempt to open the session as soon as that value becomes false, we need this branch, since `openView()` is only called when we hadn't created a GrainView yet.

(Even better, we should guard the `openSession()` call in `openView()` with a check to `shouldShowInterstitial()`.  I think I had that check in the original code, and lost it in the openView rework after Kenton did the explicit Blaze rendering, which requires we be able to get a handle to the parent DOM node.)

zarvox:
Navbar and sidebar mean the same thing.  I prefer the term navbar, since on mobile it's not on the side, and its functional purpose is navigation, whereas the "topbar" is functionally a menu bar primarily for the current grain.  The functionality of the two are collapsed into a single list on mobile.  I didn't realize this at first, and had already used "sidebar" in most places, since that was what we had talked about previously, but I'd prefer to use "navbar" moving forward.

(Also everything that says "topbar" really means "topbar and sidebar" aka "menubar and navbar".  I'd love a better noun for the combination of the two.  Chrome?)

Addressed in 10f24fe3035013d605331d4117c7e0ef7edab950

shell/shared/grain.js
```diff
@@ -1148,52 +1312,68 @@ Router.map(function () {
...
-          var apiToken = tokenInfo.apiToken;
-          if (!Grains.findOne({_id: apiToken.grainId, userId: Meteor.userId()}) &&
-              !ApiTokens.findOne({userId: apiToken.userId, "owner.user.userId": Meteor.userId()})) {
+    onRun: function () {
+      var token = this.params.token;
+      var path = this.params.path;
+      var query = this.params.query;
+      var hash = this.params.hash;
+
+      Tracker.autorun(function() {
```
dwrensha: Please document the reasons for adding this reactive context. My understanding is that you need it to wait for the `TokenInfo` data to arrive, but it sounds like you might also be using it to retry `openSession()`?

shell/shared/grain.js
dwrensha: I think the right thing here is for `openSession()` to block on the `GrainView._revealIdentity` variable while it waits for the user to click on an interstitial button. Then we'll only ever need to call `openSession()` once. I'm working on a patch that will do this...

shell/shared/grain.js
dwrensha: Er, maybe we don't need it for `TokenInfo`, because that subscription is in the `Route.waitOn` list.

zarvox: I need TokenInfo to arrive, and I also need Grains to arrive, and I need ApiTokens to arrive (because they may prevent me from needing to prompt the user with an interstitial), and I need to make sure that GrainView is marked active once it's created, and I need to open a session, but only if either the user has answered the interstitial or the Grain/ApiToken tables prove it's not needed.

zarvox: But mostly it's that last one - I want to automatically open the session as soon as the user answers the "reveal identity" question, or automatically as soon as data arrives that shows my identity is already revealed, or that I own the grain.

shell/client/grainview.js
```diff
@@ -0,0 +1,482 @@
...
+
+  this._grainId = grainId;
+  this._originalPath = path;
+  this._originalQuery = query;
+  this._originalHash = hash;
+  this._token = token;
+
+  this._status = "closed";
+  this._revealIdentity = undefined; // set to true or false to make explicit
+  this._dep = new Tracker.Dependency();
```
dwrensha: If you made this object's mutable members instances of `ReactiveVar`, would that eliminate the need for this manual Dependency? Would that make things simpler?

zarvox: I guess we could make each of them a `ReactiveVar`, in hindsight.  It might also make reactive recomputation more granular.

On the other hand, in a couple places it's nice to be able to batch `changed()` calls, particularly when you want certain changes to be observed atomically.  I'd want to be sure this._status and this._error change "together" as viewed from outside, and probably the same for `_status`, `_sessionId`, and `_title`.

I'm not sure what guarantees Tracker gives about recomputation order - when you emit a `changed()`, does it let the current context finish running before running any of the rescheduled computations, or does it do work/recomputations as `changed()` is called?

I'm okay switching the mutable variables to ReactiveVar so long as you won't see inconsistent states externally - it'd certainly help keep us from missing `depend()` and `changed()` calls when we add code.

PR #720, opened 2015-08-25
Continuing shell redesign 

zarvox:
I'm going to collect commits that address issues with the shell redesign here.


zarvox:
Garply, retest this please

PR #721, opened 2015-08-25
Fix incognito interstitial.

dwrensha:


dwrensha:
I've added a commit that fixes the log-in-while-visiting-a-shared-link problem. Apparently `onRun()` doesn't fire again after the login redirect. (I'm not sure how that's possible. It's perhaps related to https://github.com/iron-meteor/iron-router/issues/1219 .) So I've switched to `onBeforeHook()`, which has the advantage of meaning that we don't need to wait for `this.ready()`, but has the disadvantage of requiring use to manual enforce the run-only-once semantics we want.

PR #723, opened 2015-08-26
onBeforeAction() -> onRun() in the /grain route, for now

dwrensha:
Using `onBeforeAction()` here somehow breaks multigrain. This is a quick fix while I investigate the cause.

### DIFF
--- a/shell/shared/grain.js
+++ b/shell/shared/grain.js
@@ -1267,11 +1267,7 @@ Router.map(function () {
       return subscriptions;
     },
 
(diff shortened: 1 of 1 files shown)


dwrensha:
Apparently the problem here was that the `RouteController`'s `state` member can be *shared* between multiple runs of a route! Like, we navigate from `/grain/abc` to `/grain/xyz`, and that counts as a new *run* of the route, but the `RouteController` is reused, and thus the `state` member is not reset.

PR #725, opened 2015-08-26
consistent run-only-once strategies in the /grain/<grainId> and /shared/<token> routes

dwrensha:


shell/shared/grain.js
```diff
@@ -1300,6 +1304,7 @@ Router.map(function () {
     },
 
     onStop: function () {
+      this.state.set("beforeActionHookRan", undefined);
```
paulproteus: Is there idea on this line to have onStop() set this value to (basically) false? If so it's arguably cleaner to use a real boolean of `true` and `false`, and then where you call `this.state.get()` check if the return value is truthy.

I can live with this as-is, but mostly just wanted to say that intentionally using `undefined` as a value in Javascript weirds me out a little. (-:

dwrensha: As far as I know, this is the closest possible approximation to removing an entry from the `ReactiveDict`. We use this pattern elsewhere. https://github.com/sandstorm-io/sandstorm/blob/25796cffb46f9a8c77fb270c92aa7363c113af5e/shell/shared/shell.js#L785

It would work just as well to set it to `false`. I have no strong opinion one way or the other.

paulproteus: Yeah, I don't ​know if I have a strong opinion either way. Consider me +0
on this commit as-is!

kentonv:
I was surprised to learn recently that `onBeforeAction()` runs _after_ `data()`, FWIW.

dwrensha:
> I was surprised to learn recently that onBeforeAction() runs after data(), FWIW.

Yeah, `data()` gets run every time. `action()` only gets run once `this.ready()` is true.

PR #727, opened 2015-08-26
Add a grain container div above the grain frame... 

dwrensha:
...so that errors, loading spinners, and interstitials only get shown when their corresponding grain is selected.

Also, display "unknown grain" in the navbar for /shared/ links that are still on the incognito interstitial and thus have no title yet.

### DIFF
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -109,28 +109,30 @@ limitations under the License.
         sure that adding or removing grains from teh open grain list does not cause other grains'
         iframes to be re-rendered, losing state. --}}
   {{#with unpackedGrainState}}
-  {{#if error}}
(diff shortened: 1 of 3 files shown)


kentonv:
@zarvox Please review and merge if/when you're happy with this.

dwrensha:
> Also, display "unknown grain" in the navbar for /shared/ links that are still on the incognito interstitial and thus have no title yet.

Note that without this, the navbar links have zero height and are thus unclickable.

zarvox:
@dwrensha I'm surprised to hear that the navbar link is zero height without a title.  It should be `display: block` and `height: 32px` and clickable everywhere in the rectangle (that isn't the close button).

PR #729, opened 2015-08-26
Only show admin links list ifAdmin

paulproteus:
Testing done:

* Manually tested with a non-admin dev account: got no admin link.

* Manually tested with an admin dev account: got the admin link.

### DIFF
--- a/shell/packages/sandstorm-accounts-ui/login_buttons.js
+++ b/shell/packages/sandstorm-accounts-ui/login_buttons.js
@@ -5,6 +5,9 @@ var helpers = {
   isDemoUser: function () {
     return this._db.isDemoUser();
   },
+  isAdmin: function () {
+    return this._db.isAdmin();
+  },
   demoTimeLeft: function () {
     var ms = Meteor.user().expires.getTime() - Date.now();
     var sec = Math.floor(ms / 1000) % 60;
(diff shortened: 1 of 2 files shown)


shell/packages/sandstorm-accounts-ui/login_buttons.js
```diff
@@ -5,6 +5,9 @@ var helpers = {
   isDemoUser: function () {
     return this._db.isDemoUser();
   },
+  isAdmin: function () {
+    return Meteor.user().isAdmin;
```
paulproteus: Question to e.g. Drew: I notice we're access `this._db` but it seems more idiomatic to access `Meteor.user()`, so I went with what I think is more idiomatic.

Should I do things more like `isDemoUser` and access `this._db` instead?

zarvox: Prefer `return this._db.isAdmin()` here.

zarvox: To be more clear: in the longer future, we intend to have SandstormDb wrap all accesses to the collections, and that includes Meteor's internal collections, so we prefer to wrap that here.

It should make it easier for us to 1) mock out SandstormDb in unit tests and 2) eventually shift away from Mongo someday, as described in [The Great Frontend Refactor](https://oasis.sandstorm.io/grain/kj9fopjyT2CdqC3i7fZN2z).

paulproteus: Such +1. Thanks for taking the time to explain

paulproteus:
@zarvox I believe this is ready to merge, as soon as Jenkins says +1.

paulproteus:
Also yay!​

PR #734, opened 2015-08-26
Fix grain title issues for shared grains and public grains.

dwrensha:
1. Fixes the Share button for old-sharing-model (i.e. public) grains.
  2. Fixes grain-title-setting for shared grains, and adds a prompt that better explains what's happening.
  3. Gets rid of the `Grains.allow()` rule, in favor of Meteor methods `privatizeGrain()` and `updateGrainTitle()`
  4. Shows the correct title for public grains opened by non-owners.
  5. Deletes some obsolete `showMenu` stuff.

### DIFF
--- a/shell/shared/grain.js
+++ b/shell/shared/grain.js
@@ -21,15 +21,6 @@ var DEFAULT_TITLE = "Sandstorm";
 if (Meteor.isServer) {
   var Crypto = Npm.require("crypto");
 
-  Grains.allow({
-    update: function (userId, grain, fieldNames) {
-      // Allow owner to rename or privatize grain.
-      return userId && grain.userId === userId &&
-          ((fieldNames.length === 1 && fieldNames[0] === "title")
-           || (fieldNames.length === 1 && fieldNames[0] === "private"));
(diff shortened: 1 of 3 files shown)


shell/shared/grain.js
```diff
@@ -303,11 +319,15 @@ if (Meteor.isClient) {
...
-      var title = window.prompt("Set new title:", this.title);
-      if (title) {
-        var g = getActiveGrain(globalGrains.get());
-        if (g) {
-          g.setTitle(title);
+      var grain = getActiveGrain(globalGrains.get());
+      if (grain) {
+        var prompt = "Set new title:";
+        if (!grain.isOwner()) {
+          prompt = "Set a new personal title: (does not change the owner's title for this grain)";
```
zarvox: That might be a little verbose, but at least it sets expectations better.

I'm mildly worried how we'll express this if we move to making this an edit-inline field.  But I guess we can cross that bridge when we get there. :)

paulproteus:
Huge yippee for setting expectations better!​

PR #742, opened 2015-08-27
Fix grainlog scrolling on Firefox. Don't scroll the titlebar.

dwrensha:


dwrensha:
Garply, retest this please

PR #744, opened 2015-08-27
Changes for non-logged-in users

paulproteus:
Goal:

- [x] When not logged in, see a message saying "Don't lose this link! Sign up for free >>" to encourage people to sign up.

- [x] When not logged in, disable the navbar.

- [x] When not logged in, disable clicking the Sandstorm button in the top-left because normally that would enable/disable the nav bar, but now it would do nothing, since non-logged-in users can't see the nav bar.

Status:

* Ready for review.

### DIFF
--- a/shell/packages/sandstorm-accounts-ui/login_buttons.html
+++ b/shell/packages/sandstorm-accounts-ui/login_buttons.html
@@ -16,7 +16,7 @@
       {{/if}}
     {{/if}}
   {{else}}
-    {{#if currentRoute 'shared'}}
+    {{#if isCurrentRoute 'shared'}}
       <a>Don't lose this link! Sign up for free ▾</a>
     {{else}}
       <a>Sign in ▾</a>
--- a/shell/packages/sandstorm-accounts-ui/login_buttons.html
+++ b/shell/packages/sandstorm-accounts-ui/login_buttons.html
@@ -16,7 +16,11 @@
       {{/if}}
     {{/if}}
   {{else}}
-    <a>Sign in ▾</a>
+    {{#if currentRoute 'shared'}}
+      <a>Don't lose this link! Sign up for free ▾</a>
+    {{else}}
+      <a>Sign in ▾</a>
+    {{/if}}
   {{/if}}
 </template>
 
--- a/shell/packages/sandstorm-accounts-ui/login_buttons.js
+++ b/shell/packages/sandstorm-accounts-ui/login_buttons.js
@@ -2,6 +2,9 @@
 var loginButtonsSession = Accounts._loginButtonsSession;
 
 var helpers = {
+  currentRoute: function (routeName) {
+    return Router.current().route.getName() == routeName;
+  },
   isDemoUser: function () {
     return this._db.isDemoUser();
   },
--- a/shell/packages/sandstorm-accounts-ui/login_buttons.js
+++ b/shell/packages/sandstorm-accounts-ui/login_buttons.js
@@ -2,7 +2,7 @@
 var loginButtonsSession = Accounts._loginButtonsSession;
 
 var helpers = {
-  currentRoute: function (routeName) {
+  isCurrentRoute: function (routeName) {
     return Router.current().route.getName() == routeName;
   },
   isDemoUser: function () {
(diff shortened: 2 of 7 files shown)


shell/packages/sandstorm-accounts-ui/login_buttons.js
```diff
@@ -2,6 +2,9 @@
 var loginButtonsSession = Accounts._loginButtonsSession;
 
 var helpers = {
+  currentRoute: function (routeName) {
```
zarvox: Please name helpers that return booleans `isCondition`, to help make its behavior in the context where it's being called.

For instance, I'd expect a function named `currentRoute` to return `Router.current().route` as a string.

zarvox: (I recognize I goofed this up in the sandstorm-ui-topbar package, and you were probably just following that bad example.  I should probably fix it there too.)

paulproteus: I like that idea but I did copy-paste this thing from elsewhere.

Two options:

1. Keep it this name, to be consistent with the other route, or

2. Adjust the name here *and* everywhere

3. Adjust the name here, but leave everywhere else in its current
non-idiomaticness



I'll go for option 2. I don't mind a little refactoring while making
changes.

zarvox: Thanks :)

zarvox:
:+1:

shell/packages/sandstorm-accounts-ui/login_buttons.html
```diff
@@ -16,7 +16,11 @@
       {{/if}}
     {{/if}}
   {{else}}
-    <a>Sign in ▾</a>
+    {{#if isCurrentRoute 'shared'}}
+      <a>Don't lose this link! Sign up for free ▾</a>
```
kentonv: "for free" seems confusing on personal servers.

shell/packages/sandstorm-accounts-ui/login_buttons.js
```diff
@@ -2,6 +2,9 @@
 var loginButtonsSession = Accounts._loginButtonsSession;
 
 var helpers = {
+  isCurrentRoute: function (routeName) {
+    return Router.current().route.getName() == routeName;
```
kentonv: This introduces a reference to a deprecated global into a package, which we were trying not to do...

shell/packages/sandstorm-accounts-ui/login_buttons.html
paulproteus: Yeah, I was thinking it might makes sense to make this an admin setting.

paulproteus: On personal servers, fwiw, it's not super-duper clear to a random visitor that it's a "personal server". Imagine me giving a link to rose.sandcats.io to a friend who has no idea what Sandstorm is, but now knows that rose.sandcats.io is some kind of thing that has collaborative text documents. Might as well encourage them to sign in there, too.

So IMHO emphasizing that signing up is "free" is likely a good idea even there.

shell/packages/sandstorm-accounts-ui/login_buttons.js
paulproteus: What's the best way to do this instead?

paulproteus: FWIW (not really an excuse) I am copy-pasta-ing from `topbar.js`. I am happy to clean up both in the same commit.

zarvox: I think I wrote the original code, so I'm probably the guilty party here, sorry.

I am curious what we should be doing instead - inject an instance of the Router into everything that might care about it?  (I've never actually seen an instance of iron router, only the global Router.)

kentonv: Well, we're aiming to move off of iron router, so ideally we design some abstract interface that provides the functionality we want and pass that down -- something that we can re-implement in terms of some other router later.

PR #746, opened 2015-08-27
Remove color rule from %hamburger-menu-item

paulproteus:
This commit removes "color: black;" from hamburger-menu-item.

Only a few items remain black, such as one's username, since other
items seem to have extra styling that overrides this rule.

But black text on a (nearly) black background doesn't make much sense
anyway. So this commit simply removes that style rule.


zarvox:
This is fine.

From looking at Nena's latest screens for mobile, we're going to switch to the light-on-black theme for the menu/nav there, which will be consistent with the desktop styling.  We can adjust the styling of the individual nav items next.

PR #747, opened 2015-08-27
Use max-width not width on form.account-profile-editor

paulproteus:
This allows it to scale nicely when the browser window is not very wide.

@kentonv it's usually best to only add `width:` rules if you know that the browser is going to be that wide (`@media desktop...`) or to use `max-width`. I prefer the latter generally.

### DIFF
--- a/shell/client/_account.scss
+++ b/shell/client/_account.scss
@@ -583,7 +583,7 @@ body>.main-content>.account {
 }
 
 form.account-profile-editor {
-  width: 500px;
+  max-width: 500px;
   height: auto;
   background-color: white;
   padding: 0 16px;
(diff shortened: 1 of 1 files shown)


zarvox:
@kentonv mind reviewing/merging this one, since you worked on the accounts page last?  Thanks!

kentonv:
I don't agree: `max-width` is basically only safe when the content is purely paragraph-form text and can therefore wrap at arbitrary places without looking bad. With a carefully laid-out form like we have here, if you try to contract it you end up with all kinds of broken layout. I think we need explicit mobile styling here.

I'll work on it.

kentonv:
This is fixed now.

PR #748, opened 2015-08-27
Display 'Invalid token' if a share link is invalid.

dwrensha:


dwrensha:
The `<pre>` is for consistency with errors in the "grainView" template.

zarvox:
LGTM.

PR #751, opened 2015-08-27
Use DIV and background-image instead of IMG

paulproteus:
This ensures Internet Explorer can rescale the SVGs.

### DIFF
--- a/shell/packages/sandstorm-ui-applist/applist.html
+++ b/shell/packages/sandstorm-ui-applist/applist.html
@@ -8,7 +8,7 @@
       <h2 class="">Most used</h2>
       {{#each popularActions }}
       <button class="app-button app-action {{#if searching}}hide{{/if}}">
-        <img class="app-icon" src="{{ iconSrc }}">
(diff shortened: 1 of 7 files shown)


shell/packages/sandstorm-ui-applist/applist.html
```diff
@@ -16,23 +16,23 @@ <h2 class="">Most used</h2>
     </div>
     <h2 class="all-apps">All apps</h2>
     <a class="app-button install-button {{#if searching}}hide{{/if}}" href="https://apps.sandstorm.io/?host={{ origin }}">
-      <img class="app-icon pseudoapp" src="/install.svg">
+      <div class="app-icon pseudoapp" style="background-image: url('/install.svg');"></div>
```
zarvox: I suppose we could theoretically define CSS classes for each of these, but I'm honestly not sure that's worth the effort here.

If we really want to clean it up, we could probably make these be background-images for `install-button`, `upload-button`, and `restore-button` and drop the <div> entirely.

zarvox:
yaaaaaaaaaay!

paulproteus:
Garply, retest this please

PR #755, opened 2015-08-27
Purple install icon, pronoun icon

neynah:
@paulproteus


kentonv:
We actually don't need to submit a purple icon, because we should instead use a make rule that searches-and-replaces the color to whatever we need with the existing icon as input.

As for the pronoun icon, it appears to be an outline rather than a silhouette like all the others. Was that intended?

kentonv:
(Closing for now to remove from the queue.)

kpreid:
> We actually don't need to submit a purple icon, because we should instead use a make rule that searches-and-replaces the color to whatever we need with the existing icon as input.

With the proper use of SVG, it should be possible to make color variants of an SVG icon by having one SVG include another. Or at the worst, by _inserting_ a stylesheet, as opposed to find-and-replace.

neynah:
@kentonv Yes that's intended for the pronoun icon. It looks the most balanced and speech bubble-like.

@paulproteus so I think this means I won't need to upload different colours of existing SVGs.

paulproteus:
+1 to no need to upload different colors of existing SVGs; I didn't realize
how cool our tooling is.​

PR #757, opened 2015-08-27
Hide a header that is no longer true while you are searching

paulproteus:


zarvox:
I love merging usability improvements.

paulproteus:
: D​

I love making them!

PR #759, opened 2015-08-27
Fix styling of incognito interstitial.

dwrensha:


zarvox:
@dwrensha does this also fix the issue with the incognito interstitial conflicting with the first-login accounts interstitial?

shell/client/shell.scss
```diff
@@ -190,22 +190,21 @@ img {
...
-  border-radius: 4px;
-  border: none;
-  color: white;
-  cursor: pointer;
+  background-color: $default-content-background-color;
+  & button {
+    font-family: inherit;
+    font-size: inherit;
+    font-weight: 600;
+    background-color: #428bca;
```
zarvox: We should probably ask @neynah to write up a color scheme for buttons like these some time, but that time need not be now.

dwrensha:
#762 fixes that.

PR #761, opened 2015-08-27
Add install-purple.svg to list of SVGs in Makefile

paulproteus:
In #755 @neynah adds a new SVG. This does the right thing (I believe!) for adding it to the Makefile, so we actually distribute it.


kentonv:
install-purple.svg should be generated from install.svg rather than submitted as a separate file.

PR #762, opened 2015-08-27
Don't use the main-content class for first signup, as grainview special-cases it.

dwrensha:
Currently, if you still need to confirm your profile but you visit a grain or a shared link, the grainview gets rendered below the profile prompt (scroll down and you'll see it!) and then gets destroyed once the profile prompt is dismissed. The problem is that grainviews attach themselves to the first `.main-content` element they see, and the profile prompt was also using `.main-content`. This somewhat hacky fix works around the problem. With this change, we render the grain *behind* the profile prompt, where it is not visible. I've verified that even when grains are very tall (e.g. a MediaWiki with lots of content on the front page) they don't peek out from behind the profile prompt. (I don't fully understand why not, but hey, it works!)

### DIFF
--- a/shell/client/_account.scss
+++ b/shell/client/_account.scss
@@ -592,7 +592,8 @@ body>.main-content>.account {
   }
 }
 
-#first-sign-in.main-content {
+.first-sign-in-main-content {
+  z-index: 1;
   h1 {
     text-align: center;
   }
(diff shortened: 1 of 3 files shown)


zarvox:
:+1: for not having more than one `.main-content` running around.

PR #763, opened 2015-08-28
Dismiss menu on mobile after clicking grain action buttons

jparyani:


kentonv:
Underscore prefixes are supposed to indicate private fields. You want to use the reset() method here.

jparyani:
Fixed

PR #779, opened 2015-08-28
Fix @-sign in 'handle' field of account-settings page.

dwrensha:
In Firefox, this currently looks like:
broken-at

After this patch, it looks like:
unbroken-at

I've also verified that it looks good in Chrome and Safari.

### DIFF
--- a/shell/client/_account.scss
+++ b/shell/client/_account.scss
@@ -654,21 +654,20 @@ form.account-profile-editor {
         border: 1px solid #c00;
       }
 
-      &.handle {
-        label::after {
-          content: "@";
-          float: right;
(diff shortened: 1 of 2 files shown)


dwrensha:
Garply, retest this please

kentonv:
Oh fine. :P

dwrensha:
I'm guessing that you wanted to do this entirely in CSS?  For what it's worth, I still don't understand how the `margin-right: -20px;` thing works. If the margin is negative enough it wraps up to the previous line?

kentonv:
Without the negative right margin, the @ would have appeared to the left of the input box.

PR #780, opened 2015-08-28
Delete some obsolete code and clean up some comments.

dwrensha:
The grainSize subscription is now handled by GrainView.


dwrensha:
Garply, retest this please

PR #781, opened 2015-08-28
Fix tests by using a more precise CSS selector.

dwrensha:
It seems that somehow the .app-action in the popular-container was interfering with these tests.


dwrensha:
I think the second commit here, https://github.com/dwrensha/sandstorm/commit/f455e45c33a9fd254d5a701652001d178f4a63c9, might not be necessary, and that the test failure on the first commit was due to something transient. However, in any case, the second commit doesn't hurt anything.

Issue #776, opened 2015-08-28
New UI: Stripe in Edge Browser


ocdtrekkie:
Might be out of your control:

![stripeinedge](https://cloud.githubusercontent.com/assets/4399499/9547174/f9c97b66-4d5d-11e5-981e-16eaf67c2f7d.png)

kentonv:
It seems to do that in IE11 as well. Clicking "remember me" causes it to suddenly fix itself (even if you then un-check it).

I suspect this has something to do with our hack that keeps Stripe's javascript out of our origin. When the invisible iframe is clicked, it opens the stripe window and resizes itself at the same time -- maybe stripe records the small frame size too soon and messes up its own style in an attempt to fit? Maybe we can solve this by giving the iframe's inner document an explicitly large size (even when the viewport is small)?

jparyani:
Fixed by #783

jparyani:
Thanks for the report @ocdtrekkie!

kentonv:
testrock updated

kentonv:
oops, didn't mean to hit re-open

PR #783, opened 2015-08-28
Fix stripe iframe display on Internet Explorer

jparyani:
Also change to always displaying the iframe over the clickable elements,
even on desktop


jparyani:
Fixes #776

PR #785, opened 2015-08-28
Get setPath() to work once again.

dwrensha:
Also, openSession was referencing some non-existent fields like `self._query`.

### DIFF
--- a/shell/client/grainview.js
+++ b/shell/client/grainview.js
@@ -18,6 +18,7 @@ GrainView = function GrainView(grainId, path, query, hash, token, parentElement)
 
   this._grainId = grainId;
   this._originalPath = path;
+  this._path = this._originalPath;
   this._originalQuery = query;
   this._originalHash = hash;
   this._token = token;
@@ -208,6 +209,14 @@ GrainView.prototype.setTitle = function (newTitle) {
   this._dep.changed();
 }
 
+GrainView.prototype.setPath = function (newPath) {
+  this._path = newPath;
+  if (this.isActive()) {
+    window.history.replaceState({}, "", this.route());
+  }
+  this._dep.changed();
+}
+
 GrainView.prototype.depend = function () {
   this._dep.depend();
 }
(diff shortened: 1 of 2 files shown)


dwrensha:
Garply, retest this please

dwrensha:
To save you a click-through to Jenkins, this is the error from the first build attempt:

```
[FAILED ] compile: libseccomp/arch.c
/ekam-provider/canonical/libseccomp/arch.c:43:10: fatal error: 'arch-s390.h' file not found
#include "arch-s390.h"
         ^
1 error generated.
[FAILED ] compile: libseccomp/arch-syscall-dump.c
/ekam-provider/canonical/libseccomp/arch-syscall-dump.c:43:10: fatal error: 'arch-s390.h' file not found
#include "arch-s390.h"
         ^
1 error generated.
```

shell/client/grainview.js
```diff
@@ -401,9 +410,9 @@ GrainView.prototype.sessionStatus = function () {
 GrainView.prototype.route = function () {
   this._dep.depend();
   if (this._token) {
-    return "/shared/" + this._token;
+    return "/shared/" + this._token + this._path;
```
zarvox: We should probably also include the query string and fragment in both of these too?

dwrensha: Hm. We could include the original query and the original fragment. But what if someone calls setPath() with a new query and a new fragment? I don't think we've nailed down exactly the semantics we want there. Should we ignore any query and fragment that comes in on setPath()?

dwrensha: It seems clear that we do want `setPath()` to be able to change the query and fragment. I guess the question is: should they be passed as separate arguments, or should they be part of the single `path` argument? I think we want separate arguments, but in any case it's a new feature and I think it should be a separate pull request.

dwrensha:
Garply, retest this please

shell/client/grainview.js
kentonv: I think we should probably keep it as just a string that the app provides. Attempting to parse the query could break some apps that do stupid things like use the same query parameter name multiple times, and I don't think it gains us anything.

We my want to consider what happens if the app sends "../foo" as the path... but not now.

dwrensha: Oops, I just realized that as this is now, the query and fragment never get through at all, even on the first run.

PR #793, opened 2015-08-29
Include the query and fragment directly in the path.

dwrensha:
This forwards the query and fragment to the grain iframe.


dwrensha:
I've verified that this works with Lychee, which uses URL's like `http://sandstorm-utopic.dwrensha.ws/shared/GtAS_DKLR0pmhm2PN0M75t-8RvR25jU0XzNNofJkJft/#s/14408847447061`

PR #795, opened 2015-08-29
Add caret to navbar-toggling button when logged in

zarvox:
In the longer future, this could perhaps be the small arrow below the logo.  My
attempts to produce something that looked good with the arrow below the logo
were unsuccessful - the space is cluttered, and the arrow isn't recognizable as a
"toggle something" dropdown.


kentonv:
I was able to make this work per Nena's design in 4e5d868, so let's go with that.

PR #807, opened 2015-08-31
Replace remaining `legacy-bridge` instances with `sandstorm-http-bridge`

zarvox:


paulproteus:
Thank you _so_ much for this!

PR #810, opened 2015-08-31
Add support for webdav to the proxy/legacy bridge

mnutt:
As discussed in https://groups.google.com/forum/#!topic/sandstorm-dev/TcnV64xbzng.

This PR allows for a desktop/mobile client to communicate with a sandstorm grain through its API using WebDAV. It does the following:

* Adding the methods PROPFIND, PROPPATCH, MKCOL, COPY, MOVE, LOCK, UNLOCK, ACL, REPORT, and HEAD 
* Adding webdav-specific response headers DAV and ETag
* Creating a whitelist of allowed request headers
* Adding response codes 204, 206, 207, and 304. The extra noContent codes don't really fit the existing shouldResetForm heuristics so I ended up having to switch them over to just specifying the successCode in the schema. 
* Making OPTIONS requests hit the grain, as the "DAV" header can potentially be part of the response and there is no way for the proxy to determine the correct value otherwise. 
* Adding a whitelist of legacy clients that require basic auth to use web keys, and adding mirall to the list


garply:

jparyani:
Garply, ok to test

shell/server/proxy.js
```diff
@@ -1604,7 +1598,7 @@ Proxy.prototype.handleRequest = function (request, data, response, retryCount) {
     var path = request.url.slice(1);  // remove leading '/'
     var session = self.getSession(request);
 
-    if (request.method === "GET") {
+    if (request.method === "GET" || request.method === "HEAD") {
```
paulproteus: Silly (?) question: if `request.method == HEAD`, then shouldn't we do a `session.head()` instead?

shell/server/proxy.js
```diff
@@ -1620,8 +1614,96 @@ Proxy.prototype.handleRequest = function (request, data, response, retryCount) {
       }, context);
     } else if (request.method === "DELETE") {
       return session.delete(path, context);
+    } else if (request.method === "PROPFIND") {
+      return session.propfind(path, {
+        mimeType: request.headers["content-type"] || "application/octet-stream",
```
paulproteus: I'm curious why it's important to set a default Content-Type, and wondering if we can lose this logic.

shell/server/proxy.js
```diff
@@ -1620,8 +1614,96 @@ Proxy.prototype.handleRequest = function (request, data, response, retryCount) {
...
+        content: data,
+        encoding: request.headers["content-encoding"]
+      }, context);
+    } else if (request.method === "PROPPATCH") {
+      return session.proppatch(path, {
+        mimeType: request.headers["content-type"] || "application/octet-stream",
```
paulproteus: If we're going to do this repeatedly, it'd be nice to extract `mimeType` into a local variable. Not quite sure why it's needed, though; would love to hear more about that.

mnutt:
Thanks, these are good feedback. I'm going to take a look tonight at how I can clean this up.

shell/server/proxy.js
mnutt: Good point. My recollection of the webdav spec is that the client SHOULD pass a content-type header for all requests with bodies (most likely application/xml) but I don't think we need to default to application/octet-stream. We should probably let the grain handle unknown content-types instead. I've extracted all of it into its own method so it's a little easier to read.

shell/server/proxy.js
mnutt: The HTTP spec says that HEAD is supposed to be exactly the same as a GET, but the body shouldn't be sent back. I thought I'd save having to add HEAD to web-session, but if it's clearer to do so then that'd be fine as well.

zarvox: The distinction is valuable when sending the body is an expensive operation, like when a HEAD would retrieve the size of a large file, but the GET would fetch the whole 4GB contents.  It'd be nice to not make the grain attempt to stream all that back to the proxy for no particular reason. :)

mnutt: That makes sense; I'll add `session.head()`.

kentonv:
Hey, just wanted to let you know, this is still on my list, but I'm still trying to dig out from the mountain of email that built up during the launch cycle... so, soon

shell/server/proxy.js
```diff
@@ -36,6 +36,18 @@ SANDSTORM_ALTHOME = Meteor.settings && Meteor.settings.home;
...
 SANDSTORM_VARDIR = (SANDSTORM_ALTHOME || "") + "/var/sandstorm";
 
+// User-agent strings that should be allowed to use http basic authentication.
+// These are regex matches, so ensure they are escaped properly with double
+// backslashes.
+BASIC_AUTH_USER_AGENTS = [
+  "git\\/",
+  "GitHub-Hookshot\\/",
+  "mirall\\/",
+  "Mozilla\\/5\.0 \\([^\\]]*\\) mirall\\/",
```
kentonv: I think the `\` in `5\.0` should be doubled as well, otherwise Javascript interprets "5\.0" the same as "5.0" which is a regex that will also match "500", etc.

I don't quite understand the middle part of the regex. You're matching a string of any characters except `]`? Why `]`?

shell/server/proxy.js
```diff
@@ -36,6 +36,18 @@ SANDSTORM_ALTHOME = Meteor.settings && Meteor.settings.home;
 SANDSTORM_LOGDIR = (SANDSTORM_ALTHOME || "") + "/var/log";
 SANDSTORM_VARDIR = (SANDSTORM_ALTHOME || "") + "/var/sandstorm";
 
+// User-agent strings that should be allowed to use http basic authentication.
+// These are regex matches, so ensure they are escaped properly with double
+// backslashes.
```
kentonv: Comment should explain: It's important that we do _not_ match any user agents that might render HTML and execute embedded scripts.

shell/server/proxy.js
```diff
@@ -1649,24 +1643,84 @@ Proxy.prototype.handleRequest = function (request, data, response, retryCount) {
...
+      return session.lock(path, optionsForBody(), context);
+    } else if (request.method === "UNLOCK") {
+      return session.unlock(path, optionsForBody(), context);
+    } else if (request.method === "ACL") {
+      return session.acl(path, optionsForBody(), context);
+    } else if (request.method === "REPORT") {
+      return session.report(path, optionsForBody(), context);
+    } else if (request.method === "OPTIONS") {
+      return session.options(path, context).then(function (rpcResponse) {
+
```
kentonv: You removed the comment `// Reply to CORS preflight request.`, but I think it should still be here to give context.

shell/server/proxy.js
```diff
@@ -1649,24 +1643,84 @@ Proxy.prototype.handleRequest = function (request, data, response, retryCount) {
...
-    if (request.method === "GET") {
+    function optionsForBody() {
+      return {
+        content: data,
+        encoding: request.headers["content-encoding"],
+        mimeType: request.headers["content-type"]
+      };
+    }
+
+    if (request.method === "GET" || request.method === "HEAD") {
```
kentonv: Does Node actually ever pass in "HEAD" requests? I was under the impression that it would generate a "GET" request and then chop off the body itself.

src/sandstorm/web-session.capnp
```diff
@@ -111,6 +147,48 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
     encoding @2 :Text;  # Content-Encoding header (optional).
   }
 
+  struct PropfindContent {
```
kentonv: Since all these `Content` structs turn out to be the same, I think we should probably define a common `Content` struct instead. For now, use it just for the new methods; don't remove `PostContent` nor `PutContent`. Add a comment:

    // TODO(apibump): Remove PostContent and PutContent, replacing them with Content.

Then when we do an API revision we can fix this problem.

src/sandstorm/web-session.capnp
```diff
@@ -97,6 +107,32 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
...
+      "depth",                 # webdav PROPFIND
+      "destination",           # webdav COPY/MOVE
+      "overwrite",             # webdav COPY/MOVE
+      "oc-total-length",       # Owncloud client
+      "oc-chunk-size",         # Owncloud client
+      "x-oc-mtime",            # Owncloud client
+      "oc-fileid",             # Owncloud client
+      "oc-chunked",            # Owncloud client
+    ];
+    # Non-standard request headers which are whitelisted for backwards-compatibility
```
kentonv: This comment should go on the line after `headerWhitelist` is declared, i.e. inside the [], like Python doc comments:

    const headerWhitelist :List(Text) = [
      # Non-standard request headers which are whitelisted for backwards-compatibility
      # ...

      "if-match",              # webdav etag
      ...
    ];

src/sandstorm/web-session.capnp
```diff
@@ -97,6 +107,32 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
...
+    # Additional headers present in the request. Only whitelisted headers are
+    # permitted.
+
+    struct Header {
+      name @0 :Text;  # lower-cased name
+      value @1 :Text;
+    }
+
+    const headerWhitelist :List(Text) = [
+      "if-match",              # webdav etag
```
kentonv: `If-Match` and `If-Not-Match` are standard HTTP headers which we really need to support explicitly rather than pass through. Passing in these headers will affect non-webdav apps too, and cause them to possibly return new response codes that they didn't before, which we may not be prepared to handle. So, I don't think they belong on the whitelist, which is intended to be for non-standard things.

I think arguably we should add this to `Context`:

    union etagPrecondition {
      none @4 :Void;  # No precondition.
      exists @5 :Void;  # If-Match: *
      matchesOneOf @6 :List(Text);  # If-Match
      matchesNoneOf @7 :List(Text);  # If-None-Match
    }

src/sandstorm/web-session.capnp
kentonv: You know, maybe what we really want here is a single method called `dav()`, whose first parameter is an enum specifying the specific method. Otherwise we have a lot of repeated code which seems pretty redundant.

OPTIONS should still be handled specifically, though.

src/sandstorm/web-session.capnp
```diff
@@ -76,7 +76,17 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
...
+  propfind @7 (path :Text, content :PropfindContent, context :Context) -> Response;
+  proppatch @8 (path :Text, content :ProppatchContent, context :Context) -> Response;
+  mkcol @9 (path :Text, content :MkcolContent, context :Context) -> Response;
+  copy @10 (path :Text, context :Context) -> Response;
+  move @11 (path :Text, context :Context) -> Response;
+  lock @12 (path :Text, content :LockContent, context :Context) -> Response;
+  unlock @13 (path :Text, content :UnlockContent, context :Context) -> Response;
+  acl @14 (path :Text, content :AclContent, context :Context) -> Response;
+  report @15 (path :Text, content :ReportContent, context :Context) -> Response;
+  options @16 (path :Text, context :Context) -> Response;
```
kentonv: I'm somewhat concerned that passing through OPTIONS could have security problems independent of DAV. OPTIONS is a general HTTP method that has been overloaded for a lot of uses. As an example, we certainly wouldn't want to allow an app to respond to an OPTIONS request with `Access-Control-*`; Sandstorm has strong opinions on what those headers should contain. Of course, we don't pass through those headers, so that's not a problem. I am worried, though, that there might be bad things you can do with OPTIONS that could slip through; it's hard to say since it's so overloaded.

Can we narrow down the precise ways in which OPTIONS is used by WebDAV apps and support them more directly? It looks like it's only for the "DAV" header? Perhaps we could instead have an `isDav()` method for this?

src/sandstorm/web-session.capnp
```diff
@@ -154,12 +232,16 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
       created  @1 $httpStatus(id = 201, title = "Created");
       accepted @2 $httpStatus(id = 202, title = "Accepted");
 
+      noContent      @3 $httpStatus(id = 204, title = "No Content");
+      partialContent @4 $httpStatus(id = 206, title = "Partial Content");
+      multiStatus    @5 $httpStatus(id = 207, title = "Multi-Status");
+
+      # This seems to fit better here than in the 3xx range
+      notModified    @6 $httpStatus(id = 304, title = "Not Modified");
```
kentonv: How about we have a completely separate entry in the union -- as a sibling of `content` -- like:

    preconditionFailed @22 :Void;

We use this both for the traditional "precondition failed" and "not modified" responses -- we can determine which one was meant in the front-end by checking the request method and precondition: for the specific combination of GET + If-None-Match, we return a 304, otherwise we return a 412.

I think this make a lot of sense because `preconditionFailed` is semantically different from other error cases, and `notModified` is semantically different from success cases. HTTP's taxonomy of status codes is kind of ad hoc and we don't necessarily need to follow it here. And the only reason why HTTP uses status code 304 rather than always 412 is for backwards-compatibility with the time-based caching in HTTP/1.0.

src/sandstorm/web-session.capnp
```diff
@@ -208,6 +291,11 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
         language @3 :Text;  # Content-Language header (optional).
         mimeType @4 :Text;  # Content-Type header.
 
+        dav @18 :Text;
```
kentonv: Is the DAV header only returned for OPTIONS queries? That is what the standard implies. If so, then if we create an `isDav()` method like I suggested above rather than passing through OPTIONS, then we don't need this field here anymore.

If it is returned for other request types, then I wonder if it is really only returned in success cases and not error cases. I wonder if it would make sense to promote to the top level of `Response`.

src/sandstorm/web-session.capnp
```diff
@@ -225,9 +313,16 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
       noContent :group {
         # Return successful, but with no content (status codes 204 and 205)
 
+        statusCode @17 :SuccessCode;
```
kentonv: Sharing this enum with `content` creates ambiguity: What happens if I send `noContent` with `statusCode = ok`?

It looks to me like this isn't needed, particularly if `notModified` becomes a completely different member of the union. There are some places in your PR that suggest that "207 Multi Status" can come in both `content` and `noContent` varieties, but looking at the WebDAV standard I don't see this -- it looks to me like it should always have content, right? I think it belongs under `content`, and if the body is empty, then the body is empty. `noContent` was meant specifically for status codes where having an entity body would be illegal.

src/sandstorm/web-session.capnp
```diff
@@ -225,9 +313,16 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
...
+
         shouldResetForm @15 :Bool;
         # If this is the response to a form submission, should the form be reset to empty?
         # Distinguishes between HTTP response 204 (False) and 205 (True)
+
+        dav @20 :Text;
+        # Supported webdav modes
+
+        etag @21 :Text;
```
kentonv: Hmm, I think we may want:

    struct ETag {
      value @0 :Text;  # does not include quotes
      weak @1 :Bool;
    }

Otherwise it's weird that this field has to be formatted in a specific way (in quotes, with a possible "W/" prefix). The Cap'n Proto interface is supposed to encode all structure.

I guess for the preconditions, we'll want them to be `List(ETag)` rather than `List(Text)`.

src/sandstorm/web-session.capnp
```diff
@@ -97,6 +107,32 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
...
+    struct Header {
+      name @0 :Text;  # lower-cased name
+      value @1 :Text;
+    }
+
+    const headerWhitelist :List(Text) = [
+      "if-match",              # webdav etag
+      "if-not-match",          # webdav etag
+      "depth",                 # webdav PROPFIND
+      "destination",           # webdav COPY/MOVE
```
kentonv: Hmm, I'm concerned that this header is specified as containing an absolute URL to which the resource is to be moved. This interacts weirdly with Sandstorm hosts being generally unspecified.

I wonder if we want to explicitly encode this header in the parameters to copy() and move(), and specify that it should be a locally-absolute path but not specify a scheme nor hostname?

src/sandstorm/web-session.capnp
kentonv: Though I suppose having a single `dav()` method would be problematic for special handling of the "destination" header as mentioned above, and possibly other DAV headers which you aren't supporting yet but which we may want to support in the future... so maybe not.

src/sandstorm/web-session.capnp
dwrensha: > If-Match and If-Not-Match are standard HTTP headers

It's `If-None-Match`. https://tools.ietf.org/html/rfc7232#section-3.2

shell/server/proxy.js
zarvox: It does.

That said, we should probably be passing the head request as a head request, not as a get, because it's unnecessary work for the backend to generate and send e.g. a 2GB file if the frontend is going to discard it (and then potentially request the whole thing as a GET, now that it knows some metadata).  But that can be a separate bug.

shell/server/proxy.js
mnutt: It was a mistake (that happened to match); I fixed it.

shell/server/proxy.js
mnutt: Good point, I went ahead and moved HEAD into web-session.

src/sandstorm/web-session.capnp
mnutt: Would it be bad form for us to treat OPTIONS like a regular method (as it is now) in web-session.capnp but have the proxy only pick up the DAV header? Or is that what you're referring to?

kentonv: I think it would be bad form to include things in the interface that won't be honored if we can avoid it.

It might make sense to have an "options()" method that doesn't operate like the other methods, but instead returns a "struct Options" which initially will only have a boolean "isDav" field. That way if there are other options we want to support in the future, we can explicitly add them rather than add new methods.

kentonv:
Hey, not sure what state this is in -- it looks like some comments were addressed but some are not.

If you'd like, I could pick this up and finish it off to my liking. That might be easier than going back and forth. I could probably do it early next week.

mnutt:
Hi Kenton, I had intended to address the rest but was a bit behind with some other stuff this week and had to put it on the backburner. If you have a pretty clear idea of how you'd like to finish it, that'd be awesome. Otherwise I can get back into it sometime next week.

kentonv:
Working on this now!

kentonv:
Hey, I'm working on this in the "webdav" branch:

https://github.com/sandstorm-io/sandstorm/tree/webdav

I haven't really tested these changes since I'm not sure how to set up the test case. I guess I'll finish coding and then either ask you to test it or give me instructions for how I can test it. :)

I've also created a Wekan tracking tasks:

https://oasis.sandstorm.io/shared/5GRQllbB0DGBLFgUo8oMUf_6ecpTwbhIyQQ5gVbKM86

kentonv:
See PR #968 for the continuation of this story!

zarvox:
Closing, since #968 is merged :)

PR #817, opened 2015-08-31
Change download backup button to a spinner while waiting on server

### LINKED ISSUE #802: Downloading a really large backup gives no feedback while building the zip
The user would probably click the button multiple times, which would be sad for the server.

Also the really large zip may not download successfully since the download is not resumable.

jparyani:
Fixes #802

### DIFF
--- a/shell/shared/grain.js
+++ b/shell/shared/grain.js
@@ -375,11 +375,23 @@ if (Meteor.isClient) {
     },
   });
 
+  Template.grainBackupButton.onCreated(function () {
+    this.isLoading = new ReactiveVar(false);
+  });
+
+  Template.grainBackupButton.helpers({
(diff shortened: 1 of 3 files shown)


shell/shared/grain.js
```diff
@@ -375,11 +375,23 @@ if (Meteor.isClient) {
...
+  Template.grainBackupButton.helpers({
+    isLoading: function () {
+      return Template.instance().isLoading.get();
+    }
+  });
+
   Template.grainBackupButton.events({
-    "click button": function (event) {
+    "click button": function (event, template) {
+      template.isLoading.set(true);
```
zarvox: Perhaps you also want to prevent subsequent clicks from triggering further `backupGrain` calls here too?

PR #818, opened 2015-08-31
Approximately s/package/pkg/ in *.js

zarvox:
"package" is an ES6 keyword, so let's avoid using locals named precisely that.

### DIFF
--- a/shell/server/installer.js
+++ b/shell/server/installer.js
@@ -36,7 +36,7 @@ var installers = {};
 Meteor.methods({
   deleteUnusedPackages: function (appId) {
     check(appId, String);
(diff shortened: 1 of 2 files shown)


kentonv:
It appears `package` is reserved in strict mode only (and as yet is not actually used for any particular purpose). It has actually been so reserved since strict mode was introduced in ES5. It would be cool to start using strict mode, but that's a lot of work.

I'm OK with proactively renaming but it looks like this PR now has merge conflicts so will need to be redone. I'm also OK with just leaving it as-is until we actually decide to use strict mode.

zarvox:
Sounds good.  I'll close this for now, and make a new PR when the time comes.

PR #819, opened 2015-08-31
Grainlist: also allow filtering by app title

zarvox:
Some users were concerned that they could no longer filter their grains by app.
While the search and filter logic might not be optimal when filtering with
multiple search keys, this restores the majority of that functionality.


kentonv:
As implemented it looks like this only matches if either all my search terms match the app title or all my search terms match the grain title. I think it makes sense to support combining the app title with some grain title terms, e.g. "etherpad oasis" should find all my etherpad docs that have "Oasis" in the title.

zarvox:
Good call.  Adjusted.

PR #823, opened 2015-08-31
Add an ESC handler for topbar popups

jparyani:


kentonv:
Hmm, seems like it would be cleaner to register this as a permanent handler globally, rather than dynamically register it when a popup is present. Calling topbar.closePopup() should have no effect if there's no popup...

jparyani:
Done.

PR #825, opened 2015-09-01
Add a missing `.defaultText`.

dwrensha:


dwrensha:
This is broken. Do not merge yet.

dwrensha:
The problem is that `action` can be one of two things:
  1. an entry in the `UserActions` actions table, in which case `action.nounPhrase` is a string.
  2. a JSON-encoded Manifest.Action, in which case 'action.nounPhrase` is a `Util.LocalizedText`.

zarvox:
Hmmm, the entry in the `UserActions` table should probably also be a `Util.LocalizedText`.  Are we dereferencing it too soon when we insert it into the UserActions table?

kentonv:
Yes, probably we shouldn't be dereferencing this in the first place. Not sure whether it would be better to do a migration to fix this or simply check `typeof` and handle both cases.

kentonv:
Closing so that I don't accidentally merge. Please re-open when ready, or file an issue to track the bug.

PR #827, opened 2015-09-01
Vagrantfile: detect CPUs/RAM on Windows

zarvox:
This change makes the Vagrantfile detect the host's CPU count and RAM size
using powershell, which is available by default on Win7 and newer.  If
powershell is not available, we'll fall back to the old 1-CPU, 512MB-RAM
behavior.


zarvox:
cc @paulproteus

paulproteus:
If you've tested this on Windows, +1, let me know, and I'm happy to merge it as-is.

If you want me to test it, let me know and I can merge it.

Beyond that, looks totally reasonable!

PR #834, opened 2015-09-03
Construct proxies lazily on request

kentonv:
This change refactors proxy.js such that `Proxy` objects are constructed lazily on first request rather than proactively.

This will make it easier to run multiple front-ends, since it's not necessary to specifically hit the one that has the Proxy object you're looking for. It would generally be nice if all requests for the same session arrive at the same front-end instance (to avoid creating multiple WebSessions for the same grain host), but this is not strictly necessary.

@dwrensha, please review. Note that I have attempted to organize my commits into bite-sized chunks. :)

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -599,6 +599,16 @@ var getProxyForHostId = function (hostId) {
           return undefined;
         }
 
+        var apiToken;
+        if (session.hashedToken) {
+          apiToken = ApiTokens.findOne({_id: session.hashedToken});
+          // We don't have to fully validate the API token here because if it changed the session
+          // would have been deleted.
+          if (!apiToken) {
+            throw new Meteor.Error(410, "ApiToken has been deleted");
+          }
+        }
+
         var grain = Grains.findOne(session.grainId);
         if (!grain) {
           // Grain was deleted, I guess.
@@ -615,6 +625,7 @@ var getProxyForHostId = function (hostId) {
 
         var proxy = new Proxy(grain._id, grain.userId, session._id, hostId,
                               user && user._id === grain._id, user, null, false);
+        if (apiToken) proxy.apiToken = apiToken;
 
         proxiesByHostId[hostId] = proxy;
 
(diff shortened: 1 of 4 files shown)


kentonv:
Garply, retest this please.

shell/server/proxy.js
```diff
@@ -578,29 +581,69 @@ function gcSessions() {
...
+        if (!grain) {
+          // Grain was deleted, I guess.
+          throw new Meteor.Error(410, "Resource has been deleted");
+        }
+
+        // Note that we don't need to call mayOpenGrain() because the existence of a session
+        // implies this check was already performed.
+        //
+        // TODO(perf): The Proxy is going to recompute the user's permissions. Maybe those should
+        //   be cached in the Session?
```
dwrensha: `Proxy.prototype._callNewSession()` does cache a `viewInfo` and `permissions` field.

shell/server/proxy.js
```diff
@@ -304,16 +313,10 @@ function openSessionInternal(grainId, user, title, apiToken) {
 
   var isOwner = grainInfo.owner === userId;
 
-  var sessionId = Random.id();
-  var proxy = new Proxy(grainId, grainInfo.owner, sessionId, null, isOwner, user, null, false);
-  proxy.apiToken = apiToken;
-  proxies[sessionId] = proxy;
```
dwrensha: Looks like you are no longer setting `proxies[sessionId]` anywhere. Either you should eradicate that map completely (it's currently used by `getGrainSize()`) or you should update it when you update `proxiesByHostId`.

dwrensha:
OK, looks good to me aside from the concerns I raise in my previous comments.

shell/server/proxy.js
kentonv: cool, removing comment

shell/server/proxy.js
kentonv: Hah, indeed, grain sizes aren't working. Doh.

shell/server/proxy.js
dwrensha: To be clear: these values are already written to `Sessions`, but we currently do *not* take advantage of that fact to avoid repeated permissions computations.

PR #835, opened 2015-09-03
Make periodic server-side actions replica-safe.

kentonv:
Most uses of Meteor.setInterval were registering "cleanup" events, e.g. removing expired tokens. For a multi-replica system, we probably want these events to fire more often, because more data is being generated. So we stagger them according to replica number.

A couple uses of setInterval were non-cleanup, e.g. stat collection and daily mail limit enforcement. These need to be limited to running only on the first replica.

This change also fixes the mail rate limiter code to store limits in Mongo rather than in-memory, since the latter won't work when there are multiple replicas.

@jparyani, please review.

### DIFF
--- a/shell/server/drivers/mail.js
+++ b/shell/server/drivers/mail.js
@@ -272,7 +272,7 @@ hackSendEmail = function (session, email) {
       update: {$inc: {dailySentMailCount: 1}},
       fields: {dailySentMailCount: 1}
     });
-    if (user.dailySentMailCount > DAILY_LIMIT) {
+    if (user.dailySentMailCount >= DAILY_LIMIT) {
       throw new Error(
           "Sorry, you've reached your e-mail sending limit for today. Currently, Sandstorm " +
           "limits each user to " + DAILY_LIMIT + " e-mails per day for spam control reasons. " +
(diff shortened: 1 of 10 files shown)


shell/server/drivers/mail.js
```diff
@@ -260,11 +267,12 @@ hackSendEmail = function (session, email) {
       });
     }
 
-    if (!(this.userId in dailySendCounts)) {
-      dailySendCounts[this.userId] = 0;
-    }
-    var sentToday = ++dailySendCounts[this.userId];
-    if (sentToday > DAILY_LIMIT) {
+    var user = Meteor.users.findAndModify({
```
jparyani: By default, findAndModify returns the old document (see http://docs.mongodb.org/manual/reference/command/findAndModify/). It's not a huge deal, but technically people can send one email over their DAILY_LIMIT. You can add `new: true` to the `findAndModify` or change the `if` statement, if you care.

jparyani:
Otherwise, LGTM. Merge when ready.

shell/server/drivers/mail.js
kentonv: Indeed, the condition should be >=. Thanks!

jparyani:
Garply, retest this please

PR #842, opened 2015-09-03
Perform all installer tasks reactively on the main replica.

kentonv:
@dwrensha please review.

### DIFF
--- a/shell/server/installer.js
+++ b/shell/server/installer.js
@@ -48,7 +48,7 @@ Meteor.methods({
 
 deletePackage = function (packageId) {
   // Mark package for possible deletion;
-  Packages.update({_id: packageId}, {$set: {shouldCleanup: true}});
+  Packages.update({_id: packageId, status: "ready"}, {$set: {shouldCleanup: true}});
 }
 
 var deletePackageInternal = function (package) {
@@ -92,8 +92,12 @@ startInstall = function (packageId, url, retryFailed) {
   if (retryFailed) {
     Packages.update({_id: packageId, status: "failed"}, {$set: fields});
   } else {
-    fields._id = packageId;
-    Packages.insert(fields);
+    try {
+      fields._id = packageId;
+      Packages.insert(fields);
+    } catch (err) {
+      console.error("Simultaneous startInstall()s?", err.stack);
+    }
   }
 }
 
@@ -133,9 +137,7 @@ if (!Meteor.settings.replicaNumber) {
 
   Meteor.startup(function () {
     // Restart any deletions that were killed while in-progress.
-    Packages.find({status: "delete"}).forEach(function (pkg) {
-      deletePackageInternal(pkg._id);
-    });
+    Packages.find({status: "delete"}).forEach(deletePackageInternal);
 
     // Watch for new installation requests and fulfill them.
     Packages.find({status: {$in: ["download", "verify", "unpack", "analyze"]}}).observe({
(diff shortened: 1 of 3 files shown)


shell/server/installer.js
```diff
@@ -102,6 +128,28 @@ cancelDownload = function (packageId) {
   }
 }
 
+if (!Meteor.settings.replicaNumber) {
+  installers = {};
+
+  Meteor.startup(function () {
+    // Restart any deletions that were killed while in-progress.
+    Packages.find({status: "delete"}).forEach(function (pkg) {
+      deletePackageInternal(pkg._id);
```
dwrensha: `deletePackageInternal()` takes a package, not an ID.

shell/server/installer.js
```diff
@@ -102,6 +128,28 @@ cancelDownload = function (packageId) {
...
+
+    // Watch for new installation requests and fulfill them.
+    Packages.find({status: {$in: ["download", "verify", "unpack", "analyze"]}}).observe({
+      added: startInstallInternal,
+      removed: cancelDownloadInternal
+    });
+
+    // Watch for new cleanup requests and fulfill them.
+    Packages.find({status: "ready", shouldCleanup: true}).observe({
+      added: deletePackageInternal
```
dwrensha: I think this needs to be `changed`, not `added`, to work as you intend. You want this to be called immediately after any replica calls `deletePackage()`, right?

dwrensha: Oh, wait. Nevermind. I think this is okay. The package gets "added to the result set" , even though it is updated (not inserted) in the collection.

dwrensha: Here's a case I'm worried about: What happens if `deleteUnusedPackages(appId)` gets called on an app that has a package-install in progress? It looks like `shouldCleanup` will get set to `true` and then once the package is "ready" then `deletePackageInternal()` will be called, perhaps before `addUserActions()` has had a chance to run!

Possible alternative strategy: don't include `status` in the query here, and then in `AppInstaller.prototype.done()` clear then value of `shouldCleanup`. Then if a delete happens during an install, the delete gets ignored, but can be retried later.

shell/server/installer.js
```diff
@@ -41,6 +47,15 @@ Meteor.methods({
...
+  Packages.update({_id: packageId}, {$set: {shouldCleanup: true}});
+}
+
+var deletePackageInternal = function (package) {
+  verifyIsMainReplica();
+
+  var packageId = package._id;
+
   if (packageId in installers) {
     return;
```
dwrensha: I suppose it might make sense to unset `shouldCleanup` here, rather than in `AppInstaller.done()` per my previous suggestion.

shell/server/installer.js
```diff
@@ -70,28 +84,40 @@ deletePackage = function (packageId) {
...
+startInstall = function (packageId, url, retryFailed) {
+  // Mark package for possible installation.
+
+  var fields = {status: "download", progress: 0, url: url};
+
+  if (retryFailed) {
+    Packages.update({_id: packageId, status: "failed"}, {$set: fields});
+  } else {
+    fields._id = packageId;
+    Packages.insert(fields);
```
dwrensha: It looks like this could throw an exception if two people try to install the same package at approximately the same time. It's an unlikely case, and this is not a catastrophic way to handle it, but perhaps there is a better way?

dwrensha:
OK, looks good to me aside from the concerns I raise in my previous comments.

shell/server/installer.js
kentonv: I've changed deletePackage to only match status: "ready", so it won't set shouldCleanup if currently installing somewhere.

shell/server/installer.js
kentonv: wrapped in a try/catch that logs server-side. I think on the client side the user will just see the install appear to start normally, so this should work...

PR #847, opened 2015-09-04
migrate UserActions to use LocalizedText

dwrensha:


shell/packages/sandstorm-db/migrations.js
```diff
@@ -154,6 +154,22 @@ function removeKeyrings() {
...
+function useLocalizedTextInUserActions() {
+  function toLocalizedText(t) {
+    if (typeof t === "string") {
+      return { defaultText: t };
+    } else {
+      return t;
+    }
+  }
+  UserActions.find({}).forEach(function (userAction) {
+    var fields = { appTitle: toLocalizedText(userAction.appTitle),
```
kentonv: Hmm, I feel like we should only overwrite fields that exist, whereas with this code if the fields don't exist already they'll be set to `undefined`, which is almost but not quite the same thing. Thoughts?

dwrensha: OK, I've fixed this.

PR #848, opened 2015-09-04
Uninstall apps from app list.

zarvox:
Also switch the search icons to using a <span> with a background-image, since
SVG s don't work well in Edge.

Also make the search icon a label for the search input.


shell/packages/sandstorm-ui-applist/applist-client.js
```diff
@@ -182,6 +185,13 @@ Template.sandstormAppList.events({
       launchAndEnterGrainByActionId(actionId);
     });
   },
+  "click .uninstall-action": function(event) {
+    var actionId = this._id;
+    var appId = UserActions.findOne(actionId).appId;
+    console.log("Uninstalling action " + actionId);
```
kentonv: Did you mean to leave in this log?

shell/packages/sandstorm-ui-grainlist/grainlist.html
```diff
@@ -3,7 +3,10 @@
     <div class="non-usage-info">
     <h1>Grains</h1>
     <div class="search-row">
-      <img title="Search" class="search-icon" src="/search.svg"><input class="search-bar" type="text" placeholder="search" value="{{ searchText }}">
+      <label>
+        <span title="Search" class="search-icon"></span>
```
kentonv: Hmm, is this a legal way to label inputs? I wonder if instead the `<span>` should have "Search" as its text content, but with `font-size: 0` in the css...

zarvox: Yes.  The structure matches the first example on MDN: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

If you were asking about `title`: `title` is a Global attribute, which may be specified on all HTML elements: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes

(for bonus points, it's a `<span>` rather than a `<div>` because `<div>`s aren't allowed as children of `<label>`s)

shell/packages/sandstorm-ui-applist/applist-client.js
zarvox: Didn't really care one way or the other, since it's a single line triggered by a user action.  But I'll remove it, since it probably adds little value.

kentonv:
Hmm, looks like the uninstall button CSS doesn't clear the OS styles, so it looks pretty weird for me.

PR #857, opened 2015-09-09
When user logs out, clear open grain list and redirect to root page

zarvox:
A suggested change from @neynah per user studies. 

Now you won't see "403 Unauthorized" in grain frames, and instead you'll see
the expected login UI.


paulproteus:
+1

PR #867, opened 2015-09-10
Resumable sessions

jparyani:
This will allow users' grains to not refresh on a session being removed from the database (eg. when their laptop suspends and resumes after a Session has been removed from the DB).

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -647,11 +647,23 @@ Meteor.startup(function() {
     var users = [];
     var tokenIds = [];
     downstream.forEach(function (token) {
+      var proxy = proxiesByApiToken[token._id];
+      if (proxy) {
+        proxy.close();
+      }
       delete proxiesByApiToken[token._id];
       tokenIds.push(token._id);
       if (token.owner && token.owner.user){
         users.push(token.owner.user.userId);
       }
+    });
+    Sessions.find({grainId: token.grainId, $or: [{userId: {$in: users}},
+      {hashedToken: {$in: tokenIds}}]}, {fields: {hostId: 1}}).forEach(function (session) {
+      var proxy = proxiesByHostId[session.hostId];
+      if (proxy) {
+        proxy.close();
+      }
+      delete proxiesByHostId[session.hostId];
     });
     Sessions.remove({grainId: token.grainId, $or: [{userId: {$in: users}},
                                                    {hashedToken: {$in: tokenIds}}]});
@@ -950,6 +962,7 @@ function Proxy(grainId, ownerId, sessionId, hostId, isOwner, user, userInfo, isA
   this.isOwner = isOwner;
   this.isApi = isApi;
   this.hasLoaded = false;
+  this.isClosed = false;
   if (sessionId) {
     if (!hostId) throw new Error("sessionId must come with hostId");
     if (isApi) throw new Error("API proxy shouldn't have sessionId");
@@ -1045,6 +1058,10 @@ function Proxy(grainId, ownerId, sessionId, hostId, isOwner, user, userInfo, isA
       socket.destroy();
     });
   };
+}
+
+Proxy.prototype.close = function () {
+  this.isClosed = true;
 }
 
 Proxy.prototype.getConnection = function () {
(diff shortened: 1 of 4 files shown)


shell/server/proxy.js
```diff
@@ -310,14 +319,25 @@ function openSessionInternal(grainId, user, title, apiToken) {
...
   updateLastActive(grainId, userId);
 
-  var isOwner = grainInfo.owner === userId;
+  var sessionId = generateSessionId(grainId, userId, cachedSalt);
+  var session = Sessions.findOne({_id: sessionId});
+  if (session) {
+    // TODO(someday): also do some more checks for anonymous sessions (sessions without a userId).
+    if ((session.userId && session.userId !== userId) ||
+        (session.grainId !== grainId)) {
+      cachedSalt = Random.id(22);
```
kentonv: It looks like this branch can only run if we've found a collision in SHA-256, which is a major cryptographic event. Either that, or we have a serious bug. I think you should log and throw an exception rather than try a new salt.

shell/server/proxy.js
```diff
@@ -140,7 +140,7 @@ Meteor.methods({
     return grainId;
   },
 
-  openSession: function (grainId) {
+  openSession: function (grainId, cachedSalt) {
```
kentonv: Please add a type check for the new parameter.

shell/server/proxy.js
```diff
@@ -149,10 +149,10 @@ Meteor.methods({
       throw new Meteor.Error(403, "Unauthorized", "User is not authorized to open this grain.");
     }
 
-    return openSessionInternal(grainId, Meteor.user(), null);
+    return openSessionInternal(grainId, Meteor.user(), null, null, cachedSalt);
   },
 
-  openSessionFromApiToken: function(params) {
+  openSessionFromApiToken: function(params, cachedSalt) {
```
kentonv: type check here too

shell/server/proxy.js
kentonv: Also, is this the only place where a new cachedSalt is actually generated? Shouldn't a new cachedSalt always be generated if the client passed `undefined`?

shell/shared/grain.js
```diff
@@ -1007,7 +1007,7 @@ if (Meteor.isClient) {
             //   Hmm... Actually this may not be a real problem since the grain will be restarted
             //   on the next request. The only real problem is if the proxy has been removed on the
             //   server side, so perhaps check for that.
-            console.error("Session seems to have died.  Please reload to fix.");
+            console.error("Session seems to have died.  It should restart automatically");
```
kentonv: Seems like the comment before this line is obsolete. Maybe the whole callback can be deleted?

jparyani:
Done

shell/server/proxy.js
```diff
@@ -310,14 +321,27 @@ function openSessionInternal(grainId, user, title, apiToken) {
 
   updateLastActive(grainId, userId);
 
-  var isOwner = grainInfo.owner === userId;
+  cachedSalt = cachedSalt || Random.id(22);
+  var sessionId = generateSessionId(grainId, userId, cachedSalt);
+  var session = Sessions.findOne({_id: sessionId});
```
dwrensha: It looks like all anonymous sessions to a single grain are sharing the same entry in the `Sessions` collection. What if there are many tokens, each with a different set of permissions? And happens when a token gets revoked?

jparyani: The `cachedSalt` is randomly generated if not supplied. Check out `generateSessionId ` to see how it is used as part of the hash to create sessionId. There is a slight concern that if someone gained your salt, they could (maybe?) impersonate your session, but that would require a DB leak or a compromise of the client's browser.

Token revocation is the same as before. `clearSessionsAndProxies` will delete the Session from the DB, and a new Session will be established by the client which will go through the normal permissions flow.

dwrensha: D'oh. I was somehow forgetting the fact that the salt is randomly generated by the client.

kentonv:
Merge conflicts.

jparyani:
Done

PR #868, opened 2015-09-10
Add stats

jparyani:


jparyani:
Done. Also added stats for packages.

jparyani:
Done (went with the 2nd method of storing appIds on appDemo account creation).

kentonv:
FYI: I'm changing the computeTime to be at the recordStats level rather than the computeStats level.

kentonv:
Also renaming `packages` to `apps` since it's a table by app ID, not package ID. :)

kentonv:
See a0223ef -- let me know if there's anything wrong.

PR #871, opened 2015-09-10
Make a more ARIA-compliant tab widget for the Share popup

zarvox:


kentonv:
@dwrensha please review.

shell/shared/grain.js
```diff
@@ -540,12 +546,59 @@ if (Meteor.isClient) {
     },
   });
 
+  Template.shareWithOthers.onRendered(function () {
+    console.log("rendered shareWithOthers template, should focus widget");
```
zarvox: this log line should go

zarvox:
Garply, retest this please

zarvox:
Jenkins seems to be failing in the email-sending part of the testsuite, which 1) is unrelated to this change, and 2) has previously been reported to be flaky. :/

zarvox:
Garply, retest this please

dwrensha:
@zarvox: it looks to me like the email error is secondary and the main error is:

```
- Test grain anonymous user
     Timed out while waiting for element <section.shareable-link>h5> to be present for 5000 milliseconds. - Expected "visible" but got: "not found"
```

zarvox:
Oh, I guess I should adjust that test then <_<

zarvox:
There we go.

dwrensha:
I find it distracting that the tabs have a persistent thick blue border after I click on them. It only goes away once I've clicked on something else.

bluetab

dwrensha:
Huh, interesting. My previous screenshot was from Chrome. This one is from Firefox, where the focus box is thin, dotted, and black:
dottedtab

I find this version distracting too.

dwrensha:
I'm seeing another such box in the account drop down menu:

account-settings

As far as I can tell, "account settings" is always highlighted.

zarvox:
The blue glow in Chrome/dotted black line in Firefox is the browser default styling for an element with :focus.  For what it's worth, making focus visible is also important for accessibility - if I'm keyboard-restricted, I need to know what element I'm about to trigger if I'm about to press "Enter".

Furthermore, managing focus is one of the better ways we can indicate that something on the page has changed to unsighted users - otherwise, they have to build a mental model of the page, walk through it again looking for changes, and so on.  It's so much more useable to e.g. focus menu items when a menu is shown and is likely the next thing you'll want to interact with on the page.

If you have ideas for what visible indication we could offer that would preserve the usability critera but be less distracting, I'm happy to implement different :hover styles for the things you pointed out!

dwrensha:
> The blue glow in Chrome/dotted black line in Firefox is the browser default styling for an element with :focus. For what it's worth, making focus visible is also important for accessibility - if I'm keyboard-restricted, I need to know what element I'm about to trigger if I'm about to press "Enter".

Makes sense. I'm on board with this. I suppose that part of what's disorienting to me is that in most cases (topbar buttons, navbar items, ...) the Sandstorm shell does *not* display a box around the element that's in focus. The inconsistency is somewhat jarring. Do you think it would make sense to always show the focus?

shell/shared/grain.js
```diff
@@ -540,12 +546,58 @@ if (Meteor.isClient) {
...
   Template.shareWithOthers.events({
-    "click .shareable-link": function (event, instance) {
-      instance.find(".share-tabs").setAttribute("data-which-tab", "shareable-link");
-    },
-    "click .send-invite": function (event, instance) {
-      instance.find(".share-tabs").setAttribute("data-which-tab", "send-invite");
+    "click #send-invite-tab-header": activateTargetTab,
+    "click #shareable-link-tab-header": activateTargetTab,
+    "keydown [role=tab]": function(event, template) {
+      if (event.keyCode == 38 || event.keyCode == 40) { // up and down arrows
```
dwrensha: It seems sad that we have to write all of this special keyboard handling logic. I suppose there's no betters way to make it easy to focus on either tab?

zarvox:
I think the ideal thing to do would be to always show some sort of focus indicator that fits well with the application style.  This can be accomplished by writing `:hover` CSS rules (and indeed, having `:hover { outline: none; }` is why we don't see focus elsewhere today - I assume someone didn't like the look of focus applied to buttons, removed it, and didn't know about the accessibility implications).

I guess I can style focused elements with a gentler purple box-shadow, or a dotted purple outline, and then things would be both consistent and accessible?  @neynah do you have any opinion on what we should show to indicate when each of the following are focused:

- button
- tab header
- menu item

shell/shared/grain.js
zarvox: Not really.  Also, the IDs exist because the aria-labelfor and aria-labelledby attributes requires IDs.

I'd like to (at some point) make this into a package that handles things more declaratively, and has an array of tabs and tab headers, and can operate over them more generically.  But I figured we could build that when we have more than one tab widget.

zarvox:
Leaving notes to myself: in chrome, the default focus ring is `outline: -webkit-focus-ring-color auto 5px;` on :focus'd buttons.  In Firefox, there are `::-moz-focus-outer` and `::-moz-focus-inner` pseudo-elements.  Wonderfully, this appears to cause potential sizing issues for buttons in Firefox.

I guess we'll add a CSS reset for these, and then add whatever we think is a reasonable focus indicator as the default style and override where appropriate.

dwrensha:
In Chrome, pressing tab allows me to cycle through the elements of the account dropdown menu.
In Firefox, pressing tab only allows me to switch between the first and the last element in the menu.

zarvox:
```
12:27 < zarvox> dwrensha: so, is it okay to merge that branch with the accessibility changes?
...
13:18 < dwrensha> zarvox: hm. messing with the tabindex of the <a> elements in the accounts dropdown doesn't seem to help
13:18 < dwrensha> in Firefox
13:20 < dwrensha> zarvox: anyway, I have no further comments about the pull request. I think it's fine to merge.
```

So I'm merging this now.

PR #874, opened 2015-09-11
Move permissions.js into its own package and add some tests for it.

dwrensha:


shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -15,6 +15,18 @@
...
+
+SandstormPermissions = function (db) {
+  function cleanupSelfDestructing() {
+    var now = new Date();
+    db.collections.apiTokens.remove({expiresIfUnused: {$lt: now}});
+  };
+
+  // Make self-destructing tokens actually self-destruct, so they don't
+  // clutter the token list view.
+  SandstormDb.periodicCleanup(5 * 60 * 1000, cleanupSelfDestructing);
```
kentonv: Hmm, it worries me that if you constructed two SandstormPermissions objects you'd end up with multiple cleanups registered. Maybe do this is a Meteor.startup() instead?

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -357,10 +370,11 @@ downstreamTokens = function(root) {
     grainId = root.grain._id;
   }
 
-  var grain = Grains.findOne(grainId);
+  var grain = db.collections.grains.findOne(grainId);
```
kentonv: In theory we should be moving away from using db.collections and instead adding methods to db itself covering our needs, so that we can clearly see what kinds of queries we do by reviewing the SandstormDb interface. In this case a getGrain() method makes a lot of sense.

shell/packages/sandstorm-permissions/permissions-tests.js
```diff
@@ -0,0 +1,137 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+var globalDb = new SandstormDb();
+
+globalDb.collections.grains.remove({});
```
kentonv: // TODO(cleanup): Use mock DB.

(I think a mock DB should be easy to implement on top of minimongo. Basically just pass null for all of the collection names.)

shell/packages/sandstorm-permissions/permissions.js
dwrensha: But then how would it get the database object?

shell/packages/sandstorm-permissions/permissions-tests.js
dwrensha: What does a "mock DB" mean? Note that `meteor test-packages` uses a fresh isolated Mongo instance.

shell/packages/sandstorm-permissions/permissions.js
dwrensha: I could just define `cleanupSelfDestructing()` here and move the registration to db-deprecated.js, but I'm not sure that's any better.

kentonv: I guess what I don't like here is largely aesthetic, but basically I think you should only have an object if it encapsulates some meaningful state. So, yeah, I think the right model is indeed to have a function called "SandstormPermissions.registerCleanups(db)", which is called from db-deprecated.js (which perhaps we should consider renaming to "init-server.js" or something at some point, though not yet).

I guess the alternative is to say that SandstormPermissions represents a "component" of the system. It's only intended that you have one instance of it per database, and all its functionality is exposed as methods of that instance rather than "static" functions. This allows replacing it with a mock instance and could also allow adding some state someday, like perhaps a "configuration" that somehow affects the logic. I guess this is what you were going for originally, from what you said in meetings.

So I guess I'd say, if all the methods are "static", then there shouldn't be a constructor. But if the methods are methods, then it makes sense to have a constructor that registers the cleanup. (Ideally there'd also be some sort of "destructor" that unregisters cleanups... but... >_>)

shell/packages/sandstorm-permissions/permissions-tests.js
kentonv: If it's a fresh isolated mongo then why do you need to clear it? :)

I mean that each test case ought to create its own DB instance which probably should not actually be backed by mongo, but instead be a fake in-memory thing that behaves like a real db. I guess in testing terminology the right word is "fake" rather than "mock" -- a "mock" object is one where your test actually tells it exactly what sequence of method calls to expect, whereas a "fake" is a lightweight object that behaves similarly to the "real thing" but for testing purposes.

dwrensha:
Comments addressed.

dwrensha:
Rebased, to make it easier for me to build further changes on top of this while I'm waiting for it to get merged. (The further changes will come as a separate pull request.)

PR #878, opened 2015-09-14
Instructions to share /home/vagrant on Windows

neynah:
Full Credit: @paulproteus

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -88,3 +88,50 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     libvirt.memory = assign_ram_mb
   end
 end
+
+### If you're on Windows, and you want to SMB share the
+### /home/vagrant directory in the guest with your Windows
+### machines, run "vagrant ssh" then do the following:
+###
+### sudo apt-get install samba pwgen
+###
+### Then sudo nano -w /etc/samba/smb.conf and make it contain the
+### following, only remove the ### from the beginning of every line.
+###
+### [global]
+### workgroup = WORKGROUP
+### dns proxy = no
+### bind interfaces only = no
+### syslog only = yes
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -88,3 +88,50 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
+### create mask = 0750
+### directory mask = 0700
+### read only = no
+### valid users = vagrant
+### path = /home/vagrant
+###
+### Then set a password for the vagrant user on Linux and enable it for
+### samba use by running:
+###
+### pwgen -1  # generate a password
```
zarvox: `pwgen` may not be installed by default; maybe add `pwgen` to the `apt-get install` line above?

neynah:
I added pwgen, thanks.

paulproteus:
+1 @neynah can you add pwgen to the apt-get install line?​

neynah:
Did I add it correctly in 5ce7d15a159ae47e4ccf2e1a13114ff6d44be5da

paulproteus:
Yes! I'll merge now.

paulproteus:
A couple of random notes:

* The fact that you tested these instructions is very important, so you don't have to give me full credit without at _least_ taking credit yourself for the part you did do!

* To update the pull request, I notice you modified the previous commit and did a force push. One thing you could do instead (maybe you already know this) is to make one change on top, and then push the branch again, but without doing force. When I'm reviewing things, I like to see a commit on top in response to a requested change, rather than the whole commit to change. Different reviewers might have different opinions on this.

Anyway, many many thanks for verifying that these instructions work!

neynah:
Ahhh. OK, I will keep that in mind. Thank you. :)

It was a good exercise!

PR #880, opened 2015-09-14
Work around adblock rule that hides sharing-links

jparyani:


zarvox:
+1 what a great bug

paulproteus:
Totally agree with @zarvox -- great bug.

Thanks @jparyani for the crystal clear comment in the template, to avoid us causing this in the future!

PR #886, opened 2015-09-15
Never show the interstitial for /grain/ routes.

### LINKED ISSUE #809: "Reveal your identity" prompt is sometimes flashed quickly and disappears for a previously shared/revealed grain
Probably due to a subscription not being ready.

dwrensha:
Fixes #809.


zarvox:
+1 from me!

kentonv:
Does the interstitial still flash for /share routes when you are already in the owner's contacts?

dwrensha:
> Does the interstitial still flash for /share routes when you are already in the owner's contacts?

In that case, the problem only occurred after the page had redirected to a /grain/ route, so it is covered by this fix, I think. And after applying this fix, I don't observe the problem.

PR #889, opened 2015-09-16
Sort apps in the app list case-insensitively

### LINKED ISSUE #846: App sorting should be case-insensitive
Poor draw.io is last on the list.

zarvox:
Fixes #846.

I took the liberty of cleaning up the methods I'd introduced in db.js to be
less Mongo-specific and to do matching and sorting in client code, rather than
using minimongo's limited capabilities for this (especially given the
complicated backwards-compatibility logic for appTitle and friends).

### DIFF
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -681,17 +681,8 @@ if (Meteor.isServer) {
 // Below this point are newly-written or refactored functions.
 
 _.extend(SandstormDb.prototype, {
-  // TODO(cleanup): These methods shouldn't take a raw mongo queries or "aggregations" as inputs.
-  //   We want methods corresponding to each kind of query that is performed in practice. Letting
-  //   the caller pass an arbitrary query and aggregations is problematic because:
-  //   - We can't type-check them to prevent Mongo injections.
-  //   - It defeats the purpose of clearly seeing what kinds of queries we need to support, and
-  //     therefore what kind of indexes we might need.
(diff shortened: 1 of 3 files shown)


shell/packages/sandstorm-db/db.js
```diff
@@ -690,13 +690,12 @@ _.extend(SandstormDb.prototype, {
   //   - It doesn't make it any easier for us to substitute a non-Mongo database in the future.
   //   - If the caller is just going to pass in a match-all query, we may be forcing Mongo into
   //     a slower path by using $and.
-  userGrains: function userGrains (user, query, aggregations) {
-    var filteredQuery = { $and: [ { userId: user}, query ] };
-    return this.collections.grains.find(filteredQuery, aggregations);
+  userGrains: function userGrains (user) {
```
kentonv: Is the comment above this line now obsolete?

shell/packages/sandstorm-db/db.js
```diff
@@ -690,13 +690,12 @@ _.extend(SandstormDb.prototype, {
...
+    return this.collections.grains.find({ userId: user}, {});
```
kentonv: You can omit the empty second parameter (also on line 715).

shell/packages/sandstorm-db/db.js
zarvox: Indeed.  Removed.

shell/packages/sandstorm-db/db.js
zarvox: Done.

PR #893, opened 2015-09-16
Argh, apparently the transitiveShares() method was broken too.

dwrensha:
This fixes some more fallout from #874.


paulproteus:
Since part of the joy of #874 was the goal of adding tests, does it makes sense to add tests to this method as part of this "fixup the broken world" pull request?

PR #896, opened 2015-09-17
split user ids into accounts ids and identity ids

dwrensha:
This is not quite ready for merging; I'm posting it early because @kentonv wanted to take a look.

### DIFF
--- a/src/sandstorm/hack-session.capnp
+++ b/src/sandstorm/hack-session.capnp
@@ -65,45 +65,9 @@ interface HackSessionContext @0xe14c1f5321159b8f
 
   generateApiToken @3 (petname :Text, userInfo :Grain.UserInfo, expires :UInt64 = 0)
       -> (token :Text, endpointUrl :Text, tokenId :Text);
-  # Generates a new API token which can be used to access an HTTP API exported by this application.
-  # The method also returns the URL at which the API is exported.
-  #
-  # To access the API, a client may send requests to `endpointUrl` (or sub-paths thereof) and pass
-  # the following header, replacing <token> with the returned `token`:
-  #
-  #     Authorization: Bearer <token>
-  #
-  # The request will be delivered to the app like a regular web request. However, the request will
-  # contain no cookies, and any cookies in the response will be ignored. Also note that the system
-  # will arrange for `endpointUrl` to accept cross-origin request from any origin, so that
-  # third-party web sites can use XMLHttpRequest to communicate with this API.
-  #
-  # By convention, if you wish to present `endpointUrl` and `token` to the user (e.g. to copy/paste
-  # into a client app), you should do so in the format: "<endpointUrl>#<token>" -- that is,
-  # separate the two by a '#' character, as if the token is a URL "hash" or "fragment". If this
-  # combined URL is loaded directly in the browser, Sandstorm may be able to display something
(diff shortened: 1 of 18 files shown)


dwrensha:
After this change, we have account IDs, which refer to the `_id` field in the `Users` collection, and "identity IDs", which are global, stable identifiers tied to authentication methods. It's mostly true that, after the migration in this pull request, anything that still calls itself a "user ID" is now an account ID. One annoying exception is the X-Sandstorm-User-Id header that we pass to apps, which actually contains an identity ID. I think we need to do something about the potential for confusion here. One option might be to rename that header to X-Sandstorm-Identity-Id; apps could deal with it when they next update their version of sandstorm-http-bridge. Another option might be to rename the existing instances (e.g. in the Grains collection) of "userId" to "accountId"; this would make the migration somewhat less tidy, and still wouldn't deal with the fact that "user ID" sounds like it should refer to the `_id` field in `Users`.

paulproteus:
Does this mean that if I:

* Create an account with GitHub, then

* Visit an Etherpad grain, then

* Sign into the same account with email token, then

* Visit the same Etherpad grain

Etherpad will make two different colors for me? If so, that seems like
surprising behavior, fwiw.

ocdtrekkie:
@paulproteus I would assume/hope that your account will use the identity you have logged into that grain with before, or the identity that the grain was shared with you at, regardless of which login method you use to log into Sandstorm.

@dwrensha Is there someway to mark a preferred/default identity to use when creating and opening grains?

paulproteus:
David, maybe one way to answer our questions is to document some of this in
a new docs file, perhaps docs/using/multiple-identities.md and make that be
part of this pull request.

That may sound snarky but I do mean it earnestly -- that way when you
answer Jake's & my questions, you can do so in a way that's useful to other
people who are going to have similar questions possibly.

dwrensha:
To be clear, this doesn't implement any of the multiple-identity logic. It just updates our data schema to be compatible with allowing multiple identities.

@ocdtrekkie yes, there will be a notion of "default" identity.

ocdtrekkie:
@paulproteus +1 on documentation with PR. (At least, when there's UI for it.) There's a bunch of recent features that never got documented.

dwrensha:
This is ready for review.

dwrensha:
Closing for now because my further progress on top of this makes further changes to the schema (Namely: a `Users.identities` array rather than a `Users.identityIds` array) and I think it'll be cleanest to fold migration that into the migration performed here.

kentonv:
I was all ready to finally review this!

I'm going to take a look anyway; hopefully comments will be useful.

src/sandstorm/hack-session.capnp
```diff
@@ -65,45 +65,9 @@ interface HackSessionContext @0xe14c1f5321159b8f
 
   generateApiToken @3 (petname :Text, userInfo :Grain.UserInfo, expires :UInt64 = 0)
       -> (token :Text, endpointUrl :Text, tokenId :Text);
-  # Generates a new API token which can be used to access an HTTP API exported by this application.
```
kentonv: Should this deleted comment be replaced with a comment saying this method is no longer supported?

Maybe we should rename the method to obsoleteGetApiToken?

kentonv:
I skimmed and decided it all looks reasonable.

PR #897, opened 2015-09-18
Support listening on multiple ports, speaking HTTPS if requested

paulproteus:
Changes:

* Adjust run-bundle to listen on multiple ports and pass their sockets over.

* Adjust meteor-bundle-main.js to optionally rebind http.createServer() to
  return a HTTPS server.

* Adjust pre-meteor.js to handle secondary HTTP services.

* Adjust default ROOT_URL to be https:// if HTTPS is enabled.

* Adjust run-bundle & install.sh to fix permissions problems in https dir.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1524,9 +1524,15 @@ sandcats_configure_https() {
   # adjusts the hostname, the Sandstorm code can easily figure out to
   # not use these certificates without having to parse the actual
   # X.509 certificate data.
-  local HTTPS_CONFIG_DIR="var/sandcats/https/$SS_HOSTNAME"
+  local HTTPS_BASE_DIR="var/sandcats/https"
+  local HTTPS_CONFIG_DIR="$HTTPS_BASE_DIR/$SS_HOSTNAME"
   mkdir -p -m 0700 "$HTTPS_CONFIG_DIR"
   chmod 0700 "$HTTPS_CONFIG_DIR"
+
+  # Make this readable by Sandstorm.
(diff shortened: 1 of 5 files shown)


paulproteus:
Here are some things this code does _not_ do yet:

* Request renewal of certificates.

* Gracefully restart just the HTTPS service when we need to reload HTTPS to start using the renewed certificates. (Note to self: I think it should be possible to handle this with `SNICallback`.)

* Use particularly good HTTPS default ciphers.

* Ensure slowloris resistance (since we're disabling a `setTimeout` behavior of Meteor, after all).

* Support HTTPS for static publishing.

I think it's OK to merge this now, and add those in follow-up pull requests, if @kentonv is OK with it. The HTTPS behavior is only enabled if the user sets `HTTPS_PORT` in their config file, and we haven't told anyone to do that yet.

paulproteus:
Jenkins found a refactoring that I only partially completed, which I've now pushed a fix for, to this branch.

paulproteus:
Closing for now because of the issues Jenkins finds, plus I noticed in local testing this patch seems to break `run-dev.sh`.

paulproteus:
Now doesn't break `run-dev.sh`.

paulproteus:
TODO (before merge):

* Fix it so it doesn't crash if the https certs have the wrong permission

* Fix install.sh so the certs etc. are owned by the the Sandstorm user, not root

paulproteus:
Closing now because it's still broken.

PR #900, opened 2015-09-18
Support serving HTTPS if configured to do so and listening on multiple ports

paulproteus:
Changes:

* Adjust run-bundle to listen on multiple ports and pass their sockets over.

* Adjust meteor-bundle-main.js to optionally rebind http.createServer() to
  return a HTTPS server.

* Adjust pre-meteor.js to handle secondary HTTP services.

* Adjust default ROOT_URL to be https:// if HTTPS is enabled.

* Adjust run-bundle & install.sh to fix permissions problems in https dir.

* Adjust install.sh to create the HTTPS keys with permissions that are readable by Sandstorm.

* Serve static publishing on the first HTTPS port as well as the first HTTP port, even for custom user domain names.

* Adjust run-dev.sh to do something reasonable when multiple ports are defined.

### DIFF
--- a/meteor-bundle-main.js
+++ b/meteor-bundle-main.js
@@ -97,7 +97,7 @@ function monkeypatchHttpAndHttps() {
       // timeout. Since the method is not available on the nodejs
       // v0.10.x https server object, we ignore it entirely for now.
       //
-      // TODO(soon): Run slowloris against this to make sure it is
+      // TODO(security): Run slowloris against this to make sure it is
       // safe to ignore.
       //
       // Note that upon actually receiving a connection, Meteor
--- a/meteor-bundle-main.js
+++ b/meteor-bundle-main.js
@@ -12,7 +12,7 @@ var https = require('https');
 var net = require('net');
 var url = require('url');
 
-function sandstorm_main() {
+function sandstormMain() {
   monkeypatchHttpAndHttps();
 
   // Delegate to Meteor.
@@ -116,4 +116,4 @@ function monkeypatchHttpAndHttps() {
   http.createServer = fakeHttpCreateServer;
 }
 
-sandstorm_main();
+sandstormMain();
--- a/meteor-bundle-main.js
+++ b/meteor-bundle-main.js
@@ -66,8 +66,9 @@ function monkeypatchHttpAndHttps() {
         try {
           var metadata = JSON.parse(fs.readFileSync(metadataFilename));
         } catch (e) {
-          // If somehow the metadata has the wrong permissions,
-          // attempt to continue.
+          // If the key metadata has the wrong permissions due to an
+          // installer bug that went out during September, then skip
+          // those keys and attempt to continue.
           if (e.code === 'EACCES') {
             console.log("Skipping unreadable HTTPS key information file:", metadataFilename);
             continue;
--- a/meteor-bundle-main.js
+++ b/meteor-bundle-main.js
@@ -20,8 +20,11 @@ function sandstormMain() {
 }
 
 function monkeypatchHttpAndHttps() {
-  // Monkey-patch the HTTP object's listen() function so if Meteor
-  // calls it, then we listen on FD #3.
+  // Two very different monkey-patchings here.
+  //
+  // 1. Monkey-patch HTTP in the smallest way -- if someone calls
+  // listen() but doesn't provide an FD, assume they are Meteor and
+  // they want to bind to FD #3.
   var oldListen = http.Server.prototype.listen;
   http.Server.prototype.listen = function (port, host, cb) {
     // Overridable by passing e.g. {fd: 4} as port.
@@ -31,16 +34,15 @@ function monkeypatchHttpAndHttps() {
     return oldListen.call(this, {fd: 3}, cb);
   }
 
-  // When Meteor calls createServer(), if we are in HTTPS mode, give it a HTTPS server.
-  //
-  // Passing a second argument of `true` allows us to call the real
-  // createServer directly in pre-meteor.js.
-  var originalHttpCreateServer = http.createServer;
-  var fakeHttpCreateServer = function(requestListener, calledBySandstorm) {
-    if (calledBySandstorm) {
-      return originalHttpCreateServer(requestListener);
-    }
+  // 2. If we are in HTTPS mode, monkey-patch HTTP in a large way:
+  // return a HTTPS server, not a HTTP server, so that Meteor gets a
+  // HTTPS server on FD #3 without Meteor being aware of the
+  // complexity.
 
+  // Stash the original function in createServerForSandstorm(), since
+  // in pre-meteor.js we sometimes need to bind HTTP sockets.
+  http.createServerForSandstorm = http.createServer;
+  var fakeHttpCreateServer = function(requestListener) {
     function getHttpsOptions() {
       var basePath = '/var/sandcats/https/' + (
         url.parse(process.env.ROOT_URL).hostname);
@@ -109,7 +111,9 @@ function monkeypatchHttpAndHttps() {
       }
       return httpsServer;
     } else {
-      return originalHttpCreateServer(requestListener);
+      // Call http.createServerForSandstorm(), knowing that .listen()
+      // has been monkey-patched separately.
+      return http.createServerForSandstorm(requestListener);
     }
   }
 
--- a/meteor-bundle-main.js
+++ b/meteor-bundle-main.js
@@ -49,18 +49,30 @@ function monkeypatchHttpAndHttps() {
       // metadata about the key is available in e.g. 0.csr. So find
       // the most recent numbered file, then pull metadata out.
 
+      var reverseIntComparator = function (a, b) {
+        return (parseInt(a) < parseInt(b));
+      };
       var keyFilesDescending = files.filter(function (filename) {
         return filename.match(/^[0-9]*$/);
-      }).sort(function (a, b) {
-        return (parseInt(a) < parseInt(b));
-      }).reverse();
+      }).sort(reverseIntComparator);
 
       var result = {};
       var nowUnixTimestamp = new Date() / 1000;
       for (var i = 0; i < keyFilesDescending.length; i++) {
         var keyFilename = basePath + '/' + keyFilesDescending[i];
         var metadataFilename = keyFilename + '.response-json';
-        var metadata = JSON.parse(fs.readFileSync(metadataFilename));
+        try {
+          var metadata = JSON.parse(fs.readFileSync(metadataFilename));
+        } catch (e) {
+          // If somehow the metadata has the wrong permissions,
+          // attempt to continue.
+          if (e.code === 'EACCES') {
+            console.log("Skipping unreadable HTTPS key information file:", metadataFilename);
+            continue;
+          } else {
+            throw e;
+          }
+        }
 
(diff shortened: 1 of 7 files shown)


paulproteus:
Here are some things this code does _not_ do yet:

* Request renewal of certificates.

* Gracefully restart just the HTTPS service when we need to reload HTTPS to start using the renewed certificates. (Note to self: I think it should be possible to handle this with `SNICallback`.)

* Use particularly good HTTPS default ciphers.

* Ensure slowloris resistance (since we're disabling a `setTimeout` behavior of Meteor, after all).

I think it's OK to merge this now, and add those in follow-up pull requests, if @kentonv is OK with it. The HTTPS behavior is only enabled if the user sets `HTTPS_PORT` in their config file, and we haven't told anyone to do that yet.

(The above is copy-pasted from #897 which was an earlier version of this pull request, with one small change.)

meteor-bundle-main.js
```diff
@@ -1,12 +1,119 @@
...
+// so that, along with pre-meteor.js, we create the the right
+// HTTP+HTTPS services.
+
+var fs = require('fs');
 var http = require('http');
+var https = require('https');
 var net = require('net');
+var url = require('url');
+
+function sandstorm_main() {
```
kentonv: camelCase, please.

meteor-bundle-main.js
```diff
@@ -1,12 +1,119 @@
...
+    // Overridable by passing e.g. {fd: 4} as port.
+    if (typeof port == 'object') {
+      return oldListen.call(this, port, host, cb);
+    }
+    return oldListen.call(this, {fd: 3}, cb);
+  }
+
+  // When Meteor calls createServer(), if we are in HTTPS mode, give it a HTTPS server.
+  //
+  // Passing a second argument of `true` allows us to call the real
```
kentonv: What if Node adds a second argument at some point? Any value passed for it would probably appear truthy.

An alternative would be to add `http.createServerForSandstorm` which is called preferentially when available, or maybe at least use the word "sandstorm" as the second parameter so that we can detect unexpected signature changes.

meteor-bundle-main.js
```diff
@@ -1,12 +1,119 @@
...
+      var result = {};
+      var nowUnixTimestamp = new Date() / 1000;
+      for (var i = 0; i < keyFilesDescending.length; i++) {
+        var keyFilename = basePath + '/' + keyFilesDescending[i];
+        var metadataFilename = keyFilename + '.response-json';
+        try {
+          var metadata = JSON.parse(fs.readFileSync(metadataFilename));
+        } catch (e) {
+          // If somehow the metadata has the wrong permissions,
+          // attempt to continue.
```
kentonv: Is this specifically to cover for people who ran the installer script last week? If so, comment on that, otherwise it seems pretty weird to ignore EACCES errors.

meteor-bundle-main.js
```diff
@@ -1,12 +1,119 @@
...
+      }
+    };
+
+    if (process.env.HTTPS_PORT) {
+      var httpsServer = https.createServer(getHttpsOptions(), requestListener);
+      // Meteor calls httpServer.setTimeout() to set a default socket
+      // timeout. Since the method is not available on the nodejs
+      // v0.10.x https server object, we ignore it entirely for now.
+      //
+      // TODO(soon): Run slowloris against this to make sure it is
```
kentonv: TODO(security)

shell/packages/sandstorm-db/db.js
```diff
@@ -582,6 +582,12 @@ matchWildcardHost = function(host) {
 
   var prefix = wildcardHost[0];
   var suffix = wildcardHost[1];
+
+  // We ignore everything after the first : character to be agnostic
+  // as to what port a request came in on.
+  suffix = suffix.replace(/:.*/, "");
```
kentonv: How about: `.split(":")[0]`

That way I don't have to figure out whether this really needs to be `/:[^:]*$/` or whatever. :)

shell/packages/sandstorm-db/db.js
```diff
@@ -582,6 +582,12 @@ matchWildcardHost = function(host) {
 
   var prefix = wildcardHost[0];
   var suffix = wildcardHost[1];
+
+  // We ignore everything after the first : character to be agnostic
```
kentonv: Hmm. Technically, someone out there could be running sandstorm on `example.com:6080` and using it to web-publish to `example.com`, and I think this would break them?

shell/server/pre-meteor.js
```diff
@@ -375,11 +376,71 @@ Meteor.startup(function () {
...
+  // This function serves responses on Sandstorm's main HTTP/HTTPS
+  // port.
+  var serveMeteorOrStaticPublishing = function(req, res, next) {
+    return dispatchToMeteorOrStaticPublishing(req, res, next, false, true);
+  }
+
+  // Bind listeners to FD #4 and higher, if we are supposed to be
+  // listening on multiple ports.
+  function getNumberOfAlternatePorts() {
+    var numCommas = (process.env.PORT.match(/,/g) || {}).length;
```
kentonv: did you mean `|| []`, not `|| {}`?

kentonv: Also maybe it would be easier to say: `var numPorts = process.env.PORT.split(",").length`

src/sandstorm/run-bundle.c++
```diff
@@ -205,6 +205,33 @@ struct UserIds {
...
+  KJ_IF_MAYBE(portNumber, httpsPort) {
+    numHttpsPorts = 1;
+    result = kj::heapArray<uint>(numHttpsPorts + numHttpPorts);
+    result[0] = *portNumber;
+  } else {
+    numHttpsPorts = 0;
+    result = kj::heapArray<uint>(numHttpsPorts + numHttpPorts);
+  }
+
+  for (size_t i = 0; i < portsSplitOnComma.size(); i++) {
```
kentonv: This is fine, but FYI here are some fancier options:

    for (auto i: kj::indices(portsSplitOnComma)) {

or avoid an index altogether:

    auto result = kj::heapArrayBuilder<uint>(size);
    for (auto& portStr: portsSplitOnComma) {
      KJ_IF_MAYBE(portNumber, parseUInt(trim(portsSplitOnComma[i]), 10)) {
        result.add(*portNumber);
      } else {
        KJ_FAIL_REQUIRE("invalid config value PORT", portList);
      }
   }

    return result.finish();

src/sandstorm/run-bundle.c++
```diff
@@ -1213,12 +1245,14 @@ private:
         } else {
           KJ_FAIL_REQUIRE("invalid config value SERVER_USER", value);
         }
-      } else if (key == "PORT") {
-        KJ_IF_MAYBE(p, parseUInt(value, 10)) {
-          config.port = *p;
-        } else {
-          KJ_FAIL_REQUIRE("invalid config value PORT", value);
+      } else if (key == "HTTPS_PORT") {
+          KJ_IF_MAYBE(p, parseUInt(value, 10)) {
```
kentonv: bad indentation here (and following 3 lines)

src/sandstorm/run-bundle.c++
```diff
@@ -1695,11 +1738,20 @@ private:
...
       KJ_SYSCALL(listen(sockFd, 511)); // 511 is what node uses as its default backlog
 
-      if (sockFd != 3) {
+      if (sockFd != targetFdNum) {
         // dup socket to correct fd.
-        KJ_SYSCALL(dup2(sockFd, 3));
+        KJ_SYSCALL(dup2(sockFd, targetFdNum));
         KJ_SYSCALL(close(sockFd));
       }
+}
```
kentonv: indent

src/sandstorm/run-bundle.c++
```diff
@@ -1671,9 +1716,7 @@ private:
     return result;
   }
 
-  pid_t startNode(const Config& config) {
-    Subprocess process([&]() -> int {
-      // Create a listening socket for the meteor app on fd=3
+  void bindSocketToFd(const Config& config, uint port, uint targetFdNum) {
```
kentonv: The body of this method is over-indented... which admittedly helps make the diff clearer but should probably be fixed before merge.

src/sandstorm/run-bundle.c++
```diff
@@ -1730,11 +1786,21 @@ private:
         KJ_SYSCALL(setenv("MAIL_URL", config.mailUrl.cStr(), true));
       }
       if (config.rootUrl == nullptr) {
-        if (config.port == 80) {
-          KJ_SYSCALL(setenv("ROOT_URL", kj::str("http://", config.bindIp).cStr(), true));
+        kj::String scheme;
```
kentonv: You can use `kj::StringhPtr` here and remove the `kj::str()` wrappers to avoid redundant allocation and copying.

kentonv:
OK that's all my comments for now.

shell/packages/sandstorm-db/db.js
paulproteus: Wow, what a fascinating way that would be to use Sandstorm.

shell/server/pre-meteor.js
paulproteus:
```
> {}.length
undefined
> [].length
0
```

Clearly I did mean `|| {}`. Also +1 to `.split()`

shell/packages/sandstorm-db/db.js
paulproteus: I think my plan is to actively break that use-case. Here's my rationale:

* `nginx` by default (in my testing on my laptop, with nginx on `:80` and a Flask app on `:5000`) when I use `proxy_pass`, sends a `Host` header of `localhost:5000` because it's making an HTTP request to `localhost:5000`.
* You can configure a line of `    proxy_set_header Host $host;` and that sends `Host: localhost`.
* If you try really hard, it's probably possible to get nginx to send `Host localhost:80` but I guess my point is that's a pretty rare configuration and probably zero users are using that configuration.

So I think I'm going to break this, and if someone complains, then I'll figure out what to do. Maybe that means we should blog a short note about it, or something, or I should send a note to sandstorm-dev about the breaking change.

src/sandstorm/run-bundle.c++
paulproteus: I think I will go with the non-fancy one for now, but these are interesting.

src/sandstorm/run-bundle.c++
paulproteus: fixed!

src/sandstorm/run-bundle.c++
paulproteus: fixed!

src/sandstorm/run-bundle.c++
paulproteus: Fixed.

paulproteus:
These changes (I believe) address all essential feedback. Waiting for Jenkins to be happy.

It also adds commented-out documentation, which I needed in order to make sure I was making good decisions.

paulproteus:
@kentonv don't merge quite yet. Needs more manual testing.

paulproteus:
OK @kentonv I thought this was broken in my manual testing, but I was not doing `make update`. After a `make update` all is well, and I submit this for your review.

I'll be doing more testing now, but curious for your review.

PR #901, opened 2015-09-19
Verify installs using PGP.

kentonv:
This change arranges to sign all releases with a particular PGP key. The installer script, upon downloading a release bundle, will verify its signature.

I (and maybe other Sandstorm employees) will also sign statements (using our personal PGP keys) that the release-signing key is trustworthy. We will then extend the installation instructions with a "secure install" option that includes instructions for verifying our PGP keys, following the web of trust to the release-signing key, and then verifying the installer script using that key.

This change does *not* address verifying auto-udpates. That will happen separately, and will probably employ libsodium/ed25519 rather than PGP.

This change was written while not entirely sober. Needs careful review.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1038,12 +1038,12 @@ download_latest_bundle_and_extract_if_needed() {
     WORK_DIR="$(mktemp -d --tmpdir sandstorm-installer.XXXXXXXXXX)"
     local URL="https://dl.sandstorm.io/sandstorm-$BUILD.tar.xz"
     echo "Downloading: $URL"
-    curl -A "$CURL_USER_AGENT" -f "$URL" > $WORK_DIR/sandstorm-$BUILD.tar.xz
-    curl -s -A "$CURL_USER_AGENT" -f "$URL.sig" > $WORK_DIR/sandstorm-$BUILD.tar.xz.sig
+    curl -A "$CURL_USER_AGENT" -f "$URL" > "$WORK_DIR/sandstorm-$BUILD.tar.xz"
+    curl -s -A "$CURL_USER_AGENT" -f "$URL.sig" > "$WORK_DIR/sandstorm-$BUILD.tar.xz.sig"
 
     if which gpg > /dev/null; then
       # Regenerate with: gpg --armor --export 160D2D577518B58D94C9800B63F227499DA8CCBD
-      gpg --dearmor > $WORK_DIR/sandstorm-keyring.gpg << __EOF__
+      gpg --dearmor > "$WORK_DIR/sandstorm-keyring.gpg" << __EOF__
 -----BEGIN PGP PUBLIC KEY BLOCK-----
 Version: GnuPG v1
 
@@ -1089,8 +1089,8 @@ __EOF__
       echo "WARNING: gpg not installed; not verifying signatures (but it's HTTPS so you're probably fine)" >&2
     fi
 
-    tar Jxof $WORK_DIR/sandstorm-$BUILD.tar.xz
-    rm -rf $WORK_DIR
+    tar Jxof "$WORK_DIR/sandstorm-$BUILD.tar.xz"
+    rm -rf "$WORK_DIR"
 
     if [ ! -e "$BUILD_DIR" ]; then
       fail "Bad package -- did not contain $BUILD_DIR directory."
--- a/install.sh
+++ b/install.sh
@@ -1095,7 +1095,8 @@ __EOF__
       fail "Bad package -- did not contain $BUILD_DIR directory."
     fi
 
-    if [ $(stat --printf=%Y $BUILD_DIR/buildstamp) -lt $(( $(date +%s) - 30*24*60*60 )) ]; then
+    if [ ! -e "$BUILD_DIR/buildstamp" ] || \
+       [ $(stat --printf=%Y "$BUILD_DIR/buildstamp") -lt $(( $(date +%s) - 30*24*60*60 )) ]; then
       rm -rf "$BUILD_DIR"
       fail "The downloaded package seems to be more than a month old. Please verify that your" \
            "computer's clock is correct and try again. It could also be that an attacker is" \
--- a/install.sh
+++ b/install.sh
@@ -1081,9 +1081,8 @@ __EOF__
           grep -q '^\[GNUPG:\] VALIDSIG 160D2D577518B58D94C9800B63F227499DA8CCBD '; then
         echo "GPG signature is valid."
       else
-        echo "GPG signature is NOT valid! Please report to security@sandstorm.io immediately!" >&2
         rm -rf sandstorm-$BUILD
-        exit 1
+        fail "GPG signature is NOT valid! Please report to security@sandstorm.io immediately!"
       fi
     else
       echo "WARNING: gpg not installed; not verifying signatures (but it's HTTPS so you're probably fine)" >&2
--- a/install.sh
+++ b/install.sh
@@ -1095,6 +1095,14 @@ __EOF__
     if [ ! -e "$BUILD_DIR" ]; then
       fail "Bad package -- did not contain $BUILD_DIR directory."
     fi
+
+    if [ $(stat --printf=%Y $BUILD_DIR/buildstamp) -lt $(( $(date +%s) - 30*24*60*60 )) ]; then
+      rm -rf "$BUILD_DIR"
+      fail "The downloaded package seems to be more than a month old. Please verify that your" \
+           "computer's clock is correct and try again. It could also be that an attacker is" \
+           "trying to trick you into installing an old version. Please contact" \
+           "security@sandstorm.io if the problem persists."
+    fi
   }
 
   if [ -e $BUILD_DIR ]; then
--- a/install.sh
+++ b/install.sh
@@ -1037,7 +1037,26 @@ download_latest_bundle_and_extract_if_needed() {
     rm -rf $BUILD_DIR
     local URL="https://dl.sandstorm.io/sandstorm-$BUILD.tar.xz"
     echo "Downloading: $URL"
-    curl -A "$CURL_USER_AGENT" -f "$URL" | tar Jxo
+    curl -A "$CURL_USER_AGENT" -f "$URL" > /var/tmp/sandstorm-$BUILD.tar.xz
+    curl -s -A "$CURL_USER_AGENT" -f "$URL.sig" > /var/tmp/sandstorm-$BUILD.tar.xz.sig
+    curl -s -A "$CURL_USER_AGENT" -f "https://dl.sandstorm.io/keyring.gpg" > /var/tmp/sandstorm-keyring.gpg
+
+    if which gpg > /dev/null; then
+      if gpg --no-default-keyring --keyring /var/tmp/sandstorm-keyring.gpg --status-fd 1 \
+             --verify /var/tmp/sandstorm-$BUILD.tar.xz{.sig,} 2>/dev/null | \
+          grep -q '^\[GNUPG:\] VALIDSIG 160D2D577518B58D94C9800B63F227499DA8CCBD '; then
+        echo "GPG signature is valid."
+      else
+        echo "GPG signature is NOT valid! Please report to security@sandstorm.io immediately!" >&2
+        rm -rf sandstorm-$BUILD
+        exit 1
+      fi
+    else
+      echo "WARNING: gpg not installed; not verifying signatures (but it's HTTPS so you're probably fine)" >&2
+    fi
+
+    tar Jxof /var/tmp/sandstorm-$BUILD.tar.xz
+    rm -f /var/tmp/sandstorm-$BUILD.tar.xz{,.sig}
 
     if [ ! -e "$BUILD_DIR" ]; then
       fail "Bad package -- did not contain $BUILD_DIR directory."
--- a/install.sh
+++ b/install.sh
@@ -1035,15 +1035,16 @@ download_latest_bundle_and_extract_if_needed() {
 
   do-download() {
     rm -rf $BUILD_DIR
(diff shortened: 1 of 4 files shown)


kentonv:
@wtogami What do you think?

kentonv:
Now that I think of it, the keyring should not be in the bundle. I originally did this because I was thinking the auto-updater would be using it too, but then I decided the updater should use ed25519 where no keyring is needed. (Yay elliptic curves.)

The installer can download the keyring separately, and this will allow it to verify the tarball before unpacking it, which avoids the possibility of a malicious tarball leaving a mess under /opt/sandstorm.

Working on this now...

kentonv:
Yeah, that's much better.

jedisct1:
Consider [Minisign](https://jedisct1.github.io/minisign/) maybe?

kentonv:
@jedisct1 For the installer, we need to bootstrap off of tools that people already have, which unfortunately means GPG.

For the purpose of verifying updates (to be implemented separately), I plan to use libsodium, since the verification will be done in our own C++ code. I may very well use minisign to generate the signatures during our release process. :)

install.sh
```diff
@@ -1037,7 +1037,26 @@ download_latest_bundle_and_extract_if_needed() {
     rm -rf $BUILD_DIR
     local URL="https://dl.sandstorm.io/sandstorm-$BUILD.tar.xz"
     echo "Downloading: $URL"
-    curl -A "$CURL_USER_AGENT" -f "$URL" | tar Jxo
+    curl -A "$CURL_USER_AGENT" -f "$URL" > /var/tmp/sandstorm-$BUILD.tar.xz
```
zarvox: `/var/tmp/` is a shared namespace, so another user on the same machine could place their own content at `/var/tmp/sandstorm-$BUILD.tar.xz`, or over the .sig or even the keyring files.

I'd suggest making a folder under `/var/tmp` with something like:

```
mktemp --directory --tmpdir=/var/tmp -t sandstorm-install.XXXXXXXX
```

and then doing all the other work under that folder.

paulproteus: <bikeshed importance="low">I suggest `/tmp` as these files are transient. /var/tmp is persistent across reboots.</bikeshed>

release.sh
```diff
@@ -81,9 +81,21 @@ git push origin "$TAG_NAME"
...
 echo "**** Pushing build $BUILD ****"
 
+rm -f $TARBALL.sig install.sh.sig
+
+# Sign the tarball and the install script. Note that we don't sign the channel build number because
+# it wouldn't accomplish anything: If an attacker wanted to provide an old number, they could
+# provide the old signature to match. If an attacker provided a number that hasn't been used
+# before, they would not be able to provide a matching package because no such signed package
+# exists.
+gpg -u 9DA8CCBD --detach-sig $TARBALL
```
paulproteus: Consider specifying the full GPG fingerprint here. I _think_ there's no workable attack if you take care of development workstation's keyring, but I figure, better safe than sorry.

install.sh
zarvox: I think he chose `/var/tmp` since `/tmp` is likely tmpfs and on some systems, limited to something tiny like 1MB.

install.sh
```diff
@@ -1037,7 +1037,26 @@ download_latest_bundle_and_extract_if_needed() {
...
+    curl -s -A "$CURL_USER_AGENT" -f "$URL.sig" > /var/tmp/sandstorm-$BUILD.tar.xz.sig
+    curl -s -A "$CURL_USER_AGENT" -f "https://dl.sandstorm.io/keyring.gpg" > /var/tmp/sandstorm-keyring.gpg
+
+    if which gpg > /dev/null; then
+      if gpg --no-default-keyring --keyring /var/tmp/sandstorm-keyring.gpg --status-fd 1 \
```
paulproteus: Massive +1 to use of detached signatures

I can't seem to come up with a way to break this. (Inline signatures have problems that you have successfully avoided.)

install.sh
paulproteus: +1 to the "Don't use a predictable filename" suggestion, mostly because that means the logic on retry gets harder.

I'm still hopeful we can gracefully handle re-downloading the bundle -- I've seen people's Internet connection flake out mid-download, and using a non-predictable filename would help with that.

install.sh
paulproteus: ...however, maybe people can convert a detached sig to an inline sig and do the same attack. Let me see.

paulproteus: Yes, you can do the "append arbitrary data" attack against a detached sig if you convert it to an inline sig, as far as I can tell.

But it _looks_ like gpg will insist on a detached signature in the two-argument `gpg --verify x.sig x` invocation:


```
[GNUPG:] NODATA 1
gpg: not a detached signature
```

So I guess this is reasonable. `man gpg` agrees. so I'm +1.

But if we know anyone who's more of a GPG expert than me, I'd be very happy to get their review on this. (e.g. @diracdeltas?) But I'm also OK with shipping this as-is, and flagging that for being done "eventually".

zarvox: Oh, if an attacker could spoof both the keyring and the signature, then they could make the content read `\n[GNUPG:] VALIDSIG 160D2D577518B58D94C9800B63F227499DA8CCBD` and sign with a different key, which would defeat this.

I guess we're assuming that HTTPS is protecting the keyring?  Maybe we should just inline the keyring into the installer script, since you're already trusting that?

install.sh
```diff
@@ -1037,7 +1037,26 @@ download_latest_bundle_and_extract_if_needed() {
     rm -rf $BUILD_DIR
     local URL="https://dl.sandstorm.io/sandstorm-$BUILD.tar.xz"
     echo "Downloading: $URL"
-    curl -A "$CURL_USER_AGENT" -f "$URL" | tar Jxo
+    curl -A "$CURL_USER_AGENT" -f "$URL" > /var/tmp/sandstorm-$BUILD.tar.xz
+    curl -s -A "$CURL_USER_AGENT" -f "$URL.sig" > /var/tmp/sandstorm-$BUILD.tar.xz.sig
+    curl -s -A "$CURL_USER_AGENT" -f "https://dl.sandstorm.io/keyring.gpg" > /var/tmp/sandstorm-keyring.gpg
```
paulproteus: If running as root, symlink attack mumble.

install.sh
zarvox: Never mind, I stand corrected.

Might still be worth inlining the keyring in the installer script though - one fewer call, and then you also have the keyring signed in your git tag, rather than tying it in later.

install.sh
```diff
@@ -1037,7 +1037,26 @@ download_latest_bundle_and_extract_if_needed() {
     rm -rf $BUILD_DIR
     local URL="https://dl.sandstorm.io/sandstorm-$BUILD.tar.xz"
     echo "Downloading: $URL"
-    curl -A "$CURL_USER_AGENT" -f "$URL" | tar Jxo
+    curl -A "$CURL_USER_AGENT" -f "$URL" > /var/tmp/sandstorm-$BUILD.tar.xz
+    curl -s -A "$CURL_USER_AGENT" -f "$URL.sig" > /var/tmp/sandstorm-$BUILD.tar.xz.sig
```
paulproteus: Same suggestion re: don't use predictable filename. Partly because using a predictable filename means that if two local users on the same machine try to install Sandstorm (e.g. planning to play with it in their $HOME), the second one will get "Permission denied" and I have run into that installing `pip` and it makes me so rageful to think that the `pip` developers back then didn't use multi-user UNIX systems and I hope we can be better than that.

install.sh
zarvox: The rust folk also appear to do this: https://github.com/rust-lang/rustup/blob/master/rustup.sh#L118

release.sh
wtogami: agreed

paulproteus:
I think that's all the comments I have. To summarize:

* IMHO avoid use of predictable filenames for temp files, unless there's some reason that the way you're using doesn't mess up multiple users installing Sandstorm and doesn't lead to symlink-based attacks.

* Everything else looks perfectly reasonably good.

* Also I can't find `160D2D577518B58D94C9800B63F227499DA8CCBD` on keys.gnupg.net, so I assume you haven't uploaded the public key anywhere. I'm happy to sign it (we can discuss details face to face) and then we can link to e.g. http://pgp.cs.uu.nl/ as one way (in addition to keybase) people can verify they should trust this key.

paulproteus:
One other feedback item here:

* This is downgrade-attackable, in that we seem to trust these signatures forever. We might want to only trust signatures from (e.g.) the past week/month, and then if we want things to stay valid, re-sign them. I'm honestly not sure what policy I'd want on that, but we should set one. We could also decide to punt on it for now.

wtogami:
> This is downgrade-attackable, in that we seem to trust these signatures forever. 

One approach is to trust only particular hashes instead of anything signed by a particular key.  On a case-by-case basis you can think about the wisdom of trusting a hash or a key.

release.sh
```diff
@@ -81,9 +81,21 @@ git push origin "$TAG_NAME"
...
+gpg -u 9DA8CCBD --detach-sig install.sh
+
```
diracdeltas: might want to do `--digest-algo SHA512` because otherwise i think this defaults to SHA1 (assuming you haven't set algo prefs in gpg.conf)

install.sh
diracdeltas: you definitely want to inline the keyring instead of downloading it every time this script is run; otherwise you haven't added an independent source of verification (an attacker who compromises the CDN or whatever that is serving the build can also replace the GPG keyring with their own).

zarvox: I agree with the sentiment, but can that CDN-compromising attacker make the following `[GNUPG:] VALIDSIG <full key fingerprint>`... get printed by providing an alternate keyring?

diracdeltas: yes, since the signature file also comes from dl.sandstorm.io

install.sh
kentonv: I actually intentionally wanted retry attempts to overwrite previous attempts rather than litter /var/tmp with files that would never be deleted, since it's annoyingly hard/fragile to make a bash script reliably clean up after itself in error cases. I guess the multiple user point is sort of plausible, in the sense of "it will probably never happen but it could theoretically happen". I suppose I could throw $USER into the filenames?

install.sh
kentonv: @diracdeltas Sorry, I don't understand. On line 1047, we check for a precise key fingerprint. How can a malicious keyring and a malicious signature cause the correct key fingerprint to appear without the signature being made with that key?

Note that we're assuming that `install.sh` itself was verified independently before execution, so an attacker cannot have changed the fingerprint we're looking for...

(That said, I think inlining the keyring seems like a generally good idea.)

install.sh
paulproteus: Auto-cleanup is why I like /tmp. I've never seen a Linux system with a
significantly limited amount of space in /tmp. Drew, have you? I can live
with /var/tmp though if people prefer it.

Also, Kenton, don't you still have symlink attack type problems, where one
user creates a symlink in /var/tmp waiting for another user to set up
Sandstorm, resulting in an arbitrary system file being overwritten with the
Sandstorm bundle and/or keyring? I could try to whip up a proof-of-concept
system DoS exploit with the current install.sh if you need a demo of this;
presumably I would make a symlink to /etc/shadow and have the Sandstorm
tarball overwrite that, and now no one can log in after someone runs the
Sandstorm installer. To make a useful exploit that escalates privileges
would be more challenging, but possibly possible.

wtogami: /tmp on Fedora and RHEL has been a ram drive for many years now.  I don't know about other distros.

zarvox: `/var/tmp` is theoretically also automatically cleaned up, FWIW.  http://0pointer.de/blog/projects/tmp.html further suggests that if you're working with large files you should avoid `/tmp`, but *shrug*.  There exist versions of Ubuntu that do the limited-size `/tmp` but I forget which ones.

Manual cleanup is maybe a little more annoying, but not really terrible:

```
WORK_DIR="$(mktemp -d)"
cleanup () {
  if [ -d "$WORK_DIR" ] ; then
    echo "Cleaning up "$WORK_DIR"
    rm -rf "$WORK_DIR"
  fi
}
trap cleanup EXIT
```

Then again, maybe you want to leave that folder around on failure so it can be inspected and maybe someone can figure out what went wrong, rather than it disappearing without a trace.

install.sh
diracdeltas: @kentonv oops, I missed the gpg fingerprint. nvm comment above.

kentonv:
> This is downgrade-attackable, in that we seem to trust these signatures forever.

My thoughts on that:
1. The statements we sign with our personal PGP keys transferring trust to the installer key will expire after one month, allowing us to rotate keys periodically. (This prevents a downgrade attack on the installer itself.)
2. If the mtime on `/opt/sandstorm/latest/buildstamp` is more than a month old, Sandstorm should complain loudly in the UI, on the assumption that we will always release more often than that. This also conveniently gives us a way to nag people who don't auto-update.
3. The installer should also refuse to install if `buildstamp` is more than a month old. I suppose I should add that now, before I sign the installer!

> One approach is to trust only particular hashes instead of anything signed by a particular key.

But then how do you prevent an attacker from serving up an old hash?

release.sh
kentonv: For this case, even if someone injected a dupe key into my keyring, the result would be pushing a good build with a bad signature, whereas an attacker probably would prefer to push a bad build with a good signature... :) But sure, I'll extend it.

install.sh
kentonv: @paulproteus Symlink attack does seem like a real problem.

Our packages are only ~20MB; perhaps worrying about available space in /tmp is a little silly.

PR #906, opened 2015-09-21
New icons: share access, app market, report issue, pronoun, source

neynah:
@zarvox


kentonv:
I think you should overwrite share.svg -- it's not used for anything else.

neynah:
Overwritten! Thanks.

ocdtrekkie:
I feel like bug.svg has some oddities on the right side. On the inside of the bug's body, near the middle leg, there is a weird thickness thing. And another one on the bottom right leg.

neynah:
@ocdtrekkie Here's a close-up. The thickness of the legs should be the same.

Maybe this will help identify what's wrong with it.

bugcloseup

ocdtrekkie:
Your PNG definitely looks much nicer than the SVG render I'm getting from Firefox.

ocdtrekkie:
This is how my Firefox is rendering it:

weirdbug

That being said, as long as the Sandstorm build process renders it correctly, I suppose that's what matters.

kentonv:
We actually switched to shipping SVG to the browser. However, this icon will be so small that I don't think you'll ever see the weirdness. :)

PR #909, opened 2015-09-22
Fix some low-hanging usability fruit

zarvox:
* Users generally interpret "Share" to mean "Share on social media",
  especially with the icon we provided.  We mean to communicate something else:
  that this is how you grant access to this grain.

* "Save this link" has no information scent for signing in.  Revert to "Sign in".

### DIFF
--- a/shell/packages/sandstorm-accounts-ui/login_buttons.html
+++ b/shell/packages/sandstorm-accounts-ui/login_buttons.html
@@ -16,11 +16,7 @@
       {{/if}}
     {{/if}}
   {{else}}
-    {{#if isCurrentRoute 'shared'}}
-      <a class="highlight">Save this link ▾</a>
-    {{else}}
-      <a>Sign in ▾</a>
-    {{/if}}
+    <a>Sign in ▾</a>
   {{/if}}
 </template>
 
--- a/shell/client/_topbar.scss
+++ b/shell/client/_topbar.scss
@@ -300,22 +300,18 @@ body>.topbar {
 
       background-image: url("/share.svg");
       background-repeat: no-repeat;
-      background-position: 8px center;
-
       @media #{$desktop} {
-        background-position: left;
-        background-color: #762f87;
+        background-position: 2px center;
         color: #cccccc; // Have the text match the icon.
                         // This is the single place the button text is visible.
-        padding-left: 32px;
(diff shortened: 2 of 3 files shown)


shell/packages/sandstorm-accounts-ui/login_buttons.html
```diff
@@ -17,7 +17,7 @@
     {{/if}}
   {{else}}
     {{#if isCurrentRoute 'shared'}}
-      <a class="highlight">Save this link ▾</a>
+      <a class="highlight">Sign in ▾</a>
```
kentonv: Do we still want to highlight this? Maybe we should remove the conditional altogether?

shell/packages/sandstorm-ui-topbar/topbar.html
```diff
@@ -41,8 +41,8 @@
       {{/each}}
     </ul>
     <ul class="navbar {{#unless showNavbar}}hide-desktop{{/unless}}">
-      <li class="navitem-open-grain {{#if isCurrentRoute 'selectGrain'}}current{{/if}}"><a href="{{pathFor route='selectGrain'}}"><span>Open...</span></a></li>
-      <li class="navitem-create-grain {{#if isCurrentRoute 'newGrain'}}current{{/if}}"><a href="{{pathFor route='newGrain'}}">New...</a></li>
+      <li class="navitem-create-grain {{#if isCurrentRoute 'newGrain'}}current{{/if}}"><a href="{{pathFor route='newGrain'}}">Apps...</a></li>
```
kentonv: I would rather hold off on this rename until the app details pages are ready. I know we debated this at length and not everyone agrees, but I think it could create additional confusion if we change it now. And anyway, it's nice for things to change together, so that when the sidebar changes it provides a hint that the UI within has changed too and users should check it out.

shell/packages/sandstorm-accounts-ui/login_buttons.html
zarvox: Good idea.

shell/packages/sandstorm-ui-topbar/topbar.html
zarvox: Okay, I'll rebase this out of this branch.

PR #910, opened 2015-09-22
Cryptographically verify updates to Sandstorm proper.

kentonv:
Updates are signed using ed25519 via libsodium. We don't use gpg for updates because it would be an extra dependency and it's hard to use correctly. (OTOH we have to use gpg for install verification since we need to bootstrap from tools people have; see pull request #901.)

The strategy is as follows:
- We maintain a list of signing keys, which for now has only one element.
- The public keys in this list are compiled directly into the Sandstorm binary.
- Private keys are kept in a keyring file.
- Every time we want to "rotate" the key, we actually add a new key to the list, and use it to sign future updates.
- Each update is shipped with a signature file containing signatures using *every* key in the list.
- Sandstorm, when updating, checks the signature file for a signature using the most-recent key it knows about.

The scheme makes it painless to rotate keys, since rotating keys will never prevent old versions from updating.

In the event of a compromise of the update signing key, the update system is reduced to relying on HTTPS for integrity until a new release is pushed with a new key added -- which would be almost a non-event, honestly.

In the event that the private signing key becomes inaccessible to us, it will become impossible to update existing clients without manual user intervention -- which would be catastrophic.

Therefore, we should err on the side of keeping many copies of the private key.

### DIFF
--- a/src/sandstorm/run-bundle.c++
+++ b/src/sandstorm/run-bundle.c++
@@ -1883,8 +1883,9 @@ private:
     // Verify signature.
     {
       context.warning("Checking signature...");
-      KJ_ON_SCOPE_FAILURE(
-          context.warning("*** SIGNATURE CHECK FAILED: please notify security@sandstorm.io ***"));
+      KJ_ON_SCOPE_FAILURE(context.warning(
+          "*** Aborting update because signature check failed! Most likely this is due to a "
+          "network glitch, but if you suspect an attack, notify security@sandstorm.io."));
 
       // Download and parse signature file for this update.
       capnp::StreamFdMessageReader signatureMessage(
--- a/src/sandstorm/run-bundle.c++
+++ b/src/sandstorm/run-bundle.c++
@@ -24,7 +24,9 @@
 #include <capnp/serialize.h>
 #include <capnp/compat/json.h>
 #include <sandstorm/package.capnp.h>
+#include <sandstorm/update-tool.capnp.h>
 #include <sodium/randombytes.h>
+#include <sodium/crypto_sign_ed25519.h>
 #include <stdlib.h>
 #include <signal.h>
 #include <limits.h>
@@ -290,7 +292,17 @@ public:
     }
   }
 
-  ~CurlRequest() {
+  explicit CurlRequest(kj::StringPtr url, int outFd): url(kj::heapString(url)) {
+    KJ_SYSCALL(pid = fork());
+    if (pid == 0) {
+      KJ_SYSCALL(dup2(outFd, STDOUT_FILENO));
+      KJ_SYSCALL(execlp("curl", "curl", isatty(STDERR_FILENO) ? "-f" : "-fs",
+                        url.cStr(), EXEC_END_ARGS), url);
+      KJ_UNREACHABLE;
+    }
+  }
+
+  ~CurlRequest() noexcept(false) {
     if (pid == 0) return;
 
     // Close the pipe first, in case the child is waiting for that.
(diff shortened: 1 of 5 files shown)


kentonv:
@jedisct1 I ended up not using minisign in this case because I control both the sender and the receiver code, so a trivial custom Cap'n Proto based format was easier to implement, especially with the requirements regarding key rotation. As usual, libsodium is a pleasure to use. :)

src/sandstorm/run-bundle.c++
```diff
@@ -290,7 +292,17 @@ public:
     }
   }
 
-  ~CurlRequest() {
+  explicit CurlRequest(kj::StringPtr url, int outFd): url(kj::heapString(url)) {
+    KJ_SYSCALL(pid = fork());
+    if (pid == 0) {
+      KJ_SYSCALL(dup2(outFd, STDOUT_FILENO));
+      KJ_SYSCALL(execlp("curl", "curl", isatty(STDERR_FILENO) ? "-f" : "-fs",
```
zarvox: Today I learned we rely on the host's copy of curl for updates.  I found this surprising, given that we tend to ship our own copy of basically everything else.

I guess it was just easier to reuse curl (which we already depended on at install time) than integrate an HTTP client?

src/sandstorm/run-bundle.c++
```diff
@@ -1861,17 +1873,55 @@ private:
...
-    int fd = download->getPipe();
-    unpackUpdate(fd, kj::mv(download), targetBuild);
+    CurlRequest(url, file);
+    KJ_SYSCALL(lseek(file, 0, SEEK_SET));
+
+    // Verify signature.
+    {
+      context.warning("Checking signature...");
+      KJ_ON_SCOPE_FAILURE(
+          context.warning("*** SIGNATURE CHECK FAILED: please notify security@sandstorm.io ***"));
```
zarvox: If I'm reading this right, this will complain loudly with a possibly-inaccurate message if simply downloading the signature file fails (think network timeout or 502 or something). I'm inclined to think that a brief network outage should be slightly less terrifying.  Maybe move the `CurlRequest` below out into the containing scope, and only complain if you successfully fetch the signature and then it fails verification?

src/sandstorm/update-tool.capnp
```diff
@@ -0,0 +1,54 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+@0x96c3fff3f4beb8fe;
+# This file contains schemas relevant to the Sandstorm self-updater.
+
+$import "/capnp/c++.capnp".namespace("sandstorm");
+
+struct PublicSigningKey {
+  key0 @0 :UInt64;
```
zarvox: Is there a good reason this is 4 `UInt64` rather than simply a `Data` or a `List(UInt8)`?  Having it not be a simple bytestring raises questions of order and endianness - is key0 the first or the last 64 bits of the key?

It looks like you're not actually using the fields on the reading side anyway; just relying on Cap'n Proto's layout to pack these together at the beginning of the data field, and to be tightly packed?

src/sandstorm/update-tool.capnp
```diff
@@ -0,0 +1,54 @@
...
+  key1 @1 :UInt64;
+  key2 @2 :UInt64;
+  key3 @3 :UInt64;
+  # ed25519 public key.
+}
+
+struct Signature {
+  sig0 @0 :UInt64;
```
zarvox: Same thing as above - I guess this is a pattern when using Cap'n Proto for smallish bytearrays that have length of some multiple of 8 bytes?

This pattern seems convenient enough and highly performant for C++ implementations, but it's a pain when you're dealing with a higher-level language that might not expose the raw buffer to client code.  For example, when I was adding the `X-Real-IP` support, it was irritating to have to manually split a bytearray into 64-bit chunks, then encode those as hex strings (because Javascript lacks a 64-bit numeric type) to represent IP addresses, because `ip.capnp` represents the 128-bit IPv6 address as two 64-bit uints.  In practice, all the C library calls expect an array of 16 8-bit uints as declared as in <linux/in6.h>, and a typed array would be a more idiomatic representation in JS.  If we wind up ever using these in JS or some other language (say we reuse these structures for signing app updates, or something), or store these keys or signatures as JSONified-capnp in mongo, they get pretty annoying to work with.

src/sandstorm/update-tool.capnp
dwrensha: I think this would ideally be a fixed-size list of type `UInt8[32]`, but unfortunately those aren't implemented yet for Cap'n Proto.

src/sandstorm/update-tool.capnp
dwrensha: I just want to point out: there's nothing stopping node-capnp from also exposing a `getDataSection()` method.

src/sandstorm/run-bundle.c++
kentonv: We also rely on `tar`, `xz`, and `id` from the host. The thing is, when we're executing outside our chroot, we can't easily run bundled binaries, because they might depend on a different libc, etc. The updater has to run outside the chroot because it needs to download a new chroot.

src/sandstorm/run-bundle.c++
kentonv: Note that the message will be followed by a more detailed (but technical) exception.

I suppose I could move the dire bits into the specific error messages below for a failed signature. OTOH, an attacker who can serve a bogus package would presumably block the signature from being downloaded rather than let it be downloaded and proved invalid. So maybe what I really need to do is have this message say: "This may be a temporary glitch, but if the problem persists, it may indicate an attack. Please notify security@sandstorm.io."

Of course given that a signature check is taking place and therefore an attacker would gain nothing from serving a bogus package, maybe it's safe to assume that the problem almost certainly isn't an attack. If the the attacker were trying to block you from updating, they'd block all communication, so the package would never have downloaded in the first place.

Hrm.

src/sandstorm/update-tool.capnp
kentonv: Yeah, eventually this will be replaced by an inline array or UInt256, whenever we decide which one we want and actually implement it.

Issue #873, opened 2015-09-11
Design for an individual app's page (new UI)


neynah:
This is the page you'll see after you click on an app under "Apps" tab.

1st draft for desktop and mobile. @zarvox 

1. Am I missing anything?
2. Thoughts on the subtle highlight of the "Create new..."?

![app-page1](https://cloud.githubusercontent.com/assets/855341/9804647/dd8fc946-57e4-11e5-8181-d47f8914437a.png)
![app-page1-mobile1](https://cloud.githubusercontent.com/assets/855341/9804648/dd946f78-57e4-11e5-9e17-67a0cc0c5acd.png)

ocdtrekkie:
Huge +1 on this. I've wanted this info in the UI directly for a while. 

We should have a link to the source repository if it's listed in the manifest so people can report bugs.

neynah:
+1 on that idea ocdtrekkie.

kentonv:
Looks pretty good.

We will probably need an intermediate view between the mobile and wide views, since I suspect at 900px (where "desktop mode" usually starts) the design as you have it won't fit. I think what we can do here is basically take the layout of the metadata from the mobile version but still have the big icon on the left.

I just had a thought: instead of "verified", maybe we should say "signed by". This more literally states what the information means, and I think helps clarify the case where the author and the packager are different people. Of course it would be the packager signing the package.

Speaking of which, we'll need a design variant for the case where the author and packager are different people. We'll need to show both names ("Packaged by:" and "Upstream author:"). I wonder if we should hide the profile picture in this case, since it seems odd to show the packager's picture when they aren't the main author.

One more note: In addition to github and twitter we should link directly to the user's keybase profile. The key fingerprint can link there, but I think we should display their keybase profile name too, and it should probably be presented as a separate item from the fingerprint, in the same style as github and twitter. Maybe we should ask Keybase for a logo to use...

neynah:
Ideally I should be making mock-ups for mobile, wide, and intermediate as you suggest. Especially in the case where the sidebar will be expanded there will also be less width for content.

I'm curious as to what value the author/packager profile picture adds versus just having their name & username.

zarvox:
Looks very attractive.

"Distributed by" might be a less-jargony way to say "signed by"/"verified".

Do we actually have pictures for packages?  Or is this assuming that we'll piggyback on e.g. GitHub or someone else's profile pictures?  There's a potential infoleak if we're fetching those assets.

[...]

PR #913, opened 2015-09-22
icons for keybase and website

neynah:
Although I think this is usable-it may be difficult to recognize? It'd take me another hour to make improvements but I'm not sure if that'd help much.


paulproteus:
I would rather see a padlock-type icon to indicate some kind of assurance, rather than the Keybase icon in particular, personally. I can live with whatever though.

neynah:
The keybase icon will accompany the keybase username and the key will accompany the key(I think). See #873

kentonv:
Right, this icon is specifically to identify keybase, not to identify the concept of keys or signatures. It's analogous to the github icon.

PR #917, opened 2015-09-23
Use export-minimal to get smallest GPG keyring

paulproteus:
Micro-benchmark results on my personal key.

```
$ gpg --export --export-options export-minimal '97225E589D6B9E01533C485EEC4B033C70096AD1' | wc -c
8118
```

```
$ gpg --export '97225E589D6B9E01533C485EEC4B033C70096AD1' | wc -c
78782
```

Realistically this only affects people (like me) who have a lot of signatures
on their key. But I think it's smart to document it the export-minimal way
to help those who do.

### DIFF
--- a/docs/developing/publishing-apps.md
+++ b/docs/developing/publishing-apps.md
@@ -27,7 +27,7 @@ If you do it correctly, `cat pgp-signature | gpg` should print out the statement
 
 To verify your signature, you also need to export your public key and include it in your app package. You can run the following command, where `<key-id>` is a PGP key ID or a username associated with the key:
 
-`gpg --export <key-id> > pgp-keyring`
+`gpg --export <key-id> --export-options export-minimal > pgp-keyring`
 
 ## Add required metadata
 


paulproteus:
I will self-merge this on Friday (hence self-assigning it now) in case people have thoughts/advice/comments/etc.

I figure it's reasonable to do this because we're using the Sandstorm app maintainer's GPG keyring as a trust anchor and we don't need the other signatures that are on the key.


```
export-minimal
Export the smallest key possible. This removes all signatures except the most recent self-signature on each user ID. This option is the same as running the --edit-key command "minimize" before export except that the local copy of the key is not modified. Defaults to no.
```

are the upstream docs for this, see https://www.gnupg.org/documentation/manuals/gnupg/GPG-Input-and-Output.html

kentonv:
How does this affect subkeys? I believe when people are using a subkey, it's important that the primary key and subkey certificate are also present in the keyring, since we need to verify back to that.

paulproteus:
My sense from reading the docs is that it doesn't affect subkeys but I
should double-check.

paulproteus:
Turns out my key _has_ a subkey so I now truly believe this is safe.

kentonv:
Is the key you are using itself a subkey, or does it merely have subkeys? The hypothetical problem I'm imagining would come up when the key with which you are signing the package is itself a subkey of your main key -- in that case it's necessary for the keyring to also include your main key and the certificate chain. If you are signing with your primary key then there's no need for the keyring to include any subkeys.

PR #920, opened 2015-09-23
Document PGP-verified install.

kentonv:
This is (not entirely surprisingly) overcomplicated. It could perhaps be less complicated if we could find a more idiomatic way to "certify" release keys. However, GPG does not seem to be designed to handle quickly-expiring, frequently-renewed certificates:
- Its UI does not want to let us issue a new certificate while an existing one is still valid.
- When verifying, an expired certificate is still displayed with only a cryptic extra 'X' character to indicate that it is expired.

For now, we are falling back to natural-language certificates instead, but this means that the user must manually verify the key fingerprint, which is itself surprisingly hard through the GPG UI, assuming we don't want to suggest that users use --status-fd and try to manually interpret the output.


kentonv:
@paulproteus Please review.

@diracdeltas Any hints on how to do expiring, frequently-renewed certificates in PGP in a reasonable way?

docs/install.md
```diff
@@ -1,38 +1,64 @@
...
+If you'd rather not trust HTTPS at all, even from Github, another option is PGP-verified install.
+
+1. Determine the PGP key of a Sandstorm developer you trust. There are several ways to do this:
+    * Web of trust (for PGP experts).
+    * Meet us in person and ask for our business cards.
+    * Use our Keybase profiles, for example: [Kenton Varda (kentonv)](https://keybase.io/kentonv), [Asheesh Laroia (asheesh)](https://keybase.io/asheesh), [Drew Fisher (zarvox)](https://keybase.io/zarvox)
+2. Download that developer's corresponding release key certificate [from the Sandstorm github repo](https://github.com/sandstorm-io/sandstorm/tree/master/keys).
+3. Verify the certificate with GPG. For example:
+
+        gpg --decrypt release-certificate.kentonv.sig
```
paulproteus: You've done something here I haven't actually seen before with GPG: Create a ascii-armored text thing that has its signature as part of the document (?). I'm curious what GPG commands you used!

Seems to verify fine, anyway.

docs/install.md
```diff
@@ -1,38 +1,64 @@
...
+curl https://install.sandstorm.io | bash
 ```
 
-Note: If installing Sandstorm under LXC / Docker, you will need to choose the option to
-install as a non-root user. Unfortunately, this means the development tools will not
-work. This is due to the interaction between Sandstorm and Docker's use of Linux
-containerization features and missing features in the Linux kernel which we
-hope will be fixed eventually. For non-development purposes, Sandstorm should run just fine
-under Docker.
+## Option 2: Github-verified install
```
paulproteus: Proposed:

s/Github-verified install/HTTPS-verified install from GitHub/

paulproteus: BTW `raw.githubusercontent.com` is on Fastly, so (hilariously) this is just the choice of one CDN (github+fastly) vs. another CDN (sandstorm's GCE machines + CloudFlare). Whatever, though.

paulproteus: Doing this fwiw results in them getting the `install.sh` from master, which is little unexpected.

If we want to do this, we should start publishing `install.sh` as part of our https://github.com/sandstorm-io/sandstorm/releases process and (ugh!??) link to the latest one at the time of writing of the docs here. Or link to https://github.com/sandstorm-io/sandstorm/releases and say "Find the latest release and find its install.sh and wget it".

docs/install.md
```diff
@@ -1,38 +1,64 @@
...
 
-## Installing the Easy Way
+Sandstorm requires Linux x86_64, with kernel version 3.13 or later.
 
-*Prerequisite:* Linux x86_64, with kernel version 3.13 or later.
+## Option 1: HTTPS-verified install
 
-To install on your own Linux machine, see
-[https://install.sandstorm.io/](https://install.sandstorm.io) or run:
+The easiest way to install Sandstorm is by running:
```
paulproteus: If this doc gets more pagerank than sandstorm.io/install/ then I'll be a little sad. Luckily (from some Googling) there's no serious risk of that right now.

Consider linking back to sandstorm.io/install/ here, but I'm not sure where. So maybe don't bother.

docs/install.md
```diff
@@ -1,38 +1,64 @@
...
 
-For reference,
-[nginx-example.conf](https://github.com/sandstorm-io/sandstorm/tree/nginx-example.conf)
-contains the http server part of nginx config used by Sandstorm Alpha.
-## Installing from Source
+This verifies that you're running our published installer script, even in the unlikely event that someone has compromised our download server or HTTPS certificate. The installer will verify signatures on all additional files it downloads.
+
+## Option 3: PGP-verified install
+
+If you'd rather not trust HTTPS at all, even from Github, another option is PGP-verified install.
```
paulproteus: s/Github/GitHub/

docs/install.md
```diff
@@ -1,38 +1,64 @@
...
+
+1. Determine the PGP key of a Sandstorm developer you trust. There are several ways to do this:
```
paulproteus: I find this item (just the first one) hard to read. Proposed alternate text:

1. Find and download the PGP key of a Sandstorm developer you trust. There are several ways to do this:
    * Web of trust (for PGP experts).
    * Meet us in person and ask for our business cards.
    * Use our Keybase profiles, for example: [Kenton Varda (kentonv)](https://keybase.io/kentonv), [Asheesh Laroia (asheesh)](https://keybase.io/asheesh), [Drew Fisher (zarvox)](https://keybase.io/zarvox)

We suggest the `keybase track` command as one way to do this.

docs/install.md
```diff
@@ -1,38 +1,64 @@
...
+    * Meet us in person and ask for our business cards.
+    * Use our Keybase profiles, for example: [Kenton Varda (kentonv)](https://keybase.io/kentonv), [Asheesh Laroia (asheesh)](https://keybase.io/asheesh), [Drew Fisher (zarvox)](https://keybase.io/zarvox)
+2. Download that developer's corresponding release key certificate [from the Sandstorm github repo](https://github.com/sandstorm-io/sandstorm/tree/master/keys).
+3. Verify the certificate with GPG. For example:
+
+        gpg --decrypt release-certificate.kentonv.sig
+
+    Or, using Keybase:
+
+        keybase decrypt -S kentonv release-certificate.kentonv.sig
```
paulproteus: Looks like you need `verify` for keybase:

```
npmer@charkha:~$ ~/node_modules/.bin/keybase decrypt -S kentonv /tmp/release-certificate.kentonv.sig error: can't perform secret-key action without a secret key
npmer@charkha:~$ ~/node_modules/.bin/keybase verify -S kentonv /tmp/release-certificate.kentonv.sig 
info: Valid signature from keybase user kentonv (not tracking)
info: Signed about an hour ago (Wed Sep 23 2015 16:20:25 GMT-0700 (PDT))
✔ public key fingerprint: 8802 23DF 25AA 25A9 433A F0FB 4067 8458 440D DCF1
✔ "kentonvarda" on twitter: https://twitter.com/kentonvarda/status/455457515073843200
✔ "kentonv" on github: https://gist.github.com/10603028
✔ "kentonv" on reddit: https://www.reddit.com/r/KeybaseProofs/comments/3hh2b3/my_keybase_proof_redditkentonv_keybasekentonv/
✔ "kentonv" on hackernews: https://news.ycombinator.com/user?id=kentonv
✔ admin of sandstorm.io via HTTPS: https://sandstorm.io/.well-known/keybase.txt
✔ admin of capnproto.org via HTTPS: https://capnproto.org/.well-known/keybase.txt
```

paulproteus: But then I don't know how to show the plaintext, hmm.

paulproteus: Also I didn't give the keybase command my private key on this machine. I presume I shouldn't need to.

paulproteus: Given the above, I guess I recommend publishing a `release-certificate.kentonv` and a `release-certificate.kentonv.sig`. Sad to have to publish two files though.

docs/install.md
paulproteus: Argh but if people `track` then this gets published to the web, which is not the most privacy-sensitive thing we can suggest. Hmm.

paulproteus: 1. Find and download the PGP key of a Sandstorm developer you trust. There are several ways to do this:
    * Web of trust (for PGP experts).
    * Meet us in person and ask for our business cards.
    * Use our Keybase profiles, for example: [Kenton Varda (kentonv)](https://keybase.io/kentonv), [Asheesh Laroia (asheesh)](https://keybase.io/asheesh), [Drew Fisher (zarvox)](https://keybase.io/zarvox)

If you use the `keybase` commands to do verification, it will help you with this process, so you can skip this step.

paulproteus: There's my new proposed text.

paulproteus:
One way we can do ​ad-hoc certificates a little nicer is to have the ad-hoc
certificate contain the full output of gpg --fingerprint.

So I would write the following, and I would sign it with my key.

<ad_hoc_certificate_plaintext>
I'm Asheesh Laroia, and I verify that the Sandstorm release key has the
following fingerprint.

$ gpg --no-default-keyring --keyring ./release-keyring.gpg --list-keys
--fingerprint
./release-keyring.gpg
---------------------
pub   2048R/9DA8CCBD 2015-09-19
      Key fingerprint = 160D 2D57 7518 B58D 94C9  800B 63F2 2749 9DA8 CCBD
uid                  Sandstorm.io (releases) <support@sandstorm.io>
sub   2048R/B8112A1F 2015-09-19
</ad_hoc_certificate_plaintext>

docs/install.md
paulproteus: You _could_ also suggest using the Debian keyring as a trust anchor here with the following bullet:

* Install the `debian-keyring` package if you have Ubuntu or Debian, then extract Asheesh Laroia's key by running `gpg --no-default-keyring --keyring /usr/share/keyrings/debian-keyring.gpg --fingerprint asheesh@asheesh.org`.

docs/install.md
kentonv: `keybase decrypt` works but you have to have fully-configured the keybase tool with your private key. This seems to be a bug in keybase, probably because _normally_ the "decrypt" command (if it were actually decrypting) would require a private key present.

paulproteus: +1​ I can live with that.

kentonv: https://github.com/keybase/node-client/issues/212

docs/install.md
kentonv: `echo "blah" | gpg --sign --armor`

docs/install.md
kentonv: Eh, I think I am OK with people taking install.sh from master.

The purpose of "github-verified install" is not to defend against our CDN but rather to defend against our server being hacked. Since github is essentially a public ledger it's much harder to push an attack script there. It seems like starting from github makes a lot of people more comfortable, in any case.

I think to anyone who cares it's clear enough that github is still HTTPS-verified so I don't think we need to complicate the heading.

kentonv:
I came up with a new approach in which release-keyrings.gpg has all of our keys, and we instruct the user to import them into a temporary gpg workspace. It turns out that `gpg --decrypt` will in fact print full fingerprints when the key is not marked as trusted, which will always be the case here since the user hasn't marked any key as trusted. So then they can compare the fingerprint text.

Of course, advanced users will know which steps they can skip / do differently.

The instructions are still ugly but I think less ugly than before.

PR #922, opened 2015-09-24
Various tweaks to make the PGP install docs easier to follow.

kentonv:
Not sure why I didn't think of these things earlier.


kentonv:
@paulproteus

paulproteus:
Generated fresh docs! Also +1

paulproteus:
https://docs.sandstorm.io/en/latest/install/ as always is your friend

PR #923, opened 2015-09-24
Add Asheesh release certificate

paulproteus:
@kentonv please merge if you like

Also I think we should specify end-dates for this with days, like "after October 31 2015".

Also I think we should show the full output of `gpg --fingerprint` in the certificate.

Also I think we should upload the release key to the keyservers (it seems not to be).


kentonv:
> Also I think we should specify end-dates for this with days, like "after October 31 2015".

Maybe. My thought was that I plan to update my statements on the first of each month, but I wanted some flex room, so I left it a bit ambiguous. Since these statements are meant to be read by a human, I imagine they can make a judgment call that if they see the statement in early October it's probably fine, but if it's almost November then something is fishy, and if it's into November then it's definitely bad.

> Also I think we should show the full output of gpg --fingerprint in the certificate.

You meant including the UID? I worry that that could mislead people into thinking that the UID is something they should look for when verifying the key, when in fact the fingerprint is everything and the UID is completely irrelevant.

FWIW, though, you are welcome to format your statements differently if you like. :)

> Also I think we should upload the release key to the keyservers (it seems not to be).

Sure, can you do that? :)

PR #929, opened 2015-09-25
Refresh cert as needed

paulproteus:
I believe this is ready to merge. It achieves the following:

* When we have either 0 valid certificates, or 1 valid certificate that expires 3 or fewer days from now, then we get a new cert.

Implementation details:

* In the face of multiple certificates being valid, pick one in a reasonable way.

* It outright denies access to all non-SNI clients by giving them an invalid cert. (@jparyani you might find that amusing.)

* It only calls `.listen()` on the HTTPS socket (fd 3) if a key is available.

Out of scope for this pull request (to be handled tomorrow):

* Deal with clock skew.

### DIFF
--- a/meteor-bundle-main.js
+++ b/meteor-bundle-main.js
@@ -152,7 +152,8 @@ function monkeypatchHttpAndHttps() {
       var result = {
         ca: validCertificates[0].ca,
         key: fs.readFileSync(validCertificates[0].keyFilename, 'utf-8'),
-        cert: validCertificates[0].cert
+        cert: validCertificates[0].cert,
+        notAfter: validCertificates[0].notAfter
       };
 
       // Calculate re-key time.
@@ -218,8 +219,25 @@ function monkeypatchHttpAndHttps() {
           process.exit(0);
         }
       };
-      global.sandcats.hasNextRekeyTime = function() {
-        return !! sandcatsState.nextRekeyTime;
+      global.sandcats.shouldGetAnotherCertificate = function() {
+        // Get a new certificate if our current cert (a) does not
+        // exist, or (b) has fewer than three days left.
+        if (! sandcatsState.key) {
+          console.log("shouldGetAnotherCertificate: There is no key, so yes get a new one.");
+          return true;
+        }
+
+        var threeDaysInMilliseconds = 1000 * 60 * 60 * 24 * 3;
+        var now = new Date();
+        var expiry = sandcatsState.notAfter;
+        var timeLeft = expiry - now.getTime();
+
+        if (timeLeft < threeDaysInMilliseconds) {
+          return true;
+        } else {
+          console.log("Since", now, "is more than three days away from",
+                      new Date(expiry), "not renewing HTTPS cert yet.");
+        }
       }
 
       // Set up initial keys. We do this directly, skipping the
--- a/meteor-bundle-main.js
+++ b/meteor-bundle-main.js
@@ -48,6 +48,64 @@ function monkeypatchHttpAndHttps() {
   // in pre-meteor.js we sometimes need to bind HTTP sockets.
   http.createServerForSandstorm = http.createServer;
   var fakeHttpCreateServer = function(requestListener) {
+    function getNonSniKey() {
+      // Call this function to get a 'ca', 'cert', 'key' for browsers
+      // that don't support a HTTPS feature called Server Name
+      // Indication ("SNI").
+
+      // If we're lucky, we already have the files.
+      var hostname = 'for-clients-without-sni.sandstorm-requires-sni.invalid';
+      var basePath = '/var/sandcats/https/' + hostname;
+      if (! fs.existsSync(basePath)) {
+        fs.mkdirSync(basePath, 0700);
+      }
+      var keyBasename = "0.key";
+      var certBasename = "0.crt";
+      var keyFilename = basePath + "/" + keyBasename;
+      var certFilename = basePath + "/" + certBasename;
+      var files = fs.readdirSync(basePath);
+      if ((files.indexOf(keyBasename) == -1) ||
+          (files.indexOf(certBasename) == -1)) {
+        // Otherwise, we generate them synchronously. This could slow down the
+        // first startup of a HTTPS-enabled Sandstorm.
+        console.log("Generating default HTTPS key for use with non-SNI clients.");
+
+        // Generate 2048-bit key.
+        var keys = forge.pki.rsa.generateKeyPair({bits: 2048});
+        var keyAsText = forge.pki.privateKeyToPem(keys.privateKey);
+        fs.writeFileSync(keyFilename, keyAsText, "utf-8");
+
+        // Generate certificate.
+        var cert = forge.pki.createCertificate();
+        cert.publicKey = keys.publicKey;
+        cert.serialNumber = '01';
+        cert.validity.notBefore = new Date("2015-09-01T00:00:00Z");
+        cert.validity.notAfter = new Date("2025-09-01T00:00:00Z");
+        var attribs = [{
+          name: 'commonName',
+          value: hostname
+        }];
+        cert.setSubject(attribs);
+        cert.setIssuer(attribs);
+
+        // Sign it with the same key.
+        cert.sign(keys.privateKey);
+
+        // Save it to disk.
+        var certAsText = forge.pki.certificateToPem(cert);
+        fs.writeFileSync(certFilename, certAsText, "utf-8");
+        console.log("Non-SNI key generation done.");
+      }
+
+      var keyAsText = fs.readFileSync(keyFilename, "utf-8");
+      var certAsText = fs.readFileSync(certFilename, "utf-8");
+      return {
+        ca: [],
+        key: keyAsText,
+        cert: certAsText
+      };
+    };
+
     function getCurrentSandcatsKeyAndNextRekeyTime() {
       // Call this function to get up-to-date sandcats https key
       // information.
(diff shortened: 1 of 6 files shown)


paulproteus:
@kentonv any review you can give this will be nice. I will be doing more testing.

meteor-bundle-main.js
```diff
@@ -73,29 +113,150 @@ function monkeypatchHttpAndHttps() {
...
           if ((e.code === 'EACCES') || (e.code == 'ENOENT')) {
-            console.log("Skipping unreadable HTTPS key information file:", metadataFilename);
+            console.log("Skipping unreadable HTTPS key information file while examining:",
+                        metadataFilename);
             continue;
           } else {
-            throw e;
+            // Sometimes the server gives us a 0-byte response,
+            // presumably due to a connection getting reset?.
+            console.error("Got exception reading JSON from file:", metadataFilename, "Dazed and confused, but trying to continue.", e);
```
kentonv: style nitpick: please wrap at 100 columns

meteor-bundle-main.js
```diff
@@ -73,29 +113,150 @@ function monkeypatchHttpAndHttps() {
...
+        validCertificates.push(metadata);
+      }
+
+      // Sort by notBefore, ascending.
+      validCertificates.sort(function(cert1, cert2) {
+        return (cert1.notBefore - cert2.notBefore);
+      });
+
+      if (! validCertificates) {
+        // TODO: Be more robust here, though it's not clear how.
```
kentonv: Maybe choose a bogus certificate so that it's more obvious to the user that the problem is with certificates?

zarvox: A couple options:

- Attempt to rekey
- Print a log suggesting what action the user can do to fix this

meteor-bundle-main.js
```diff
@@ -73,29 +113,150 @@ function monkeypatchHttpAndHttps() {
...
+          result.nextRekeyTime = secondBest.notBefore;
+        } else {
+          result.nextRekeyTime = validCertificates[0].notAfter;
+        }
+
+        console.log("Will switch to", validCertificates[1].keyFilename,
+                    "at time", new Date(result.nextRekeyTime),
+                    "via real-time SNI re-keying system.");
+      } else {
+        console.log("We have only the one key that we know about. We'll need to renew soon.");
```
kentonv: Is the strategy here to have two certificates ready at all times? Does that mean someone installing Sandstorm from scratch receives two weeks of certs instead of one? Would it be possible instead to wait until there's a day left on the current certificate, then fetch the next one?

meteor-bundle-main.js
```diff
@@ -73,29 +113,150 @@ function monkeypatchHttpAndHttps() {
...
+      var result = {
+        ca: validCertificates[0].ca,
+        key: fs.readFileSync(validCertificates[0].keyFilename, 'utf-8'),
+        cert: validCertificates[0].cert
+      };
+
+      // Calculate re-key time.
+      //
+      // - If the cert we want to switch to is not valid yet, then we
+      //   re-key at its notBefore time.
```
kentonv: Should there be a grace period to account for clock skew? (The client's clock could be off several minutes.)

meteor-bundle-main.js
```diff
@@ -73,29 +113,150 @@ function monkeypatchHttpAndHttps() {
...
+      //
+      // - If the cert we want to switch to *is* already valid, then
+      //   we re-key at the notAfter time of our current cert.
```
kentonv: Why not rekey immediately in this case?

meteor-bundle-main.js
```diff
@@ -73,29 +113,150 @@ function monkeypatchHttpAndHttps() {
...
+      // there are valid certificates for us to use. This could be a
+      // problem if BASE_URL is https but we aren't ready.
+
+      // Create a local variable with key information.
+      var sandcatsState = {};
+
+      // Create a global that others can call into if they want to
+      // re-key, such as when `sandcats.js` downloads a new
+      // certificate.
+      global.sandcats = {};
```
zarvox: Why a global here, rather than e.g. pre-meteor instantiating a Sandcats object, which carries these functions?

meteor-bundle-main.js
```diff
@@ -73,29 +113,151 @@ function monkeypatchHttpAndHttps() {
...
+            console.log("Skipping unreadable HTTPS key information file while examining:",
+                        metadataFilename);
             continue;
           } else {
-            throw e;
+            // Sometimes the server gives us a 0-byte response,
+            // presumably due to a connection getting reset?.
+            console.error("Got exception reading JSON from file:", metadataFilename,
+                          "Dazed and confused, but trying to continue.", e);
+            continue;
```
paulproteus: Addressed style nitpick (requesting reflow of very long line onto two lines).

meteor-bundle-main.js
paulproteus: I partly like the idea of choosing a bogus certificate so it's more obvious the problem is with certificates, but I don't like that this would result in users seeing a "OH NO YOUR BROWSER DETECTED SOMETHING TERRIBLE" message.

I think since re-eying only takes ~30 seconds, what I'd want to do is here is:

* Log a message indicating what's up, and
* Don't listen on port 443 yet, and
* Trigger re-keying, and
* When re-keying succeeds, restart the Sandstorm shell by doing `process.exit(99)` and allowing the auto restart process to restart the Sandstorm shell.

I want to restart at the end because the alternative is to provide more and more monkey-patching and proxy object-ing inside this monkey-patched fakeCreateServer() and I don't know that I can confidently do that reliably. It would require not just a mock `createServer` but a mock `HTTP.Server` instance that Meteor will register callbacks on, which seems a bit terrifying.

Possibly someone should tell me to use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy instead (so long as we have those in the version of Meteor we're using).

But note that having 0 key files is a pretty uncommon situation -- anyone who uses the install script will have key files generated during install. This is just for people switching HTTPS on after the fact. So I think that a lot of careful engineering effort in an unlikely codepath is not clearly a good idea.

meteor-bundle-main.js
paulproteus: The current strategy here is to have two certificates ready at all times. This is for two reasons: (1) It made implementing things easier and (2) if the Sandcats service is (somehow) down for a while, or super buggy, or people turn off their machine and then turn it back on a week later or something like that, I don't want people's certs to expire. (Of course, it _will_ auto re-key on startup, so maybe I could worry about that less.)

Alternate strategy 1: If this is your first cert, get just 1 week, after that, get 2 weeks

* Day 0: new install: have a duration of 1 week (first cert)

* Day 6: Fetch the next week (second cert)

* Day 7: Switch to second cert

* Day 7: Fetch a week after that (third cert)

Alternate strategy 2:

* Always do the renewal one day before the cert is going to expire. This scares me because it gives us a smaller window to deal with sandcats service bugs.

@kentonv do you have a preference for "Alternate strategy 1" or "Alternate strategy 2"?

meteor-bundle-main.js
paulproteus: Yeah. I was trying to think through the best algorithm here, and wanted to deal with that after submitting this basic implementation.

So thinking aloud:

* Scenario:
    * Sandstorm is already running, and
    * cert 1 expires tomorrow 12:00 PM
    * cert 2 starts today 12:00 PM
    * time right now is 12:00 PM

* Clock that is slow - thinks it's 11:55 AM
    * We keep using cert 1 until "12:00 PM" which is actually 12:05 PM
    * We switch to cert 2 at "12:00 PM" actually 12:05 PM
    * Result: This is OK (because cert 1 didn't expire yet).

* Clock that is fast - thinks it's 12:05 PM
    * We stopped using cert 1 at "12:00 PM" which was actually "11:55 AM"
    * We switched to cert 2 at "12:00 PM" actually 11:55 AM
    * Result: For five minutes, web clients see an expired certificate.
    * Proposal to fix this: Do the switch-over at the _average_ of the cert 1 expiry & cert 2 start date.
    * Alternate proposal: Do the switch-over some constant factor into cert 2 (e.g. 10 minutes) so long as cert 1 is still valid for that time.

meteor-bundle-main.js
paulproteus: Ya, I'm using a global here because I'm at the most top-level file, before Meteor exists yet, and I want to provide access to some state in this file. To try to be clean, I do it by proving a function that encapsulates the important aspect of state, rather than providing the raw state.

Other ideas welcome.

meteor-bundle-main.js
paulproteus: * Scenario:
    * Sandstorm is starting up right now, and
    * cert 1 expires tomorrow 12:00 PM
    * cert 2 starts today 12:00 PM
    * time right now is 11:58 AM

* Clock that is 5 min slow - thinks it's 11:53 AM
    * We keep using cert 1 until "12:00 PM" which is actually 12:05 PM
    * We switch to cert 2 at "12:00 PM" which is actually 12:05 PM
    * Result: This is OK (because cert 1 didn't expire yet).

* Clock that is 5 minutes fast - thinks it's 12:03 PM
    * We'll pick cert 1 since it's still valid and we prefer older certs that are still valid.
    * We will tell ourselves to re-key at tomorrow "12:00 PM" (actually 11:55 AM)
    * Result: This is OK (because we ditched cert 1 a few min early but that was OK).

meteor-bundle-main.js
paulproteus: Doing it this way maximizes how long we use each cert, but I'll probably rethink this as part of the "only renew four days before expiry" goal.

meteor-bundle-main.js
paulproteus: Per discussion Kenton prefers something more like "Alternate Strategy 2" where we pick a 2-3 day window. I'll go with a 4 day window.

meteor-bundle-main.js
paulproteus: It used to be the case that we calculated the answer to "Should I request a fresh cert from sandcats.io?" based on the rekey time. But now we don't do that. So I guess we can use the newest cert that is still valid if we want.

paulproteus: It'd be nice if we could always rekey immediately, but I prefer my current algorithm because it's easier for me to explain to myself how it operates in the face of clock skew. Can think about this more, but would prefer to do so on a separate pull request if the current version is not wrong.

meteor-bundle-main.js
paulproteus: Will handle clock skew in a separate pull request, if that's OK.

PR #930, opened 2015-09-25
Styles cleanup

zarvox:
* Remove a bunch of dead styles.
* Remove script that dealt with styles for elements that no longer exist.
* Refactor a bunch of CSS into SCSS nested blocks.


zarvox:
Yay, delete-code Friday.

PR #936, opened 2015-09-29
Fix clang 3.7 compilation errors.

dwrensha:
I get a bunch of errors like the following when I try `make continuous` with clang 3.7:

```
compile: sandstorm/sandstorm-http-bridge.c++
  In file included from /ekam-provider/canonical/sandstorm/sandstorm-http-bridge.c++:25:
  In file included from /ekam-provider/c++header/kj/async-io.h:29:
  In file included from /ekam-provider/c++header/kj/async.h:670:
  /ekam-provider/c++header/kj/async-inl.h:314:12: error: no matching function for call to 'from'
      return PtmfHelper::from<ReturnType, Decay<Func>, ParamTypes...>(
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  /ekam-provider/c++header/kj/async-inl.h:353:45: note: in instantiation of function template specialization 'kj::_::GetFunctorStartAddress<capnp::Response<sandstorm::ByteStream::
    WriteResults> &&>::apply<(lambda at /ekam-provider/canonical/sandstorm/sandstorm-http-bridge.c++:204:41) &>' requested here
              GetFunctorStartAddress<DepT&&>::apply(func)),
                                              ^
  /ekam-provider/c++header/kj/memory.h:344:21: note: in instantiation of member function 'kj::_::TransformPromiseNode<kj::_::Void, capnp::Response<sandstorm::ByteStream::WriteResults>, 
    (lambda at /ekam-provider/canonical/sandstorm/sandstorm-http-bridge.c++:204:41), kj::_::PropagateException>::TransformPromiseNode' requested here
    return Own<T>(new T(kj::fwd<Params>(params)...), _::HeapDisposer<T>::instance);
                      ^
  /ekam-provider/c++header/kj/async-inl.h:757:7: note: in instantiation of function template specialization 'kj::heap<kj::_::TransformPromiseNode<kj::_::Void, capnp::Response<sandstorm:
    :ByteStream::WriteResults>, (lambda at /ekam-provider/canonical/sandstorm/sandstorm-http-bridge.c++:204:41), kj::_::PropagateException>, kj::Own<kj::_::PromiseNode>, (lambda at /
    ekam-provider/canonical/sandstorm/sandstorm-http-bridge.c++:204:41), kj::_::PropagateException>' requested here
        heap<_::TransformPromiseNode<ResultT, _::FixVoid<T>, Func, ErrorFunc>>(
        ^
  /ekam-provider/canonical/sandstorm/sandstorm-http-bridge.c++:204:36: note: in instantiation of function template specialization 'kj::Promise<capnp::Response<sandstorm::ByteStream::
    WriteResults> >::then<(lambda at /ekam-provider/canonical/sandstorm/sandstorm-http-bridge.c++:204:41), kj::_::PropagateException>' requested here
          taskSet.add(request.send().then([](auto x){}));
                                     ^
  /ekam-provider/c++header/kj/async-inl.h:277:21: note: candidate function [with R = void, C = (lambda at /ekam-provider/canonical/sandstorm/sandstorm-http-bridge.c++:204:41), P = <
    capnp::Response<sandstorm::ByteStream::WriteResults> &&>] not viable: no overload of 'operator()' matching 'void ((lambda at /ekam-provider/canonical/sandstorm/sandstorm-http-bridge
    .c++:204:41)::*)(NoInfer<capnp::Response<sandstorm::ByteStream::WriteResults> &&>)' for 1st argument
    static PtmfHelper from(R (C::*p)(NoInfer<P>...)) { BODY; }
                      ^
  /ekam-provider/c++header/kj/async-inl.h:279:21: note: candidate function [with R = void, C = (lambda at /ekam-provider/canonical/sandstorm/sandstorm-http-bridge.c++:204:41), P = <
    capnp::Response<sandstorm::ByteStream::WriteResults> &&>] not viable: no overload of 'operator()' matching 'void ((lambda at /ekam-provider/canonical/sandstorm/sandstorm-http-bridge
```

After this patch, the compilation succeeds.

I'm puzzled about the root cause here. How could adding these return statements change anything?

I think that, if possible, KJ should be updated so that this patch isn't necessary. I'm too intimidated to play around with the template magic there, though.

### DIFF
--- a/src/sandstorm/backend.c++
+++ b/src/sandstorm/backend.c++
@@ -240,7 +240,7 @@ kj::Promise<void> BackendImpl::deleteGrain(DeleteGrainContext context) {
   if (iter != supervisors.end()) {
     shutdownPromise = iter->second.promise.addBranch()
         .then([context](Supervisor::Client client) mutable {
-      return client.shutdownRequest().send().then([](auto) {});
+      return client.shutdownRequest().send().then([](auto) { return; });
     }).then([]() -> kj::Promise<void> {
       return KJ_EXCEPTION(FAILED, "expected shutdown() to throw disconnected exception");
     }, [](kj::Exception&& e) -> kj::Promise<void> {
--- a/src/sandstorm/sandstorm-http-bridge.c++
+++ b/src/sandstorm/sandstorm-http-bridge.c++
@@ -201,7 +201,7 @@ public:
         auto request = responseStream.writeRequest();
         auto dst = request.initData(body.size());
         memcpy(dst.begin(), body.begin(), body.size());
-        taskSet.add(request.send().then([](auto x){}));
+        taskSet.add(request.send().then([](auto x){ return; }));
         body.resize(0);
       }
 
@@ -420,7 +420,7 @@ private:
         KJ_FAIL_ASSERT("Failed to parse HTTP response from sandboxed app.", error);
       } else if (messageComplete || actual == 0) {
         // The parser is done or the stream has closed.
-        taskSet.add(responseStream.doneRequest().send().then([](auto x){}));
+        taskSet.add(responseStream.doneRequest().send().then([](auto x){ return; }));
         return kj::READY_NOW;
       } else {
         taskSet.add(pumpStreamInternal(kj::mv(stream)));
@@ -554,7 +554,7 @@ private:
       auto request = responseStream.writeRequest();
       auto dst = request.initData(data.size());
       memcpy(dst.begin(), data.begin(), data.size());
-      taskSet.add(request.send().then([](auto x){}));
+      taskSet.add(request.send().then([](auto x){ return; }));
     } else {
       body.addAll(data);
     }
@@ -626,7 +626,7 @@ public:
     auto request = clientStream.sendBytesRequest(
         capnp::MessageSize { data.size() / sizeof(capnp::word) + 8, 0 });
     request.setMessage(data);
-    tasks.add(request.send().then([](auto response) {}));
+    tasks.add(request.send().then([](auto response) { return; }));
   }
 
 protected:
--- a/src/sandstorm/supervisor.c++
+++ b/src/sandstorm/supervisor.c++
@@ -1590,7 +1590,7 @@ public:
     req.setOwner(owner);
     req.setRequirements(requirements);
     return req.send().then([context](auto args) mutable {
-      context.getResults().setSturdyRef(args.getToken());
+      return context.getResults().setSturdyRef(args.getToken());
     });
   }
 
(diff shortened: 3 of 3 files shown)


kentonv:
This seems like a clang bug. Is 3.7 released?

dwrensha:
> Is 3.7 released?

Yes.

```
$ clang++ --version
clang version 3.7.0 (tags/RELEASE_370/final)
Target: x86_64-unknown-linux-gnu
Thread model: posix
```

http://llvm.org/releases/download.html

kentonv:
https://llvm.org/bugs/show_bug.cgi?id=24989

I think we should stick with Clang 3.6 until this is fixed. Adding redundant returns everywhere is ugly.

dwrensha:
A perhaps somewhat nicer workaround is to add `-> void` annotations to the lambdas, like this:
https://github.com/dwrensha/sandstorm/commit/e2f31bab559aa2fcdaf1c3118cab824be944e49d

zarvox:
So, it's been a month and upstream doesn't appear to have looked at the bug yet.  I posit that merging one of the workarounds is preferable to the build being broken if someone tries to build with clang-3.7, because usually when install instructions specify a version, they mean "greater than or equal to this version".

dwrensha:
Yeah, I'm in favor of merging #1132.

kentonv:
I am inevitably going to add more instances of this, though. I also wonder whether the problem indicates that Clang 3.7 is miscompiling lambdas, which would be a good reason to avoid it. I'd somewhat prefer adding a note to the readme saying "Do not use 3.7 due to [this known bug]"...

paulproteus:
One alternative, fwiw, is that the build could fail early & loud with a
message about the compiler version being broken, telling you to get
clang-3.5 or something.

sprin:
I ran into this after upgrading to Fedora 23 (where clang 3.7 is the default). The patch resolved the compilation issues.

I also tried removing the Fedora clang packages and compiling the [3.6.2 release](http://llvm.org/releases/download.html#3.6.2) ([GPG pubkey](http://lists.llvm.org/pipermail/llvm-dev/2015-May/085517.html) for the signatures is not easy to find). This also resolved the errors. However `make continuous` seems to take several times longer, but I didn't time it. I'll stick with 3.6 for the time being.

Thanks for the good notes on the issue and patch here!

sprin:
I fixed the clang 3.6 performance issues. It turns out that clang compiles a debug build by default. To compile a standard "release" build:
```
cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release 
```
http://llvm.org/docs/GettingStarted.html#local-llvm-configuration

From a clean start, Sandstorm build times are now back to around 4 minutes.

PR #939, opened 2015-09-29
Account for clock skew

paulproteus:


paulproteus:
@kentonv requesting your review + merge.

paulproteus:
@jparyani remarks that 15 minutes isn't a huge amount of time. I'll do something else.

paulproteus:
I'm going to stick with this 15 minute algorithm since it's only a small tweak to what I have now. I'd like to simplify it further but I don't think I have time to make it significantly better before the launch.

@kentonv re-requesting your review.

paulproteus:
Thanks!

Issue #941, opened 2015-09-30
Get Sandstorm's built-in HTTPS to A or A+


paulproteus:
Filing this bug so I don't forget about it.

kentonv:
It appears that Node 0.10.x does not support either DHE or ECDHE. Later versions support both, but Meteor is stuck on 0.10.x. This is sad. We need to convince them to update.

https://github.com/meteor/meteor/issues/3666

ocdtrekkie:
I got an A when testing a Sandcats server. However, it states that "This server supports TLS 1.0 and TLS 1.1. Grade will be capped to B from January 2020." So we may wish to disallow TLS 1.0 and TLS 1.1. A lot of entities have done so this past year anyways, so people relying on it are going to have a bad time in a lot of places.

ocdtrekkie:
As previously noted, I just tested again and got a B. So we should look at disabling TLS 1.0/1.1 support ASAP, and then close this issue.

ocdtrekkie:
In 0.263, Sandstorm once again gets an A. This is a moving target, so we need not keep this open further without a specific action item.

PR #942, opened 2015-09-30
Stop using RC4; pick basically-reasonable ciphers

paulproteus:
This gets us an A- on the Qualys labs checks. I will investigate further
later to get a higher score.

### DIFF
--- a/meteor-bundle-main.js
+++ b/meteor-bundle-main.js
@@ -317,6 +317,18 @@ function monkeypatchHttpAndHttps() {
       // Configure options for httpsServer.createServer().
       var httpsOptions = getNonSniKey();
 
+      // Set default ciphers & tell HTTPS clients we really like our ciphers.
+      //
+      // This ciphers list is from the nginx configuration for
+      // oasis.sandstorm.io.
+      var ciphers = ('ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:' +
+                     'ECDH+AES128:DH+AES:ECDH+3DES:DH+3DES:RSA+AESGCM:' +
+                     'RSA+AES:RSA+3DES:!aNULL:!MD5:!DSS');
+      var honorCipherOrder = true;
+
+      httpsOptions.ciphers = ciphers;
+      httpsOptions.honorCipherOrder = honorCipherOrder;
+
       // The SNICallback option is a function that nodejs will call on
       // every inbound request with the inbound hostname. We get to
       // return an object of ca & key & cert to use.
@@ -346,6 +358,8 @@ function monkeypatchHttpAndHttps() {
         }
 
         return crypto.createCredentials({
+          ciphers: ciphers,
+          honorCipherOrder: honorCipherOrder,
           ca: sandcatsState.ca,
           key: sandcatsState.key,
           cert: sandcatsState.cert


paulproteus:
I'm going to manually test this on `rose.sandcats.io` now.

zarvox:
Consider using https://wiki.mozilla.org/Security/Server_Side_TLS as your reference ciphersuite source.

paulproteus:
Ya, I did that first, but then found that I still couldn't get PFS stuff to
work, so am using the oasis config for now, and will re-examine later.​

paulproteus:
Of note, I am not sure how to integrate the need for `dhparam` with the nodejs `https.createServer()` API in node 0.10.x. I filed a bug to remind me to think more about that later, #941

paulproteus:
https://www.ssllabs.com/ssltest/analyze.html?d=test6.rose.sandcats.io&hideResults=on is our `A-` report card, which I am happy enough with for now, and request review/merge from @kentonv .

Note that, as in all things in our little HTTPS world, this patch was slightly more complicated than I expected. We have to set this parameter not just for the default `createServer()` but also in the `SNICallback` or else we only use good certs for the `*.invalid` hostname. (-:

paulproteus:
Garply, retest please

kentonv:
This doesn't appear to match Oasis' config. Our cipher list is:

    ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-DSS-AES128-GCM-SHA256:kEDH+AESGCM:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA:ECDHE-ECDSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-DSS-AES128-SHA256:DHE-RSA-AES256-SHA256:DHE-DSS-AES256-SHA:DHE-RSA-AES256-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:AES:CAMELLIA:DES-CBC3-SHA:!aNULL:!eNULL:!EXPORT:!DES:!RC4:!MD5:!PSK:!aECDH:!EDH-DSS-DES-CBC3-SHA:!EDH-RSA-DES-CBC3-SHA:!KRB5-DES-CBC3-SHA;

I think it's important that we support PFS. ECDHE should work without the need for DH parameters. I am OK with not supporting DHE (which requires DH parameters to be secure); this means older browsers will not get PFS but new browsers will.

paulproteus:
I must have found the wrong line to copy.

I will look into this in a bit, presumably today. If Jason's still possibly
interested, that would be way cool. I'll ping Jason before I dig into this,
but it won't be for a little while.

kentonv:
I merged this, but I think I accidentally made it the subject of a rebase where I didn't change anything, but this caused the hash to change, so github doesn't recognize it as a merge. Oh well, it's merged. I also made a subsequent commit changing the cipher suite to actually match oasis, although this doesn't help as it turns out Node 0.10.x simply does not support DHE nor ECDHE. :( Meteor is stuck on node 0.10.x so we can't do anything about it for now.

Luckily our certs rotate every week?

PR #946, opened 2015-09-30
Adjust install script for HTTPS

paulproteus:
Changes:

* Rename a WARNING to a NOTE, since it did not indicate anything that is
  broken.

* Suggest `sudo sandstorm admin-token` by default.

* In the case of HTTPS setup, advise Chrome. (We also indicate that OCSP is
  the reason Firefox and other browsers might show an error message. Huge
  thanks to @ndarilek for noticing and reporting this issue.)

* Spin-loop for up to 90 seconds waiting for a HTTPS server to come online.
  (On my laptop, it only takes 7 seconds.)

* Enable HTTPS by default, and respect the `-i` parameter to let users opt
  out. (Mostly this is for the installer tests to opt out.)

* Only do HTTPS if ports 443 and 80 are available.

* For HTTPS servers, store HTTPS_PORT=443 in the config and use port 80 as
  the HTTP port.


paulproteus:
I am basically sure that the installer-tests will fail on this branch, or at least be massively slowed down. Going to go through them now.

But other than that this is ready for e.g. @kentonv 's review (although I guess I do own the install script).

paulproteus:
Happy to have some post-commit review on this fwiw. I'll be fixing up tests (even more, sorry about the noise) in a follow-up pull request.

kentonv:
Main issue I have is that if we decide not to obtain a cert due to port 80 or 443 being unavailable we should make it really clear to the user that this happened.

paulproteus:
Yeah totally agreed. Can see about fixing that today. Should be a simple
fix technically; I just want to make sure I also write a test for it and
mull over the string.

PR #956, opened 2015-10-01
Actually bind to port 80

paulproteus:
We were checking if port 80 is available, but never actually using it.

This fixes that.


paulproteus:
I've run the two tests I changed locally, and I think this is safe to merge. I'll merge it, ask Jenkins to verify, then ask Kenton to ship a new installer.

PR #958, opened 2015-10-02
split user IDs into account IDs and identity IDs: take 2

dwrensha:
This changeset performs the database migration to allow us to support multiple identities per account. It also updates all places (I think!) in the code that used to assume there was only one identity per account. Finally, it partially implements account merging and unmerging (with no UI yet), but disables those features for now, until they are more fully implemented and tested.

### DIFF
--- a/shell/packages/sandstorm-accounts-merge/accounts-merge-server.js
+++ b/shell/packages/sandstorm-accounts-merge/accounts-merge-server.js
@@ -123,6 +123,7 @@ Meteor.methods({
 });
 
 SandstormAccountsMerge.registerObservers = function (db, backend) {
+  // TODO(before enabling merging): make these handlers robust to exceptions.
 
   Meteor.users.find({"merging.status": "pending"}).observe({
     added: function(losingUser) {
@@ -185,6 +186,7 @@ SandstormAccountsMerge.registerObservers = function (db, backend) {
       // Transfer grain storage to new owner.
       // Note: We don't parallelize this because it can cause some contention in the Blackrock
       //   back-end.
+      // TODO(before enabling): make this robust to mid-transfer server crashes.
       grains.forEach(function (grain) {
         return waitPromise(backend.cap().transferGrain(grain.userId, grain._id, winningUserId));
       });
@@ -230,6 +232,7 @@ SandstormAccountsMerge.registerObservers = function (db, backend) {
       // Transfer grain storage to new owner.
       // Note: We don't parallelize this because it can cause some contention in the Blackrock
       //   back-end.
+      // TODO(before enabling): make this robust to mid-transfer server crashes.
       grains.forEach(function (grain) {
         return waitPromise(backend.cap().transferGrain(grain.userId, grain._id, destUserId));
       });
--- a/shell/packages/sandstorm-accounts-merge/accounts-merge-server.js
+++ b/shell/packages/sandstorm-accounts-merge/accounts-merge-server.js
@@ -28,6 +28,8 @@ waitPromise = function (promise) {
 
 Meteor.methods({
   mergeWithAccount: function (token) {
+    throw new Meteor.Error(400, "Accounts merging turned off for now. Needs more testing.");
+
     check(token, String);
     if (!this.userId) {
       throw new Meteor.Error(403, "Cannot merge accounts if not logged in.");
@@ -72,6 +74,8 @@ Meteor.methods({
   },
 
   unmergeToAccount: function (destUserId) {
+    throw new Meteor.Error(400, "Accounts unmerging turned off for now. Needs more testing.");
+
     check(destUserId, String);
     if (!this.userId) {
       throw new Meteor.Error(403, "Cannot merge accounts if not logged in.");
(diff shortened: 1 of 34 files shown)


shell/packages/sandstorm-db/db.js
```diff
@@ -33,22 +33,33 @@ if (Meteor.isServer && process.env.LOG_MONGO_QUERIES) {
...
+//       id: The globally-stable SHA-256 ID of this identity. This field must be present.
+//       service: String identifying the authentication scheme used by this identity, e.g. "github"
+//                or "google".
 //       name: String containing the display name of the user. Default: first part of email.
 //       handle: String containing the user's preferred handle. Default: first part of email.
 //       picture: _id into the StaticAssets table for the user's picture. Default: identicon.
 //       pronoun: One of "male", "female", "neutral", or "robot". Default: neutral.
+//       unverifiedEmail: Email address specified by the user.
+//       verifiedEmail: Only provided by some services. Cannot be directly edited by the user.
+//       main: True is this is the user's main identity.
```
kentonv: is -> if

shell/packages/sandstorm-db/migrations.js
```diff
@@ -191,6 +160,79 @@ function verifyAllPgpSignatures() {
...
+                     {$set: {identityId: identity.id}},
+                     {multi: true});
+    ApiTokens.update({"owner.user.userId": user._id},
+                     {$set: {"owner.user.identityId": identity.id}},
+                     {multi: true});
+    ApiTokens.update({"owner.grain.introducerUser": user._id},
+                     {$set: {"owner.grain.introducerIdentity": identity.id}},
+                     {multi: true});
+    ApiTokens.update({"requirements.permissionsHeld.userId": user._id},
+                     {$set: {"requirements.$.permissionsHeld.identityId": identity.id}},
```
kentonv: Whoa, what is this $ in the middle of the path? Can you add a comment explaining?

shell/packages/sandstorm-db/migrations.js
```diff
@@ -191,6 +160,79 @@ function verifyAllPgpSignatures() {
...
+                     {multi: true});
+  });
+
+  ApiTokens.remove({userInfo: {$exists: true}});
```
kentonv: multi: true?

shell/packages/sandstorm-accounts-merge/accounts-merge-server.js
```diff
@@ -0,0 +1,241 @@
...
+        newAction.userId = winningUserId;
+        db.collections.userActions.upsert({userId: winningUserId, packageId: action.packageId,
+                                           title: action.title, nounPhrase: action.nounPhrase,
+                                           command: action.command},
+                                          {$set: newAction});
+      });
+
+      var grains = Grains.find({userId: losingUser._id}).fetch();
+
+      db.collections.grains.update({userId: losingUser._id},
```
kentonv: I like how you made account merging resume in case of server failure. I think you'll need to do the same at the grain level, though. Currently if there is a crash immediately after this update runs, then the grains will end up all broken, and we wont retry moving them when starting back up because their user IDs have already changed.

shell/packages/sandstorm-db/migrations.js
dwrensha: Not needed in the case of `remove()`:
http://docs.meteor.com/#/full/remove
http://docs.mongodb.org/manual/reference/method/db.collection.remove/#query-considerations

shell/packages/sandstorm-accounts-merge/accounts-merge-server.js
```diff
@@ -0,0 +1,241 @@
...
+                                            sourceServices: oldServices}},
+                         $unset: servicesToRemove,
+                         $pull: {identities: {id: {$in: Object.keys(oldIdentities)}},
+                                 mergedUsers: destUserId} });
+  }
+});
+
+SandstormAccountsMerge.registerObservers = function (db, backend) {
+
+  Meteor.users.find({"merging.status": "pending"}).observe({
```
kentonv: This is pretty great in that if the server dies mid-merge it will retry on next start.

However, the function is currently not very robust against exceptions thrown within it. It would probably be a good idea to add a lot of try/catch blocks, not just around the whole thing but also individual parts, like maybe transferring of each grain. So if one grain fails to transfer, we don't get left in a totally inconsistent state -- only that one grain is affected. We can then inform the user of this and maybe fire off an alert email to ourselves (on Oasis at least).

shell/packages/sandstorm-db/migrations.js
dwrensha: Oh shoot, this doesn't actually do what I thought it did. The "$" only gets the first matched element. See http://docs.mongodb.org/manual/reference/operator/update/positional/#up._S_

Note that I use the "$" operator elsewhere. I'm not sure that this would be the right place to comment about what it means.

dwrensha:
Garply, retest this please

dwrensha:
I've addressed the comments.

PR #966, opened 2015-10-02
Add Piwik to docs.sandstorm.io

### LINKED ISSUE #965: Add e.g. piwik to understand docs traffic
Right now, `docs.sandstorm.io` has no way for us to know which pages are popular, etc. So maybe we can track traffic on it via Piwik.

Filing this & self-assigning so I don't lose track.

(Thanks @jadeqwang for the idea.)

paulproteus:
Close #965


paulproteus:
Seems to work in local testing!

Also Privacy Badger seems to block it. Sad. Oh well.

PR #968, opened 2015-10-03
Support webdav in WebSession

kentonv:
@mnutt, this is based on your PR, but I added a bunch of changes. Can you review my commits (which are hopefully clear and organized) and also test that this actually works? You'll need to rebuild your spk with the new http-bridge, of course.

If I broke stuff -- which I'd estimate 75% likely that I did -- then can you either: (a) debug, or (b) give me precise instructions on how I can build your app and test it myself?

Thanks!


kentonv:
Oh, I probably should have brought this up earlier, but also @mnutt can you sign the CLA:

https://sandstorm.io/cla/

Note that this is the same CLA that the Apache Foundation uses.

kentonv:
Garply, retest this please.

mnutt:
This actually all works perfectly. I tested it with my webdav client and backend and everything worked.  Unfortunately my backend is only Class 1, so there are Class 2 things I can't test, but it works great for my use case. Thanks for finishing it!

The way I tested it was to use this spk: https://s3.amazonaws.com/mnutt-misc/davros-0.8.0.spk and litmus (webdav unit test suite) : http://www.webdav.org/neon/litmus/ And then after building litmus run

```
litmus "http://api.local.sandstorm.io:6080/remote.php/webdav/" bearer <WEBKEY_HASH>
```

(unfortunately early on OwnCloud hard-coded remote.php in; it's something I just have to live with for interoperability)

kentonv:
OK, I'm going to go ahead and merge this so it can go through our testing cycle.

I sort of suspect that we'll turn up some bugs later on, but we can submit fixes then.

paulproteus:
Oh my goodness, glorious day!​

kentonv:
@mnutt I am seeing a failure with litmus and the spk you linked:

    2. options............... FAIL (server does not claim WebDAV compliance)

It looks like the app is in fact returning `false` for all `davClass` booleans in response to the `options()` call, for the path `/remote.php/webdav/litmus/`. Any idea why that might be?

All the other tests pass, though.

PR #974, opened 2015-10-05
Remove unused jparker:crypto-sha256 dependencies.

dwrensha:
After #958, we are no longer actually using the jparker:crypto-sha256 package. Note also that if we just want a function to return a hex-encoded SHA256 in the browser, then the meteor sha package (see https://www.meteor.com/utilities and https://atmospherejs.com/meteor/sha) might be a better choice. (We're already using it in our accounts-email-token package.)

This change also updates some documentation to specify that some hashes in our database are base64-encoded.


dwrensha:
Garply, retest this please

PR #977, opened 2015-10-05
Fix a typo that broke handle autocreation for dev accounts

zarvox:
I usually bitwise-or my handle with things ;)


paulproteus:
@zarvox I do too!

But only because of this bug.

PR #981, opened 2015-10-07
Support different netcats

paulproteus:


kentonv:
If nc fails, won't the whole script quit out because of `set -e`? Or does only the function fail? But that will break your code too, won't it?

paulproteus:
&& is permitted in set -e mode. Example:

asheesh@laptop:~ $ set -e
asheesh@laptop:~ $ false

(bash exits)

(start bash again)

asheesh@laptop:~ $ set -e
asheesh@laptop:~ $ false || echo sadness
sadness
asheesh@laptop:~ $

(bash is still around)

zarvox:
To be specific: exit codes used in conditional statements do not trigger quitting when `set -e` is active; only simple commands, pipelines, lists, or compound commands will trigger immediate exit on failure.  This means subcommands triggered by `if`, `elif`, `while`, `until`, `||`, `&&` and `!` all do not exit immediately.

kentonv:
Ah, I knew that exit values used as *inputs* to conditionals (like `if`) were excepted, but I didn't know that *outputs* of boolean operators were excepted even if they weren't used as inputs to anything. This offends my language-designer instincts but I guess that's the way it is.

kpreid:
Er, it looks like Kenton's taste is satisfied; only the _left hand side_ of `&&` or `||` is exempt, so it's just like `if`:

    $ bash; echo 'exited'
    $ set -e
    $ true && false
    exited
    $ bash; echo 'exited'
    $ set -e
    $ false || true
    $ false || false
    exited

In zarvox's example, `echo` _succeeded_, so there's no reason for `set -e` mode to exit.

kentonv:
@kpreid No, what you you seem to be observing is that only the left input to `||` or `&&` is exempted, while the right input is not. I am fine with this: the left input to a short-circuit logic operator is much like the input to `if`: clearly the error is being handled, therefore should not cause an exit.

However, what upsets me is that when the left side of `&&` fails, then this also means that _the overall expression_ evaluates to a failure, yet this failure does not trigger an exit even if the overall expression is not used as the input to any conditional.

    $ bash; echo 'exited'
    $ set -e
    $ false && true   # doesn't exit!
    $ echo $?
    1
    $ false && false   # still doesn't exit!!??!?!?!!
    $ echo $?
    1
    $ true && false
    exited

zarvox:
Yeah.  Even if you add `set -o pipefail` above, the behavior is the same.

I agree the behavior is tricky and probably crufty, and we would probably do well to avoid depending on it.

PR #982, opened 2015-10-07
Add code to fetch Keybase profile information.

kentonv:
We maintain a cache of this information on the Sandstorm server. We force a cache update when a new package is installed by the same author. All fetches are server-to-server and do not identify the user who triggered the fetch (and the cache is shared between users).

@zarvox, care to review?

### DIFF
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -1220,6 +1220,9 @@ if (Meteor.isServer) {
           // Remove potentially Mongo-incompatible stuff. (Currently Keybase returns nothing that
           // this would filter.)
           for (var field in proof) {
+            // Don't allow field names containing '.' or '$'. Also don't allow sub-objects mainly
+            // because I'm too lazy to check the field names recursively (and Keybase doesn't
+            // return any objects anyway).
             if (field.match(/[.$]/) || typeof(proof[field]) === "object") {
               delete proof[field];
             }
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -93,6 +93,8 @@ Packages = new Mongo.Collection("packages");
 //   url:  When status is "download", the URL from which the SPK can be obtained, if provided.
 //   isAutoUpdated: This package was downloaded as part of an auto-update. We shouldn't clean it up
 //     even if it has no users.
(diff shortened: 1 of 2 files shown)


shell/packages/sandstorm-db/db.js
```diff
@@ -1153,4 +1179,74 @@ if (Meteor.isServer) {
...
+          displayName: (profile.profile || {}).full_name,
+          handle: (profile.basics || {}).username,
+          proofs: (profile.proofs_summary || {}).all || []
+        };
+
+        record.proofs.forEach(function (proof) {
+          // Remove potentially Mongo-incompatible stuff. (Currently Keybase returns nothing that
+          // this would filter.)
+          for (var field in proof) {
+            if (field.match(/[.$]/) || typeof(proof[field]) === "object") {
```
zarvox: Huh, today I learned that MongoDB can't store arbitrary compliant JSON.

shell/packages/sandstorm-db/db.js
```diff
@@ -1153,4 +1179,74 @@ if (Meteor.isServer) {
...
+
+          // Indicate not verified.
+          // TODO(security): Asynchronously verify proofs. Presumably we can borrow code from the
+          //   Keybase node-based CLI.
+          proof.status = "unverified";
+        });
+
+        KeybaseProfiles.update(keyFingerprint, {$set: record}, {upsert: true});
+      } else {
+        KeybaseProfiles.update(keyFingerprint, {$unset: {profile: ""}}, {upsert: true});
```
zarvox: Not sure where this `profile` field that you're unsetting is coming from - it's not declared in `db.js`.  Perhaps this should be `displayName`?

kentonv: Oops, this was vestigial. What I meant to do was unset any existing cached data because Keybase reported no matches for the key. Fixed.

zarvox:
LGTM, code-wise.

I assume we'll only want to show proofs that are in the "verified" state in the author details section of the app details page?  Or did you intend to just punt to trusting keybase for V1?

Second question: do we prefer the keybase display name?  We should be able to pull the public key's display name out of the `pgpKeyring` as well.

zarvox:
As discussed in person, answer are:

1. Trust keybase for baseline info and link to keybase profile for now; add throbber/additional checkmarks if/when client-side verification is complete
2. Just use keybase display name for now.  Code to extract the display name from the keyring doesn't exist yet.

PR #983, opened 2015-10-07
Move dev and emailToken login data into user.identities.

dwrensha:


shell/packages/sandstorm-db/db.js
```diff
@@ -56,6 +55,11 @@ if (Meteor.isServer && process.env.LOG_MONGO_QUERIES) {
 //       verifiedEmail: Only provided by some services. Cannot be directly edited by the user.
 //       main: True if this is the user's main identity.
 //       noLogin: True if the user does not trust this identity for account authentication.
+//       devName: Only present when `service` is "dev". A string containing the name used to create
+//                this dev account.
+//       emailTokens: Only present when `service` is "emailToken". An array of objects of the form
```
kentonv: Note that this establishes a precedent that `identities` contains potentially-sensitive information (even though it's only a hash) next to information which is commonly publicly visible. The problem with this is that it could be tempting for someone to publish the whole identities array when they really only care about the public profile data.

If we're careful to provide common ways to access this data correctly then this could be avoided. But we should also consider whether the "profile" parts should be in a sub-object that can be more explicitly published.

Another possibility would be to keep the email tokens in the "services" object, since that's where OAuth tokens are currently stored and it's generally understood to contain secrets. But arguably some other fields of `identities` (e.g. `main` and `noLogin`) should not be public either so this doesn't totally solve the problem.

shell/packages/sandstorm-db/migrations.js
```diff
@@ -238,10 +238,28 @@ function splitUserIdsIntoAccountIdsAndIdentityIds() {
...
 }
 
-function appUpdateSettings() {
-  Settings.insert({_id: "appMarketUrl", value: "https://apps.sandstorm.io"});
-  Settings.insert({_id: "appIndexUrl", value: "https://app-index.sandstorm.io"});
-  Settings.insert({_id: "appUpdatesEnabled", value: true});
+function appUpdateSettings() {}
+// This migration was written to populate initial values for the "appMarketUrl", "appIndexUrl",
+// and "appUpdatesEnabled" settings. We've replaced it with a no-op becuase it's more convenient
+// to lazily initialize these settings, returning the canonical values as defaults.
```
kentonv: There's a problem with this: If the setting is not present in the database *on the client side*, it's possibly because it isn't set (and we should use the default) but it's also possibly because the subscription isn't ready (in which case using the default is dangerous, as some paranoid users may explicitly not want to connect to the app market under any circumstances).

Is there actually a good reason for this change or is it just aesthetics?

kentonv: If this is to make running tests easier, arguably tests should not be using the live app market / index anyway.

shell/packages/sandstorm-db/db.js
dwrensha: Proposal: we could make each identity be of the following form:
```
{  "id" : "a1bf13e1da9a356086d8c4fe5d52fd096d9ac32eddf95bec1e1f492e310afc82",
   "service" : {"emailToken": {"email": "dwrenshaw@gmail.com", "tokens": []}},
   "main" : true,
   noLogin: false,
   "profile": {"name" : "David Renshaw",
                  "handle" : "dwrensha",
                  "pronoun" : "male",
                  "verifiedEmail" : "dwrenshaw@gmail.com",
                  "unverifiedEmail" : "david@example.com" }
}
```
 
Then identities.profile could be freely shared with contacts.

If you agree with this proposal, do you think it makes sense to roll it into the present pull request?

dwrensha:
Thanks for the help @jparyani. After https://github.com/sandstorm-io/sandstorm/commit/b318b329381477c238bc696cc3918caf8f8e8ff8, the Incoming Mail test succeeds.

shell/packages/sandstorm-db/db.js
kentonv: I like that.

Though I'd say we shouldn't share e-mail addresses with contacts by default. Possibly arguably those fields shouldn't be rolled into profile? Not sure.

If we're going to do this, we should do it before the next release to avoid redundant migrations. Not sure what to do about our personal installs, though.

dwrensha: I'm right now working on rewriting the `moveDevAndEmailLoginDataIntoIdentities()` migration to include this change.

shell/packages/sandstorm-db/migrations.js
dwrensha: Ok, I've reverted back to the way things were and I've worked around the problem in the tests.

shell/packages/sandstorm-db/db.js
dwrensha: I've mostly finished this (see latest push), but it needs more testing and the email test in particular is still flaky. I'll take a further look later tonight.

dwrensha: Ugh. A problem with representing the service as capnp-style union like this:

```
"service" : {"emailToken": {"email": "dwrenshaw@gmail.com", "tokens": []}}
```

is that Mongo apparently doesn't make it easy to project out just the keys of an object. So we couldn't say "just return the service name" without explicitly specifying fields to exclude.

As such, I'm not super happy with this right now. I'll take another look in the morning.

dwrensha: Ah, maybe it's not so bad. We could do something like:

```
db.users.find({}, {"identities.service.emailToken._dummy": 1, "identities.service.dev._dummy": 1})
```

where the `_dummy` field need not exist in the actual data.

dwrensha:
I consider this ready to merge.

kentonv:
@zarvox let me know when you've taken a look and/or decided you don't care to take a look.

shell/client/grainview.js
```diff
@@ -251,41 +254,54 @@ GrainView.prototype.revealIdentity = function () {
...
     }
   }
-  this._userIdentity.set(identity);
+  this._userIdentityId.set(identity.id);
   this._dep.changed();
 }
 
 GrainView.prototype.doNotRevealIdentity = function () {
-  this._userIdentity.set(false);
+  this._userIdentityId.set(false);
```
zarvox: `null` or `undefined` seem more appropriate than `false` here

shell/packages/accounts-email-token/token_server.js
```diff
@@ -1,22 +1,21 @@
...
-  Meteor.users.update({"services.emailToken.tokens": {$exists: true}}, {
-    $pull: {
-      "services.emailToken.tokens": {
-        createdAt: {$lt: new Date(Date.now() - TOKEN_EXPIRATION_MS)}
-      }
-    }
-  }, { multi: true });
+  while (0 < Meteor.users.update(
+      {"identities.service.emailToken.tokens.createdAt": {$lt: new Date(Date.now() - TOKEN_EXPIRATION_MS)}},
+      {$pull: {
```
zarvox: TIL `$pull`

shell/client/grainview.js
dwrensha: As explained in the comments above: `false` means incognito; `undefined` means we still need to decide whether to reveal an identity.

shell/packages/accounts-email-token/token_server.js
```diff
@@ -53,8 +57,9 @@ Accounts.registerLoginHandler("emailToken", function (options) {
     };
   }
 
-  if (!user.services || !user.services.emailToken ||
-      !(user.services.emailToken.tokens)) {
+  var identity = user.identities[0];
```
zarvox: Does this mean I can't log in with my email token identity, if it's not the first identity for my user?

Also, consider using `identity.` instead of `user.identities[0]` below if you're going to pull it out into a variable.

dwrensha: The projection `{fields: {"identities.$": 1}` will only populate the `identities` array with the first element that matched the query. http://docs.mongodb.org/manual/reference/operator/projection/positional/#proj._S_

shell/packages/sandstorm-db/db.js
```diff
@@ -732,9 +733,12 @@ _.extend(SandstormDb.prototype, {
 
   getIdentity: function getIdentity (identityId) {
     check(identityId, String);
-    var user = Meteor.users.findOne({"identities.id": identityId}, {fields: {"identities.$": 1}});
+    var user = Meteor.users.findOne({"identities.id": identityId},
+                                    {fields: {"identities.$": 1, "services": 1}});
+    // TODO(cleanup): Minimongo doesn't support "$", so this actually doesn't work on the client.
```
zarvox: If that's the case, then consider moving these inside an `if (Meteor.isServer)` block?

shell/packages/sandstorm-db/migrations.js
```diff
@@ -305,5 +341,4 @@ function migrateToLatest() {
   }
 }
 
-// Apply all migrations on startup.
-Meteor.startup(migrateToLatest);
+SandstormDb.prototype.migrateToLatest = migrateToLatest;
```
zarvox: Don't worry about this right now, but this appears to work because `SandstormDb` is defined in a file alphabetically before "migrations.js" - we should probably make the ordering more explicit.

I could see a SandstormDb getting declared in `lib/db.js`, and methods on that object or on that prototype being defined in `common/` or `client/` or `server/` depending on where they're safe to use, since we have a number of methods that are client-only or server-only.

shell/packages/accounts-email-token/token_server.js
zarvox: Thanks for that link.  I tried a bunch of variations of google searches on "mongodb $ field" and never managed to get to that page.

zarvox:
@kentonv +1 from me.

shell/packages/sandstorm-db/migrations.js
dwrensha: The load order in a package is controlled by package.js. See https://github.com/meteor/meteor/issues/2731#issuecomment-58969066

Issue #984, opened 2015-10-07
Requesting certificates too often


paulproteus:
Tentative analysis by @kentonv 

- getCurrentSandcatsKeyAndNextRekeyTime() returns the oldest cert that is still valid.
- shouldGetAnotherCertificate() bases its decision based on the current key's validity, not any future key. Therefore as long as the old key is valid it will keep returning true.
- Also of note: rekey() will not select a newer certificate until the old one actually expires, even though the rekey time was carefully chosen to be before that.

I think getCurrentSandcatsKeyAndNextRekeyTime() probably needs to select the *newest* cert that has been valid for at least an hour?

paulproteus:
I can confirm the sandcats codebase hotfix works in that `sandcats.io` blocks the `/getcertificate` request.

I'll now work on fixing the underlying logic error in the sandstorm codebase so we can push a Sandstorm release.

paulproteus:
I _think_ what I originally wanted to do, for what it's worth, was for `rekey()` to switch to the newer certificate sooner rather than later, which means that `getCurrentSandcatsKeyAndNextRekeyTime()` should be changed to choose the newest certificate that has been valid for at least an hour.

PR #987, opened 2015-10-07
Calculate if to refresh certs based on newest valid cert

paulproteus:
Refs #984 

I _think_ this fix is correct. I'm going to think about it and also test it more now.

### DIFF
--- a/meteor-bundle-main.js
+++ b/meteor-bundle-main.js
@@ -115,52 +115,13 @@ function monkeypatchHttpAndHttps() {
       };
     };
 
-    function getCurrentSandcatsKeyAndNextRekeyTime() {
-      // Call this function to get up-to-date sandcats https key
-      // information.
-      //
-      // This returns an object with keys [ca, key, cert] which are
-      // valid options for https.createServer(). The object we return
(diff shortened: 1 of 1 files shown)


kentonv:
Once you're happy with this, can you squash the commits for easier cherry-picking?

paulproteus:
OK @kentonv I've squashed this and I am happy with it. Testing done:

paulproteus:
* On my laptop (against sandcats dev), verify that using this causes it to stop requesting certs when it has a cert that would work fine.

* On my laptop (against sandcats dev), verify that if I remove all certs, Sandstorm attempts to get a fresh cert. (It happens that sandcats blocks this request because my laptop has requested too many certs.)

* On my laptop (against sandcats dev), verify that if I remove recent certs leaving only one that expires on Saturday morning, my laptop attempts to renew. (It happens that sandcats blocks this request because my laptop has requested too many certs.)

* On rose.sandcats.io (against sandcats prod), verify that using this causes rose to use a cert that expires ~10 days from now, and stop auto-renewing needlessly.

Issue #659, opened 2015-08-04
Forever stuck in "Downloading..." state


campadrenalin:
When I try to download apps on my standalone instance, the app is not actually downloaded. I get the "Downloading..." page, stuck at 0% indefinitely.

I can actually load the SPK URLs via curl on the same machine. I know it's not a network issue, not on that level anyways. And the log shows nothing about this stuff. I have to conclude it's something broken in Sandstorm, or Firefox.

http://sandstorm.roaming-initiative.com:6080/install/3b04dd588ab7dbdf5a0299838fb3230d?url=http://sandstorm.io/apps/etherpad10.spk

paulproteus:
Thanks for filing this!

Interestingly, you're not the only person to report a problem like this.​

From what I can tell, you're hosted on Linode, which has nothing special
going on by default on their network. Maybe I can take a quick look at the
download code.

paulproteus:
First question: When you do that, does this message get logged to the
Javascript console?

Downloading app: http://sandstorm.io/apps/etherpad10.spk

​From what I can tell, it should be, based on
https://github.com/sandstorm-io/sandstorm/blob/master/shell/server/installer.js#L355

campadrenalin:
Hmm. Definitely glad I made an official issue, then.

Yeah, hosted on Linode. Should have nothing interesting in the network config.

Nope, no such message. The last thing I see (other than static resources like inbox.svg) is:

    GET http://sandstorm.roaming-initiative.com:6080/sockjs/043/tfv43q9j/websocket

And I kinda suspect this is a websocket connectivity problem, although I don't know how exactly to test that, since I haven't delved that deep into the Sandstorm internals.

paulproteus:
I'll dig into this, hopefully soon.

campadrenalin:
Awesome, thank you! Obviously let me know if there's anything I can do to
help you reproduce and debug.

kentonv:
> From what I can tell, it should be, based on
https://github.com/sandstorm-io/sandstorm/blob/master/shell/server/installer.js#L355

That's server-side, so it should be logged to the server log visible in the admin panel (or at `/opt/sandstorm/var/log/sandstorm.log`).

kentonv:
(The server log could definitely contain more info here. Take a look and let us know what you find.)

campadrenalin:
[...]

PR #994, opened 2015-10-08
Fix ownership and permissions /var/sandstorm/socket to allow reinstalling.

### LINKED ISSUE #660: Installing over existing install
When the installer is directed to install over an existing install, we should assume that the user intent is:
- Wants to keep data.
- Wants to change their answers to the installer's questions -- i.e. modify sandstorm.conf.
- Wants to re-download the package, in case it was corrupted somehow.

We should make sure to test this case.

In #659, a user who had installed over an existing install found that permissions were wrong on `/opt/sandstorm/var/sandstorm/socket` -- the directory somehow had gid=root, even though its parent and children all correctly had gid=sandstorm. This prevented the Sandstorm frontend from being able to open the backend socket. It's unclear whether this problem was in fact caused by over-installing.

dwrensha:
Fixes #660. See also #659.

### DIFF
--- a/src/sandstorm/run-bundle.c++
+++ b/src/sandstorm/run-bundle.c++
@@ -1674,6 +1674,13 @@ private:
         // Make socket available to server user.
         KJ_SYSCALL(chmod(socketPath.cStr(), 0770));
         KJ_SYSCALL(chown(socketPath.cStr(), 0, config.uids.gid));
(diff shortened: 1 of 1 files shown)


paulproteus:
Ooh. I guess I should honor your pull request by writing a test.

PR #995, opened 2015-10-09
Move fallback logic about packages to SandstormDb static methods

zarvox:
We have a decent amount of logic dealing with backwards-compatibility with app
titles and actions, and guessing values for new manifest fields based on older
ones.  This commit attempts to centralize that logic as some methods on the data
model in the sandstorm-db package.

This commit also add an appId field to devApps, to make accessing them more
like accessing Packages.  This makes dealing with packages that may be dev
packages or installed packages a bit simpler.

### DIFF
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -112,6 +112,7 @@ DevApps = new Mongo.Collection("devapps");
 //
 // Each contains:
 //   _id:  The application ID string (as with Packages.appId).
+//   appId: The same as _id, included to mimic Packages.appId.
 //   packageId:  The directory name where the dev package is mounted.
 //   timestamp:  Time when the package was last updated. If this changes while the package is
 //     published, all running instances are reset. This is used e.g. to reset the app each time
@@ -801,6 +802,10 @@ _.extend(SandstormDb.prototype, {
     return this.userActions(Meteor.userId());
   },
 
+  iconSrcForPackage: function iconSrcForPackage (pkg, usage) {
+    return Identicon.iconSrcForPackage(pkg, usage, this.makeWildcardHost("static"));
+  },
+
   getPlan: function (id) {
     check(id, String);
     var plan = Plans.findOne(id);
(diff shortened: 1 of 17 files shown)


shell/packages/sandstorm-db/db.js
```diff
@@ -112,6 +112,7 @@ DevApps = new Mongo.Collection("devapps");
 //
 // Each contains:
 //   _id:  The application ID string (as with Packages.appId).
+//   appId: The same as _id, included to mimic Packages.appId.
```
kentonv: This feels odd to me. We're relying on DevApps and Packages having some similar fields, but not all... Would it be cleaner to translate DevApps into Packages at the locations where we want to mix them?

zarvox: Maybe, though perhaps if we were being ORM-y we'd want some third entity that is the data shared between the two.  The fields that Packages have that DevApps don't have mostly relate to installation lifecycle, as far as I can tell:

* `status` would always be "ready" for devApps
* `progress` would be -1 (N/A)
* `error` is not applicable since `status` is always `ready`
* `shouldCleanup` relates to uninstalling apps, which you don't do with devApps
* `url` is only relevant when `status` is `download`, so N/A
* `isAutoUpdated` would be false, though maybe semantically we would want dev apps to not get cleaned up

That leaves:

* `authorPgpKeyFingerprint`, which we should maybe add to DevApps and attempt to populate from spk dev
* `timestamp`, which is unique to DevApps
* `packageId`, which is unique to DevApps (perhaps we should replace DevApps._id with packageId, and do `DevApps.findOne({appId: appId})` anywhere we relied on _id being an appId?)

In some ways it's like Packages and DevApps both extend some base that has an appId, a packageId, and a manifest.  Not sure how to canonically represent that in Mongo, though.

kentonv: Hmm, I think I would feel more comfortable if DevApps' _id became the package ID. Then all of the fields that exist in both have the same meaning in both, and we aren't repeating information in DevApps.

zarvox: All right, I'll look into doing that too.

zarvox: How would you feel about renaming the collection to DevPackages instead, to hint more at the analogous structure?

kentonv: I'm OK with renaming the variable. You may want to leave the collection name unchanged in Mongo... though I guess all members of this collection are ephemeral so maybe it doesn't matter, but I worry that there's some problem I'm not thinking of with changing the collection name.

zarvox: It'd possibly break if you've got an `spk dev` running across an upgrade, but that's going to be true if we switch _id to mean "package id" as well, so I can't think of any additional bad cases that also renaming the collection would trigger.

zarvox:
Rebased, but don't merge yet - still needs to address the comments above.

zarvox:
Manual tests look good.  Merge if Jenkins is also happy with it.

PR #996, opened 2015-10-09
Stop using transient <input> elements for app, backup, and picture uploads.

### LINKED ISSUE #851: Uploading backups and profile pictures fails in Chrome
Roughly 25% of the time, nothing happens after I select a file in the file picker.

The problem goes away if I delete these lines:
https://github.com/sandstorm-io/sandstorm/blob/5f6de637f6420d4c4c051f08bf7dadd5f197b33e/shell/shared/shell.js#L762
https://github.com/sandstorm-io/sandstorm/blob/5f6de637f6420d4c4c051f08bf7dadd5f197b33e/shell/packages/sandstorm-accounts-ui/account-settings.js#L184

It seems that the input element is getting garbage-collected before it can fire its "change" event.

dwrensha:
Also, remove the unused "uploadForm" template.

Fixes #851.


shell/packages/sandstorm-accounts-ui/account-settings.js
```diff
@@ -139,7 +139,15 @@ Template.sandstormAccountSettings.events({
       instance._profileSaved.set(true);
     });
   },
-  "change": function () { Template.instance()._profileSaved.set(false); },
+  "change": function (event, instance) {
+    if (event.target == instance.find("input[name='picture']")) { return; }
+    // Pictures get saved right away.
```
kentonv: Usually we place comments describing code above the code (unlike comments describing declarations).

zarvox:
hooray!

shell/packages/sandstorm-accounts-ui/account-settings.js
dwrensha: OK, done.

PR #1001, opened 2015-10-09
Address feedback from Meteor packaging user-test

paulproteus:


paulproteus:
Self-merging this now.

PR #1005, opened 2015-10-12
shell: link to log tab on server errors

### LINKED ISSUE #952: "see <install-location>/var/log/sandstorm.log for details" -> link to log tab instead
When testing SMTP on the admin panel, we sometimes report: "For internal server errors, see <install-location>/var/log/sandstorm.log for details."

Instead, we should link directly to the log tab, since we have that now.

zarvox:
Fixes #952.


paulproteus:
+1, requesting review/merge by @kentonv

PR #1019, opened 2015-10-13
Link to Sandcats terms & privacy

paulproteus:


paulproteus:
The sandcats ToS is now live, so merging this + rebuilding docs

PR #1024, opened 2015-10-14
By default, export SANDSTORM=1 so apps can detect running under Sandstorm

zarvox:


src/sandstorm/spk.c++
```diff
@@ -1093,7 +1093,12 @@ private:
         "  argv = [", argv, "],\n"
         "  environ = [\n"
         "    # Note that this defines the *entire* environment seen by your app.\n"
-        "    (key = \"PATH\", value = \"/usr/local/bin:/usr/bin:/bin\")\n"
+        "    (key = \"PATH\", value = \"/usr/local/bin:/usr/bin:/bin\"),\n"
+        "    (key = \"SANDSTORM\", value = \"1\"),\n"
+        "    # Apps that are adapted for Sandstorm may benefit from enabling additional\n"
```
paulproteus: I find this a little hard to read, so I propose the following:

```
Export SANDSTORM=1 into the environment, so that apps running within Sandstorm
can detect if $SANDSTORM == "1" at runtime, switching UI and/or backend to use
the app's Sandstorm-specific integration code.
```

src/sandstorm/spk.c++
```diff
@@ -1093,7 +1093,12 @@ private:
         "  argv = [", argv, "],\n"
         "  environ = [\n"
         "    # Note that this defines the *entire* environment seen by your app.\n"
-        "    (key = \"PATH\", value = \"/usr/local/bin:/usr/bin:/bin\")\n"
+        "    (key = \"PATH\", value = \"/usr/local/bin:/usr/bin:/bin\"),\n"
+        "    (key = \"SANDSTORM\", value = \"1\"),\n"
```
paulproteus: +1 to this.

src/sandstorm/spk.c++
zarvox: Wow, you're right, that's way better.

paulproteus: You're very kind!

paulproteus:
It may sound like I'm asking for paperwork/busywork but I think we should
see this change as an opportunity for a community-wide transition and
document what we want to do to make the transition as smooth as possible.
If you don't prefer to work on that, I probably look into that early next
week.

I think there are a few things we can do on top of this, and you're already
one step ahead of me:

* Make sure this variable gets passed-through into the various "platform
stacks" (you've already begun this, it seems, which is great)

* Submit pull requests to people's existing packages to get them to
consider adding these lines, with a link to a blog post (?) explaining why
we're doing this (if not blog post, then wiki page)

* Make sure docs.sandstorm.io mentions this variable in relevant places (
packaging-tutorial.md, packaging-tutorial-meteor.md, the platform-stacks
docs?).

I want to make sure that we don't forget anything, is the idea. Maybe the
above is the total list. If you're +0 on this, I can copy the above into a
pad that we can use as an ad-hoc todo list.

ocdtrekkie:
@paulproteus Re: "Submit pull requests to people's existing packages to get them to
consider adding these lines" - The use of this thing as something developers can detect has an obvious benefit, but if someone doesn't need it to make their app work, is there a particular reason they'd need to implement code?

paulproteus:
My theory is that a pull request on a package seems like a useful way to
alert Sandstorm app developers that this functionality is easily-gettable
in their existing packages, and I figure they can use the notification to
choose to either merge the change to their sandstorm-pkgdef.capnp with one
click on GitHub, or choose not to do so and click "Close".

I think a sizeable number of people don't read every pull request to
sandstorm-io/sandstorm so this might be the only way they hear about the
new option for detecting Sandstorm at runtime, and this can help people get
their app forks possibly upstreamed.

zarvox:
I'm +1 on adjusting platform stacks and docs.  I'm only +0 on (me personally) making pull requests to change existing packages - they'll likely need to switch from patching out functionality entirely to switching at runtime, which is probably better done by the app author themselves.

This is particularly valuable for apps that are:

1. Maintained as non-Sandstorm apps upstream
2. Need functionality changes to integrate well with Sandstorm
3. The Sandstorm integration is not currently upstreamable because the packager's changes would cause the non-Sandstorm version to regress.

Adapting the long tail of apps we've changed to use this so the changes can be upstreamed is more than I'm willing to commit to right here, but the intent is that whoever maintains the apps can use this pattern to make it possible to upstream their Sandstorm integrations.

You comments on outreach are valid, though I'm not sure making a pull request with a change that will not impact the function of the app in any way is a great use of time.

ocdtrekkie:
@paulproteus I suppose that would depend on the likelihood of getting upstreamed. Arguably building WordPress this way would be silly, because the chance WordPress is going to accept it into upstream is very low. (At least, in the short term.) There may also be apps being developed exclusively for Sandstorm, where this is also silly.

For developers directly supporting Sandstorm who also support standalone installation (EtherCalc, Wekan), it definitely makes sense to notify them of it.

My thoughts, anyways. I just wanted to ask because you made it sound like "something everyone should do differently" and I was kind of confused how this was that. Rather than just a useful thing some people would want to use.

zarvox:
The other `vagrant-spk` stacks need no modifications, as far as I can tell.  The other stacks are fine unmodified.

paulproteus:
+1 to this code & the discussion above. Requesting review/merge by @kentonv whenever works.

kentonv:
I don't think we should proactively send PRs to app authors. Technically we haven't introduced anything new here -- they could have done this all along. But if we send a PR I think they'll assume that the change is a new requirement of some sort, and they'll thus be confused. And if they don't actually need this, we may be wasting their time.

If the intent is that SANDSTORM=1 will be _required_ to be set for our tools to work, then it doesn't belong in apps' pkgdefs, it belongs somewhere in vagrant-spk where it will be set for all apps.

PR #1026, opened 2015-10-14
Fix notification styling

jparyani:


kentonv:
I improved the app ID check in 10202bf.

PR #1030, opened 2015-10-15
run-bundle: bind /etc to /etc.host RO and symlink interesting files

### LINKED ISSUE #895: In the case of DNS issues, try restarting Sandstorm
Right now, on my laptop, I see this message logged to the Sandstorm log:

```
Couldn't send UDP sandcats ping { [Error: getaddrinfo ESRCH] code: 'ESRCH', errno: 'ESRCH', syscall: 'getaddrinfo' }

events.js:72
              ^
Error: getaddrinfo ESRCH
    at errnoException (dns.js:37:11)
    at Object.onanswer [as oncomplete] (dns.js:124:16)
```

It's being logged in a loop, so presumably it causes Sandstorm to never fully bring up the interface.

We should fix this by being more resilient to DNS failures in the sandcats UDP ping code, but the real issue here is that `/etc/resolv.conf` changed on my system.

`sudo service sandstorm stop ; sleep 2 ; sudo service sandstorm start`

worked fine to make the error go away. This is seemingly because `run-bundle` will freshly copy `resolv.conf` in? (But I don't find evidence for that in `run-bundle`'s source code, in a quick skimming, but can look into this more in the future.)

A number of people have reported this sort of problem. Maybe it makes sense for there to be an RPC that the shell can do, so it can ask Sandstorm to make sure the right thing happens.

zarvox:
This way, when e.g. `resolv.conf` changes, we won't be using an out-of-date copy.

Fixes #895

### DIFF
--- a/src/sandstorm/run-bundle.c++
+++ b/src/sandstorm/run-bundle.c++
@@ -1190,16 +1190,22 @@ private:
   }
 
   void linkEtc() {
+    // We will create a symlink for the first child of /etc named in each line of etc.list to
+    // symlink that file or folder from the host into the /etc/ tmpfs.
     auto files = splitLines(readAll("etc.list"));
-
-    // Now bind in each file.
     for (auto& file: files) {
-      KJ_REQUIRE(file.startsWith(kj::str("/etc/")), "asked to link in file outside of /etc/");
-      auto source = kj::str("etc.host/", file.slice(5));
-      auto sourceAsSeenByTarget = kj::str("/", source);
-      if (access(source.cStr(), R_OK) == 0) {
-        auto target = kj::str(".", file);
-        KJ_SYSCALL(symlink(sourceAsSeenByTarget.cStr(), target.cStr()));
+      auto pathElements = split(file, '/');
+      KJ_REQUIRE(pathElements.size() >= 3, kj::str("Invalid path ", file));
+      KJ_REQUIRE(pathElements[0].size() == 0, kj::str("Relative path given in etc.list: ", file));
+      KJ_REQUIRE(kj::str(pathElements[1]) == "etc", "etc.list asked to symlink in file outside of /etc/");
+      auto etcChild = pathElements[2];
+      auto linkTargetAsSeenByLink = kj::str("/etc.host/", etcChild);
+      auto linkToCreate = kj::str("./etc/", etcChild);
+
+      // Only attempt to create the symlink if we haven't created it already.
+      struct stat stats;
+      if (lstat(linkToCreate.cStr(), &stats) < 0 && errno == ENOENT) {
+        KJ_SYSCALL(symlink(linkTargetAsSeenByLink.cStr(), linkToCreate.cStr()));
       }
     }
   }
--- a/src/sandstorm/run-bundle.c++
+++ b/src/sandstorm/run-bundle.c++
@@ -1195,9 +1195,9 @@ private:
     auto files = splitLines(readAll("etc.list"));
     for (auto& file: files) {
       auto pathElements = split(file, '/');
-      KJ_REQUIRE(pathElements.size() >= 3, kj::str("Invalid path ", file));
-      KJ_REQUIRE(pathElements[0].size() == 0, kj::str("Relative path given in etc.list: ", file));
-      KJ_REQUIRE(kj::str(pathElements[1]) == "etc", "etc.list asked to symlink in file outside of /etc/");
+      KJ_REQUIRE(pathElements.size() >= 3, "invalid path", file);
+      KJ_REQUIRE(pathElements[0].size() == 0, kj::str("relative path given in etc.list", file));
+      KJ_REQUIRE(kj::str(pathElements[1]) == "etc", "etc.list asked to symlink in file outside of /etc/", file);
       auto etcChild = pathElements[2];
       auto linkTargetAsSeenByLink = kj::str("/etc.host/", etcChild);
       auto linkToCreate = kj::str("./etc/", etcChild);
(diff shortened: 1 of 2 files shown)


src/sandstorm/run-bundle.c++
```diff
@@ -1183,18 +1189,17 @@ private:
     KJ_SYSCALL(sigprocmask(SIG_SETMASK, &sigset, nullptr));
   }
 
-  void copyEtc() {
+  void linkEtc() {
     auto files = splitLines(readAll("etc.list"));
```
paulproteus: As a side note: @zarvox you may know more about this than I do - is there a _directory_ of resolv.conf entries (`/etc/resolv.conf.d` ?) on some systems? If so we should think about adding them to `etc.list`.

Not necessarily on-topic for this pull request, but thought I'd mention.

paulproteus:
+1 fwiw, so excited to see this happen

src/sandstorm/run-bundle.c++
zarvox: Oh, do you have resolvconf installed?  can you look to see if /etc/resolv.conf is a symlink on your machine, and if it is, tell me what it points to?

paulproteus: Before installing `resolvconf`:

```
➜  ~  ls -l /etc/resolv.conf 
-rw-r--r-- 1 root root 52 Oct 14 18:11 /etc/resolv.conf
```

After:

```
➜  ~  ls -l /etc/resolv.conf         
lrwxrwxrwx 1 root root 31 Oct 14 18:34 /etc/resolv.conf -> /etc/resolvconf/run/resolv.conf
```

paulproteus: Presumably adding `/etc/resolvconf/run/resolv.conf` to `etc.list` will do the right thing, in that case.

(also uh did I just break my own Sandstorm install?)

paulproteus: FWIW:

```
➜  ~  ls -l /etc/resolvconf/run/resolv.conf 
-rw-r--r-- 1 root root 52 Oct 14 18:11 /etc/resolvconf/run/resolv.conf
```

So I think that's a good sign.

zarvox: Darn, this patchset can't handle that setup.  I'm also not sure how we would, short of handling that case specifically (by, say, readlink -f each of the things in `etc.list`, and making *that* the target instead, but that will probably break in other setups)  `/etc/resolv.conf` points to a folder under `/etc/`, but we need it to point to that folder under `/etc.host`.

paulproteus: If /etc/resolv.conf is a symlink to /etc.host/resolv.conf and
/etc.host/resolv.conf is a symlink to /etc/resolvconf/run/resolv.conf and
/etc/resolvconf/run/resolv.conf is a symlink to
/etc.host/resolvconf/run/resolv.conf , then I think we're golden.

Since we control /etc/ we can make the above happen, I believe.
​
The simplest hard-coded way to do this (IMHO) is to add
/etc/resolvconf/run/resolv.conf to the whitelist, and all will magically
Just Work. If /etc.host/resolvconf/run/resolv.conf doesn't exist, then we
create a broken symlink, but that's not a huge deal.

zarvox: Oh, you're right.  I'd say to add `/etc/resolvconf` to the whitelist, and then path traversal will take care of the rest.

Also we only create each symlink if its target exist at startup time, which I think is reasonable, but maybe we want to make dangling symlinks just in case the paths are added later?

zarvox: One possible issue is that the whitelist is generated dynamically at build time based on what Kenton's machine detects as the set of files needed to perform DNS resolution on his system, which may differ from that of other systems.  For instance, I suspect he does not use resolvconf, which leads to its absence from `etc.list` in our published packages.

Perhaps we should add some explicit whitelist entries, and autodetect the rest.

kentonv: Am I right in guessing that before this change, resolveconf actually did not break Sandstorm because we were copying each file, and therefore we accidentally dereference any symlinks?

zarvox: Sort of.  It was fine for the initial run because we copy `/etc/resolv.conf`, which derefs to a file with content, but as soon as the `resolv.conf` contents changed, we'd be still using a stale copy.

zarvox:
Okay, I think this approach will also work for `resolvconf` users!

zarvox:
KJ is too cool.  It would be amazing if we could generate some docs from the source, so these things could be linked about and indexed.  :D

I think this PR is ready to be submitted!

src/sandstorm/run-bundle.c++
```diff
@@ -1121,10 +1121,16 @@ private:
     KJ_SYSCALL(mount("/dev/urandom", "dev/urandom", nullptr, MS_BIND, nullptr));
     KJ_SYSCALL(mount("/dev/fuse", "dev/fuse", nullptr, MS_BIND, nullptr));
 
-    // Mount a tmpfs at /etc and copy over necessary config files from the host.
+    // Bind in the host's /etc as /etc.host.
+    // As noted in backup.c++, MS_BIND does not respect mount flags on the initial bind, and
+    // we have to issue a remount to set them.
+    KJ_SYSCALL(mount("/etc", "etc.host", nullptr, MS_BIND, nullptr));
+    KJ_SYSCALL(mount("/etc", "etc.host", nullptr, MS_BIND | MS_REMOUNT | MS_RDONLY, nullptr));
```
kentonv: Oh,  you'll want to throw `| MS_NOSUID | MS_NODEV | MS_NOEXEC` on here, otherwise this line will fail if these flags happen to be set on the source, because mount() thinks you're trying to remove the flags.

zarvox:
Added nosuid,nodev,noexec mount flags and updated comment.

kentonv:
Since this change is currently a regression for some users and Drew is on vacation, I'm reverting it for now. We should retry in the future with the regressions fixed.

PR #1039, opened 2015-10-16
WORK IN PROGRESS: Pass --repair to mongod

paulproteus:
General strategy to fix this:

* Create a doMongoRepair boolean param to startMongo()

* Every even-numbered time, doMongoRepair is false.

* Every odd-numbered time, doMongoRepair is true. So that way, we start Mongo without
  trying to do a repair, but if Mongo _does_ fail, then we should try starting it in
  repair mode.

Unfortunately I ended up with a large number of yaks to shave:

* I turned the args for /bin/mongod into a kj::Vector, rather than a big
  constant array. I seem to have introduced a segfault in the process (?)

* I noticed that runServerMonitor() would fail with an assertion error if
  Mongo crashes, before these changes.

* I noticed that if mongo exits too soon, it causes our call to waitpid() to
  crash, because there is no such PID.

* We can't call kill on the PID since it doesn't exist, either.

Probably other problems.

This commit is definitely not in a workable state, but I realize I am out of my
league for a "quick" fix. Help requested.

### DIFF
--- a/src/sandstorm/run-bundle.c++
+++ b/src/sandstorm/run-bundle.c++
@@ -1441,7 +1441,8 @@ private:
     uint64_t backendStartTime = getTime();
 
     context.warning("** Starting MongoDB...");
-    pid_t mongoPid = startMongo(config);
+    bool doMongoRepair = false;
+    pid_t mongoPid = startMongo(config, doMongoRepair);
     int64_t mongoStartTime = getTime();
 
     // Create the mongo user if it hasn't been created already.
@@ -1511,8 +1512,14 @@ private:
           backendStartTime = getTime();
         }
         if (mongoDied) {
+          // If Mongo died, let's try doing a repair. This way, every second mongo starting is a
+          // repair.
+          //
+          // Mongo's --repair argument causes Mongo to exit once the repair is over. This will
+          // re-trigger this branch, and we'll then re-start Mongo not in repair mode.
+          doMongoRepair = ! doMongoRepair;
           maybeWaitAfterChildDeath("MongoDB", mongoStartTime);
-          mongoPid = startMongo(config);
+          mongoPid = startMongo(config, doMongoRepair);
           mongoStartTime = getTime();
         }
         if (nodeDied) {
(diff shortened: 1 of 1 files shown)


paulproteus:
I suspect the best way forward is to have @kentonv take a look at this and fix it so my changes actually work. But if Kenton doesn't want to, I might request other help.

In general, it's clear we haven't tested the following cases, I think:

* node exits immediately.

* mongod exits immediately.

* mongod exits with a non-zero status code.

Those are bigger problems that are worth fixing IMHO.

src/sandstorm/run-bundle.c++
```diff
@@ -1556,36 +1563,65 @@ private:
...
-          "--auth", "--nohttpinterface", "--noprealloc", "--nopreallocj", "--smallfiles",
-          "--replSet", "ssrs", "--oplogSize", "16",
-          EXEC_END_ARGS));
+      kj::Vector<const char*> args;
+      args.add("/bin/mongod");
+      args.add("--fork");
+      args.add("--bind_ip");
+      args.add("127.0.0.1");
+      args.add("--port");
+      args.add(kj::str(config.mongoPort).cStr());
```
kentonv: kj::str here returns a temporary string which, because it's not being stored anywhere, is destroyed when the line finishes. But you've stored a pointer to said string into the vector, so that pointer is now dangling, possibly causing the segfault.

kentonv: One solution is to do:

    kj::String mongoPortStr = kj::str(config.mongoPort);
    args.add(mongoPortStr.cStr());

src/sandstorm/run-bundle.c++
```diff
@@ -1511,8 +1512,14 @@ private:
           backendStartTime = getTime();
         }
         if (mongoDied) {
+          // If Mongo died, let's try doing a repair. This way, every second mongo starting is a
```
kentonv: It would be nice to detect more precisely whether Mongo is in need of repair. There are a lot of other reasons why mongo sometimes dies, e.g. out of disk space, and doing a repair in those cases seems risky.

src/sandstorm/run-bundle.c++
```diff
@@ -1556,36 +1563,65 @@ private:
...
+      // immediately on startup, meaning the last one is two seconds in. So, we'll sleep for 3
+      // seconds to be safe.
+      // TODO(cleanup): There must be a better way...
+      int n = 3;
+      while (n > 0) n = sleep(n);
+    } else {
+      context.warning(kj::str("** Mongo exited with exitCode: ", exitCode));
+    }
 
     return KJ_ASSERT_NONNULL(parseUInt(trim(readAll("/var/pid/mongo.pid")), 10));
```
kentonv: The reason you're running into trouble with waitpid() and kill() later is because this line is returning a bogus PID in the repair case, since no Mongo daemon is started in that case. Probably it's returning the PID from the previous run.

Note that before your change, startMongo() always threw an exception if Mongo failed to start, but now you've changed the code to print a warning on error rather than throw.

Fundamentally, the problem here is that the "calling convention" of mongod is completely different when you use --repair. Normally, the process we exec forks a daemon and exits quickly, hence we wait for it to exit before continuing, and then we monitor the daemon process. But with --repair there is no daemon; the process we need to "monitor" is the one *we* forked. So perhaps what you want to do is, in the repair case, don't call waitForExit() at all, but just return the PID of the direct child process.

However, I think fundamentally the whole approach here -- of trying to use the same code to launch mongod and repair mongod -- is probably too hacky. When we want to do a repair, we should probably do it synchronously.

src/sandstorm/run-bundle.c++
paulproteus: Fair enough! re: out of disk space

Then I guess there are two ways to detect this condition:

1. Rely on mongo to detect it by reading its textual log. Unfortunately,
there isn't an exit code we can rely on:
http://docs.mongodb.org/manual/reference/exit-codes/

2. Look for the file ourselves, and if it is non-empty, then always do
--repair and check that --repair exits with status code 0.

So I vote for option 2.

We could optionally do the thing where we check that the PID in that file
doesn't exist. Might as well do that, but I don't think it matters much.

src/sandstorm/run-bundle.c++
paulproteus: +1

src/sandstorm/run-bundle.c++
paulproteus: +1, doing the repair synchronously strikes me as very reasonable, given that the frontend will break in half during that time.

paulproteus:
I think I should be able to get this pull request working, given the feedback above. I'll spend a short while on it today before trying to abandon it to someone else.

kentonv:
Assuming this is obsoleted by #1051.

paulproteus:
Yup thanks. Deleting branch.

PR #1050, opened 2015-10-19
If Sandstorm fails, don't kill the update process

paulproteus:
This commit has an upside:

* If Sandstorm crashes because e.g. `mongod` crashes, then we continue looking for updates.

This commit has a downside:

* Sandstorm now has no way to signal to a process monitor (e.g. systemd) that it has failed. This means that errors like binding to a port that is already in use prevent Sandstorm don't clearly alert the admin to a problem, which is pretty sad.

So I'm not sure I actually am +1 on this change. @kentonv your feedback very much requested.

### DIFF
--- a/src/sandstorm/run-bundle.c++
+++ b/src/sandstorm/run-bundle.c++
@@ -1406,8 +1406,7 @@ private:
           context.warning("** Updater died; restarting it");
           updaterPid = startUpdater(config, true);
         } else if (sandstormDied) {
-          context.exitError("** Server monitor died. Aborting.");
-          KJ_UNREACHABLE;
+          context.warning("** Sandstorm died. Keeping the update monitor alive just in case.");
         }
       } else if (siginfo.ssi_signo == SIGINT) {
         // Pass along to server monitor.


kentonv:
Did Sandstorm actually crash from things like "port already in use" before? Or did it go into an infinite loop of trying to start node and failing?

paulproteus:
It would actually crash, which has the upside that systemd monitoring shows the process as FAILED. You can repro by setting PORT=22 on a Sandstorm on a systemd Sandstorm install.

kentonv:
For advanced sysadmins, having systemd be aware of failure may be useful, but for users like me who know nothing about systemd, it's most useful if Sandstorm tries its best to self-recover. Consider a user who launched a VM with a one-click launcher and doesn't know how to open a shell, much less use it.

That said, going into emergency update mode when the port is in-use does not seem right...

Perhaps we should hold off on this until someone has time to change it so that update checks happen more often in this mode with "type=emerg" or something, so that we can get alerts when this is happening.

paulproteus:
I like that type=emerg thing, and it "shouldn't be that hard" so ​I'm cool
abandoning this until we have that.

PR #1058, opened 2015-10-21
Add app details page

zarvox:
Numerous changes here:

* Added a bunch of icons to the bundle
* Extracted a reusable grain table component
* Created reusable button styles, and applied their usage throughout the UI
* Renamed "New..." to "Apps" and "Open..." to "Grains"
* Switched "Apps" to show apps, not actions
* Added app-specific app details page with a grain list and action buttons
* Updated tests to launch grains from the app details page.
* Refactored various methods and styles

Still TODO:

* Switch the install flow to use the app details page instead


shell/packages/sandstorm-ui-app-details/app-details-common.js
```diff
@@ -0,0 +1,14 @@
+SandstormAppDetails = function(db, quotaEnforcer, appId) {
```
kentonv: This object is only constructed on the client, right? Why have a "-common.js" at all?

shell/packages/sandstorm-ui-app-details/app-details-common.js
```diff
@@ -0,0 +1,14 @@
...
+    this._db = db;
+    this._quotaEnforcer = quotaEnforcer;
+    this._appId = appId;
```
kentonv: nitpick: 2-space indents, not 4...

shell/packages/sandstorm-ui-app-details/package.js
```diff
@@ -0,0 +1,28 @@
...
+Package.describe({
+  summary: "Sandstorm UI app details page",
+  version: "0.1.0"
+});
+
+Package.onUse(function (api) {
+  api.use(["check", "reactive-var", "reload", "templating", "tracker", "underscore", "sandstorm-identicons"], "client");
+  api.addFiles(["app-details-common.js"], ["client","server"]);
+  api.addFiles(["app-details.html", "app-details-client.js"], "client");
+  api.addFiles(["app-details-server.js"], "server")
```
kentonv: Since the server file is empty, and -common.js contains code only used on the client, I'd go ahead and make this a client-only package. I think that's probably the right thing for all our UI packages, really.

shell/packages/sandstorm-ui-app-details/app-details-common.js
zarvox: Huh, good point.  I guess in theory the other sandstorm-ui-* probably deserve the same change.

shell/packages/sandstorm-ui-app-details/package.js
zarvox: Good call.  I probably wrote that a long time ago when I thought I needed to provide other subscriptions, and then wound up not needing any.

I'll fix it up.

shell/client/_topbar.scss
```diff
@@ -498,7 +501,19 @@ body>.topbar {
           // Make the "Open" and "New" verbs align with the start
```
zarvox: Update/remove comment.

shell/client/shell.html
```diff
@@ -564,14 +564,19 @@ <h4 class="share-with-others">Share with others</h4>
...
 
-<template name="selectGrain">
+<template name="grains">
   {{>title "Grains"}}
-  {{> sandstormGrainList}}
+  {{> sandstormGrainListPage}}
+</template>
+
+<template name="appDetails">
+  {{>title appTitle}}
```
zarvox: Move into sandstormAppDetails where appTitle is actually defined.

Add package dependency on sandstorm-ui-topbar.

paulproteus:
After an in-person conversation with @kentonv I'm +1 on keeping the "Report issue" link going to GitHub if there is a GitHub source code link, and falling back to email if there isn't a GitHub link, and in the near term we can provide a `sandstorm-pkgdef.capnp` property people can change to customize that.

kentonv:
Mostly looks great! I'm excited to launch this.

UI comments:

- I think "publisher" should be changed to "signed by". I understand that not everyone will know what this means, but the main point of this UI element is to communicate that the package is signed and who signed it. For people who know what it means, the label "signed by" is more useful, and for those who don't, we should be teaching them about the meaning rather than obfuscating it.

- The twitter link should go to the twitter profile page, not to the specific proof tweet. Similarly for Github. If we want to link to the proofs (which seems reasonable), we should add a second link like "[proof]".

- Maybe don't linkify the key fingerprint since it's redundant with the keybase profile link. Actually the best thing to do on clicking the fingerprint is probably to select it for copy/pasting. I'd also argue that the fingerprint should be first in the list.

- External links should probably open in a new tab to avoid losing Sandstorm state.

- The "publisher" drop-down responds kind of strangely to window width shrinkage.

- Let's make "report issue" go directly to github issues for code on github. So far I've been unable to find any apps that don't have issues enabled. (We'll plan on making the issues link customizable in metadata later.)

- I'm not thrilled with the purple create buttons. I think they are probably easily missed as-is. Hopefully @neynah can come up with some ideas here.

(Anything you don't finish today, I or someone else can probably work on while you're out.)

kentonv:
The more I think about it, the more I really like the in-table "create new" button.

I have in the past struggled to find how to create a new document on Google Docs. Yes, it sure seems pretty obvious once you have discovered that the button is in the upper-left, but I'm pretty sure it took me a few times of searching around the page before I got that. I feel like the in-table buttons as Nena originally designed them would be impossible to miss. Granted, this is intuition and I don't have data, but it's a strong feeling to me.

paulproteus:
w/r/t "signed by" vs. "packaging by": My perspective here is, I'm fine with whatever, but if we're unsure of something, let's do one of (a) defer it to @neynah or (b) get it in @neynah 's testing queue. Seems like we're happy with (b) for this word choice.

neynah:
If implementation weren't an issue I would highly prefer the in-table "create new" button. However, I'm also more than happy to spend some time and try to figure out a different display for the "create a new grain" button so it's less confusing.

kentonv:
I think in-table should be entirely doable.

Use `<td colspan=4>` or whatever to make a cell that spans all columns, then fill it with a button.

zarvox:
How do you left-align the text in the button with the text in the rest of the rows if you've done a colspan?

kentonv:
@zarvox The leftmost column looks to be fixed-width so you could use a fixed padding-left I suppose.

kentonv:
@zarvox Alternatively, have two columns, only have the `<button>` in the right column, but have the left also be clickable.

zarvox:
I'll take a second pass at implementing the in-table buttons.  I still think if you highlight them with an outline or bright colors in the grain list, it'll be distracting, and if you don't, then they'll be easily overlooked/feel uninteractive.  But I'm perfectly happy submitting this to user testing, or giving the original design a shot first.  (It would be cool to take some new users and time them at tasks like "create a spreadsheet", or do the "look at page for two seconds, then answer questions about it" test.)

For comparison: in Drive, the "New" button is uniquely saturated and commands the most attention.  It is approximately the only shape or object I notice on the page:
screenshot

In Docs, I actually had a harder time parsing the top row, but it looks like:

screenshot

I'm not sure which you were referring to above, but I subjectively find the Drive experience clearer.

zarvox:
Test failed because Jenkins' disk is full again.

kentonv:
@zarvox FWIW, Drive is the one where I have actually struggled to find the "new" button, but in my case it was full of content, which I think makes a pretty big difference compared to your screenshot.

zarvox:
@neynah and I adjusted things for attractiveness and usability and came up with this.

zarvox:
Oh, of course the tests fail - they were looking for the wrong elements after the action buttons moved around.  Fix soon.

zarvox:
Garply, retest this please

kentonv:
Garply, retest this please

paulproteus:
/usr/bin/yes

PR #1059, opened 2015-10-21
Upgrade nightwatch and capture screenshots on test/assertion failure

zarvox:
@jparyani I would love it if we integrated screenshots with Jenkins, so when tests fail, we can see what the page looked like for the failed assertion.  Any thoughts on:

* if upgrading nightwatch is risky or not
* how might we organize/clean out the screenshots folder?
* how might we make Jenkins capture these things as build artifacts?

### DIFF
--- a/tests/nightwatch.json
+++ b/tests/nightwatch.json
@@ -20,6 +20,8 @@
       "silent": true,
       "screenshots" : {
         "enabled" : true,
+        "on_failure": true,
+        "on_error": true,
         "path" : "screenshots"
       },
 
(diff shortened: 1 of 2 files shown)


paulproteus:
re: organize/clean out screenshots folder: I would propose generating it
once per build, and then having Jenkins archive the artifacts, and then
only keeping the last N builds. That way, no need to explicitly clean
things out. (Unless I misunderstand.)

Capturing things as build artifacts is "easy". It hinges on knowing the
file path where they generated.​

I suggest doing this; even if it's risky, let's bite the bullet, so long as
this passes in Jenkins. We can revert if needed. It's not going to get
_less_ risky over time, I think.

jparyani:
LGTM. I'll setup jenkins to handle build artifacts/cleaning up.

zarvox:
:fireworks:

PR #1060, opened 2015-10-21
Only log certificateSigningRequest if there is one

paulproteus:
This avoids a crash where, if the server IP address changed, we would
crash rather than actually update the IP address.

Tested by maurer.


paulproteus:
:heart_decoration:

PR #1061, opened 2015-10-21
Add a failing test to check the build server

jparyani:
Do not merge this

This is just to check that the build server is producing artifacts correctly for failing builds.


jparyani:
This works now (see https://build.sandstorm.io/job/Sandstorm%20PRs/1115/ for an example of artifacts).

paulproteus:
This blew my mind when I clicked on the screenshot link. So much yes.

PR #1070, opened 2015-10-22
Revert "run-bundle: bind /etc to /etc.host RO and symlink interesting…

kentonv:
Reverts sandstorm-io/sandstorm#1030

This breaks some systems, e.g. ones where resolveconf is under /run. We need a more robust strategy here.


paulproteus:
I was thinking about this. It seems we'd do well to have a way to test Sandstorm on various OSs/configurations, not just test the installer. But I could re-use some of that infrastructure probably very effectively.

PR #1071, opened 2015-10-22
Make resumable sessions a bit more robust

jparyani:
This makes it so that incoming requests to a former session (identified by the "sandstorm-sid" cookie) which no longer has a Session object in the DB will block requests for up to 15s waiting for the client to recreate a Session.

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -422,19 +422,17 @@ var getProxyForHostId = function (hostId, isAlreadyOpened) {
         if (!session) {
           if (isAlreadyOpened) {
             return new Promise(function (resolve, reject) {
-              inMeteorInternal(function () {
-                var observer;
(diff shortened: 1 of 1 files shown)


shell/server/proxy.js
```diff
@@ -416,8 +420,27 @@ var getProxyForHostId = function (hostId) {
       return inMeteor(function () {
         var session = Sessions.findOne({hostId: hostId});
         if (!session) {
-          // Does not appear to be a valid session host.
-          return undefined;
+          if (isAlreadyOpened) {
+            return new Promise(function (resolve, reject) {
+              inMeteorInternal(function () {
```
kentonv: Hmm, inMeteorInternal isn't meant to be called directly.

Is it actually needed? I thought that the function passed into Promise's constructor was called synchronously, so would already be in the parent's context.

PR #1075, opened 2015-10-24
Typo in offer template data

mnutt:
If I'm understanding it correctly, it looks like `expires` has been sent as `exires` such that `data.expires < Date.now()` never matches and expired keys never get removed from sessionStorage.


garply:

dwrensha:
Garply, ok to test

kentonv:
Lack of type-safety is sadness.

Luckily if I'm reading right, the worst case here is that some obsolete data sticks around in the memory of a long-running browser window, but as soon as you exit your browser everything should be cleaned up...

PR #1081, opened 2015-10-26
Update the first-user-admin-promotion code to account for the new schema

dwrensha:
The first dev user was erroneously becoming an admin.


paulproteus:
Is the only dangerous (historic) instance of `devName`? Just thought I'd ask for a grep.

dwrensha:
I grepped. The only other uses are in migrations.js and in some inaccessible code in accounts-merge-server.js.

paulproteus:
Yay, thanks!

paulproteus:
Argh wrong button. Meant to click **Comment** not **Close and comment**!

dwrensha:
(The test failed the first time with an error about a grain title. I think that indicates some slight lurking flakiness in the tests.)

PR #1084, opened 2015-10-27
Allow up to 60 seconds (not 20) for /getcertificate response

paulproteus:
This fixes what I believe is the one remaining flakiness of installer-tests.


paulproteus:
Self-merging to kick off the Jenkins build.

PR #1085, opened 2015-10-27
Install explain why not port 80

paulproteus:
@kentonv feel free to review.

I might push another test for this before I merge it, but I have manually verified that the port 80 & 443 message does print. If you have feedback on the message, I'm happy to hear it.


install.sh
```diff
@@ -763,6 +768,23 @@ full_server_install() {
       ACCEPTED_FULL_SERVER_INSTALL=no
     fi
 
+    if [ "yes" = "$SHOW_MESSAGE_ABOUT_NEEDING_PORTS_OPEN" ] ; then
+      echo ""
+      echo "NOTE: Your system has port 80 and/or 443 bound, so we won't automatically"
```
kentonv: Hmm, _probably_ anyone who gets this far knows what ports are, but I think we could make the language a bit less technical without harm. For example, we could say something like::

> It looks like your system already has some other web server installed (port 80 and/or 443 are taken), therefore Sandstorm cannot act as your main web server. If you continue, we'll set up Sandstorm to run on port $DEFAULT_PORT instead, and we **won't** set up HTTPS. This makes sense if you plan to use your other web server as a reverse proxy, or if you're OK with typing the port into your browser whenever you access Sandstorm (and you don't need security). Alternatively, you can quit now, uninstall your other web server, and try again.

kentonv:
ping

paulproteus:
Will revise today & self-merge

paulproteus:
@kentonv ready for re-review. Apologies for delay.

paulproteus:
Moved some tests around. Self-merging to make sure Jenkins likes it.

PR #1091, opened 2015-10-28
App details cleanups

zarvox:
* Make rows in the app-details page clickable anywhere in the row, not just on the link
* Fix dangling link pointing to `/grain/new`
* Redirect `/grain/new` to `/apps`
* Remove a `console.log`


paulproteus:
+1, also I claim this merits a dot-release, since it fixes an important
part of the new user UI.

Before this pull request, if you are a new user, and you have no grains,
and you click **Grains** then you click **Install apps now to start
creating** you end up at a 403 message.

PR #1093, opened 2015-10-29
spk: don't generate files with trailing whitespace

zarvox:
cc @paulproteus 

Addresses sandstorm-io/vagrant-spk#92


paulproteus:
You're too kind.

+1

PR #1094, opened 2015-10-29
Change fallback nounPhrase from "instance" to "grain"

zarvox:
Per @neynah's request.

Also remove a (related) dead function.


shell/packages/sandstorm-ui-applist/applist-client.js
```diff
@@ -91,38 +91,6 @@ var matchApps = function (searchString) {
   return matchingApps;
 };
 
-var nounFromAction = function (action, appTitle) {
```
paulproteus: Wow! We can delete `nounFromAction`!

neynah:
Great thanks! :dancer:

PR #1096, opened 2015-10-29
Don't highlight newly-installed apps in the Most Used row

### LINKED ISSUE #1095: after uploading spk, "Most Used" apps get a truncated second row
looks like this:
cutoff

zarvox:
They're most interesting in the main listing, and we're going to scroll to make sure that one's visible anyway, so calling attention to exactly one thing is preferable.

Fixes #1095.


paulproteus:
Yay!​

Issue #62, opened 2014-07-26
Server-side outgoing DNS fails if resolv.conf wasn't ready when Sandstorm started


vitch:
I'm probably doing something else dumb...

I've accessed my self hosted sandstorm instance over https proxied by nginx. I first set up Github authentication. When I try to log in a github window opens and closes and closes then the message "Internal server error" appears in the signin overlay at the top right of the screen.

Looking at `/opt/sandstorm/var/log/sandstorm.log` I see the following message:

```log
{"line":"230","file":"oauth_server.js","message":"Error in OAuth Server: Failed to complete OAuth handshake with Github. getaddrinfo ENOTFOUND","time":{"$date":1406409017886},"level":"warn"}
Exception while invoking method 'login' Error: Failed to complete OAuth handshake with Github. getaddrinfo ENOTFOUND
    at getAccessToken (packages/github/github_server.js:46)
    at Object.userAgent [as handleOauthRequest] (packages/github/github_server.js:5)
    at OAuth._requestHandlers.(anonymous function) (packages/oauth2/oauth2_server.js:8)
    at middleware (packages/oauth/oauth_server.js:89)
    at packages/oauth/oauth_server.js:62
```

I thought I'd try Google instead...

When I click "Sign in with google" after choosing my account (my browser is signed into two google apps accounts) I see the following:

![screen shot 2014-07-26 at 22 05 15](https://cloud.githubusercontent.com/assets/9898/3712288/a1679030-1509-11e4-9835-99c0aca8e003.png)

Nothing about this shows up in the logs.

Any ideas?

zarvox:
I believe this is fixed by #1097, which is exciting because now I can close the fourth-oldest open issue. :)

Issue #704, opened 2015-08-21
Cannot install libreboard, error while downloading


wascrum:
hello,
I have a virtual machine with ubuntu 14.04 and sandstorm installed. When I want to install the app 'libreboard' the following error occours:

"Failed to install app!
getaddrinfo ENOTFOUND"

Log says following:
"...** Starting Sandstorm at: Thu Aug 20 15:13:38 2015
** Starting back-end...
** Starting MongoDB...
note: noprealloc may hurt performance in many applications
about to fork child process, waiting until server is ready for connections.
forked process: 7
child process started successfully, parent exiting
[object Object]
Successfully added user: {
	"user" : "sandstorm",
	"roles" : [
		"readWriteAnyDatabase",
		"userAdminAnyDatabase",
		"dbAdminAnyDatabase"
	]
}
** Back-end and Mongo started; now starting front-end...
WriteResult({ "nRemoved" : 0 })
Migrations applied: 0/6
Applying migration 1
Applied migration 1
Applying migration 2
Applied migration 2
Applying migration 3
Applied migration 3
Applying migration 4
Applied migration 4
Applying migration 5
Applied migration 5
Applying migration 6
Applied migration 6
Admin token deleted.
Downloading app: http://git.libreboard.com/libreboard/releases/raw/master/0.8.0/libreboard.spk
Failed to install app: Error: getaddrinfo ENOTFOUND
    at errnoException (dns.js:37:11)
    at Object.onanswer [as oncomplete] (dns.js:124:16)
Downloading app: http://git.libreboard.com/libreboard/releases/raw/master/0.8.0/libreboard.spk
Failed to install app: Error: getaddrinfo ENOTFOUND
    at errnoException (dns.js:37:11)
    at Object.onanswer [as oncomplete] (dns.js:124:16)
** Updater died; restarting it
...
** Updater died; restarting it"

Can someone help me with this? Maybe a dns problem? We are behind a firewall.
Is there a way to download the files manually and put it in sandstorm?

paulproteus:
Hi! One thing you can do is manually download the SPK (since you have the
URL) and use "Upload an app" to upload it from your own computer.

You might find the ENOTFOUND error goes away if you restart Sandstorm,
also, since it copies /etc/resolv.conf into its container at the time it
starts.

If you need HTTP proxy support, that's something we're working on. See
https://github.com/sandstorm-io/sandstorm/issues/693 for the status of that
[...]

PR #1097, opened 2015-10-30
Second attempt at making DNS work when `resolv.conf` is updated

### LINKED ISSUE #895: In the case of DNS issues, try restarting Sandstorm

### LINKED ISSUE #1047: cannot log in: getaddrinfo ESRCH
After applying #1030, I am unable to log in with Github or Email authentication. I see the following error in /opt/sandstorm/var/log/sandstorm.log:

```
Exception while invoking method 'login' Error: Failed to complete OAuth handshake with Github. getaddrinfo ESRCH
    at getAccessToken (packages/github/github_server.js:46:1)
    at Object.userAgent [as handleOauthRequest] (packages/github/github_server.js:5:1)
    at OAuth._requestHandlers.(anonymous function) (packages/oauth2/oauth2_server.js:8:1)
    at middleware (packages/oauth/oauth_server.js:173:1)
    at packages/oauth/oauth_server.js:146:1
```

zarvox:
We now also bind in `/run` from the host as `/run.host` to deal with recent Ubuntu releases, where `/etc/resolv.conf` is a symlink to `/run/resolvconf/resolv.conf`.

To keep names coherant, `etc.list` is renamed to `host.list` and now allows paths under both `/etc` and `/run`.

`/run/resolvconf` is added to `host.list`.

Fixes #895.
Fixes #1047.

### DIFF
--- a/make-bundle.sh
+++ b/make-bundle.sh
@@ -41,7 +41,7 @@ copyDep() {
   elif [[ "$FILE" == /etc/* ]]; then
     # We'll want to copy configuration (e.g. for DNS) from the host at runtime.
     if [ -f "$FILE" ]; then
-      echo "$FILE" >> tmp/etc.list
+      echo "$FILE" >> tmp/host.list
     fi
   elif [ -h "$FILE" ]; then
     # Symbolic link.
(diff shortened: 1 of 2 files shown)


zarvox:
Oh, looks like this should also fix #62 from ages ago.  Fancy that.

zarvox:
And also the root cause of the original issue in #704.

kentonv:
So have we decided not to fall back to copying resolve.conf when it links elsewhere? (That would theoretically avoid any possibility of regression from this change.)

zarvox:
It would, but it'd be a bit more complicated, would probably be untested over time (since it's trying to support theoretical systems that none of us use regularly), and would have the same limitations as the current resolv.conf logic, which would still break if resolv.conf doesn't exist yet but would appear later, as in #62.  I'm not convinced the "save a copy of resolv.conf, then if later we think the symlink to the host's copy is dead, replace it with this file" logic is going to be more robust than just figuring out how to link in the host config for reasonable cases.

Since we know there *are* users with the latter problem, but the hypothesized regressions remain theoretical, I felt like this tips the balance in favor of providing a fix for the known problem, rather than committing to never fixing it to avoid possibly regressing on setups that may or may not exist (and would be the result of server-specific modifications.

If you're still strongly of the opinion we should implement the fallback logic, I'll write it.  Not breaking users is pretty important to me too. :)

The approach described was:

* copy `/etc/resolv.conf` from host into `/etc`'s tmpfs before `pivot_root` (say, /etc/resolv.conf.host-initial),
* after `pivot_root`, `stat()` `/etc/resolv.conf` and if ENOENT, replace `/etc/resolv.conf` with the copy made

zarvox:
I believe this is ready for review/merge.  I manually tested the fallback logic and it appears to work as desired.

paulproteus:
I just need to say: fixing this _extremely_ subtle bug is completely amazing.

I don't know of another pull request that closed 4 bugs and caused untold mystifying breakage for self-install users.

PR #1105, opened 2015-10-31
Add missing cursor:pointer to some buttons on the app-details page.

dwrensha:
This makes the "create new grain" and the "signed by" buttons more discoverably clickable.

### DIFF
--- a/shell/client/_app-details.scss
+++ b/shell/client/_app-details.scss
@@ -107,6 +107,7 @@
           }
           >button.show-authorship-button {
             @extend %unstyled-button;
+            cursor: pointer;
             padding-left: 8px;
             padding-right: 24px;
             background-repeat: no-repeat;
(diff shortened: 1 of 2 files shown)


zarvox:
Perhaps this should go on the base `%unstyled-button`!

dwrensha:
I dunno, looks like style to me. :P

zarvox:
Hehehe. :)

In a broader sense, whenever we're trying to do something interactive with one of these buttons (and we wouldn't be using unstyled-button if we didn't want interactivity!), we're going to want to indicate that it's interactive with `cursor: pointer`.

I checked all of our uses of `unstyled-button`, and literally every single one of them gets `cursor:pointer` from somewhere (either that element or a `&:hover`), so I'd say it's *probably* the right thing, to have it on the base class.

dwrensha:
@zarvox: yeah, I think I like #1110 better.

kentonv:
Agreed #1110 is better.

PR #1110, opened 2015-11-02
Add 'cursor: pointer;' to unstyled-button.

dwrensha:
Alternative to #1105.


zarvox:
:+1:

Not sure what's going on with the tests - maybe the test server ran out of disk space again?

PR #1124, opened 2015-11-04
Allow apps to request that the "Share Access" menu be shown via postMessage

zarvox:
@jacksingleton noted at the recent meetup that it's nice to have a button in
the app to start the sharing flow, rather than showing a screenshot with
instructions for users that might become out-of-date.

This adds a postMessage handler the current grain can call that simply shows
the sharing popup.

### DIFF
--- a/shell/shared/grain.js
+++ b/shell/shared/grain.js
@@ -1129,9 +1129,8 @@ if (Meteor.isClient) {
         // Only show this popup if no other popup is currently active.
         // TODO(security): defend against malicious apps spamming this call, blocking all other UI.
         var currentGrain = getActiveGrain(globalGrains.get());
-        if (senderGrain === currentGrain && !globalTopbar._expanded.get()) {
-          globalTopbar._expanded.set("share");
-          globalTopbar._menuExpanded.set(false);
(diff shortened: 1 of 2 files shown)


paulproteus:
Yay, glad to see this getting worked on!

shell/shared/grain.js
```diff
@@ -1124,6 +1124,15 @@ if (Meteor.isClient) {
         check(path.charAt(0), '/');
         // TODO(security): More sanitization of this path. E.g. reject "/../../".
         senderGrain.setPath(path);
+      } else if (event.data.startSharing) {
+        // Allow the current grain to request that the "Share Access" menu be shown.
+        // Only show this popup if no other popup is currently active.
+        // TODO(security): defend against malicious apps spamming this call, blocking all other UI.
+        var currentGrain = getActiveGrain(globalGrains.get());
+        if (senderGrain === currentGrain && !globalTopbar._expanded.get()) {
```
kentonv: A leading underscore indicates a private member. Can you add public methods to SandstormTopbar.prototype to use here? Maybe like `isPopupOpen()` and `openPopup(name)`?

zarvox: Sure, will do.

zarvox:
Changes made.

PR #1132, opened 2015-11-05
Fix clang 3.7 compilation errors.

dwrensha:
This seems like a nicer workaround than the original proposal in #936.

### DIFF
--- a/src/sandstorm/backend.c++
+++ b/src/sandstorm/backend.c++
@@ -243,7 +243,7 @@ kj::Promise<void> BackendImpl::deleteGrain(DeleteGrainContext context) {
   if (iter != supervisors.end()) {
     shutdownPromise = iter->second.promise.addBranch()
         .then([context](Supervisor::Client client) mutable {
-      return client.shutdownRequest().send().then([](auto) {});
+      return client.shutdownRequest().send().then([](auto) -> void {});
     }).then([]() -> kj::Promise<void> {
       return KJ_EXCEPTION(FAILED, "expected shutdown() to throw disconnected exception");
     }, [](kj::Exception&& e) -> kj::Promise<void> {
--- a/src/sandstorm/sandstorm-http-bridge.c++
+++ b/src/sandstorm/sandstorm-http-bridge.c++
@@ -211,7 +211,7 @@ public:
         auto request = responseStream.writeRequest();
         auto dst = request.initData(body.size());
         memcpy(dst.begin(), body.begin(), body.size());
-        taskSet.add(request.send().then([](auto x){}));
+        taskSet.add(request.send().then([](auto x) -> void {}));
         body.resize(0);
       }
 
@@ -458,7 +458,7 @@ private:
         KJ_FAIL_ASSERT("Failed to parse HTTP response from sandboxed app.", error);
       } else if (messageComplete || actual == 0) {
         // The parser is done or the stream has closed.
-        taskSet.add(responseStream.doneRequest().send().then([](auto x){}));
+        taskSet.add(responseStream.doneRequest().send().then([](auto x) -> void {}));
         return kj::READY_NOW;
       } else {
         taskSet.add(pumpStreamInternal(kj::mv(stream)));
@@ -592,7 +592,7 @@ private:
       auto request = responseStream.writeRequest();
       auto dst = request.initData(data.size());
       memcpy(dst.begin(), data.begin(), data.size());
-      taskSet.add(request.send().then([](auto x){}));
+      taskSet.add(request.send().then([](auto x) -> void {}));
     } else {
       body.addAll(data);
     }
@@ -692,7 +692,7 @@ public:
     auto request = clientStream.sendBytesRequest(
         capnp::MessageSize { data.size() / sizeof(capnp::word) + 8, 0 });
     request.setMessage(data);
-    tasks.add(request.send().then([](auto response) {}));
+    tasks.add(request.send().then([](auto response) -> void {}));
   }
 
 protected:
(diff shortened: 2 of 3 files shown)


kentonv:
This is a nicer solution but still sad. Is there any pressing reason to update to 3.7? Can we stick with 3.6 for now?

dwrensha:
I can live with the status quo, but it is somewhat annoying to have to keep rebasing this patch whenever I `git pull` on my Arch Linux box, and I don't really want to maintain an old Clang install there. 

I also don't know enough about the C++ standard to be sure that this actually is a bug on Clang's side (though this documentation seems to support the case that it is indeed a bug on Clang's side: http://en.cppreference.com/w/cpp/language/lambda http://en.cppreference.com/w/cpp/language/function#Return_type_deduction).

zarvox:
So.  I upgraded to Fedora 23 over the weekend.  Fedora 23 ships clang 3.7 by default, and there's not a convenient way to install an older clang.  This means that without these patches, Sandstorm no longer compiles on my laptop.  (Of course I can build in a VM, and all that, but that's less convenient/performant.)

I'd *really* like it if we can merge this patch (or a similar one) that means that I don't have to carry a diff against master.  I'm happy to take on the work of writing similar patches as the backend changes; what I'm less okay with is having to maintain this diff locally for my own productivity, while knowing that the current release of at least two major distributions (Fedora and Arch) will not be able to build Sandstorm from source out of the box.

dwrensha:
For what it's worth, the problem is still present on clang/llvm TRUNK as of a week ago, which was in fact approximately when the 3.8 branch was made.

zarvox:
As a bonus, I also get an ICE when attempting to compile `run-bundle.c++`.  Looking into that now.

zarvox:
ICE seems to be from infinite recursion in name-mangling.  The following 11 frames loop back-to-back for thousands and thousands of frames:

```
#18867 0x0000559ff7227cbd in (anonymous namespace)::CXXNameMangler::mangleNestedName(clang::NamedDecl const*, clang::DeclContext const*, llvm::SmallVector<llvm::StringRef, 4u> const*, bool, bool) ()
No symbol table info available.
#18868 0x0000559ff7228851 in (anonymous namespace)::CXXNameMangler::mangleNameWithAbiTags(clang::NamedDecl const*, llvm::SmallVector<llvm::StringRef, 4u> const*, bool) ()
No symbol table info available.
#18869 0x0000559ff7223688 in (anonymous namespace)::CXXNameMangler::mangleName(clang::NamedDecl const*, bool) ()
No symbol table info available.
#18870 0x0000559ff7220266 in (anonymous namespace)::CXXNameMangler::mangleType(clang::QualType) ()
No symbol table info available.
#18871 0x0000559ff7224281 in (anonymous namespace)::CXXNameMangler::mangleFunctionEncoding(clang::FunctionDecl const*, bool) ()
No symbol table info available.
#18872 0x0000559ff7227e86 in (anonymous namespace)::CXXNameMangler::mangleLocalName(clang::Decl const*, llvm::SmallVector<llvm::StringRef, 4u> const*, bool) ()
No symbol table info available.
#18873 0x0000559ff7228877 in (anonymous namespace)::CXXNameMangler::mangleNameWithAbiTags(clang::NamedDecl const*, llvm::SmallVector<llvm::StringRef, 4u> const*, bool) ()
No symbol table info available.
#18874 0x0000559ff7223688 in (anonymous namespace)::CXXNameMangler::mangleName(clang::NamedDecl const*, bool) ()
No symbol table info available.
#18875 0x0000559ff7220266 in (anonymous namespace)::CXXNameMangler::mangleType(clang::QualType) ()
No symbol table info available.
#18876 0x0000559ff72269cd in (anonymous namespace)::CXXNameMangler::mangleTemplateArg(clang::TemplateArgument) ()
No symbol table info available.
#18877 0x0000559ff722750c in (anonymous namespace)::CXXNameMangler::mangleTemplateArgs(clang::TemplateArgumentList const&) ()
No symbol table info available.
```

zarvox:
Fedora's clang-3.7 appears to segfault on the idParser in run-bundle.c++ : https://github.com/sandstorm-io/sandstorm/blob/master/src/sandstorm/run-bundle.c%2B%2B#L153..L202 

LLVM/Clang built from trunk does not segfault, but does still fail to handle type inference as described in https://llvm.org/bugs/show_bug.cgi?id=24989

My approach here will probably be:

* build clang from source and use that with Sandstorm to get work done
* maybe when I have time figure out what's up with clang-3.7 and if this issue is limited to the Fedora package or if it's an upstream problem, and try to get a patch backported to the appropriate place.

dwrensha:
I wonder whether Fedora is having problems similar to Arch's recent clang-packaging bug: https://bugs.archlinux.org/task/47329

zarvox:
Ahhhhh, I see what's going on now.  tl;dr: GCC made a big ABI change that's likely to break a lot of things.

Arch and Fedora (and FWIW, Ubuntu 15.10, and Debian likely will in due time) have switched to using GCC 5 for the system libraries, including libstdc++.  GCC 5 makes mandatory use of the abi_tags attribute.  Clang does not yet support GCC's mangling of abi_tags.

https://llvm.org/bugs/show_bug.cgi?id=23529#c17 in particular has exactly the stacktrace I observed, and that whole thread is full of a great deal of pretty-unhappy discussion trying to figure out what's going on and how to proceed.

kentonv:
Ouch. My Debian seems to have GCC 5.3 but I guess what you're saying is that libstdc++ and libgcc haven't been updated yet?

zarvox:
That sounds correct, per https://wiki.debian.org/GCC5

FWIW, Fedora 22 also shipped GCC5, but did not do the mass-rebuild.  The mass-rebuild *was* done for Fedora 23, which is why the breakage is appearing now for me.

So my understanding is that this will likely be a problem for anyone using an unpatched clang on any distro that ships a libstdc++ with the new ABI.

PR #1134, opened 2015-11-05
Explain how to re-key an app

paulproteus:


zarvox:
How much harder would it be to simply implement `vagrant-spk rekey` and tell people to use that instead?

paulproteus:

zarvox:
I guess sorta hard because reliably extracting the precise key from `sandstorm-pkgdef.capnp` is nontrivial without a capnproto parser and also removing the pgp signature is hard without an AST.

PR #1136, opened 2015-11-06
Install flow shows app details

### LINKED ISSUE #15: "Install this app" doesn't say which app
I visit https://alpha.sandstorm.io/install/4e1da8cec67250de8fa80371bc21f4eb?url=http://sandstorm.io/apps/ipython-notebook.spk and it says

Install this app?
Do you want to install this app?

[button]Install[/button]

But the only indication of the App's name is in the URL.

### LINKED ISSUE #873: Design for an individual app's page (new UI)

### LINKED ISSUE #1043: Improve install app confirmation page.
I think we should provide user more information about application will be installed rather than just display this simple prompt.

install this app

### LINKED ISSUE #1135: App Details page links should open in new tab
Add <code>target="_blank"</code> to the links at the top so that clicking on them doesn't automatically close all of my open grains.

zarvox:
Adapted the app details components to allow for reuse in the install flow.

If "safe", installing an app proceeds directly to the app page after successful
download and installation.  If not known safe, users are prompted with the details
of the app before confirming installation.

Fixes #15 (as of this writing, the oldest open bug!)
Fixes #1043
Fixes #1135 as part of the refactor
Closes #873

### DIFF
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -565,15 +565,15 @@ limitations under the License.
 </template>
 
 <template name="apps">
-  {{> sandstormAppList }}
+  {{> sandstormAppListPage . }}
 </template>
 
 <template name="grains">
-  {{> sandstormGrainListPage}}
+  {{> sandstormGrainListPage . }}
 </template>
 
 <template name="appDetails">
-  {{> sandstormAppDetailsPage }}
+  {{> sandstormAppDetailsPage . }}
 </template>
 
 <template name="_grainSpinner">
@@ -597,7 +597,7 @@ limitations under the License.
 </template>
 
 <template name="install">
-  {{> sandstormAppInstall }}
+  {{> sandstormAppInstallPage . }}
 </template>
 
 <template name="account">
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -599,69 +599,7 @@ limitations under the License.
 <template name="install">
   {{> sandstormAppInstall }}
 </template>
-<!--
-<template name="install">
-  {{>title "Install app"}}
 
-  {{#if error}}
-    <div id="install" class="error">
-      <div class="centered-box">
-        <p>Failed to install app!
-          {{#if step}}<button id="retry">Try Again</button>{{/if}}</p>
-        <pre>{{error}}</pre>
-      </div>
-    </div>
-  {{else}}
-    <div id="install" class="instep-{{step}}">
-      {{! TODO(cleanup):  The use of a list here is vestigial.  We now display only the step
-            that is currently running. }}
-      <ol>
-        <li id="step-wait">Waiting for write access...</li>
-        <li id="step-download">Downloading... <span class="progress">{{progress}}
-            <button id="cancelDownload">Cancel</button></span></li>
-        <li id="step-verify">Verifying... <span class="progress">{{progress}}</span></li>
-        <li id="step-unpack">Unpacking... <span class="progress">{{progress}}</span></li>
-        <li id="step-analyze">Analyzing... <span class="progress">{{progress}}</span></li>
-        <li id="step-confirm">Install this app?
-          <div class="confirm-form">
-            {{! TODO(someday):  Support editing action names (i.e. make them petnames). }}
-            {{#if hasOlderVersion}}
-              <p>You have an older version of this app installed.  Do you want to upgrade?</p>
-              <p><button id="confirmInstall">Upgrade</button></p>
-            {{else}}
-            {{#if hasNewerVersion}}
-              <p>You already have a newer version of this app installed.  Are
-                you sure you want to revert to this older version?</p>
-              <p><button id="confirmInstall">Downgrade</button></p>
-            {{else}}
-              <p>Do you want to install this app?</p>
-              <p><button id="confirmInstall">Install</button></p>
-            {{/if}}
-            {{/if}}
-          </div></li>
-        <li id="step-run">App Installed
-          <div class="done-notice">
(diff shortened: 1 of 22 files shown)


shell/client/shell.html
```diff
@@ -597,6 +597,10 @@ <h4 class="share-with-others">Share with others</h4>
 </template>
 
 <template name="install">
+  {{> sandstormAppInstall }}
+</template>
+<!--
```
paulproteus: Seems weird to leave this commented-out, fwiw!

shell/packages/sandstorm-ui-app-details/app-details-client.js
```diff
@@ -32,7 +32,7 @@ var latestAppManifestForAppId = function (db, appId) {
 
 var getAppTitle = function (appDetailsHandle) {
   var pkg = latestPackageForAppId(appDetailsHandle._db, appDetailsHandle._appId);
-  return SandstormDb.appNameFromPackage(pkg);
+  return pkg && SandstormDb.appNameFromPackage(pkg) || "<unknown>";
```
paulproteus: The `unknown` case should never happen, right? If that's accurate, then probably we should (in the future, not necessarily now) do something to log the details of this case because if we ever do return `<unknown>` it's because of a bug somewhere.

shell/client/shell.html
zarvox: oh, yeah, whoops.

shell/packages/sandstorm-ui-app-details/app-details-client.js
zarvox: It happens because we are doing subscription joins and the first render might not have an actual `pkg` object yet.

kentonv:
You requested that I write this as a comment: Let's remove the code which bypasses the interstitial when coming from the Sandstorm app market, now that the interstitial shows useful information.

kentonv:
UI: Mostly looks great.

The "downloading" progress view is now left-aligned, and I think it doesn't look very good that way. It really needs a design overhaul at some point, but perhaps for now it makes sense to leave it centered.

shell/client/shell.html
```diff
@@ -597,65 +597,7 @@ <h4 class="share-with-others">Share with others</h4>
...
-            {{else}}
-              <div id="my-files-arrow"></div>
-            {{/if}}
-            {{/if}}
-          </div></li>
-        <li id="step-delete">Server is currently deleting this app...</li>
-      </ol>
-    </div>
-  {{/if}}
+  {{> sandstormAppInstall }}
```
kentonv: I was surprised that no parameter was being passed here, but it appears this is because the data context *is* the SandstormAppInstall object. Maybe comment on that, or write `{{> sandstormAppInstall . }}` to be explicit, or something.

zarvox:
Yay.  Interstitial bypass removal in progress, along with removing the "highlight" logic, since you'll never be sent through the app list that way.  (Hopefully that's not a "how do I get back here?" UX regression?  We should consider user testing this.  I expect that people will find their way back because "Apps" is fairly suggestive text.)

Re: centering: text is generally easier to scan and read when left-aligned, and the other steps will have the heading on the left, so I figured consistency and legibility justified switching to left alignment for all the steps.  Glancing over the rest of our UI, almost nothing else is centered, save the "grain loading" spinner and `/upload`, IIRC?

I could be convinced that we should change the completion fraction to be a progress bar which spans the available width, possibly in a row with the cancel button?  I could maybe be convinced to center the download percentage if there were a slightly stronger justification than "doesn't look very good that way" - I personally always thought the centered text didn't look very good.

I think the header text should remain left-aligned as it is in the rest of the views, though.  Consistency is good.

shell/client/shell.html
zarvox: Ahhh, perhaps I should do the same with the `sandstormAppDetailsPage` and `sandstormGrainListPage`

Actually, I should probably rename this `sandstormAppInstallPage` while I'm here.

zarvox:
Apparently I was mistaken and the "confirm" step is always shown after all, which is great and exactly what we wanted anyway.  Hooray!

kentonv:
I definitely think the download dialog should be improved and the improved version could very well be left-aligned. But as-is in this PR, the feeling I get when seeing it is similar to the feeling I get when seeing raw unstyled HTML. Raw HTML is very readable, but it wouldn't be a good way to do our front page. Something something "user confidence".

> Apparently I was mistaken and the "confirm" step is always shown after all, which is great and exactly what we wanted anyway. Hooray!

If so you should be able to delete the `referredFromSandstorm()` function. :)

zarvox:
All right, I think I've addressed all of your comments, and the tests pass locally on my machine!

I've made the download page include a `<progress>` bar if we know the download size, and centered text otherwise.  Hopefully that makes it look less raw-HTML-y in the interim, until we can poke @neynah to sketch out something fancier.

I've also made the download update progress twice a second, which hopefully makes the download progress look a bit smoother without increasing server load too much.

kentonv:
Bug: If I install an unsigned SPK, I see "The app publisher has proven using PGP that they control these accounts." but there are no accounts listed. Probably the message should say "This app has no signature; its author cannot be verified." or something like that.

zarvox:
Fixed (along with some style improvements).

ocdtrekkie:
One merge button, five email notifications. :D

PR #1138, opened 2015-11-06
Update handbook to explain sharing links

paulproteus:
This also means that we link to http://www.w3.org/TR/capability-urls/
in a way that is consistent with current Sandstorm.


paulproteus:
Self-merging since 100% docs

paulproteus:
https://docs.sandstorm.io/en/latest/developing/handbook/ is refreshed

PR #1102, opened 2015-10-30
Replace svg icons with an autogenerated webfont

jparyani:
I've replaced the notification bell icon as an example of how this should be used.

There's some small stylistic things that I'm open to changing.

1. The icon classes are prefixed by default with `icon_` (eg. `icon_notification`). By convention, I think we have been using `-` instead of `_` in CSS.
2. I'm using a mixin for `webfont-icon`. This is a little inefficient, but I'm not sure how to cleanly do this with an extend instead.

PR #1140, opened 2015-11-07
Use Meteor's npm and node for webfont generation.

dwrensha:
#1102 made the build depend on installed versions of npm and node. This patch eliminates those new dependencies, using the bundled Meteor versions of those binaries.


kentonv:
good call

PR #1141, opened 2015-11-07
Whitelist the owncloud mobile client apps for HTTP Basic auth.

zarvox:
This will result in us sending a 401 to the apps when they try to connect
without providing credentials.

@mnutt do you think this change will allow the owncloud mobile apps to work with Davros on Sandstorm, or will additional changes be needed?

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -45,6 +45,8 @@ BASIC_AUTH_USER_AGENTS = [
   "GitHub-Hookshot\\/",
   "mirall\\/",
   "Mozilla\\/5\\.0 \\([^\\\\]*\\) mirall\\/",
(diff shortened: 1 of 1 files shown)


mnutt:
This would be great to do and will ensure that mobile apps work when they are fixed, however the mobile apps have a bug that _also_ needs to be fixed for everything to work.

The bug happens here: https://github.com/owncloud/ios/blob/d707f618dfd51926b9a68cf05225ff6176bb0d82/Owncloud%20iOs%20Client/Login/Login/LoginViewController.m#L1968

It won't let you continue if the server returns a 200 response, which it will in this case because davros has to put the username and password in the URL. With the username and password removed from the URL, a different check fails.

mnutt:
I've figured out why the owncloud mobile clients aren't currently working; with this change it should be possible to open a PR with them and fix it. It's just an invasive change on their end, so it'll take a while.

PR #1149, opened 2015-11-10
Workaround bug in simplesmtp. It fails to escape "\n." properly

jparyani:


dwrensha:
I think this replaces the first character of every line with ".", because within a regular expression "." means "match a single character".

jparyani:
Good catch, thanks. In testing, I was somehow looking at previous emails and didn't notice the first character getting replaced with a '.'...

dwrensha:
This appears to work correctly for me!

The comment is a bit ambiguous. It's not immediately clear that "it" refers to SimpleSmtp rather than your workaround. Maybe something like this would be more clear: "SimpleSmtp does not add leading dots, so we need to. See http://tools.ietf.org/html/rfc5321#section-4.5.2"

The test failure is not surprising: RoundCube is inserting an extra line with a dot, and now we are faithfully forwarding it!

jparyani:
Comment changed and test should be fixed. I'll revert the test back when I've updated roundcube/sandstorm-smtp-bridge

dwrensha:
Hm... it's a bit more subtle than I had originally thought. Here's maybe a better comment:

'Lines starting with a dot in the raw text of the message should have already been prepended with an additional dot in `_message.body`, as per http://tools.ietf.org/html/rfc5321#section-4.5.2 . SimpleSmtp does not validate against this requirement, so for robustness we additionally add a dot in cases where we know that the caller failed to do so. In particular, this prevents a caller from triggering a premature end-of-message by including the string "< CLRF >.< CLRF >". '

dwrensha:
One question is, does [`EmailMessage.text`](https://github.com/sandstorm-io/sandstorm/blob/master/src/sandstorm/email.capnp) already have dot-escaping applied? The data that I'm getting from RoundCube seems to indicate that the current answer is "yes", and if so then this pull request is the appropriate thing to do, I think. However, perhaps the answer should  be "no", because we should be striving to make the Cap'n Proto interfaces as simple as possible.

kentonv:
The text in the Cap'n Proto protocol should _not_ be expected to be escaped -- this is the kind of protocol detail that we want to hide, to make life easier for apps.

Currently this change only adds a `.` on lines that don't already start with two dots. That's not correct -- it should add a `.` to every line that starts with a `.`, regardless of whether another `.` follows. Otherwise if I write an email in which I intentionally start a line with two `.`s, one of them will get eaten by the SMTP receiver, which will interpret it as an escape.

PR #1154, opened 2015-11-12
update comment about dot-escaping

dwrensha:


kentonv:
Thanks for catching that.

PR #1155, opened 2015-11-12
spk: enable --details by default for `verify`

zarvox:
Leave the switch in, to preserve compatibility with callers, but it has no noticeable effect.


kentonv:
In theory this breaks callers who call `spk verify` directly for the purpose of getting just the app ID. Probably no such caller exists? OK I guess I'll take it.

paulproteus:
: D​

PR #1160, opened 2015-11-13
Add note to run `vagrant-spk dev` before packaging

simonv3:
This was a bit of a stumbling block for me this morning, which zarvox (on IRC, not sure what their GitHub handle is) helped me fix through troubleshooting that I hadn't actually built the app before packaging it. This note might prevent future people from having the same problem.

Edit: I also stole the note syntax from [the readthedocs documentation](https://read-the-docs.readthedocs.org/en/latest/versions.html). Hopefully it works!

### DIFF
--- a/docs/vagrant-spk/packaging-tutorial-meteor.md
+++ b/docs/vagrant-spk/packaging-tutorial-meteor.md
@@ -361,6 +361,10 @@ vagrant-spk pack ~/projects/package.spk
 This will take a few moments, and once it is done, there will be a
 file in `~/projects/package.spk` that contains the full app.
 
+Note
+
+Packaging your data requires running the `vagrant-spk dev` command first. `vagrant-spk pack` only copies the output from the dev command into the package. This might be an oversight when applying minor changes without testing on your local sandstorm instance.
+
 You can see how large it is by running the following command:
 
 ```bash


garply:

zarvox:
Garply, ok to test

zarvox:
Hi Simon!  I'm zarvox here too. :)

Thanks for filing this!  @paulproteus owns the docs, so I'll let him be in charge of this, but +1 from me on noting that for the meteor stack, you need to run `dev` before `pack` if you've made source changes.

zenhack:
Looks like this was just waiting on a second look from @paulproteus that never happened. It seems reasonable to me. @ocdtrekkie want to weigh in?

ocdtrekkie:
I am not sure how I feel about this. (Which may be why I never dug into this PR.)

- Main gripe: It's sole purpose is to help when changes are made "without testing", which generally, we should consider bad?
- Secondary gripe: I don't feel that "vagrant-spk pack only copies the output from the dev command" is accurate enough. It's definitely true you need setup.sh to run at some juncture, but I am not sure the language here sounds right.
- It's "note" syntax does not match how another note on the same page is marked.
- There's a lower-case Sandstorm.

ocdtrekkie:
There's another "interesting" point to make about this tutorial, somewhat unrelated to the PR: This tutorial is about the Meteor stack of vagrant-spk. I honestly need to check if it still works, every Meteor app I know of on Sandstorm just uses meteor-spk... I think there was a push for a while to get everyone using vagrant-spk, and then that kinda rolled backwards a bit?

zenhack:
Ideally the meteor stack would just *use* meteor-spk, which would obviate the issue since meteor-spk understands the dependencies more deeply than spk or vagrant-spk can. I think I'd rather do that than add docs about workarounds with the current solution that afaik nobody uses. But I suspect that may require changes to vagrant-spk itself.

I agree with your other gripes.

ocdtrekkie:
The Meteor stack *does* use meteor-spk. But I am not sure where the relationship between the two currently sits, tbh.

zenhack:
From looking at the stack, it looks like it downloads meteor-spk and fishes some stuff out of its tree, but never actually runs meteor-spk at any point. (It also looks like the version is slightly out of date)

PR #1170, opened 2015-11-17
tying up some acounts-identity loose ends

dwrensha:


dwrensha:
Garply, retest this please

dwrensha:
Garply, retest this please

dwrensha:
Garply, retest this please

jparyani:
Garply, retest this please

dwrensha:
Garply, retest this please

dwrensha:
I've tacked on a bunch more changes to this pull request:

  1. More styling.
  2. When you confirm an email identity, the link in the sent email now does the right thing.
  3. Logging out other sessions for the current user, or all users for a given service.

shell/packages/accounts-email-token/token_server.js
```diff
@@ -83,25 +89,42 @@ var ROOT_URL = Url.parse(process.env.ROOT_URL);
 var HOSTNAME = ROOT_URL.hostname;
 
 var makeTokenUrl = function (email, token) {
-  return process.env.ROOT_URL + "/_emailToken/" + encodeURIComponent(email) + "/" + encodeURIComponent(token);
+  return process.env.ROOT_URL + "/_emailLogin/" + encodeURIComponent(email) + "/" + encodeURIComponent(token);
+};
+
+var makeTokenUrl = function (email, token, linkingIdentity) {
```
kentonv: This looks like it's redefining the var defined four lines earlier. Did you mean to delete the other one?

dwrensha: Why, yes I did.

shell/packages/accounts-email-token/token_server.js
```diff
@@ -146,22 +170,47 @@ var createAndEmailTokenForUser = function (email) {
...
+    }
+    // Create user. result contains id and token.
+    var user = createAndEmailTokenForUser(email, linkingIdentity);
+  },
+
+  linkEmailIdentityToAccount: function (email, token, accountId) {
+    // Links the email identity with address `email` and login token `token` to the account with
+    // ID `accountId`, which must also be the current account.
+    check(email, String);
+    check(token, String);
```
kentonv: Please add a check() for accountId -- even though there's obviously no injection possibility since it is only used in a `!==` comparison, I'd like to as a rule check() everything so that we could, for example, have an automated script verify our checks.

Alternatively, maybe it makes sense to drop accountId as a parameter, since it's not valid to specify anything other than the caller's user ID anyway?

shell/packages/accounts-email-token/token_server.js
dwrensha: Done.

shell/packages/accounts-email-token/token_server.js
dwrensha: Done.

shell/packages/accounts-email-token/token_common.js
```diff
@@ -44,3 +44,21 @@ Router.route("/_emailToken/:_email/:_token", function () {
     }
   });
 });
+
+Router.route("/_emailLinkIdentity/:_email/:_token/:_accountId", function () {
+  this.render("Loading");
+  var self = this;
+  if (Meteor.userId() === this.params._accountId) {
+    Meteor.call("linkEmailIdentityToAccount",
+                this.params._email, this.params._token, this.params._accountId,
```
kentonv: You're still passing accountId here, which is no longer a parameter.

dwrensha: Good catch. Fixed now. Apparently Meteor didn't mind the extra parameter and was still calling the callback upon completion.

kentonv: Yeah it's standard behavior for Javascript to simply ignore extra parameters. So technically the code wasn't even wrong -- until we someday add a new parameter with a different meaning.

PR #1171, opened 2015-11-17
Protect against missing Overwrite header in WebDAV MOVE or COPY

mnutt:
The "overwrite" header seems to be somewhat optional, defaulting to "t", so we should make sure the proxy doesn't break if it doesn't get sent on a `MOVE` or `COPY` action.


garply:

PR #1179, opened 2015-11-19
Add initial version of contact autocomplete

jparyani:
This changes the sharing dialog box to search contacts, as well as allow
emails to be entered.

Includes a migration to populate the Contacts table from ApiTokens.

### DIFF
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -196,8 +196,6 @@ Contacts = new Mongo.Collection("contacts");
 //            the contact to the owner.
 //   created: Date when this contact was created.
 //   identityId: The `_id` of the user whose contact info this contains.
-//   profile: An object containing denormalized profile data for this contact. See the `Users`
-//             Collection for a list of all the fields
 
 Sessions = new Mongo.Collection("sessions");
 // UI sessions open to particular grains.  A new session is created each time a user opens a grain.
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -191,11 +191,13 @@ Contacts = new Mongo.Collection("contacts");
 //
 // Each contains:
 //   _id: random
-//   ownerId: The `_id` of the user who owns this contact.
-//   identityId:  The identity of the contacted user.
+//   ownerId: The accountId of the user account who owns this contact.
 //   petname: Human-readable label chosen by and only visible to the owner. Uniquely identifies
 //            the contact to the owner.
 //   created: Date when this contact was created.
+//   identityId: The `_id` of the user whose contact info this contains.
+//   profile: An object containing denormalized profile data for this contact. See the `Users`
+//             Collection for a list of all the fields
 
 Sessions = new Mongo.Collection("sessions");
 // UI sessions open to particular grains.  A new session is created each time a user opens a grain.
--- a/shell/packages/sandstorm-db/migrations.js
+++ b/shell/packages/sandstorm-db/migrations.js
@@ -359,7 +359,7 @@ function splitAccountUsersAndIdentityUsers() {
 }
 
 function populateContactsFromApiTokens() {
-  ApiTokens.find({"owner.user": {$exists: 1}}).forEach(function(token) {
+  ApiTokens.find({"owner.user": {$exists: 1}, accountId: {$exists: 1}}).forEach(function(token) {
     var identityId = token.owner.user.identityId;
     var profile = SandstormDb.prototype.getIdentity(identityId).profile;
     Contacts.upsert({ownerId: token.accountId, identityId: identityId}, {
--- a/shell/packages/sandstorm-db/migrations.js
+++ b/shell/packages/sandstorm-db/migrations.js
@@ -359,7 +359,8 @@ function splitAccountUsersAndIdentityUsers() {
 }
 
 function populateContactsFromApiTokens() {
-  ApiTokens.find({"owner.user": {$exists: 1}, accountId: {$exists: 1}}).forEach(function(token) {
+  ApiTokens.find({"owner.user.identityId": {$exists: 1},
+                  accountId: {$exists: 1}}).forEach(function(token) {
     var identityId = token.owner.user.identityId;
     var profile = SandstormDb.prototype.getIdentity(identityId).profile;
     Contacts.upsert({ownerId: token.accountId, identityId: identityId}, {
--- a/shell/packages/sandstorm-db/migrations.js
+++ b/shell/packages/sandstorm-db/migrations.js
@@ -358,6 +358,20 @@ function splitAccountUsersAndIdentityUsers() {
   });
 }
 
+function populateContactsFromApiTokens() {
+  ApiTokens.find({"owner.user": {$exists: 1}}).forEach(function(token) {
+    var identityId = token.owner.user.identityId;
+    var profile = SandstormDb.prototype.getIdentity(identityId).profile;
+    Contacts.upsert({ownerId: token.accountId, identityId: identityId}, {
+      ownerId: token.accountId,
+      petname: profile && profile.name,
+      created: new Date(),
+      identityId: identityId,
+      profile: profile,
+    });
+  });
+}
+
 // This must come after all the functions named within are defined.
 // Only append to this list!  Do not modify or remove list entries;
 // doing so is likely change the meaning and semantics of user databases.
@@ -377,6 +391,7 @@ var MIGRATIONS = [
   moveDevAndEmailLoginDataIntoIdentities,
   repairEmailIdentityIds,
   splitAccountUsersAndIdentityUsers,
+  populateContactsFromApiTokens,
 ];
 
 function migrateToLatest() {
--- a/shell/packages/sandstorm-db/profile.js
+++ b/shell/packages/sandstorm-db/profile.js
@@ -284,7 +284,7 @@ SandstormDb.getUserEmails = function (user) {
   return result;
 }
 
-SandstormDb.prototype.addContact = function (ownerId, identityId) {
+SandstormDb.prototype.addContact = function addContact (ownerId, identityId) {
   var db = this;
   var profile = db.getIdentity(identityId).profile;
   db.collections.contacts.upsert({ownerId: ownerId, identityId: identityId}, {
@@ -292,6 +292,5 @@ SandstormDb.prototype.addContact = function (ownerId, identityId) {
     petname: profile && profile.name,
     created: new Date(),
     identityId: identityId,
-    profile: profile,
   });
 }
--- a/shell/packages/sandstorm-db/profile.js
+++ b/shell/packages/sandstorm-db/profile.js
@@ -285,6 +285,8 @@ SandstormDb.getUserEmails = function (user) {
 }
 
 SandstormDb.prototype.addContact = function addContact (ownerId, identityId) {
+  check(ownerId, String);
+  check(identityId, String);
   var db = this;
   var profile = db.getIdentity(identityId).profile;
   db.collections.contacts.upsert({ownerId: ownerId, identityId: identityId}, {
--- a/shell/packages/sandstorm-db/profile.js
+++ b/shell/packages/sandstorm-db/profile.js
@@ -92,26 +92,6 @@ if (Meteor.isServer) {
 
   var Url = Npm.require("url");
   httpProtocol = Url.parse(process.env.ROOT_URL).protocol;
-
-  var Crypto = Npm.require("crypto");
-  Meteor.methods({
-    revealIdentity: function (token, identityId) {
-      var db = this.connection.sandstormDb;
-      if (!db.userHasIdentity(this.userId, identityId)) {
-        throw new Meteor.Error(403, "User doesn't own this indentity.");
-      }
-      var hashedToken = Crypto.createHash("sha256").update(token).digest("base64");
-      var ownerId = db.collections.apiTokens.findOne({_id: hashedToken}).accountId;
-      var profile = db.getIdentity(identityId).profile;
-      db.collections.contacts.upsert({ownerId: ownerId, identityId: identityId}, {
-        ownerId: ownerId,
-        petname: profile && profile.name,
-        created: new Date(),
-        identityId: identityId,
-        profile: profile,
-      });
-    }
-  });
 } else {
   var identiconCache = {};
 
@@ -302,4 +282,16 @@ SandstormDb.getUserEmails = function (user) {
     result[0].primary = true;
   }
   return result;
+}
+
+SandstormDb.prototype.addContact = function (ownerId, identityId) {
+  var db = this;
+  var profile = db.getIdentity(identityId).profile;
+  db.collections.contacts.upsert({ownerId: ownerId, identityId: identityId}, {
+    ownerId: ownerId,
+    petname: profile && profile.name,
+    created: new Date(),
+    identityId: identityId,
+    profile: profile,
+  });
 }
--- a/shell/packages/sandstorm-db/profile.js
+++ b/shell/packages/sandstorm-db/profile.js
@@ -92,6 +92,26 @@ if (Meteor.isServer) {
 
   var Url = Npm.require("url");
   httpProtocol = Url.parse(process.env.ROOT_URL).protocol;
+
+  var Crypto = Npm.require("crypto");
+  Meteor.methods({
+    revealIdentity: function (token, identityId) {
+      var db = this.connection.sandstormDb;
+      if (!db.userHasIdentity(this.userId, identityId)) {
+        throw new Meteor.Error(403, "User doesn't own this indentity.");
+      }
+      var hashedToken = Crypto.createHash("sha256").update(token).digest("base64");
+      var ownerId = db.collections.apiTokens.findOne({_id: hashedToken}).accountId;
+      var profile = db.getIdentity(identityId).profile;
+      db.collections.contacts.upsert({ownerId: ownerId, identityId: identityId}, {
+        ownerId: ownerId,
+        petname: profile && profile.name,
+        created: new Date(),
+        identityId: identityId,
+        profile: profile,
+      });
+    }
+  });
 } else {
   var identiconCache = {};
 
--- a/shell/packages/sandstorm-ui-autocomplete-input/autocomplete-client.js
+++ b/shell/packages/sandstorm-ui-autocomplete-input/autocomplete-client.js
@@ -150,9 +150,21 @@ Template.contactInputBox.events({
     template.currentText.set(event.target.value);
   },
   "keydown .completed-contact": function (event, template) {
-    if (event.keyCode === 8) { // Backspace
+    if (event.keyCode === 8 || event.keyCode == 46) { // Backspace or Delete
       deleteSelected(this, template);
       return false;
+    } else if (event.keyCode === 37) { // Left
+      var previous = event.target.previousElementSibling;
+      if (previous) {
+        previous.focus();
+      }
+    } else if (event.keyCode === 39) { // Right
+      var next = event.target.nextElementSibling;
+      if (next) {
+        next.focus();
+      } else {
+        template.find("input").focus();
+      }
     }
   },
   "click .closer": function (event, template) {
@@ -205,6 +217,12 @@ Template.contactInputBox.events({
         newContact = contacts[contacts.length - 1];
       }
       template.highlightedContact.set(newContact);
+      return false;
+    } else if (event.keyCode === 37) { // Left
+      var chip = template.find(".completed-contacts>li:last-child");
+      if (chip) {
+        chip.focus();
+      }
       return false;
     } else if (event.keyCode === 13) { // Enter
       var highlightedContact = template.highlightedContact.get();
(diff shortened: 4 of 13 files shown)


shell/client/grainview.js
```diff
@@ -319,6 +319,9 @@ GrainView.prototype.revealIdentity = function (identityId) {
       identity = _.findWhere(myIdentities, {_id: token.owner.user.identityId});
     }
   }
+  if (this._token) {
+    Meteor.call("revealIdentity", this._token, identity._id);
```
dwrensha: As I understand this, the user who receives a share call this method, which causes the sharer to gain them as a contact? This seems wrong, as the sharer must trust the receiving user's client to actually call this method. I think it would make more sense instead to add some logic to `SandstormPermissions.createNewApiToken()`, populating a contact whenever the owner is `user`.

shell/packages/sandstorm-db/profile.js
```diff
@@ -70,6 +70,26 @@ if (Meteor.isServer) {
 
   var Url = Npm.require("url");
   httpProtocol = Url.parse(process.env.ROOT_URL).protocol;
+
+  var Crypto = Npm.require("crypto");
+  Meteor.methods({
+    revealIdentity: function (token, identityId) {
```
dwrensha: As explained in my comment in grainview.js, I don't think this should be a Meteor method.

shell/packages/sandstorm-db/profile.js
```diff
@@ -70,6 +70,26 @@ if (Meteor.isServer) {
...
+      }
+      var hashedToken = Crypto.createHash("sha256").update(token).digest("base64");
+      var ownerId = db.collections.apiTokens.findOne({_id: hashedToken}).accountId;
+      var profile = db.getIdentity(identityId).profile;
+      db.collections.contacts.upsert({ownerId: ownerId, identityId: identityId}, {
+        ownerId: ownerId,
+        petname: profile && profile.name,
+        created: new Date(),
+        identityId: identityId,
+        profile: profile,
```
dwrensha: This will include `profile.pictureUrl`. You'll need to increment the refcount in `StaticAssets` if you want to avoid a potential 404 when the contact changes their picture.  (The refcounting is actually currently broken, but #1181 fixes that.)

shell/packages/sandstorm-db/db.js
```diff
@@ -185,11 +185,13 @@ Contacts = new Mongo.Collection("contacts");
...
 // Each contains:
 //   _id: random
-//   ownerId: The `_id` of the user who owns this contact.
-//   identityId:  The identity of the contacted user.
+//   ownerId: The accountId of the user account who owns this contact.
 //   petname: Human-readable label chosen by and only visible to the owner. Uniquely identifies
 //            the contact to the owner.
 //   created: Date when this contact was created.
+//   identityId: The `_id` of the user whose contact info this contains.
+//   profile: An object containing denormalized profile data for this contact. See the `Users`
```
dwrensha: I'm tempted to try to figure out how to avoid adding this `profile` field. I remember you were thinking about implementing a join at some point. Did you run into any problems with that approach?

dwrensha:
Some thoughts about `Contacts.profile`:

A raw identity is an entry in the Users table that has a `profile` field. A raw identity may contain private information, such as an email address or hashed login tokens.

Before we display an identity to a user, we typically do some amount of processing on the raw identity. This processing includes hiding fields that the user shouldn't see and computing values for fields that aren't present in the database.

Here some things that need to happen, in no particular order:

  1. The data needs to make it from the server to the client. This is typically done through a subscription, e.g. "accountIdentities".
  2. We need to call `SandstormDb.fillInPictureUrl(user)` (currently misnamed `fillInIdenticon(user)`). This adds a `profile.pictureUrl` field using the static asset ID in `profile.picture` if one is present, or an identicon otherwise. The identicon is only generated if this function is called on the client.
  3. We need to call `SandstormDb.fillInProfileDefaults(user)` (currently named `fillInDefaults(user)`). This fills in any missing values for `profile.name`, `profile.handle`, and `profile.pronoun` from default values computed from `user.services`. This is commonly a no-op, because the first thing a user must do is confirm their profile, writing values for these fields.
  4. We need to call `SandstormDb.fillInIntrinsicName(user)` (currently rolled into `fillInDefaults()`). This populates `user.profile.intrinsicName` and possibly `user.privateIntrinsicName` based on `user.services`. (For Google login, the email address is the private intrinsic name and the the Google display name is the public intrinsic name.)


Our problem right now is: how does a user receive identity information for their contacts?

The solution in this pull request is to maintain a `contact.profile` field that caches fully-processed identity data. Open questions for this approach: 

  1. When does the cache get updated?
  2. How do we make sure that we never display a delete StaticAsset?
  3. What if the pictureUrls become invalid due to the server getting re-configured?


An alternate approach might be to write a "contactIdentities" subscription that performs a join on the Contacts and Users tables, thus avoiding the need to denormalize. We could even make it a fully reactive subscription. Some open questions for this approach:

  1. At exactly which stage of processing are the identities that get sent on the subscription?
  2. How expensive is it to perform such a join?

dwrensha:
I also just want to say that I'm really excited that this pull request has a working implementation of share-by-identity. :)

kentonv:
I'm going to delegate this review to David. :)

jparyani:
I've gotten rid of the "revealIdentity" meteor method and stopped caching the identity.profile in Contacts. I've also rebased on top of https://github.com/sandstorm-io/sandstorm/pull/1213, so this shouldn't be merged until that has.

shell/packages/sandstorm-db/migrations.js
```diff
@@ -358,6 +358,20 @@ function splitAccountUsersAndIdentityUsers() {
   });
 }
 
+function populateContactsFromApiTokens() {
+  ApiTokens.find({"owner.user": {$exists: 1}}).forEach(function(token) {
+    var identityId = token.owner.user.identityId;
+    var profile = SandstormDb.prototype.getIdentity(identityId).profile;
+    Contacts.upsert({ownerId: token.accountId, identityId: identityId}, {
```
dwrensha: Due to a bug that we recently fixed, `token.accountId` might not exist on some tokens.
See [this comment](https://github.com/sandstorm-io/sandstorm/blob/90d86d7d2cc8b7865fb84d17076b10f010ad5069/shell/packages/sandstorm-permissions/permissions.js#L449-L452).
You should either include `{accountId: {$exists: 1}}` in the `ApiTokens` query, or you should repair these tokens.

dwrensha: Note, though, that such tokens are probably very rare. To make one, I think you'd need to have obtained a token from an anonymous offer template and then redeemed it as a /shared/ link as a logged-in user. So I recommend just ignoring them for now.

shell/packages/sandstorm-ui-autocomplete-input/autocomplete-client.js
```diff
@@ -0,0 +1,210 @@
+window.UserContacts = new Mongo.Collection("userContacts");
```
dwrensha: Why is this being added as a field of `window`? If you expect to need it outside of this package, it would be better to expose it via `api.export()`. But it looks like you aren't actually using it outside this package, so maybe it can just be `var UserContacts = ...`?

shell/packages/sandstorm-ui-autocomplete-input/autocomplete-client.js
```diff
@@ -0,0 +1,210 @@
...
+
+Template.contactInputBox.onCreated(function () {
+  var self = this;
+  this.currentText = new ReactiveVar(null);
+  this.inputActive = new ReactiveVar(false);
+  this.selectedContacts = this.data.contacts;
+  this.selectedContactsIds = new ReactiveVar([]);
+  this.highlightedContact = new ReactiveVar({_id: null});
+  this.subscribe("userContacts");
```
dwrensha: Oh, interesting. I had assumed that the Contacts subscription would be a one-time global subscription. I suppose one advantage to your approach here is that the non-reactivity of the subscription matters less; if one your contacts uploads a different profile picture, you'll see it next time you open a new contactInputBox.

shell/packages/sandstorm-ui-autocomplete-input/autocomplete-server.js
```diff
@@ -0,0 +1,16 @@
+Meteor.publish("userContacts", function () {
+  // TODO(someday): make this reactive
```
dwrensha: I think this has more than "someday" urgency. As this is written, it looks like a user could still see a 404 on profile pictures that are being concurrently changed by their owner.

The fully reactive join should be similar to, but simpler than, what I implemented [here](https://github.com/sandstorm-io/sandstorm/blob/cc46173039ab6f5e5616ce3b606df7809081a38e/shell/packages/accounts-identity/accounts-identity-server.js#L225-L283).

shell/packages/sandstorm-ui-autocomplete-input/autocomplete-client.js
dwrensha: You should document what fields are present in this pseudocollection.

shell/shared/grain.js
```diff
@@ -254,14 +254,14 @@ Meteor.methods({
...
-                                emailAddresses, message) {
+                                contacts, message) {
     if (!this.isSimulation) {
       check(origin, String);
       check(identityId, String);
       check(grainId, String);
       check(title, String);
       check(roleAssignment, roleAssignmentPattern);
-      check(emailAddresses, [String]);
+      check(contacts, [Object]);
```
dwrensha: Can you require a more specific type than `[Object]` here?

shell/packages/sandstorm-ui-autocomplete-input/autocomplete-client.js
```diff
@@ -0,0 +1,210 @@
...
+
+  var selectedContactsIds = template.selectedContactsIds.get();
+  template.selectedContactsIds.set(_.filter(selectedContactsIds, function (id) {
+    return id !== self._id;
+  }));
+  template.find("input").focus();
+}
+
+Template.contactInputBox.events({
+  // "click .contact-box": function (event) {
```
dwrensha: What is the status of this commented-out code? At the very least, it should have a comment about why it is here.

shell/shared/grain.js
```diff
@@ -273,32 +273,40 @@ Meteor.methods({
...
-            text: message.text + "\n\nFollow this link to open the shared grain:\n\n" + url +
-              "\n\nNote: If you forward this email to other people, they will be able to access " +
-              "the share as well. To prevent this, remove the link before forwarding.",
-            html: html,
-          });
-          outerResult.successes.push(emailAddress);
-        } catch (e) {
-          outerResult.failures.push({email: emailAddress, error: e.toString()});
+      contacts.forEach(function(contact) {
+        if (contact.isDefault && contact.profile.service === "email") {
```
dwrensha: I guess this means we can't yet share-by-identity to an email identity? That seems fine for now, but I think we'll eventually want to make a distinction here.

jparyani: My concern with email share by identity is that there may be some user confusion. We'd have to be very clear that sometimes we don't email and sometimes we do and why, which I'm not sure how to do easily. For now, it's at least semi-obvious that if you type an email address, you will be sending an email invite.

shell/packages/sandstorm-db/profile.js
```diff
@@ -281,3 +281,14 @@ SandstormDb.getUserEmails = function (user) {
   }
   return result;
 }
+
+SandstormDb.prototype.addContact = function addContact (ownerId, identityId) {
```
dwrensha: Please check that `ownerId` and `identityId` are strings.

jparyani:
All comments addressed. Waiting on the addition of `SandstormDb.getVerifiedEmails` in order to send out a notification.

dwrensha:
If you need a verified email you should call `_.findWhere(SandstormDb.getUserEmails(accountId), {verified: true})`. Re our discussion about github verified emails, I'm going to change some internal implementation details that you don't need to worry about.

jparyani:
It's a little more complicated than that since we only have an identity. After talking with Kenton, we're going to ignore accounts and send an email if the user has a verified email for that identity. If there's no verified email associated with said identity (ie. in rare cases with github), we'll still create the ApiToken for the share, and send back a warning to the sharer telling them to manually share the grain link.

dwrensha:
Ah, good point. So I'll need to expose my `getVerifiedEmails` function.

jparyani:
This is now dependent on #1235 being merged first.

Thanks for adding it so fast @dwrensha!

shell/packages/sandstorm-ui-autocomplete-input/autocomplete-client.js
```diff
@@ -0,0 +1,220 @@
+var UserContacts = new Mongo.Collection("userContacts");
+// A psuedo-collection used to store the results of joining Contacts with identity profiles
+//
+// Each contains:
+//   _id: the id of identity (from Meteor.users collection)
+//   profile: the profile of the identity (see db.js for fields in this object)
```
dwrensha: ... with profile defaults and the intrinsic name already filled in.

shell/packages/sandstorm-ui-autocomplete-input/autocomplete-server.js
dwrensha: This TODO has been addressed!

shell/packages/sandstorm-ui-autocomplete-input/autocomplete-server.js
```diff
@@ -0,0 +1,51 @@
...
+  var db = this.connection.sandstormDb;
+  var self = this;
+
+  // We maintain a map from identity IDs to live query handles that track profile changes.
+  var loginIdentities = {};
```
dwrensha: A less confusing name might be `contactIdentities`.

shell/shared/grain.js
```diff
@@ -273,32 +280,71 @@ Meteor.methods({
...
+              text: message.text + "\n\nFollow this link to open the shared grain:\n\n" + url +
+                "\n\nNote: If you forward this email to other people, they will be able to " +
+                "access the share as well. To prevent this, remove the link before forwarding.",
+              html: html,
+            });
+          } catch (e) {
+            outerResult.failures.push({contact: contact, error: e.toString()});
+          }
+        } else {
+          // TODO(someday): notify the user somehow?
```
dwrensha: Looks like this TODO was addressed too!

dwrensha:
I pulled this branch to my dev server and I keep getting an error like "Failed to send to: David Renshaw" when I try to share-by-identity.

jparyani:
It wasn't actually rebased on top of #1235 yet and thus didn't work. It's now fixed. I was previously testing locally with an older version of `getVerifiedEmail`.

dwrensha:
Aw... it's sad that the /grain/ link in the notification email displays just "Unauthorized [403]" with no hint that you should log in to access the grain. I guess now would be a good time to figure out a better message for that.

dwrensha:
Oh, I see that the small italic text in the email explained as much.

paulproteus:
Perhaps the link should be to

/grain/:grainId?requireLogin=true

(or something like that)

so that we immediately show a login box.

shell/shared/grain.js
```diff
@@ -273,32 +280,70 @@ Meteor.methods({
...
+                    "link, and you will need to be logged in with your " + contact.profile.service +
+                    " account with username " + contact.profile.intrinsicName,
+                  html: html,
+                });
+              } else {
+                outerResult.failures.push({contact: contact, warning: "User does not have a " +
+                  "verified email, so notification of this share was not sent to them. Please " +
+                  "manually share " + url + " with them."});
+              }
+            } catch (e) {
```
dwrensha: If something goes wrong with sending the email (e.g. the SMTP URL is not properly configured), it looks like this will display the error about the email and possibly leave the impression that the share failed, whereas it was actually just the notification that failed.

jparyani:
> Perhaps the link should be to

> /grain/:grainId?requireLogin=true

That's a good idea, but I would make the URL slightly different, since we're using the trailing part of the path to signify the path arguments inside the grain (I know there's no trailing slash in your example, but I worry it would be confusing). Something like /grainWithLogin/:grainId seems better to me. We should also consider adding the account information, so that we can display an even more custom tailored login message (e.g. /grainWithLogin/github/jparyani/:grainId).

kentonv:
I get an exception during migration:

    Error: Match error: Expected string, got undefined
        at check (packages/check/match.js:33:1)
        at [object Object].getIdentity (packages/sandstorm-db/db.js:776:1)
        at packages/sandstorm-db/migrations.js:364:1
        at [object Object]._.extend.forEach (packages/mongo/mongo_driver.js:1022:1)
        at [object Object].Cursor.(anonymous function) [as forEach] (packages/mongo/mongo_driver.js:869:1)
        at Array.populateContactsFromApiTokens [as 15] (packages/sandstorm-db/migrations.js:362:1)
        at [object Object].migrateToLatest (packages/sandstorm-db/migrations.js:433:1)
        at lib/db-deprecated.js:71:1
        at /home/kenton/src/sandstorm/shell/.meteor/local/build/programs/server/boot.js:249:5

The problem happens with ApiTokens predating multi-identity which were owned by demo users. The multi-identity migration assigned `owner.identityId`s only to tokens whose `owner.userId` matched a user that existed at the time of the migration -- but of course demo users more than an hour old had been deleted.

We'll need to work around such tokens existing. These tokens are not redeemable by anyone, so they are safe to ignore. Arguably they should be deleted, though of course deleting things in a migration is always scary.

It looks like there are also ApiTokens that have a `userId` but no `identityId` for a similar reason. Mostly these are sharing links created by demo users for their own grains, which of course have since been deleted. In theory I think a demo user could have accepted a share link for a non-demo-user's grain and then reshared it, and the effect would be that whoever accepted the reshare will now have the permissions that the demo user assigned to them even if the demo user themselves had lesser permissions on the grain. It's unlikely that this actually happened.

 @dwrensha can you verify my findings and write a cleanup migration?

kentonv:
The auto-complete chip-based editor looks pretty cool -- honestly a lot better than I was expecting!

I did run into some annoying issues, though:

* Sometimes when the autocomplete box first appears the first member is not selected (no purple outline), and so cannot be selected by pressing "enter". Other times it is immediately selected. It should always be immediately selected.
* Pressing enter when nothing is selected in the autocomplete box is treated as form submit. Clearly this is wrong independent of autocomplete, but it is exacerbated by the fact that the autocomplete box could be showing without any identity actually being selected.
* After creating one or more chips, I can apparently select the last one by pressing "backspace", and then delete it by pressing "backspace" again. However, once I have a chip selected, I cannot use the arrow keys to move back to the text input, nor to select a different chip than the last one (e.g. if I want to delete the first chip instead). I also cannot use the arrow keys to get *to* the chip list from the edit input. I intuitively expect all these to work.

Regarding the mail sent:

> Note: This was shared directly with you. No one else can open the above link, and you will need to be logged in with your google account with username Kenton Varda

Let's change this to:

> Note: You will need to log in with your Google account with address temporal@gmail.com to access this grain.

For Github, it should say "log in with your Github account with username kentonv", and for email, "log in with your email address temporal@gmail.com".

**Alternatively** I think we could remove this note entirely if the link we send automatically prompts the user to log in with the appropriate account type, which I think it should. As discussed, I think we would accomplish this by sending a sharing link where the token resolves to an ApiToken that has `owner.user` (not a webkey), and we say something like:

> Access Denied.
>
> This grain was shared with the Google account connected to temporal@gmail.com.
>
> [Log in as temporal@gmail.com with Google] [Request Access as a different user]

The first button would initiate Google login sending [`login_hint`](https://developers.google.com/identity/protocols/OpenIDConnect#login-hint) to the expected user. In the case of other login providers (like Github) which don't provide anything like `login_hint`,  then the "log in as" option should only appear if the user is not already logged in as some other identity from that provider.

The "Request Access" button initiates the request access flow, which will prompt the user to choose the identity under which to request access.

Of course, I don't expect all this to be implemented in this PR.

kentonv:
LOL I messed up your branch. Hold on let me fix that.

kentonv:
OK that's better. The commit I added fixes the migration issue for the purpose of this PR.

dwrensha:
> We'll need to work around such tokens existing. These tokens are not redeemable by anyone, so they are safe to ignore.

https://github.com/jparyani/sandstorm/commit/e7942bf52361591f52c58cdbdaa83e1d9e688734 is a good workaround.

> It looks like there are also ApiTokens that have a userId but no identityId for a similar reason. Mostly these are sharing links created by demo users for their own grains, which of course have since been deleted. In theory I think a demo user could have accepted a share link for a non-demo-user's grain and then reshared it, and the effect would be that whoever accepted the reshare will now have the permissions that the demo user assigned to them even if the demo user themselves had lesser permissions on the grain. It's unlikely that this actually happened.

1. Alice shares to Demo User. (token 1)
2. Demo User shares to Bob. (token 2)
3. Demo user gets deleted.
4. Migrations happen.

Token 1 now has an `owner.user.userId` but no `owner.user.identityId`.
Token 2 now has a `userId` but no `identityId` (nor `accountId`, which is the new field that has taken the place of `userId`).

Bob now has no access to the grain, because the token he owns has an invalid provider.

So I think we can safely delete tokens that look like either token 1 or token 2.

jparyani:
Ready to merge

kentonv:
Merge conflicts.

jparyani:
Rebased and ready to merge

PR #1183, opened 2015-11-20
Upgrade Meteor to 1.2.1

zarvox:
Specific breakage fixed:

* "package" is a keyword in ES6, so switch to using pkg everywhere.
* SCSS requires that variables be quoted if they are not valid CSS values.


zarvox:
Test failed because some globals like `check` are no longer getting automatically imported!  This is actually great.  I'll look into fixing the imports.

zarvox:
Garply, retest this please

PR #1190, opened 2015-11-23
Turn topbar into a flexbox

paulproteus:
This works around the Chrome bug that has been driving me crazy for the past six months.

Notes:

- I need to manually test on IE11 I suppose.

- This elegantly uses `priority=n` to set the order of topbar elements.

@kentonv Thanks for mentioning that a flexbox is a good workaround for this Chrome bug.

### DIFF
--- a/shell/packages/sandstorm-ui-topbar/topbar.html
+++ b/shell/packages/sandstorm-ui-topbar/topbar.html
@@ -28,7 +28,7 @@ limitations under the License.
     <ul class="menubar">
       {{#each items}}
         {{#unless onlyPopup}}
-          <li class="{{name}}">
+          <li style="{{inlineStyle}}" class="{{name}}">
             {{#if ./popupTemplate}}
               <button class="show-popup">
                 {{> template data.get}}
--- a/shell/packages/sandstorm-ui-topbar/topbar.js
+++ b/shell/packages/sandstorm-ui-topbar/topbar.js
@@ -459,6 +459,13 @@ SandstormTopbar.prototype.addItem = function (item) {
     throw new Error("duplicate top bar item name:", item.name);
   }
 
+  // We need inline style to convert the topbar "priority" into an "order" for the flexbox.
+  item.inlineStyle = "";
+  if (item.priority) {
+    item.inlineStyle = 'order: ' + item.priority + ';'
+  }
+
+
   this._items[item.name] = item;
   this._itemsTracker.changed();
 
--- a/shell/client/_topbar.scss
+++ b/shell/client/_topbar.scss
@@ -203,6 +203,7 @@ body>.topbar {
     overflow: hidden;
     @media #{$desktop} {
       height: $topbar-height-desktop;
+      display: flex;
     }
 
     >.home-button {
@@ -366,7 +367,8 @@ body>.topbar {
 
     >.notifications {
       @media #{$desktop} {
-        float: right;
+        // The notification bell and everything after it are right-aligned.
+        margin-left: auto;
       }
 
       @media #{$mobile} {
@@ -416,9 +418,6 @@ body>.topbar {
       }
 
       @media #{$desktop} {
-        float: right;
-        display: block;
-
         >button.show-popup {
           >a {
             padding: 0 8px;
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -58,16 +58,15 @@ limitations under the License.
     <a href="/">Sandstorm</a>
   {{/sandstormTopbarItem}}
   {{#if showAccountButtons}}
-    {{>sandstormTopbarItem name="account" topbar=globalTopbar data=accountButtonsData
+    {{>sandstormTopbarItem name="account" topbar=globalTopbar data=accountButtonsData priority=61
                            template="accountButtons" popupTemplate="accountButtonsPopup"}}
   {{else}}
     {{>sandstormTopbarItem name="login" topbar=globalTopbar data=globalAccountsUi
                            template="loginButtons" popupTemplate="loginButtonsPopup"}}
   {{/if}}
   {{#if currentUser}}
-    {{>sandstormTopbarItem name="notifications" topbar=globalTopbar
-                           template="notifications" popupTemplate="notificationsPopup"
-                           priority=-1}}
+    {{>sandstormTopbarItem name="notifications" topbar=globalTopbar priority=60
+                           template="notifications" popupTemplate="notificationsPopup"}}
   {{/if}}
   {{!-- admin alerts --}}
   {{#with adminAlert}}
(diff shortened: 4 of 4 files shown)


zarvox:
IE11 doesn't need vendor prefixes, but IE10 does: http://caniuse.com/#feat=flexbox .  Consider testing that as well, or declaring bankruptcy and dropping IE10 support.

paulproteus:
Hey @kentonv (and other stakeholders) I want to declare IE11 is the oldest version of IE we support. Rationale:

* Testing things comes at an engineering cost.

* Current Google Analytics says 72% of our IE users use IE11 (and 11.2% use IE9, and 11% use IE10)

* IE is 0.6% of our userbase

* So IE not-11 is 0.16% of our Google Analytics "sessions".

* FWIW Edge is counted separately.

Meanwhile Safari is 4%, and "Safari in-app" is 0.41%.

dwrensha:
For what it's worth, the Chrome bug is reportedly fixed on the next Chrome release. It looks like Chrome releases come [approximately every 6 weeks, and we're due for one in a week or so](https://en.wikipedia.org/wiki/Google_Chrome_release_history).

dwrensha:
Or, rather, if I'm remembering properly, we verified that it's fixed on version 48, but unsure about version 47, which is the imminent release.

ocdtrekkie:
Rationale in favor of dropping IE9/10: You have to have Windows 7 to even potentially not have IE11, which supports IE11 anyhow. The first OS you'd actually cut out with this change is Windows Vista. Vista is an unwanted child anyhow.

Rationale against dropping IE9/10: Some organizations are still using enterprise or government software that is only supported on IE9. If you guys want to cater to enterprise customers, you may need to contend with long support dates, such as below.

Potentially useful, draw.io's support EOL dates for IE versions:
Internet Explorer 7  - 12th January 2016
Internet Explorer 8 - 31st August 2016
Internet Explorer 9 - 30th June 2017

ocdtrekkie:
I find @davidjgraph 's EOL dates super comforting, so I wanted to give him props.

paulproteus:
To get a sense of how much I seem to want this, if this were in a build, I would immediately log into my self-install and `sudo sandstorm update` and not be willing to wait a day for auto-update.

paulproteus:
I'm going to pause here, and if there's pull request feedback, respond more tomorrow.

shell/packages/sandstorm-ui-topbar/topbar.html
```diff
@@ -28,7 +28,7 @@
     <ul class="menubar">
       {{#each items}}
         {{#unless onlyPopup}}
-          <li class="{{name}}">
+          <li {{inlineStyle}} class="{{name}}">
```
zarvox: I'd rather see this read `style="{{inlineStyle}}"` and have `inlineStyle` return a string

shell/packages/sandstorm-ui-topbar/topbar.js
```diff
@@ -427,6 +427,13 @@ SandstormTopbar.prototype.addItem = function (item) {
     throw new Error("duplicate top bar item name:", item.name);
   }
 
+  // We need inline style to conver the topbar "priority" into an "order" for the flexbox.
+  item.inlineStyle = {};
+  if (item.priority) {
+    item.inlineStyle['style'] = 'order: ' + item.priority + ';'
```
zarvox: And then this would just be `item.inlineStyle = 'order: ' + item.priority + ';'`

kentonv:
Note that as of January 12 2016, Microsoft is ending all support and security patches for non-current versions of IE. Also note that Google Apps require at least IE10. So I'm pretty OK with requiring IE10.

https://support.microsoft.com/en-us/gp/microsoft-internet-explorer

zarvox:
To be clear, we already require IE10 today, without this changeset.

The question at hand is whether we should require IE11, by not including the prefixed `-ms-flex`.

ocdtrekkie:
If IE10 is already required I see no reason not to require IE11. (Though Asheesh said 11% of IE users were on IE9.)

shell/packages/sandstorm-ui-topbar/topbar.html
paulproteus: Drew, your proposal results in style="" in the no-inline-style case, but
I'm OK with that, so I've made this change locally. (Will push for review
again shortly.)

paulproteus: Since we require IE10, and Jake Weisz says requiring IE11 is the same as
requiring IE10, then I propose we declare that we require IE11, and then I
can avoid bothering with vendor prefixing, which is good not just because
it makes my life easier but also because it saves us from work down the
road.

Semi-sorry if I sound like I'm being lazy, but I'm extremely happy whenever
I can shrink the testing matrix of things I write. And a line of code (in
this case CSS) never-written is a line tested.

paulproteus:
@kentonv re-ready for merge and/or review.

(Note that I did switch to Chrome 47 for now, but I do want to land this if possible.)

kentonv:
You know, looking at this again I'm seeing that this actually changes the topbar API and changes the ordering of HTML elements (and thus what screenreaders see and what is seen in the menu on mobile). I had assumed previously that this was only changing the layout mechanism.

I'm going to need to look a bit closer at this and see if this is really what we want.

zarvox:
I guess in practice we have several orderings we care about:

* document ordering, which corresponds to tabindex ordering and screenreader ordering
* layout order on desktop
* layout order on mobile

If each of these requires a different order, perhaps we need to be explicit about each of these properties: `documentOrder`, `desktopOrder`, `mobileOrder` could be separate properties.  We'd order things in the document by documentOrder (which is I think what `priority` did before this changeset), but specify flex-order for `desktopOrder` (and perhaps, someday we'd make `mobileOrder` also flexbox-driven - today, it simply uses document order as well.).

kentonv:
Yeah, I think it makes sense for "desktop order" to be specified in CSS. Before this change, we roughly specified desktop layout using "float: left", "float: right", and "display: inline-block" to specify things that should go on the left, on the right, or in the middle. Ordering within these groups was then based on HTML order which was based on priority. (Note that the float-rights ended up in reverse order, or in other words higher-priority things were pushed towards the outside. This was intentional.)

kentonv:
Again I really apologize for not having thought about this the first time I saw this PR. For some reason my brain turned off when I saw zarvox had comments that weren't addressed yet. I really should have thought more carefully then. I know it's really frustrating to have a redesign requested after iterating for a while on the original design... :/

paulproteus:
It's OK - thanks for the thoughtfulness now!

In this case, I have Chromium 47 for now so I'm not in as tight of a bind
on this. (-:

And I like the direction this is going. A question - when you say "desktop
order specified in the CSS" do you mean similar to how this works now, by
using inline style to set the order? Or do you think the sass files should
specify it?

My original version had the desktop order specified in the sass files, but
honestly I don't love that because it moves the expression of the order far
away from the definition of the topbar elements. But I can live with either.

kentonv:
I mean in the scss file. I consider the ordering within the topbar to be a design/layout matter, which as you know I prefer to go into the scss rather than the HTML.

paulproteus:
The main reason I did it in the HTML, for what it's worth, is that I can
easily tell that I've updated all the callsites of the sandstormTopBar, but
in CSS, some of the topbar elements are harder to find.

But I am cool with us being consistent about separating styles out into the
scss files! So let's go with that.

paulproteus:
I'm going to close this for now, since I don't plan to work on it in the near future.

PR #1196, opened 2015-11-24
Installer: Use throw-away GPGHOME

kentonv:
In theory our gpg commands shouldn't affect or be affected by GPGHOME, but one user reported that gpg failed out on their machine because the default GPG homedir didn't already exist.


paulproteus:
+1

I would prefer to set the GPG homedir with `--homedir` but life is short and I'm happy enough to just merge this as-is.

PR #1199, opened 2015-11-24
add pseudocollection for tracking other accounts that an identity is linked to

dwrensha:
This big change here is the addition of the fully reactive `LoginIdentitiesOfLinkedAccounts` pseudocollection. This allows for some simplifications in the login interstitial. The new pseudocollection also gets used for a new widget on the profile editor page; now if an identity is linked to other accounts, you can click a button to expand a list of login identities for those accounts, which looks like this (and needs some design work):
hidden

expanded

### DIFF
--- a/shell/packages/accounts-identity/accounts-identity-server.js
+++ b/shell/packages/accounts-identity/accounts-identity-server.js
@@ -45,70 +45,38 @@ Meteor.methods({
                                  "identity", function () { return { userId: accountUserId }; });
   },
 
-  getLoginAccountOfIdentity: function() {
-    // Attempts to find an account that has the current user as a login identity. If the identity
-    // is not linked to any account, creates a new account for it. Returns a value of type
-    // `OneOf({alreadyAccount: null},
-    //        {loginAccountId: String},
-    //        {nonloginAccounts: [{accountId: String, loginIdentityUser: User}]})`
-    // where the nonloginAccounts variant indicates that this identity cannot log in to any existing
-    // account, and the corresponding list has information about the accounts that this identity is
-    // linked to. The alreadyAccount variant is not an error because the client is allowed to call
-    // this method before its accountIdentities subscription is ready, and therefore it might
-    // not yet know whether the user is an identity or an account.
+  createAccountForIdentity: function() {
+    // Creates a new account for the currently-logged-in identity.
 
     var user = Meteor.user();
-    if (!user) {
-      throw new Meteor.Error(403, "Must be logged in to look up accounts.")
-    }
-    if (user.loginIdentities) return {alreadyAccount: null};
-
-    var loginAccount = Meteor.users.findOne({"loginIdentities.id": user._id},
-                                            {fields: {_id: 1, "loginIdentities.$": 1}});
-
-    if (loginAccount) {
-      return {loginAccountId: loginAccount._id};
+    if (!(user && user.profile)) {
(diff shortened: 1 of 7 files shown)


shell/packages/accounts-identity/accounts-identity-server.js
```diff
@@ -238,3 +202,62 @@ Accounts.linkIdentityToAccount = function (identityId, accountId) {
                       {$push: {"nonloginIdentities": {id: identityId}}});
 
 }
+
+Meteor.publish("accountsOfIdentity", function (identityId) {
+  check(identityId, String);
+  var hasIdentityCursor =
+      Meteor.users.find({$or: [{_id: identityId},
```
kentonv: The first clause in this `or` seems wrong -- it doesn't in any way depend on the caller's user ID, and I think it will always match at least one user record if the identity is in the system at all.

shell/packages/accounts-identity/accounts-identity-server.js
```diff
@@ -238,3 +202,62 @@ Accounts.linkIdentityToAccount = function (identityId, accountId) {
...
+                               {_id: this.userId, "loginIdentities.id": identityId},
+                               {_id: this.userId, "nonloginIdentities.id": identityId}]});
+  if (hasIdentityCursor.count() == 0) return;
+  var self = this;
+  hasIdentityCursor.observe({removed: function () { self.stop(); }});
```
kentonv: It looks like you're leaking the observe handle here.

shell/packages/accounts-identity/accounts-identity-server.js
```diff
@@ -238,3 +202,62 @@ Accounts.linkIdentityToAccount = function (identityId, accountId) {
...
+            changed: function (id, fields) {
+              self.changed("loginIdentitiesOfLinkedAccounts", id, fields);
+            }
+          });
+      }
+    });
+  }
+  var cursor = Meteor.users.find({$or: [{"loginIdentities.id": identityId},
+                                        {"nonloginIdentities.id": identityId}]});
+  cursor.forEach(addIdentitiesOfAccount);
```
kentonv: I think this is redundant. When you call `observe`, it calls the `added` callback for all current results before returning. So I think you want to delete this line, and remove `this.ready()` to happen after the `observe()` call.

dwrensha: The point here is to call `this.ready()` at the correct time. The client waits for `ready()` and then makes a decision about what to do.

kentonv: Sorry, my previous comment should have said "move this.ready()", not "remove this.ready()". If this.ready() happens after the observe, then you'll get the right behavior: observe() calls "added" callbacks for all matching documents before returning, i.e. it is equivalent to forEach().

dwrensha: Cool! I'm happy to have learned about this, and I apologize for not reading your initial comment more carefully before responding

shell/packages/accounts-identity/accounts-identity-server.js
dwrensha: Fixed.

shell/packages/accounts-identity/accounts-identity-server.js
dwrensha: Fixed.

PR #1202, opened 2015-11-25
For review/feedback, not ready for merge: Response Header Whitelist

sprin:
The idea is to allow through a header, X-Session-Id, that identifies an anonymous browser session in the context of exported HTTP APIs. Like a session cookie, a session id is generated server-side by the app, signed, and sent to the client. Client-side JavaScript receives the header, and stores it in sessionStorage. In future requests, the client sends back the X-Session-Id header, and the backend validates the signature, then uses the id to authorize the request.

Perhaps a more general solution is to use a JWT as a store for arbitrary session data, and name the header X-Session.

This patch is not yet working: the client never sees any X-Session-Id originating from the app. I also was not clear on when external-ui-view.js plays a role.

### DIFF
--- a/src/sandstorm/sandstorm-http-bridge.c++
+++ b/src/sandstorm/sandstorm-http-bridge.c++
@@ -33,6 +33,7 @@
 #include <unistd.h>
 #include <map>
 #include <unordered_map>
+#include <set>
 #include <time.h>
 #include <stdlib.h>
 #include <signal.h>
@@ -260,6 +261,42 @@ public:
           break;
       }
       cookie.setHttpOnly(cookies[i].httpOnly);
+    }
+
+    // Add whitelisted headers to additionalHeaders. With respect to security,
+    // the consumers of  WebSession::Response are responsible for making sure
+    // these headers are actually whitelisted. Since this bridge is included in
+    // the app package and runs in the grain itself, we cannot trust that the
+    // whitelist is correctly implemented here. An alternate implementation may
+    // not respect the whitelist. However, for the sake of building a Response
+    // that contains only valid headers, only whitelisted headers are added
+    // here.
+
+    // Create a set from the headerWhitelist
+    auto headerWhitelist = std::set<std::string>(
+        WebSession::Response::HEADER_WHITELIST->begin(),
+        WebSession::Response::HEADER_WHITELIST->end());
+
+    // Add whitelisted headers, and headers matching the app prefix, to a
+    // temporary vector of headers. It is possible for a header name to appear
+    // more than once.
+    kj::Vector<Header> headersMatching;
+    for (auto &header : headers) {
+      auto headerName = header.first;
+      if (headerWhitelist.find(headerName) != headerWhitelist.end() ||
+          headerName.startsWith(WebSession::APP_HEADER_PREFIX)) {
+        headersMatching.add(kj::mv(header.second));
+      }
+    }
+    // Initialize additionalHeaders once we know how many headers to include.
+    auto headerList = builder.initAdditionalHeaders(headersMatching.size());
+    // Add the headers matching the whitelist
+    int i = 0;
+    for (auto const &header : headersMatching) {
+      auto respHeader = headerList[i];
+      respHeader.setName(header.name);
+      respHeader.setValue(header.value);
+      i++;
     }
 
     switch (statusInfo.type) {
--- a/src/sandstorm/sandstorm-http-bridge.c++
+++ b/src/sandstorm/sandstorm-http-bridge.c++
@@ -1561,6 +1561,14 @@ private:
             checkIsHttpToken(headerName),
             ": ",
             checkIsHeaderValue(headerValue)));
+        } else {
+          for (auto const &prefix : *WebSession::Context::HEADER_PREFIX_WHITELIST)
+            if (headerName.startsWith(prefix.cStr())) {
+              lines.add(kj::str(
+                checkIsHttpToken(headerName),
+                ": ",
+                checkIsHeaderValue(headerValue)));
+          }
         }
       }
     }
--- a/src/sandstorm/sandstorm-http-bridge.c++
+++ b/src/sandstorm/sandstorm-http-bridge.c++
@@ -31,6 +31,7 @@
 #include <capnp/serialize.h>
 #include <arpa/inet.h>
 #include <unistd.h>
+#include <regex>
 #include <map>
 #include <unordered_map>
 #include <set>
@@ -1543,8 +1544,24 @@ private:
     }
     auto additionalHeaderList = context.getAdditionalHeaders();
     if (additionalHeaderList.size() > 0) {
+
+      // Create a set from the headerWhitelist
+      auto headerWhitelist = std::set<std::string>(
+          WebSession::Context::HEADER_WHITELIST->begin(),
+          WebSession::Context::HEADER_WHITELIST->end());
+
       for (auto header : additionalHeaderList) {
-        lines.add(kj::str(header.getName(), ": ", header.getValue()));
+        auto headerName = header.getName();
+        auto headerValue = header.getValue();
+        if (headerWhitelist.find(headerName) != headerWhitelist.end() ||
+            headerName.startsWith(WebSession::APP_HEADER_PREFIX)) {
+          // Check header is present in whitelist.
+          // Check header name and value to prevent header injection.
+          lines.add(kj::str(
+            checkIsHttpToken(headerName),
+            ": ",
+            checkIsHeaderValue(headerValue)));
+        }
       }
     }
     auto eTagPrecondition = context.getETagPrecondition();
@@ -1581,6 +1598,36 @@ private:
 
     lines.add(kj::str(""));
     lines.add(kj::str(""));
+  }
+
+  static kj::StringPtr checkIsHttpToken(kj::StringPtr val) {
+    // Allowable values for HTTP tokens per RFC 7230 ("token").
+    // Same as valid header field names ("field-name").
+    // Notably, this excludes CR and LF, thus preventing header injection.
+    // TODO(cleanup): Use kj::parse rather than std::regex.
+    static const std::regex tokenRe = std::regex("^[a-zA-Z0-9_!#$%&'*+.^`|~-]+$");
+    if (!std::regex_match(val.cStr(), tokenRe)) {
+      KJ_FAIL_ASSERT("Invalid HTTP token.", val.cStr());
+    } else {
+      return val;
+    }
+  }
+
+  static kj::StringPtr  checkIsHeaderValue(kj::StringPtr val) {
+    static const std::regex headerValRe = std::regex(
+      "^[\x21-\x7E,\x80-\xFF]+([\x20,\x09]+[\x21-\x7E,\x80-\xFF]+)*?$");
+    // Allowable values for header field values per RFC 7230 ("field-content").
+    // Does not check for obsolete line-folding ("obs-fold").
+    // Notably, this excludes CR and LF, thus preventing header injection.
+    // Character ranges explained:
+    // \x21-\x7E: visible (printing) characters ("VCHAR").
+    // \x80-\xFF: characters outside of US ASCII, treated as opaque data ("obs-text").
+    // \x20,x09: space and horizontal tab ("SP / HTAB")
+    if (!std::regex_match(val.cStr(), headerValRe)) {
+      KJ_FAIL_ASSERT("Invalid HTTP token.", val.cStr());
+    } else {
+      return val;
+    }
   }
 
   template <typename Context>
--- a/src/sandstorm/web-session.capnp
+++ b/src/sandstorm/web-session.capnp
@@ -38,6 +38,11 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
   # application's storage was written and may automatically implement etags based on hashing the
   # content.
 
+  const appHeaderPrefix :Text = "x-sandstorm-app-";
+  # Headers with this prefix will always be allowed in both the request and response.
+  # They represent custom HTTP headers that an app is intentionally exposing through
+  # Sandstorm.
+
   struct Params {
     # Startup params for web sessions.  See `UiView.newSession()`.
 
--- a/src/sandstorm/web-session.capnp
+++ b/src/sandstorm/web-session.capnp
@@ -391,6 +391,7 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
       # purposes. This whitelist exists to help avoid the need to modify code originally written
       # without Sandstorm in mind -- especially to avoid modifying client apps.
       # Feel free to send us pull requests adding additional headers.
+      "x-oc-mtime",            # Owncloud protocol
     ];
 
   }
--- a/src/sandstorm/web-session.capnp
+++ b/src/sandstorm/web-session.capnp
@@ -376,6 +376,23 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
       # TODO(someday):  Return blob directly from storage, so data doesn't have to stream through
       #   the app?
     }
+
+    additionalHeaders @19 :List(Header);
+    # Additional headers present in the reponse. Only whitelisted headers are
+    # permitted.
+
+    struct Header {
+      name @0 :Text;  # lower-cased name
+      value @1 :Text;
+    }
+
+    const headerWhitelist :List(Text) = [
+      # Non-standard response headers which are whitelisted for backwards-compatibility
+      # purposes. This whitelist exists to help avoid the need to modify code originally written
+      # without Sandstorm in mind -- especially to avoid modifying client apps.
+      # Feel free to send us pull requests adding additional headers.
+    ];
+
   }
 
   interface RequestStream extends(Util.ByteStream) {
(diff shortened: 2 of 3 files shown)


garply:

sprin:
Yes, even with the typo fix, the header is not making it through.

Thanks for the catch @mnutt !

kentonv:
We have a "whitelist" for request headers that should simply be passed through. Could we add such a whitelist for response headers, and add this to it?

Is the exact name important? If not, can we use a name that's very unlikely to collide with any other header name that might be in use by other software today or in the future? I'd hate for someone else to define some other meaning for `X-Session-Id` that has security implications incompatible with our sandbox.

Perhaps we could even whitelist `X-Sandstorm-App-*`, thus allowing apps to define whatever additional headers they need, in a way that we know will never be assigned any special meaning by browsers or proxies?

sprin:
@kentonv: sounds good. The name of the session id header is not important in the case of my app, since it is defined by the Sandstorm plugin. I think having an `X-Sandstorm-App-*` escape hatch for custom headers is a good idea.

My understanding is that the response headers need to appear explicitly in the web-session capnp schema. I'll look at how request headers are being passed through to see if the same can be done for response headers.

sprin:
Ready for more review.

My questions: 
 - I am not sure if this is how we want to handle the wildcard in the whitelist. It could simply be added to the proxy as a special case, but I like having it documented in web-session.capnp.
 - I think the wildcard should be added to the request header whitelist as well, yes?

kentonv:
Hi Steffen,

Thanks for working on this!

I think that the header whitelist needs to be applied in sandstorm-http-bridge, not in proxy.js. While proxy.js is the main consumer of WebSession capabilities, it is not the only possible consumer. Our intent is that it will eventually be possible for clients to call the Cap'n Proto interfaces directly.

> I am not sure if this is how we want to handle the wildcard in the whitelist. It could simply be added to the proxy as a special case, but I like having it documented in web-session.capnp.

I agree that it's nice to see it documented there, but the logic to support it is kind of complicated and sadly a bit slow: where previously you could iterate over the whitelist and look up each in the map for O(n lg m) time, now you have an O(n * m) loop. Maybe the performance will never matter, but the code also seems more complex.

So I would somewhat lean towards special-casing this and documenting it in web-session.capnp as a comment. However, I leave it up to your judgment.

sprin:
> I think that the header whitelist needs to be applied in sandstorm-http-bridge, not in proxy.js.

I thought that since we wanted to whitelist response headers primarily for security reasons, we need to do the validation outside of the grain.  The grain can bundle it's own HTTP bridge that does not apply the whitelist, therefore defeating our attempts to enforce the whitelist.

> Our intent is that it will eventually be possible for clients to call the Cap'n Proto interfaces directly.

Where will the whitelist be applied when clients speak Cap'n Proto natively?

I think the only viable way to enforce such a whitelist is outside the grain, in the consumer, since we can't trust the producer inside the grain, and we can't express such a constraint in the language of the schema itself. If there are multiple consumers of the WebSession Response inside Sandstorm itself, then I think there needs to be a shared validator. This is difficult if our consumers are in different languages.

Aside: While Cap'n Proto doesn't support list/regex constraints, [XML Schema](http://www.w3.org/TR/xmlschema-2/#regexs) and [JSON Schema](http://www.w3.org/TR/xmlschema-2/#regexs) do. I can see the reasons for not including extensive validation support in the schema language (speed and complexity), but it does mean the burden of validating lies on the consumers, for anything beyond type/enum validations.

> I agree that it's nice to see it documented there, but the logic to support it is kind of complicated and sadly a bit slow

Yes, I agree regarding the complexity, especially if there is more than one consumer that will need to implement this. As for big O performance, I wasn't too concerned since I think it's reasonable to expect that m and n will always be very small.

However, @zarvox had a great suggestion that I believe points the way to resolving both the complexity and speed issues. Assuming we only want a single, constant wildcard prefix (X-Sandstorm-App-), a separate list can be used for headers that should be prefixed. This means the "verbatim whitelist" operation remains O(n lg m), but we now add another O(n) step to check if any headers match the prefix. The downside is that there are now two lists in the schema for "extra headers": "additionalHeaders" for unprefixed, whitelisted headers, and a new list for prefixed headers. Another way to deal with this is to add an extra attribute to the Header struct itself, indicating if the header should be treated as one expected to be in the "verbatim whitelist" or prefixed.

I will make some changes to address the complexity issue in consumers (and incidentially, performance), following the approach of adding an extra attribute to Header.

In the meantime, I am curious to hear your thoughts on where response validations should happen (in the grain, Sandstorm, or the serialization format itself).

sprin:
I forgot to say, thanks for the feedback! I share your desire for a careful solution here, since it is in a performance and security critical area.

kentonv:
Sorry for the delay, a lot was going on and this slipped past me.

> we can't trust the producer inside the grain

Ughh, of course, that was a dumb suggestion on my part. The whitelist needs to be applied on the proxy end for security, otherwise an app could serve whatever header it wanted by constructing the response maliciously. Actually, I think the whitelist may need to be applied at *both* ends.

sprin:
No problem, I have been working on the next step in small pieces as I dust off my C++ and learn more about capnp, kj, and Sandstorm.

I started implementing the new strategy both at the proxy and the bridge. Though code inside the grain can't be trusted, it still makes sense for the bridge that Sandstorm ships to send a response with the correct headers.

Just as a learning exercise, I started a line-by-line re-implementation of the bridge, leaving out the network code and reading a response from stdin, and outputting the capnp message to stdout, so it can be piped to `capnp decode`. This also makes for easy debugging of the code that will go into the PR. Later, I'd like to look at building a similar network-independent testing interface to sandstorm-http-bridge.

kentonv:
That sounds pretty great, actually! More tests are always a good thing.

sprin:
I've added the schema, proxy, and bridge changes. I rebased onto master, started testing the front end, and noticed a showstopper bug in the JS, so it's not quite there yet.

sprin:
Ready for more review.

The rules now appear to be working correctly in proxy.js. I also added validation of header field names and values in `additionalHeaders` against [RFC 7230](https://tools.ietf.org/html/rfc7230) to prevent header injection for the case where capnp messages can be received from an untrusted source.

TODOs for other PRs:
 - I believe we need to apply this validation on all header fields that are translated from capnp messages.
 - Add additional headers to response whitelist. Only "Allow" was added, for the purpose of having something to test against.
 - Create network-independent testing interface for bridge, for both request and response handling.

mnutt:
I just found out that to get Owncloud's X-OC-MTime header working, I need to whitelist both the request and response header. Thanks for putting this together, and I'm looking forward to seeing it merged!

src/sandstorm/web-session.capnp
```diff
@@ -366,6 +375,27 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
...
+      # Indicates whether name represents a header that was prepended with appHeaderPrefix.
+      # It should be reconstructed with the prefix before being sent to the HTTP client.
+    }
+
+    const headerWhitelist :List(Text) = [
+      # Non-standard response headers which are whitelisted for backwards-compatibility
+      # purposes. This whitelist exists to help avoid the need to modify code originally written
+      # without Sandstorm in mind -- especially to avoid modifying client apps.
+      # Feel free to send us pull requests adding additional headers.
+      "allow"
```
kentonv: I'd prefer only to whitelist non-standard headers. For standard headers, we should represent the header's meaning semantically using regular struct fields.

If the purpose is just for testing, perhaps you could make up a header `X-Sandstorm-Whitelist-Test`, or maybe add `X-OC-MTime` as requested by @mnutt.

src/sandstorm/web-session.capnp
```diff
@@ -38,6 +38,11 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
   # application's storage was written and may automatically implement etags based on hashing the
   # content.
 
+  const appHeaderPrefix :Text = "x-sandstorm-app-";
```
kentonv: It feels slightly odd to specify this in lower-case when lower-casing of headers is specifically a property of Node. Perhaps we should specify it as "X-Sandstorm-App-" and expect the caller to lower-case it if the context calls for that?

kentonv: Then again I suppose we're specifying the whitelist itself in lowercased form. OK, then. Bleh, case-insensitivity is the worst.

src/sandstorm/web-session.capnp
```diff
@@ -140,6 +145,9 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
     struct Header {
       name @0 :Text;  # lower-cased name
       value @1 :Text;
+      prefixed @2 :Bool = false;
```
kentonv: Mixing non-prefixed and prefixed headers like this worries me a bit -- it feels like a pattern that could lead to injections when receiving code doesn't carefully check this boolean.

Would it make sense instead to separate the whitelisted headers and prefixed headers into two lists? Or maybe we should simply not strip the prefix from the name in Header.name, to avoid ambiguity?

kentonv: In fact, as-is, this is exploitable: Existing apps compiled before this change will not check the `prefix` field, so after this change anyone could send `X-Sandstorm-App-X-Sandstorm-User-Id` to old apps to falsify authentication info.

src/sandstorm/sandstorm-http-bridge.c++
```diff
@@ -260,6 +262,56 @@ public:
...
+    // Add the headers matching the whitelist
+    int i = 0;
+    for (auto const &header : headersMatchingWhitelist) {
+      auto respHeader = headerList[i];
+      respHeader.setName(header.name);
+      respHeader.setValue(header.value);
+      i++;
+    }
+    // Add the headers matching the app prefix, setting the "prefixed" flag.
+    size_t prefix_size = WebSession::APP_HEADER_PREFIX.toString().size();
```
kentonv: Nitpick: Please use camelCase names rather than under_scores.

src/sandstorm/sandstorm-http-bridge.c++
```diff
@@ -1238,7 +1290,19 @@ private:
...
+          // Check header name and value to prevent header injection.
+          if (header.getPrefixed()) {
+            // If the header has the "prefixed" flag set, prepend the app header prefix.
+            lines.add(kj::str(
+                  WebSession::APP_HEADER_PREFIX.toString(),
+                  header.getName(),
+                  ": ",
+                  header.getValue()));
+          } else {
+            lines.add(kj::str(header.getName(), ": ", header.getValue()));
```
kentonv: Can you add a check that the header is actually in the whitelist here?

Note that since the whitelist isn't long, it's probably fine to iterate over it.

src/sandstorm/sandstorm-http-bridge.c++
```diff
@@ -1238,7 +1290,19 @@ private:
     auto additionalHeaderList = context.getAdditionalHeaders();
     if (additionalHeaderList.size() > 0) {
       for (auto header : additionalHeaderList) {
-        lines.add(kj::str(header.getName(), ": ", header.getValue()));
+        if (checkIsHttpToken(header.getName()) && checkIsHeaderValue(header.getValue())) {
```
kentonv: Cap'n Proto string getters are not particularly efficient since they need to validate the pointer. Can you save getName() and getValue() into local vars first to avoid calling these repeatedly below?

src/sandstorm/sandstorm-http-bridge.c++
```diff
@@ -1274,6 +1338,27 @@ private:
     lines.add(kj::str(""));
   }
 
+  static bool checkIsHttpToken(kj::StringPtr val) {
+    // Allowable values for HTTP tokens per RFC 7230 ("token").
+    // Same as valid header field names ("field-name").
+    // Notably, this excludes CR and LF, thus preventing header injection.
+    static const std::regex tokenRe = std::regex("^[a-zA-Z0-9_!#$%&'*+.^`|~-]+$");
```
kentonv: Can you add:

    // TODO(cleanup): Use kj::parse rather than std::regex.

(I won't make you cater to my NIH in this PR, though. :) )

src/sandstorm/sandstorm-http-bridge.c++
```diff
@@ -1274,6 +1338,27 @@ private:
...
+    return std::regex_match(val.cStr(), tokenRe);
+  }
+
+  static bool checkIsHeaderValue(kj::StringPtr val) {
+    static const std::regex headerValRe = std::regex(
+      "^[\x21-\x7E,\x80-\xFF]+([\x20,\x09][\x21-\x7E,\x80-\xFF]+)*?$");
```
kentonv: The spec defines field-content as:

    field-content = field-vchar [ 1*( SP / HTAB ) field-vchar ]

I think the `1*` means "one or more", but your regex appears to interpret it as "exactly one".

src/sandstorm/sandstorm-http-bridge.c++
```diff
@@ -1274,6 +1338,27 @@ private:
     lines.add(kj::str(""));
   }
 
+  static bool checkIsHttpToken(kj::StringPtr val) {
```
kentonv: You have these helpers returning bool but I think it may make sense for them to throw exceptions. I can't think of a case where it's important to gracefully recover from garbage header names or values.

In fact I'd have the helpers return their parameter if it's valid and throw if not. That way we can easily add calls to these helpers in lots of other places above where we ought to have had them all along, to sanitize other input strings.

OTOH, if a header name is merely missing from the whitelist, then we should drop it rather than throw, otherwise it could be a compatibility issues.

kentonv:
Sorry for the long delay in reviewing this -- holidays, etc. I think this is all the comments I have for now. Thanks for working on this!

paulproteus:
@mnutt mentioned that this is blocking a Davros update that adds `X-OC-mtime`. @sprin , do you think that you're interested in reviving this PR?

If so, that's great. And if not, then we should figure out how to get something like this landed; perhaps @mnutt would find himself willing to take it over.

garply:

sprin:
@paulproteus, yes, I would be interested in picking this up again. I hadn't read @kentonv's feedback until now, since I was traveling at the time and forgot to come back and take a look.

mnutt:
I'd be very interested in helping get this merged, because some of the newer Owncloud clients have an off-by-one error that causes them to crash when they don't get the `x-oc-mtime` header.

sprin:
@mnutt: I am picking this up today, and will have some time to work over the next few days. I'll be reachable over IRC when I am working.

sprin:
My updates are on a different branch. All commits were rewritten to some degree:
https://github.com/sandstorm-io/sandstorm/compare/master...sprin:header_rules

Should I (force) update the branch this PR is tracking? I believe github will drop the content of the original changes from the PR thread, thus losing important information from the discussion. Is it better to make a different PR if the commit history is changed?

sprin:
> I'd prefer only to whitelist non-standard headers. For standard headers, we should represent the header's meaning semantically using regular struct fields.
>
> If the purpose is just for testing, perhaps you could make up a header X-Sandstorm-Whitelist-Test, or maybe add X-OC-MTime as requested by @mnutt.

Replaced the test header with X-OC-MTime.

> Mixing non-prefixed and prefixed headers like this worries me a bit -- it feels like a pattern that could lead to injections when receiving code doesn't carefully check this boolean.
>
> Would it make sense instead to separate the whitelisted headers and prefixed headers into two lists? Or maybe we should simply not strip the prefix from the name in Header.name, to avoid ambiguity?

Removed the prefix boolean and kept the prefix on Header.name.

> Can you add a check that the header is actually in the whitelist here?
>
> Note that since the whitelist isn't long, it's probably fine to iterate over it.

Added a check of request headers against the whitelist in the bridge.

> Cap'n Proto string getters are not particularly efficient since they need to validate the pointer. Can you save getName() and getValue() into local vars first to avoid calling these repeatedly below?

:thumbsup:

> Can you add:
> ```
> // TODO(cleanup): Use kj::parse rather than std::regex.
> ```

:thumbsup:


> The spec defines field-content as:
> ```
> field-content = field-vchar [ 1*( SP / HTAB ) field-vchar ]
> ```
> I think the 1* means "one or more", but your regex appears to interpret it as "exactly one".

Fixed the regex.

> You have these helpers returning bool but I think it may make sense for them to throw exceptions. I can't think of a case where it's important to gracefully recover from garbage header names or values.
>
> In fact I'd have the helpers return their parameter if it's valid and throw if not. That way we can easily add calls to these helpers in lots of other places above where we ought to have had them all along, to sanitize other input strings.

Changed the helpers to return the parameter if valid and throw if not.

I also saw that a request header prefix whitelist entry had been added for `x-hgarg-*`. I created a separate constant list for request header prefix whitelist entries, as this simplified the logic and improved efficiency. The behavior should be the same, except that the trailing `*` is dropped since entries in the new constant list are understood to be prefixes.

sprin:
I just force-updated this branch after @dwrensha assured me comments on the old commits would not be lost. This is ready for review now.

paulproteus:
Hi @sprin ! My guess is that @kentonv will take a fresh look tomorrow, as that's the biggest code-review day around here!

paulproteus:
@sprin I notice the pull request title still says "Not ready for merge". But it is now, right? If so please change the title accordingly so I can make sure we're respecting your wishes.

If not ready for merge, but merely ready for review, then that's fine too, but I thought I should ask.

kentonv:
I've tweaked this a bit in https://github.com/sandstorm-io/sandstorm/pull/2271 and plan to merge it after tests run. Thanks!

mnutt:
I tested it on Davros and see the `x-oc-mtime` header coming through just fine.

It merges cleanly into sandstorm/master but the annotation needs to be updated from `@19` to `@20`.

sprin:
Great! Glad to see this merged in #2271.

PR #1208, opened 2015-11-28
Docs: HTTP Communication Overview

sprin:
This adds a new section to "How It Works" with graphical overviews of communication between a web client and both native and legacy HTTP Sandstorm apps. The graphs are generated by graphviz.

This depends on the markdown-inline-graphviz plugin. To install graphviz and the plugin (Debian example):
```
sudo apt-get install graphviz
pip install markdown-inline-graphviz
```

### DIFF
--- a/docs/using/how-it-works.md
+++ b/docs/using/how-it-works.md
@@ -11,3 +11,64 @@
 * The application may store persistent state in the `/var` directory.
 * App servers are aggressively killed off as soon as the user closes the browser tab, then restarted when the user returns later.
 * Packages are cryptographically signed.  Packages signed with the same key represent versions of the same app, and are thus allowed to replace older versions -- although the user must still confirm these upgrades.
+
+HTTP Communication Overview
+===========================
+
+This diagram shows shows how communications flows from a web client, such as a
+browser, to a native Sandstorm app (one which speaks Cap'n Proto).
+
+{% dot communication_overview_native_app.svg
+    graph G {
+      rankdir=LR;
+      compound=true;
+      node [shape=box fontsize=10];
+
+      client [label="Web client\n(eg. browser)"];
+
+      subgraph cluster_sandstorm {
+        label="Sandstorm";
+        proxy [label="Proxy";];
+        websession [label="WebSession Serialization\n(HTTP over Cap'n Proto)"];
+      }
+
+      subgraph cluster_grain {
+        label="Grain";
+        app [label="Native Sandstorm App\n(speaks Cap'n Proto )"];
+      }
+
+      client -- proxy;
+      proxy -- websession;
+      websession -- app;
+    }
+%}
+
+With legacy HTTP applications, the Sandstorm HTTP bridge is used to translate
+the Cap'n Proto WebSession to HTTP.
+
+{% dot communication_overview_http_app.svg
+    graph G {
+      rankdir=LR;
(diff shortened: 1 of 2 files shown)


garply:

paulproteus:
Garply, ok to test

paulproteus:
This looks great. Some suggestions:

- Use `##` before the header, not reStructuredText-style underlines (Right now your current strategy results in a H1 tag, and I think an H2 makes more sense.)

- I would reverse the order of these two diagrams, since legacy HTTP apps is the common case. (Also I personally think calling most of the world's software "legacy" is a bit unnecessarily intense :grinning: , but since that is the way we are doing things elsewhere, I'm totally OK with this for now!)

- I gently suggest listing filenames like `proxy.js` and `sandstorm-http-bridge.c++` so that people know where to hack on things. I say "gently suggest" because I'm not _totally_ sure this is a great idea, but maybe it is.

Also curious for @mnutt 's feedback, since this is seemingly the doc that he would have liked to have in the past! But it might be hard to know without rendering the doc, which is maybe a lot to ask. So maybe it's OK to wait until it's in the docs to get feedback from @mnutt if he doesn't want to run `dot` locally.

paulproteus:
And let me say again - thanks so much for submitting this! I think this is significant enough that it's probably worth signing the Sandstorm CLA.

https://sandstorm.io/cla/ covers that. Let me know if you're OK with that; it's the same CLA as the Apache Software Foundation uses.

sprin:
I fixed the header, reordered the diagrams and added filenames to the Sandstorm components. I signed the CLA.

The built docs are [here](https://sprin.github.io/sandstorm/using/how-it-works/#http-communication-overview).

A bit of a sidetrack, but worth discussing here I think:

I agree that "legacy" is a bit of a turn-off to potential authors/porters of apps for Sandstorm. When I first read that, I wasn't clear on what benefits my HTTP app would be missing out on, and slightly discouraged. Outlining the specific benefits somewhere would be good resource for app developers to decide if they want to start using Cap'n Proto natively. Dropping the "legacy" moniker for HTTP apps while touting the benefits of Cap'n Proto would be more encouraging. No one wants to put effort into producing a "legacy app".

The section that discusses some benefits of native Cap'n Proto support is [here](https://sprin.github.io/sandstorm/developing/handbook/#platform-protocol-philosophy). To summarize this section, Cap'n Proto-native apps can:

 - run more efficiently
 - quickly take advantage of new platform features
 - native support for sharing data with other apps via capabilities

I think it would be great to elaborate:

 - How will the app be more efficient? Smaller package size? Less memory consumption? (Both, since neither the bridge nor an HTTP parser are needed. Might be good to contrast the same app speaking HTTP versus Cap'n Proto).
 - What features are only accessible via Cap'n Proto now?
 - What new Cap'n Proto-specific features are planned?

mnutt:
I'd second the legacy wording. Even starting from scratch, it's still significantly easier to figure out how to get started with the http bridge. Node-capnp doesn't work with newer node, and requires you to separately build capnp from master.

paulproteus:
I'm happy to merge this straight-away, and let the legacy/etc. discussion continue, either here or on the email list.

paulproteus:
https://docs.sandstorm.io/en/latest/using/how-it-works/ now contains this!

PR #1212, opened 2015-11-30
Add a C.UTF-8 locale descriptor, and generate a locale in make-bundle.sh

zarvox:
Before this patch, to obtain a sane locale, we would copy in locales from the
host system, then set `LANG=C.UTF-8` in `run-bundle.c++`.  This had two issues:

* Build hosts that aren't running Debian don't have a C.UTF-8 locale, so the
  generated bundle wouldn't run unless you'd manually defined one on the host
* This included *all locales* from the host system, including the whole
  locale-archive, which (if you have additional locales installed) could add
  ~100MB to the package.

With this patch, we move to standardize what locale data we ship:
`make-bundle.sh` will generate a locale similar to Debian's C.UTF-8 locale for
inclusion in the bundle.  This locale, like Debian's, is non-conforming to the
SUS, but is significantly more useful in a variety of ways that are more
important for real applications, such as:

* having actual correct character classes for the whole Unicode character set
* having a guaranteed sort order for characters beyond U+007F

To minimize duplicated effort delta from glibc upstream, this implementation
still relies on the system's UTF-8 character map and some other glibc-provided
i18n files that were available as part of the minimal install on all systems I
tested.

Closes #169.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -76,6 +76,15 @@ base64 handling in src/sandstorm/sandstorm-http-bridge.c++
     License: Public Domain
     License text: http://creativecommons.org/licenses/publicdomain/
 
+localedata-C
+    What: A C locale supporting UTF-8, based on Debian and Fedora's C.UTF-8 locales
+    Source URLs:
+      http://sources.debian.net/data/main/g/glibc/2.19-22/localedata/locales/C
+      https://lists.stg.fedoraproject.org/archives/list/glibc%40lists.fedoraproject.org/message/RAIVZBJH55SKKLPSR6AX7RJOIUG3BF57/attachment/2/0001-Add-a-C.UTF-8-locale.patch
+    Copyright: Aurelien Jarno (Debian), Mike Fabian (Fedora)
+    License: LGPL Version 2.1 or later
+    License text: http://opensource.org/licenses/LGPL-2.1
+
 Trademarks
 ==========
 
--- a/Makefile
+++ b/Makefile
@@ -305,7 +305,7 @@ shell-build: shell/lib/* shell/client/* shell/server/* shell/shared/* shell/publ
 # ====================================================================
 # Bundle
 
-bundle: tmp/.ekam-run shell-build make-bundle.sh meteor-bundle-main.js
+bundle: tmp/.ekam-run shell-build make-bundle.sh localedata-C meteor-bundle-main.js
 	@$(call color,bundle)
 	@./make-bundle.sh
 
--- a/localedata-C
+++ b/localedata-C
@@ -1,0 +1,227 @@
+escape_char /
+comment_char %
+% Locale for C locale in UTF-8
+
+LC_IDENTIFICATION
+title      "C locale"
+source     ""
+address    ""
+contact    ""
+email      "drew@sandstorm.io"
+tel        ""
+fax        ""
+language   "C"
+territory  ""
+revision   "1.0"
+date       "2015-11-30"
+%
+category  "C:2015";LC_IDENTIFICATION
+category  "C:2015";LC_CTYPE
+category  "C:2015";LC_COLLATE
+category  "C:2015";LC_TIME
+category  "C:2015";LC_NUMERIC
+category  "C:2015";LC_MONETARY
+category  "C:2015";LC_MESSAGES
+category  "C:2015";LC_PAPER
+category  "C:2015";LC_NAME
+category  "C:2015";LC_ADDRESS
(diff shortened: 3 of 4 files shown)


dwrensha:
I tested this out on the Arch Linux box that originally drove me to open #169. It seems to work well!

dwrensha:
> having actual correct character classes for the whole Unicode character set

What is a "character class" and where do these get specified in your locale definition? Is that part of what `--charmap=UTF-8` brings?

> having a guaranteed sort order for characters beyond U+007F

My understanding is that the plain C locale does have a guaranteed sort order: it just compares raw bytes. So would it be more accurate to say that your C.UTF-8 locale has a guaranteed *sane* sort order?

zarvox:
Yay, glad this also works on Arch.  Thanks for testing! :)

`localedef --help` will print a couple of paths, including the character maps and locale path, which I'll refer to later.

Character classes (approximately LC_CTYPE) are basically the tables that drive `isupper()`, `islower()`, `isdigit()` and such.  More interesting are "transliterations": when you uppercase `ß`, it actually splits into two characters `SS`.  In this implementation, I'm punting and using whatever your distro shipped with by that `copy "i18n"` line, which will basically use whatever is defined in `<your locale path>/i18n`.  It's possible that if you're building from an older distro, it may only support e.g. Unicode 5.0 instead of Unicode 7.0 or something; if we care a ton, we can add `deps/glibc` and use the character classes from upstream directly instead.

`--charmap=UTF-8` says "look inside the System's directory for character maps for a file named `UTF-8` and use that to connect codepoints to bytestrings, character names, and widths".

The plain C locale has a guaranteed sort order (LC_COLLATE) for Unicode codepoints 0x00 through 0x7f - look in `locales/POSIX` inside your "locale path".  I'm not sure what it does for the rest of the characters that fall into the UNDEFINED category - they might be sorted numerically by codepoint, or by bytestring, but I'm not actually sure if that behavior is well-defined.  Granted, I would be surprised if it did anything *but* sort by bytestring, but I can't find that written out anywhere.  (If they sort by bytestring, they also sort by codepoints, because UTF-8 is well designed like that.)  But this way, the desired behavior is explicitly written out.

zarvox:
Added a commit to add `deps/glibc` for the locale data.  I'm curious if folks are +1/-1 on that.

kentonv:
Is `localedata-C` copied from somewhere? Do we need a copyright notice?

Do we know of some specific benefit that the glibc dependency adds, or is it just a matter of cleanliness? I'd be inclined not to include it unless it is fixing something.

zarvox:
I produced `localedata-C` based on [Debian's C locale](http://sources.debian.net/data/main/g/glibc/2.19-22/localedata/locales/C) and [Fedora's still-in-development locale](https://lists.stg.fedoraproject.org/archives/list/glibc@lists.fedoraproject.org/thread/44YDNSOM4KXUY63F336ZUD75XMR5HNHV/).

My guess (since no additional license information was given) would be that both patches are LGPL2.1+, since the glibc packages that contains the patch are default-LGPL2.1+ (with numerous exceptions for specific source files).  What would the proper way to acknowledge/redistribute that be?  Inline comment in the patch with links to the two other patches upon which it is based?  Additional entry in `LICENSE`?  Additional acknowledgement in the `/about` "Third-party works used" page?

As for the glibc dependency: no great reason, but it guarantees that we're using the latest Unicode rules as provided by upstream glibc in our locale, so we'll have correct classifications of whatever emoji they think up this week.  It's also consistent across build environments, whereas who knows what i18n data gets put where in distros.  But we already bundle in whatever libc/libstdc++ comes from the build environment anyway, so **shrug**.  I guess if glibc ever decides to change the layout of the i18n data, we'd need to switch to something predictable then, which would probably be "track upstream glibc" but maybe they'll never do that.

kentonv:
Keeping in mind that I'm not a license compliance expert and could be doing this all wrong: I think both adding an entry to LICENSE and adding comments in the file seem like good things to do. `/about` already lists glibc. It probably should also list Debian (since we use it for our official builds), but I'm not sure if it makes sense to list individual patches to various distros that we mixed-and-matched into one of our source files...

kentonv:
Regarding pulling glibc into deps, let's not for now. It seems to me that if we're using `localedata` from source then we should be building `localedef` from source in case of version skew, but I don't think we want to do that right now. Willing to revisit if we observe actual problems that would be solved.

zarvox:
Dropped the deps/glibc commit, fixup'd updates to the LICENSE.  This should be good to merge.

PR #1213, opened 2015-11-30
move identity switcher into account buttons dropdown

dwrensha:
This rebases and adds one commit (https://github.com/sandstorm-io/sandstorm/commit/7ef09579d34335d239f48473072a54ae0e623f62) to #1199, moving the identity switcher into the account buttons dropdown menu. I'll rebase away the duplicated commits if #1199 ends up getting merged first.

The new commit also gets rid of the `SandstormDb.getUserIdentities()` function, replacing it with `SandstormDb.getUserIdentityIds()`, which doesn't need to make a database query.

### DIFF
--- a/shell/packages/sandstorm-db/profile.js
+++ b/shell/packages/sandstorm-db/profile.js
@@ -18,49 +18,59 @@ var makeIdenticon;
 var httpProtocol;
 
 if (Meteor.isServer) {
+  Meteor.publish("identityProfile", function (identityId) {
+    check(identityId, String);
+
+    // Dummy query handle for the case where this.userId === identityId
+    var hasIdentityHandle = {stop: function () {} };
+
+    if (this.userId !== identityId) {
+      var hasIdentityCursor =
+          Meteor.users.find({$or: [{_id: this.userId, "loginIdentities.id": identityId},
+                                   {_id: this.userId, "nonloginIdentities.id": identityId}]});
+      if (hasIdentityCursor.count() == 0) return;
+      hasIdentityHandle = hasIdentityCursor.observe({removed: function () { self.stop(); }});
+    }
+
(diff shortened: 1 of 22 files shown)


dwrensha:
I've rebased this against master and added some small updates. Now the identity switcher button is not shown unless the user has more than one identity.

shell/packages/sandstorm-db/profile.js
```diff
@@ -18,49 +18,59 @@ var makeIdenticon;
...
 if (Meteor.isServer) {
+  Meteor.publish("identityProfile", function (identityId) {
+    check(identityId, String);
+
+    // Dummy query handle for the case where this.userId === identityId
+    var hasIdentityHandle = {stop: function () {} };
+
+    if (this.userId !== identityId) {
+      var hasIdentityCursor =
+          Meteor.users.find({$or: [{_id: this.userId, "loginIdentities.id": identityId},
```
kentonv: This code appears in multiple places now, right? Can we factor it out into a helper function?

shell/packages/sandstorm-db/profile.js
```diff
@@ -206,49 +216,13 @@ SandstormDb.fillInPictureUrl = function(user) {
     makeIdenticon(user._id);
 }
 
-SandstormDb.getUserIdentities = function (user) {
-  // Given a user object, return an array containing all of the user's identities. Always returns
```
kentonv: Can we keep this comment?

kentonv: Rather, can we add a doc comment to the new method. :)

dwrensha: Done.

shell/packages/sandstorm-db/profile.js
dwrensha: Done.

PR #1214, opened 2015-12-01
Fix bug that prevented offer templates from creating child tokens.

dwrensha:


zarvox:
Ahhhh, I see.  The generated apiToken would be assigned an accountId, even though it wasn't provided by an identity, which meant that on the redemption path, the redeemer didn't hold that accountId, so it couldn't use that token.

dwrensha:
The issue here was a match failure on token creation.

zarvox:
:+1:

PR #1219, opened 2015-12-01
Ensure that expectSize() request gets sent.

dwrensha:
On IRC, mortehu noted that `expectSizeTask` is unused. When it gets dropped, the `expectSize()` request could (and probably does) get cancelled.

This patch ensures that the request goes through and ignores any errors, as recommended by the documentation in [util.capnp](https://github.com/sandstorm-io/sandstorm/blob/master/src/sandstorm/util.capnp).

Note: that documentation may have been written before `Exception.Type.Unimplemented` existed. A more complete solution might be to only ignore errors if they are of type `Unimplemented`. Indeed, there is some code to that effect in blackrock.

### DIFF
--- a/src/sandstorm/supervisor.c++
+++ b/src/sandstorm/supervisor.c++
@@ -2009,7 +2009,7 @@ public:
         req.setSize(stats.st_size);
         auto expectSizeTask = req.send();
         auto inStream = kj::heap<kj::FdInputStream>(kj::mv(*fd));
-        return pump(*inStream, kj::mv(stream)).attach(kj::mv(inStream));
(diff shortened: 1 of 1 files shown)


kentonv:
Eek, that does look like a bug.

Instead of `detach()`, can we attach the promise to the `pump()` call, as we do with `inStream`? Note that `attach()` accepts multiple arguments.

In general I treat any use of `detach()` as suspicious, because using detach makes it hard to reason about ownership. In this particular case I don't think there's anything wrong, but I prefer avoiding `detach()` whenever possible.

dwrensha:
Done.

PR #1220, opened 2015-12-01
Add v1 of referral program code

paulproteus:
Specifically:

- Store referral information on Identities & Accounts, per @kentonv

- Track creation & use of sharing links to record who referred whom

- Create a basic page to see info about the referral program

- Use referral bonus when calculating quota

- Limit max referral bonus

- Use pathFor when creating links, per @zarvox

### DIFF
--- a/shell/packages/sandstorm-accounts-ui/accounts-ui-server.js
+++ b/shell/packages/sandstorm-accounts-ui/accounts-ui-server.js
@@ -23,11 +23,12 @@ Meteor.users.deny({
 });
 
 Meteor.publish("getMyUsage", function () {
+  var db = this.connection.sandstormDb;
   if (this.userId) {
     // TODO(someday): Make this reactive.
     var user = Meteor.users.findOne(this.userId);
-    var usage = this.connection.sandstormDb.getMyUsage(user);
-    var referralBonus = this.connection.sandstormDb.getMyReferralBonus(user);
+    var usage = db.getMyUsage(user);
+    var referralBonus = db.getMyReferralBonus(user);
     this.added("users", this.userId, {
       pseudoUsage: usage,
       pseudoReferralBonus: referralBonus,
--- a/shell/packages/sandstorm-accounts-ui/accounts-ui-server.js
+++ b/shell/packages/sandstorm-accounts-ui/accounts-ui-server.js
@@ -25,8 +25,13 @@ Meteor.users.deny({
 Meteor.publish("getMyUsage", function () {
   if (this.userId) {
     // TODO(someday): Make this reactive.
-    var usage = this.connection.sandstormDb.getMyUsage(Meteor.users.findOne(this.userId));
-    this.added("users", this.userId, {pseudoUsage: usage});
+    var user = Meteor.users.findOne(this.userId);
+    var usage = this.connection.sandstormDb.getMyUsage(user);
+    var referralBonus = this.connection.sandstormDb.getMyReferralBonus(user);
+    this.added("users", this.userId, {
+      pseudoUsage: usage,
+      pseudoReferralBonus: referralBonus,
+    });
   }
   this.ready();
 });
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -587,10 +587,17 @@ var calculateReferralBonus = function(accountId, plan) {
               maxPaidStorageBonus)};
   } else {
     var maxFreeStorageBonus = 2 * 1e9;
-    return {grains: successfulReferralsCount * Infinity,
-            storage: Math.min(
-              successfulReferralsCount * 50 * 1e6,
-              maxFreeStorageBonus)};
+    var bonus = {
+      storage: Math.min(
+        successfulReferralsCount * 50 * 1e6,
+        maxFreeStorageBonus)
+    };
+    if (successfulReferralsCount >= 0) {
+      bonus.grains = Infinity;
+    } else {
+      bonus.grains = 0;
+    }
+    return bonus;
   }
 }
 
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -57,7 +57,7 @@ if (Meteor.isServer && process.env.LOG_MONGO_QUERIES) {
 //                an identicon will be used.
 //       pronoun: One of "male", "female", "neutral", or "robot".
 //   unverifiedEmail: If present, a string containing an email address specified by the user.
-//   referredBy: The Account that referred this Identity.
+//   referredBy: ID of the Account that referred this Identity.
 //
 // Account users additionally contain the following fields:
 //   loginIdentities: Array of identity objects, each of which may include the following fields.
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -640,6 +640,7 @@ isUserExcessivelyOverQuota = function (user) {
 
   var quota = getUserQuota(user);
 
+  // quota.grains = Infinity means unlimited grains. IEEE754 defines Infinity == Infinity.
   if (quota.grains < Infinity) {
     var count = Grains.find({userId: user._id}, {fields: {}, limit: quota.grains * 2}).count();
     if (count >= quota.grains * 2) return "outOfGrains";
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -84,9 +84,9 @@ if (Meteor.isServer && process.env.LOG_MONGO_QUERIES) {
 //   dailySentMailCount: Number of emails sent by this user today; used to limit spam.
 //   referralCompletedBy: The Account that referred this Account. If this is set, we
 //                        stop writing new referredBy values onto Identities for this account.
-//   referralCompletedOn: The Date at which the completed referral occurred.
-//   identitiesReferred: List of Identity IDs that this Account has referred. This is used for
-//                       reliably determining which Identity's names are safe to display.
+//   referredCompleteDate: The Date at which the completed referral occurred.
+//   referredIdentityIds: List of Identity IDs that this Account has referred. This is used for
+//                        reliably determining which Identity's names are safe to display.
 //   stashedOldUser: A complete copy of this user from before the accounts/identities migration.
 //                   TODO(cleanup): Delete this field once we're sure it's safe to do so.
 
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -57,6 +57,7 @@ if (Meteor.isServer && process.env.LOG_MONGO_QUERIES) {
 //                an identicon will be used.
 //       pronoun: One of "male", "female", "neutral", or "robot".
 //   unverifiedEmail: If present, a string containing an email address specified by the user.
+//   referredBy: The Account that referred this Identity.
 //
 // Account users additionally contain the following fields:
 //   loginIdentities: Array of identity objects, each of which may include the following fields.
@@ -81,6 +82,11 @@ if (Meteor.isServer && process.env.LOG_MONGO_QUERIES) {
 //   appDemoId: If this is an appdemo user (see above), the app ID they started out demoing.
 //   payments: Object defined by payments module, if loaded.
 //   dailySentMailCount: Number of emails sent by this user today; used to limit spam.
+//   referralCompletedBy: The Account that referred this Account. If this is set, we
+//                        stop writing new referredBy values onto Identities for this account.
+//   referralCompletedOn: The Date at which the completed referral occurred.
+//   identitiesReferred: List of Identity IDs that this Account has referred. This is used for
+//                       reliably determining which Identity's names are safe to display.
 //   stashedOldUser: A complete copy of this user from before the accounts/identities migration.
 //                   TODO(cleanup): Delete this field once we're sure it's safe to do so.
 
@@ -560,6 +566,52 @@ isSignedUpOrDemo = function () {
   return false;
 }
 
+var calculateReferralBonus = function(accountId, plan) {
+  // This function returns an object of the form:
+  //
+  // - {grains: 0, storage: 0}
+  //
+  // which are extra resources this account gets as part of participating in the referral
+  // program. (Storage is measured in bytes, as usual for plans.)
+
+
+  // Authorization note: Only call this if accountId is the current user!
+  var isPaid = (plan && plan !== "free");
+
+  successfulReferralsCount = countReferrals(accountId);
+  if (isPaid) {
+    var maxPaidStorageBonus = 30 * 1e9;
+    return {grains: 0,
+            storage: Math.min(
+              successfulReferralsCount * 2 * 1e9,
+              maxPaidStorageBonus)};
+  } else {
+    var maxFreeStorageBonus = 2 * 1e9;
+    return {grains: successfulReferralsCount * Infinity,
+            storage: Math.min(
+              successfulReferralsCount * 50 * 1e6,
+              maxFreeStorageBonus)};
+  }
+}
+
+var countReferrals = function (accountId) {
+  var referredIdentityIds = Meteor.users.findOne(
+    {_id: accountId},
+    {fields: {referredIdentityIds: 1}}).referredIdentityIds;
+  return (referredIdentityIds && referredIdentityIds.length || 0);
+}
+
+getUserQuota = function (user) {
+  // Re-fetch the user, since typically Meteor gives is an object with just an _id.
+  var user = Meteor.users.findOne({_id: user._id});
+  var plan = Plans.findOne(user.plan || "free");
+  var referralBonus = calculateReferralBonus(user._id, plan);
+  var userQuota = {};
+  userQuota.storage = (plan.storage + referralBonus.storage);
+  userQuota.grains = (plan.grains + referralBonus.grains);
+  return userQuota;
+}
+
 isUserOverQuota = function (user) {
   // Return false if user has quota space remaining, true if it is full. When this returns true,
   // we will not allow the user to create new grains, though they may be able to open existing ones
@@ -569,8 +621,7 @@ isUserOverQuota = function (user) {
 
   if (!Meteor.settings.public.quotaEnabled || user.isAdmin) return false;
 
-  var plan = Plans.findOne(user.plan || "free");
-
+  var plan = getUserQuota(user);
   if (plan.grains < Infinity) {
     var count = Grains.find({userId: user._id}, {fields: {}, limit: plan.grains}).count();
     if (count >= plan.grains) return "outOfGrains";
@@ -587,14 +638,14 @@ isUserExcessivelyOverQuota = function (user) {
 
   if (!Meteor.settings.public.quotaEnabled || user.isAdmin) return false;
 
-  var plan = Plans.findOne(user.plan || "free");
+  var quota = getUserQuota(user);
 
-  if (plan.grains < Infinity) {
-    var count = Grains.find({userId: user._id}, {fields: {}, limit: plan.grains * 2}).count();
-    if (count >= plan.grains * 2) return "outOfGrains";
+  if (quota.grains < Infinity) {
+    var count = Grains.find({userId: user._id}, {fields: {}, limit: quota.grains * 2}).count();
+    if (count >= quota.grains * 2) return "outOfGrains";
   }
 
-  return plan && user.storageUsage && user.storageUsage >= plan.storage * 1.2 && "outOfStorage";
+  return quota && user.storageUsage && user.storageUsage >= quota.storage * 1.2 && "outOfStorage";
 }
 
 isAdmin = function() {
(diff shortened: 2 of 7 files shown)


shell/shared/shell.js
```diff
@@ -154,6 +166,49 @@ if (Meteor.isServer) {
...
+       "profile.name": {$exists: true}},
+      {fields: {
+        "_id": 1,
+        "referredBy": 1,
+        "profile.name": 1}});
+    return notCompletedReferralIdentities;
+  });
+
+  Meteor.publish("myReferrals", function() {
+    if (!this.userId) {
```
paulproteus: Question for @jparyani - is this more idiomatic as a `check(this.userId, String)`?

shell/packages/sandstorm-db/db.js
```diff
@@ -843,6 +894,30 @@ _.extend(SandstormDb.prototype, {
     return user && Plans.findOne(user.plan || "free");
   },
 
+  getMyReferralBonus: function(user) {
```
jparyani: This function taking a user parameter feels weird to me. The name of the function is get**My**ReferralBonus, which implies that it only works for the current user. Also, the logic to  ignore the parameter client-side seems messy and inconsistent to me.

shell/shared/shell.js
jparyani: I think it's better the way you have it. Typically checks are used on user input, whereas here you really are only testing for falseyness of `this.userId`.

shell/packages/sandstorm-db/db.js
paulproteus: FWIW I'm doing my best to copy the logic of `getMyUsage`! Maybe we should discuss this bit IRL, but I can also take detailed suggestions here.

shell/packages/sandstorm-accounts-ui/accounts-ui-server.js
```diff
@@ -25,8 +25,13 @@ Meteor.users.deny({
 Meteor.publish("getMyUsage", function () {
   if (this.userId) {
     // TODO(someday): Make this reactive.
-    var usage = this.connection.sandstormDb.getMyUsage(Meteor.users.findOne(this.userId));
-    this.added("users", this.userId, {pseudoUsage: usage});
+    var user = Meteor.users.findOne(this.userId);
+    var usage = this.connection.sandstormDb.getMyUsage(user);
+    var referralBonus = this.connection.sandstormDb.getMyReferralBonus(user);
```
jparyani: Small sylistic thing, consider doing `var db = this.connection.sandstormDb` to save on repetition. I've started doing that at the top of all my meteor methods, even if I only use the db once.

shell/packages/sandstorm-db/db.js
```diff
@@ -587,14 +638,14 @@ isUserExcessivelyOverQuota = function (user) {
 
   if (!Meteor.settings.public.quotaEnabled || user.isAdmin) return false;
 
-  var plan = Plans.findOne(user.plan || "free");
+  var quota = getUserQuota(user);
 
-  if (plan.grains < Infinity) {
-    var count = Grains.find({userId: user._id}, {fields: {}, limit: plan.grains * 2}).count();
-    if (count >= plan.grains * 2) return "outOfGrains";
+  if (quota.grains < Infinity) {
```
jparyani: This was slightly confusing to me at first, since I'm not all that familiar with the rules for comparing numbers to null/undefined. It might be better to make your check explicitly be `quota.grains !== undefined`.

shell/packages/sandstorm-accounts-ui/accounts-ui-server.js
paulproteus: +1, will fix

shell/packages/sandstorm-db/db.js
paulproteus: +1, will push a commit that clarifies this

paulproteus:
I still have this test failure to fix.

```
S: tinytest - permissions: parentToken : !!!!!!!!! FAIL !!!!!!!!!!!
Error: Meteor.userId can only be invoked in method calls. Use this.userId in publish functions.
    at AccountsServer.userId (accounts_server.js:80:13)
    at AccountsServer.user (accounts_common.js:53:23)
    at Object.Meteor.user (accounts_common.js:230:19)
    at Object.SandstormPermissions.createNewApiToken (packages/sandstorm-permissions/permissions.js:489:1)
    at packages/local-test_sandstorm-permissions/permissions-tests.js:125:1
    at [object Object].func (packages/tinytest/packages/tinytest.js:644:1)
    at packages/tinytest/packages/tinytest.js:414:1
    at [object Object]._.extend.withValue (packages/meteor/packages/meteor.js:1013:1)
    at packages/meteor/packages/meteor.js:407:1
    at runWithEnvironment (packages/meteor/packages/meteor.js:1067:1)
```

shell/packages/sandstorm-db/db.js
paulproteus: Well, will add a comment.

shell/server/proxy.js
```diff
@@ -257,8 +341,22 @@ Meteor.methods({
       if (identityId != apiToken.identityId && identityId != grain.identityId &&
           !ApiTokens.findOne({'owner.user.identityId': identityId, parentToken: hashedToken })) {
         var owner = {user: {identityId: identityId, title: title}};
-        SandstormPermissions.createNewApiToken(globalDb, {rawParentToken: token}, apiToken.grainId,
-                                               apiToken.petname, {allAccess: null}, owner);
+
+        // Create a new API token for the identity redeeming this token.
+        var result = SandstormPermissions.createNewApiToken(
```
paulproteus: @dwrensha this hopefully avoids the need to modify the data model for ApiTokens.

paulproteus:
Ready for your review & merge, @kentonv.

shell/packages/sandstorm-db/db.js
```diff
@@ -560,6 +566,52 @@ isSignedUpOrDemo = function () {
...
+  successfulReferralsCount = countReferrals(accountId);
+  if (isPaid) {
+    var maxPaidStorageBonus = 30 * 1e9;
+    return {grains: 0,
+            storage: Math.min(
+              successfulReferralsCount * 2 * 1e9,
+              maxPaidStorageBonus)};
+  } else {
+    var maxFreeStorageBonus = 2 * 1e9;
+    return {grains: successfulReferralsCount * Infinity,
```
kentonv: Zero times Infinity is NaN. I'm guessing you actually wanted `grains` to come out zero when there are no successful referrals. You'll need to use a ternary operator for that.

shell/packages/sandstorm-db/db.js
```diff
@@ -560,6 +566,52 @@ isSignedUpOrDemo = function () {
...
+
+var countReferrals = function (accountId) {
+  var referredIdentityIds = Meteor.users.findOne(
+    {_id: accountId},
+    {fields: {referredIdentityIds: 1}}).referredIdentityIds;
+  return (referredIdentityIds && referredIdentityIds.length || 0);
+}
+
+getUserQuota = function (user) {
+  // Re-fetch the user, since typically Meteor gives is an object with just an _id.
```
kentonv: Eh? I've never known that to be true.

`Meteor.user()` on the client side will only return published fields, but on the server side it should return the whole document.

shell/server/proxy.js
```diff
@@ -36,6 +36,85 @@ SANDSTORM_ALTHOME = Meteor.settings && Meteor.settings.home;
...
+  // Bail out early if quota enforcement is disabled.
+  if (! Meteor.settings.public.quotaEnabled) {
+    return;
+  }
+
+  var aliceAccountId = apiTokenAccountId;
+  var bobAccountId = accountId;
+  var bobIdentityId = identityId;
+
+  if (Meteor.users.find({_id: bobAccountId,
```
kentonv: This find() seems redundant since the update() below already guards against updating a user who already has `referredBy` set.

shell/server/proxy.js
```diff
@@ -36,6 +36,85 @@ SANDSTORM_ALTHOME = Meteor.settings && Meteor.settings.home;
...
+                         referredByComplete: {$exists: true}}).count() > 0) {
+    return;
+  }
+
+  // Only actually update Bob's Identity ID if there is no referredBy.
+  var updatedCount = Meteor.users.update(
```
kentonv: You store `updatedCount` to a variable but don't use it. Did you mean to?

shell/packages/sandstorm-db/db.js
```diff
@@ -560,6 +566,52 @@ isSignedUpOrDemo = function () {
...
+  // which are extra resources this account gets as part of participating in the referral
+  // program. (Storage is measured in bytes, as usual for plans.)
+
+
+  // Authorization note: Only call this if accountId is the current user!
+  var isPaid = (plan && plan !== "free");
+
+  successfulReferralsCount = countReferrals(accountId);
+  if (isPaid) {
+    var maxPaidStorageBonus = 30 * 1e9;
```
kentonv: This logic is very specific to Oasis, but quota is likely to be used much more broadly than that. This is e.g. why we have the `Plans` collection rather than hard-coding them. Would it make sense to store the referral parameters in `Plans`? For each plan type, you could store the max bonus and the per-referral bonus as parameters. If they aren't set, perhaps the referral program should turn itself off?

shell/server/proxy.js
kentonv: Oh, I misread -- this one is querying the account, while the one below is querying the identity.

That said, I think I would argue that after a referral is complete, you should leave `referredBy` set on the identity, and then in this code here you can do just one query rather than two. It feels easier to reason about that way, IMO, and probably saves queries / database fragmentation overall.

kentonv: > probably saves queries / database fragmentation overall.

(Admittedly, the difference is very small.)

shell/server/proxy.js
```diff
@@ -36,6 +36,85 @@ SANDSTORM_ALTHOME = Meteor.settings && Meteor.settings.home;
...
+  var now = new Date();
+  Meteor.users.update({_id: bobAccountId, referredByComplete: {$exists: false}},
+                      {$set: {referredByComplete: bobIdentity.referredBy,
+                              referredCompleteDate: now}});
+
+  // Update Alice's Account.referredIdentityIds.
+  Meteor.users.update({_id: aliceAccountId},
+                      {$push: {referredIdentityIds: bobIdentity._id}});
+
+  // Remove now-redundant Bob identity referredBy.
```
kentonv: If I understand correctly, there's no particular need to remove this, so as mentioned above I'd leave it in and avoid unnecessary rewrites of the userdoc, plus you can save a query above.

shell/server/proxy.js
```diff
@@ -257,8 +341,22 @@ Meteor.methods({
...
+
+        // If the parent API token is forSharing and it has an accountId, then the logged-in user (call
+        // them Bob) is about to access a grain owned by someone (call them Alice) and save a reference
+        // to it as a new ApiToken. (For share-by-link, this occurs when viewing the grain. For
+        // share-by-identity, this happens immediately.)
+        if (result.parentApiToken) {
+          var parentApiToken = result.parentApiToken;
+          if (parentApiToken.forSharing && parentApiToken.accountId) {
+            storeReferralProgramInfoApiTokenCreated(
+              globalDb, this.userId, owner.user.identityId, parentApiToken.accountId);
```
kentonv: I think `owner.user.identityId` is the same thing as `identityId`, right?

shell/server/proxy.js
kentonv: OK, I see later on that maybe leaving the referredBy fields in would be annoying when you want to display the UI showing incomplete referrals.

I sort of feel like it would be more robust to add an additional field marking the referral complete, or maybe replace `referredBy` with a sentinel value like `"completed"`, in order to feel comfortable that it won't accidentally be set again later.

Hmm, there may be a real problem, or at least weirdness, here in the case where Bob's identity is connected to multiple accounts: Alice could end up seeing Bob as an incomplete referral even though Bob's main account has already completed referral, because somehow Bob's secondary account accepted a sharing link from Alice and therefore `referredBy` got set again on Bob's identity. Blocking `referredBy` from possibly being set again would stop this.

In fact, it seems like once Bob's account has been successfully referred, *all* of Bob's identities should be marked such that they cannot be referred again, otherwise Alice could get stuck seeing Bob's secondary identity as an incomplete referral even though it is impossible for Bob to complete it because he has already completed a different referral. I guess even if you do set `referredBy` on all of his identities, that doesn't cover new identities added later... arg...

shell/shared/shell.js
```diff
@@ -154,6 +166,49 @@ if (Meteor.isServer) {
...
+    // identity IDs, and make sure the client has that list of identity IDs. Separately we make sure
+    // the names names of the identities are sent, so the client can show useful information.
+    var completedIdentityIdsCursor = Meteor.users.find(
+      {_id: this.userId},
+      {fields: {"referredIdentityIds": 1}});
+    return completedIdentityIdsCursor;
+  });
+
+  Meteor.publish("myReferralsNames", function() {
+    var user = Meteor.users.findOne({_id: this.userId},
```
kentonv: Check that `this.userId` is non-null before passing to query.

shell/shared/shell.js
```diff
@@ -154,6 +166,49 @@ if (Meteor.isServer) {
...
+                                    {fields: {"referredIdentityIds": 1}});
+    if (user && user.referredIdentityIds) {
+      // Completed referral IDs are not reactive at the moment.
+      return Meteor.users.find(
```
kentonv: I think that here you're revealing all of Bob's identities to Alice, not just the specific identity she referred. This could be bad for some of the user cases we had in mind for multiple identities. Perhaps this argues for keeping `referredBy` set on the identity, still referring to Alice (not a sentinel value), so that you can query for the exact identity that Alice actually referred.

shell/shared/shell.js
```diff
@@ -935,6 +1008,16 @@ Router.map(function () {
     }
   });
 
+  this.route("referrals", {
+    path: "/referrals",
+
+    waitOn: function() {
+      return [Meteor.subscribe("myReferrals"),
```
kentonv: Note that a publish can actually return multiple cursors. So in theory you could combine these three subscribes into one big subscribe. However, last I knew Meteor had a known bug that it couldn't support a single publish returning multiple cursors from the same collection -- and all your cursors are from the users collection. Does that restriction still exist? I believe if you try it it will give a clear error if so.

Hmm, another thing to consider here is whether you really ought to be subscribing to a pseudo-collection of some sort. It's mildly disconcerting to me that we're returning cursors into the users table, matching users other than the caller. You've been careful to select only safe fields, but it still makes me worry since this is the most sensitive table in the database.

Taking into consideration all the comments above, perhaps it would actually make a lot of sense to have a single publish which publishes a pseudo-collection which is populated by the identities Alice has referred and a boolean "completed" flag for each one. On the server side, one query is performed to find all identities with `referredBy` equals Alice (and we stop removing `referredBy` from identities on completion, so this matches completed referrals too), and then individual queries are done (still on the server side) to each account to see if it is completed, in order to fill in the boolean fields.

shell/packages/sandstorm-db/db.js
paulproteus: Fixed in a follow-up commit.

shell/packages/sandstorm-db/db.js
paulproteus: Fair enough. I guess I'm a bit freaked-out by the fact that this function is available (as far as I can tell?) on both the server & client. I might not know the full server vs. client rules, though. And certainly querying against `Meteor.users` on the client will always get exactly what is published, so re-fetching against that is meaningless on the client.

Fixed in a follow-up commit.

shell/server/proxy.js
paulproteus: FWIW in a more heavily `console.log`-y version of this code, I used to log `updatedCount`. Agreed that the variable is superflous; fixed in a follow-up commit.

shell/packages/sandstorm-db/db.js
paulproteus: In the future, I can imagine we might tweak the referral program. Having the code depend on data in the database is error-prone and I can't say I love that pattern.

For now, I'd call this referral program experimental. I am already imagining how we might change it:

- Instead of "invite someone to a grain" being the thing that triggers the start of a referral program, we could do "suggest an app to someone and have them start a private grain of it".

So I'd rather keep this in the code and not move this into the database, but if you feel it's essential, I can do that.

paulproteus: Also `Plans` would be a little weird to me since that is ForeignKey'd to from `Users`, but that's not a showstopper.

shell/shared/shell.js
paulproteus: Yeah, a publish can't return >1 cursor into the same collection, which is why I made this multiple subscriptions.

I'll work on making it a pseudo-collection. I've never made a pseudo-collection before!

shell/packages/sandstorm-db/db.js
kentonv: Hmm... but once we launch this logic we must honor any referral bonuses awarded, so these parameters can't just go away, right?

paulproteus: That's true; they can't totally go away. But if we want to remove old details like this from the code, we could move these facts into a migration, compute the referral bonus the user has earned, and store that in some new parameter. We'll need a user property something akin to that for the Indiegogo bonuses anyway.

shell/packages/sandstorm-db/db.js
```diff
@@ -57,6 +57,7 @@ if (Meteor.isServer && process.env.LOG_MONGO_QUERIES) {
 //                an identicon will be used.
 //       pronoun: One of "male", "female", "neutral", or "robot".
 //   unverifiedEmail: If present, a string containing an email address specified by the user.
+//   referredBy: The Account that referred this Identity.
```
dwrensha: s/The Account/The ID of the account/ ?

shell/packages/sandstorm-db/db.js
```diff
@@ -81,6 +82,11 @@ if (Meteor.isServer && process.env.LOG_MONGO_QUERIES) {
 //   appDemoId: If this is an appdemo user (see above), the app ID they started out demoing.
 //   payments: Object defined by payments module, if loaded.
 //   dailySentMailCount: Number of emails sent by this user today; used to limit spam.
+//   referralCompletedBy: The Account that referred this Account. If this is set, we
```
dwrensha: "ID of the Account that referred this Account"?

shell/packages/sandstorm-db/db.js
```diff
@@ -81,6 +82,11 @@ if (Meteor.isServer && process.env.LOG_MONGO_QUERIES) {
...
+//                        stop writing new referredBy values onto Identities for this account.
+//   referralCompletedOn: The Date at which the completed referral occurred.
+//   identitiesReferred: List of Identity IDs that this Account has referred. This is used for
```
dwrensha: Looks like you renamed this to "referredIdentityIds".

shell/packages/sandstorm-db/db.js
dwrensha: This is apparently now "referredByComplete"

shell/packages/sandstorm-db/db.js
```diff
@@ -81,6 +82,11 @@ if (Meteor.isServer && process.env.LOG_MONGO_QUERIES) {
 //   appDemoId: If this is an appdemo user (see above), the app ID they started out demoing.
 //   payments: Object defined by payments module, if loaded.
 //   dailySentMailCount: Number of emails sent by this user today; used to limit spam.
+//   referralCompletedBy: The Account that referred this Account. If this is set, we
+//                        stop writing new referredBy values onto Identities for this account.
+//   referralCompletedOn: The Date at which the completed referral occurred.
```
dwrensha: This is apparently now "referredCompleteDate"

shell/packages/sandstorm-db/db.js
paulproteus: +1, fixed in latest push

shell/packages/sandstorm-db/db.js
paulproteus: +1, fixed in latest push

shell/packages/sandstorm-db/db.js
paulproteus: +1, fixed in latest push

paulproteus:
@kentonv for now, I plan to keep erasing `referredBy` from identities when their half-referral gets turned into a real referral. Here's my summary of the above conversation:

Question: When a referral is complete, should we leave referredBy on the identity?

Reasons for yes:

- Slightly less code & slightly less DB workload.
- The Account does store the list of identities successfully referred, so we can use that to infer if the referral is complete or not.

Reasons for no:

- If an identity is accessible from two accounts, we can't tell on the second account that it's not half-referred.

The "no" reason seems compelling to me, so I'm going to _not_ leave referredBy on the identity.

However, there was lots of discussion above. Let's continue it here, rather than as a comment on a line, so that it doesn't become hard to find as the code in this review changes.

@kentonv I'm hungry for your +1 or -1 on my analysis above!

shell/packages/sandstorm-db/db.js
paulproteus: Fixed in latest push (I said `ID of the Account` not `The ID of the account`; I figure Account is basically as much of a proper noun as User even though it's not quite).

shell/shared/shell.js
paulproteus: +1, will fix in follow-up commit

shell/shared/shell.js
paulproteus: FWIW, I don't think this is true, but if it is, I hope you'll let me know.

I'm only getting all the `users` whose `_id` is within `referredIdentityIds`, which should be just the identity for Bob that Alice referred.

paulproteus:
@kentonv I believe this is fully reactive, addresses all your comments, and is ready for re-review/merge.

Thanks to @jparyani and @dwrensha for their help with this, too.

Also all the complex server-side join stuff I'm doing in this seems like a lot of work just to get the referrals page to be reactive, but whatever, I guess I did the work, so feel free to merge it. (-:

paulproteus:
I believe the above changes fix all the known issues. Let me know!

PR #1221, opened 2015-12-01
Add imperative sentence to excessivelyOverQuota message

paulproteus:
I think this improves clarity.


kentonv:
Note that probably 95% of the time it will be the owner seeing this message. Should we detect that case and show a slightly different message or just leave it like this?

kentonv:
Ping. I'm happy with "yeah that's fine, let's go with this" -- just wanted to make sure you are aware.

paulproteus:
Thanks for the ping!

I'm fine with this string, let's go with this.

(Rationale: the owner will be able to figure out what that means, and I don't feel like writing the conditional logic. Semi-lazy but I think a small improvement is OK for now.)

PR #1222, opened 2015-12-02
Implement sidebar shrink/expand

### LINKED ISSUE #872: Proposal for new sidebar design
The 3rd sidebar is my choice for the expanded version. The 4th sidebar is a compact version which displays by clicking the arrow.

shelldoc-sidebar-toggle1

kentonv:
Fixes #872.

### DIFF
--- a/shell/client/_topbar.scss
+++ b/shell/client/_topbar.scss
@@ -459,9 +459,38 @@ body>.topbar {
 
   >.navbar {
     @media #{$desktop} {
+      >button.navbar-shrink {
+        @extend %unstyled-button;
+        display: block;
+        width: 100%;
+        height: $navbar-shrink-button-height;
+        border: 0px solid #303030;
+        border-left-width: ($navbar-width-desktop - $navbar-width-desktop-shrunk);
+        background-color: #606060;
+        color: #ddd;
+        text-align: center;
+        line-height: $navbar-shrink-button-height;
+        padding: 0;
+        outline: none;
+
+        transition: border-left-width 0.2s;
+      }
+
       &.hide-desktop {
         // !important used to override other classes that might assign e.g. display: block
-        display: none !important;
+        width: $navbar-width-desktop-shrunk !important;
+
+        >button.navbar-shrink {
+          border-left-width: 0;
+        }
+      }
+
+      overflow: hidden;
+      transition: width 0.2s;
+    }
+    @media #{$mobile} {
+      >button.navbar-shrink {
+        display: none;
       }
     }
     @media #{$desktop} {
@@ -485,7 +514,7 @@ body>.topbar {
           display: block;
           box-sizing: border-box;
           min-height: 32px;
-          padding-left: 32px;
+          padding-left: $navbar-width-desktop-shrunk;
           text-decoration: none;
           text-overflow: ellipsis;
           overflow: hidden;
@@ -496,7 +525,7 @@ body>.topbar {
         }
         &.navitem-open-grain, &.navitem-create-grain {
           background-repeat: no-repeat;
-          background-position: 4px center;
+          background-position: (($navbar-width-desktop-shrunk - 24px) / 2) center;
           background-size: 24px 24px;
           font-size: 150%;
           line-height: 48px;
@@ -517,6 +546,7 @@ body>.topbar {
           &.current, &.current:hover {
             background-image: url('/open-grain-m.svg');
           }
+          border-bottom: none;
         }
         &.navitem-create-grain {
           background-image: url('/apps.svg');
@@ -558,7 +588,7 @@ body>.topbar {
       margin: 0;
       padding: 0;
       position: absolute;
-      top: 96px;
+      top: (96px + $navbar-shrink-button-height);
       bottom: 0px;
       left: 0px;
       width: 200px;
@@ -567,6 +597,10 @@ body>.topbar {
       >li.navitem-grain {
         a {
           padding-right: 32px;
+
+          // Stack above .app-icon so the icon doesn't block clicks.
+          position: relative;
+          z-index: 1;
         }
         background-color: $topbar-background-color;
         color: $topbar-foreground-color;
@@ -581,11 +615,10 @@ body>.topbar {
         .app-icon {
           @extend %pseudo-img-tag;
           position: absolute;
-          top: 2px;
-          left: 2px;
+          top: 4px;
+          left: (($navbar-width-desktop-shrunk - 24px) / 2);
           width: 24px;
           height: 24px;
-          margin: 2px;
         }
         .close-button {
           @extend %unstyled-button;
@@ -603,6 +636,7 @@ body>.topbar {
           &:hover {
             border: 1px solid $topbar-foreground-color-hover;
           }
+          z-index: 2;  // Stack above <a>.
         }
       }
     }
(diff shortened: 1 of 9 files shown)


kentonv:
I just realized that this causes the sidebar to be visible to non-logged-in users. Working on another commit to fix that...

kentonv:
Fixed.

shell/packages/sandstorm-ui-topbar/topbar.html
```diff
@@ -40,9 +40,15 @@
         {{/unless}}
       {{/each}}
     </ul>
-    <ul class="navbar {{#unless showNavbar}}hide-desktop{{/unless}}">
+    <ul class="navbar {{#if hideNavbar}}hide-desktop{{else}}{{#if shrinkNavbar}}shrink-desktop{{/if}}{{/if}}">
       <li class="navitem-create-grain {{#if isCurrentRoute 'apps'}}current{{/if}}"><a href="{{pathFor route='apps'}}">Apps</a></li>
       <li class="navitem-open-grain {{#if isCurrentRoute 'grains'}}current{{/if}}"><a href="{{pathFor route='grains'}}"><span>Grains</span></a></li>
+
+      <!-- TODO(cleanup): ul's can only contain li's, so stuff below here is not really valid -->
+      <button class="navbar-shrink">
```
zarvox: Any reason you can't wrap this in an appropriately-sized `<li>`?

shell/packages/sandstorm-ui-topbar/topbar.html
```diff
@@ -40,9 +40,15 @@
...
+        {{#if shrinkNavbar}}&gt;{{else}}&lt;{{/if}}
```
zarvox: I'd prefer if the < and > were part of a background-image and the text was something like "collapse sidebar button" and "expand sidebar button" for our screenreader users

If you don't want to put the text in the button itself, consider adding a title attribute.

zarvox:
There's some weirdness in behavior for keyboard users - when I tab through the focusable elements and reach the "close grain" button, the browser ensures that button is visible, even though I can't reach it through direct click actions.

The sides of the focus ring get cut off when using the keyboard with , since we're not actually adjusting the size of the elements that are getting "collapsed".  Prefer making the clickable things actually shrink.

The focus ring is not visible at all when the navbar-collapse button is focused.  It should be.

zarvox:
The account interstitial seems to also need to have its size adjusted when the sidebar collapses:
account1

zarvox:
The "Apps" link also appears to have a completely-hidden focus ring now too.

zarvox:
This is more of a @neynah question, but the additional padding on both sides for the icons makes the grain link look unbalanced when the sidebar is expanded, which is the default state, which is a little sad.  It also reduces the number of pixels available for the grain title in the sidebar.

That said, I think our click targets should ideally be on the order of 48px per side, especially when they're tightly packed, so I don't really want to make the collapsed sidebar click targets any narrower.  While theoretically, if your browser window is maximized, targets on the sides are infinitely wide, many users probably won't have a maximized browser window.

Any ideas on how to make more efficient use of the space?  Maybe we could have the icons and text slide over a couple pixels to the right when the navbar is collapsed, and keep the current look and space usage when it's expanded?

shell/packages/sandstorm-ui-topbar/topbar.html
kentonv: ... Apparently not. Not sure why I didn't. Fixed.

(I think I was distracted by the non-wrapped `<ul>` below it, or something.)

shell/packages/sandstorm-ui-topbar/topbar.html
kentonv: I agree but I need asserts for that. I've added a title and inserted a TODO.

kentonv:
* The close buttons are now actually `display: none` when collapsed, so they don't become tab stops.
* I managed to avoid `overflow: hidden`, so now this change does not hide any focus rings that were visible before. However, I personally find the focus rings extremely difficult to see as-is. It's 1px wide, low contrast, and on many elements doesn't even "ring" the elements but rather covers only two or three sides. I think we need something better here. I think a slight change to background color (like "hover" usually gets) would make a lot of sense. Would like to get @neynah's input on this. For the moment I haven't added a focus indicator on the collapse button but I agree there should be one.
* Regarding padding on the icons: The lack of padding before actually made me a bit uncomfortable. There's something disconcerting about icons or text that push right up against the left edge of the window. Looking back at Nena's designs I saw that padding was supposed to be there, and it made this change easier, so I added it. Happy to implement whatever Nena prefers, though.

shell/client/_topbar.scss
```diff
@@ -459,9 +436,43 @@ body>.topbar {
...
+        width: 100%;
+        height: $navbar-shrink-button-height;
+        border: 0px solid #303030;
+        border-left-width: ($navbar-width-desktop - $navbar-width-desktop-shrunk);
+        background-color: #606060;
+        color: #ddd;
+        text-align: center;
+        line-height: $navbar-shrink-button-height;
+        padding: 0;
+        outline: none;
```
zarvox: If you remove this `outline: none`, you'll have a focus ring like everything else interactive.

zarvox:
Getting there!

+1 on "have better focus indicators than the existing focus ring."  I'm given to understand that box shadows may work well for this.

Theoretically, the purple focus ring should be visible for all `<a>`, `<button>`, `<input>`, and `<textarea>` elements, per `client/_focus.scss` - you appear to be explicitly removing it from the collapse button?

I now have a scrollbar from this `<a>` when the sidebar is collapsed, which I think isn't quite what you had in mind:

account1

shell/client/_topbar.scss
kentonv: Hmm, I forgot I did that. I guess I probably added it because I personally find it really obnoxious when these outlines appear around a button when I click on it. It's not particularly rational but it triggers my OCD senses -- like leaving a room with the light still on. I find I'm compelled to click somewhere neutral to make the outline go away.

I've removed this line for now, but I hope we can find a solution that doesn't involve outlines or box shadows. Or maybe we can find a solution where clicking on a thing doesn't cause its focus indicator to turn on, while keyboarding to it does.

kentonv:
Heh, I had noticed that the `<a>` was extending outside of the navbar but I hadn't noticed that this caused a scrollbar at the bottom. Whoops.

Fixed now.

zarvox:
Ahhh, you rebased.  And added something about demo mode.  Yay.

kentonv:
Rebased and dealt with the new demo expiration timer.

zarvox:
Looks good.  Thanks for iterating with me.  Merge when tests pass.  (I think they failed because build is out of disk space - paging @jparyani ? you can probably delete deps/glibc, which probably got created from one of my PRs-under-test.)

Future thoughts: I'm mildly sad about all these `!important` running around, which tend to be a CSS code smell.  I'm hopeful that we can find a way to reduce the number of these floating around, rather than sprinkling it everywhere styles care about their parents' styles.  I thought selector specificity was supposed to keep us from needing `!important`?  Hrm.

kentonv:
@zarvox - I have honestly never used `!important` before this change, but I saw some already there, found myself needing to add some others nearby, and then sort of started sticking them on everything in that block by default. I suspect we could get rid of them with some refactoring.

kentonv:
Garply, retest this please.

PR #1229, opened 2015-12-03
Split /account route into /account/profile and /account/billing.

dwrensha:
This adds a "Profile" and a "Billing" tab to the /account page when payments are enabled. The design follows the pattern of the tabs in our /admin page.

### DIFF
--- a/shell/client/_account.scss
+++ b/shell/client/_account.scss
@@ -387,6 +387,29 @@ body>.main-content>.account {
     padding: 0;
   }
 
+  ul.account-tabs {
+    position: absolute;
+    top: 5px;
+    left: 300px;
+    font-size: 18px;
+    >li {
+      float: left;
+      display: block;
+      >a {
+        padding: 10px;
+        border-radius: 4px;
+        &:hover {
+          background-color: lighten($sandstorm-purple-color, 40%);
+        }
+      }
+      &.active >a {
+        color: white;
+        background-color: $sandstorm-purple-color;
+      }
+
+    }
+  }
+
   .identities-editor {
     position: relative;
     white-space: nowrap;
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -608,15 +608,32 @@ limitations under the License.
 
 <template name="account">
   <div class="account">
-    {{#if .}}
-      {{>sandstormAccountSettings .}}
+    <h2>My Account</h2>
+    {{#if isPaymentsEnabled}}
+    <ul class="account-tabs">
+      <li class="{{#if profileActive}}active{{/if}}">
+        {{#linkTo route="accountProfile"}}Profile{{/linkTo}}
+      </li>
+      <li class="{{#if billingActive}}active{{/if}}">
+        {{#linkTo route="accountBilling"}}Billing{{/linkTo}}
+      </li>
+    </ul>
     {{/if}}
+    {{> Template.dynamic template=accountTab}}
   </div>
 </template>
 
-<template name="accountUsage">
-  {{> billingUsage db=globalDb}}
-  {{> billingSettings db=globalDb}}
+<template name="accountProfile">
+  {{#if .}}
+    {{>sandstormAccountSettings .}}
+  {{/if}}
+</template>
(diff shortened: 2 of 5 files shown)


paulproteus:
This will probably require me to refresh my referral program pull request.
I am OK with that! Glad this is happening.

kentonv:
Hmm, the admin page is the one page that Nena hasn't restyled, and I think it definitely could use a facelift. But I'll accept this for now and we can improve styling later.

kentonv:
I'm reverting this:
- The "change plan" button is broken.
- The "add a new card" button is broken.
- Styling of the payment methods section was lost.
- I'm not really sure it makes sense to separate the account settings into tabs. The fact that there are tabs seems fairly easy to miss (at least as currently designed), whereas scrolling is more obvious. Meanwhile the billing page has so little data on it that it honestly looks kind of weird. I think we should instead aim to redesign the "payment methods" section to have a similar form factor to the "usage" section, so that they can sit next to each other, taking up really not that much space.

kentonv:
I'll work on a stopgap solution here so that we can release soon.

kentonv:
Doh, apparently the payment methods table and "add a new card" button were broken before this change.

kentonv:
Hmm, it occurs to me that you must have some changes sitting in the Blackrock repo that were supposed to go along with you changes to the accounts page, but never got sent as a pull request.

I've managed to fix things up at my end to look like this, which I'm pretty happy with:

screenshot from 2015-12-02 23-03-34

Probably these changes conflict with any changes you might have made under blackrock, though.

I guess we'll need to see what Nena thinks.

zarvox:
The Billing section looks pretty good, but I think we can do better on the identities+emails section.  It feels a bit cluttered.  I think this is probably because we have like five unique vertical dividers going on here, which are close enough that they clash and intersect other elements awkwardly:

lines

I think we can reduce the number of vertical divisions.  Concrete suggestions (would love to hear @neynah's +1/-1 on these):

### Account editor
* Either move the toggle login checkbox and the "unlink this identity" button to the right so they fit in that column with the rest of the form fields, or add some horizontal rule (I remember a "danger zone" section in some mockups?) 
* Expand the input fields to have their right edge aligned with the right edge of the "Save" button.
* Expand the whole white-backgrounded area to make full use of the horizontal space on the page.

### Email notification section:
* Either a horizontal rule or more margin between the account editor and the email notifications section.  The grid clearly is intended to have a horizontal line here, but it's not very distinct.  Alternately, extend the white background of the "Primary e-mail address for..." line as far right as the right edge of the Billing section.
* Align the baseline of the email address text and the "primary" pip.  The latter appears slightly below the baseline of the former.
* Make the email section be a single column that spans the entire width of the frame (modulo padding), and place the "add new e-mail address" form below the email/isprimary/setprimary grid.
* Make the "primary" pip's background color be not the same as the default action button.  It demands more attention than I think that piece of information deserves.
* Round the "primary" pip's corners.  Most of our other rectangle things are slightly rounded.

### Billing

* Make the "CHANGE PLAN" text follow the standard elsewhere in the UI of capitalizing the first letter of the first word and no others: "Change plan".

Combined, these changes should give us a grid that has substantially fewer competing vertical lines, clear alternation between vertical and horizontal divisions, and neater alignment.

dwrensha:
Y'all must have really tall screens. :)

> Hmm, it occurs to me that you must have some changes sitting in the Blackrock repo that were supposed to go along with you changes to the accounts page, but never got sent as a pull request.

That's very charitable of you, @kentonv. There are in fact no such pending changes. The "change plan" button is broken because I forgot to pass the topbar into the `billingUsage` template. The payment methods section is broken because I never got the Stripe stuff setup on my machine, and I assumed that the controls there would work in an environment where Stripe was properly configured.

It looks like you managed to get things working. Since I don't actually have any conflicts pending in Blackrock, do you want to just go ahead with your version? I don't have a terribly strong opinion on tabbed-vs-vertical-scroll.

kentonv:
I've merged my version (earlier today).

PR #1230, opened 2015-12-03
Make demo timer much more obvious.

kentonv:
We were already planning to do this, but also a Twitter conversation with @hintjens today made it really clear to me that we have a problem here: people don't realize that the demo is time-bound, and don't see the countdown timer in the upper-right. This change makes it much bigger and eye-catching on the sidebar, with a "create account" button that opens the login menu.

@hintjens also gave feedback that:
* If we had asked him to register upfront before doing anything, he probably would have refused.
* Once he created some data he wanted it to be permanent, and to that end he would have registered if prompted.
* The demo / non-demo dichotomy is confusing and unexpected.

So far I haven't thought of a way to satisfy all three constraints: We can't let people create permanent data without registering, since we wouldn't know how to authenticate them later. So it seems like demo mode is necessary and we have to live with it being confusing, but maybe it will be less confusing after this change?

The styling in this change is my own. I assume @neynah will come up with something better when she has a chance but IMO this seems good enough to ship for now.

screenshot from 2015-12-02 16-39-56

(TODO: make visible in mobile view)

### DIFF
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -75,12 +75,13 @@ limitations under the License.
   {{/if}}
   {{#if currentUser}}
     {{>sandstormTopbarItem name="notifications" topbar=globalTopbar
-                           template="notifications" popupTemplate="notificationsPopup"}}
+                           template="notifications" popupTemplate="notificationsPopup"
+                           priority=-1}}
   {{/if}}
   {{!-- admin alerts --}}
   {{#with adminAlert}}
     {{#if adminAlertIsTooLarge}}
-      {{#sandstormTopbarItem name="admin-alert" topbar=globalTopbar}}
+      {{#sandstormTopbarItem name="admin-alert" topbar=globalTopbar priority=-2}}
         {{!-- TODO(cleaup): Awkwardly, the event handlers on Template.layout actually work
             on this topbar item solely because the whole topbar itself is embedded inside the
             "layout" template. But, we shouldn't be relying on that! Move this to its own
--- a/shell/packages/sandstorm-accounts-ui/login_buttons.html
+++ b/shell/packages/sandstorm-accounts-ui/login_buttons.html
@@ -1,7 +1,7 @@
 <template name="loginButtons">
   {{!-- Data context is an instance of AccountsUi. --}}
   {{#if isDemoUser}}
-    <a class="{{#if expiringSoon}}expiring-soon{{/if}}">Demo {{demoTimeLeft}} ▾</a>
+    <a>Create Account ▾</a>
   {{else}}
     <a>Sign in ▾</a>
   {{/if}}
@@ -57,11 +57,11 @@
 </template>
 
 <template name="_loginButtonsLoggedOutDropdown">
-  <h4>Sign in</h4>
-
   {{#if isDemoUser}}
-    <p class="demo-note">Sign in to make your demo data permanent.</p>
+    <h4>Create Account</h4>
+    <p class="demo-note">Your demo data will be transferred to your new account.</p>
   {{else}}
+    <h4>Sign in</h4>
     {{#if isCurrentRoute 'shared'}}
       <p class="demo-note">Sign in to save this link.</p>
     {{/if}}
--- a/shell/packages/sandstorm-ui-topbar/topbar.html
+++ b/shell/packages/sandstorm-ui-topbar/topbar.html
@@ -51,6 +51,17 @@ limitations under the License.
           <button class="close-button" title="Close {{title}}"></button>
         </li>
         {{/each}}
+
+        {{#with accountExpires}}
+          <li><div class="demo-notice">
+            <p>This demo expires in: <span class="countdown {{#if urgent}}urgent{{/if}}">{{countdown}}</span></p>
+
+            <p>To make your data permanent, create an account:
+              <button class="sign-in">Create account »</button></p>
+
+            <p class="self-host"><a href="https://sandstorm.io/install" target="_blank">Or, get Sandstorm on your own server (it's open source) »</a></p>
+          </div></li>
+        {{/with}}
       </ul>
     </ul>
   </ul>
--- a/shell/packages/sandstorm-ui-topbar/topbar.html
+++ b/shell/packages/sandstorm-ui-topbar/topbar.html
@@ -59,7 +59,7 @@ limitations under the License.
             <p>To make your data permanent, create an account:
               <button class="sign-in">Create Account »</button></p>
 
-            <p class="self-host"><a href="https://sandstorm.io/install" target="_blank">Or, get Sandstorm on your own server (it's open source) »</a></p>
+            <p class="self-host"><a href="https://sandstorm.io/install/" target="_blank">Or, get Sandstorm on your own server (it's open source) »</a></p>
           </div></li>
         {{/with}}
       </ul>
--- a/shell/packages/sandstorm-ui-topbar/topbar.html
+++ b/shell/packages/sandstorm-ui-topbar/topbar.html
@@ -57,7 +57,7 @@ limitations under the License.
             <p>This demo expires in: <span class="countdown {{#if urgent}}urgent{{/if}}">{{countdown}}</span></p>
 
             <p>To make your data permanent, create an account:
-              <button class="sign-in">Create account »</button></p>
+              <button class="sign-in">Create Account »</button></p>
 
             <p class="self-host"><a href="https://sandstorm.io/install" target="_blank">Or, get Sandstorm on your own server (it's open source) »</a></p>
           </div></li>
--- a/shell/client/_topbar.scss
+++ b/shell/client/_topbar.scss
@@ -606,6 +606,53 @@ body>.topbar {
         }
       }
     }
+
+    .demo-notice {
+      p {
+        color: $topbar-foreground-color;
+        margin: 8px;
+        line-height: normal;
+        font-size: 14px;
+
+        a {
+          color: white;
+          padding: 0;
+          white-space: normal;
+          &:hover {
+            text-decoration: underline;
+          }
+        }
+      }
+
+      .countdown {
+        display: block;
+        font-size: 30px;
+        font-weight: bold;
+        color: #bb0;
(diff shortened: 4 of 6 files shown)


paulproteus:
This resulting design is *vastly* more reasonable than I expected! I had
tried earlier and could not come up with anything borderline reasonable.

And thank you hugely for asking @hintjens for feedback!

shell/client/shell.html
```diff
@@ -75,12 +75,13 @@
...
     {{>sandstormTopbarItem name="notifications" topbar=globalTopbar
-                           template="notifications" popupTemplate="notificationsPopup"}}
+                           template="notifications" popupTemplate="notificationsPopup"
+                           priority=-1}}
   {{/if}}
   {{!-- admin alerts --}}
   {{#with adminAlert}}
     {{#if adminAlertIsTooLarge}}
-      {{#sandstormTopbarItem name="admin-alert" topbar=globalTopbar}}
+      {{#sandstormTopbarItem name="admin-alert" topbar=globalTopbar priority=-2}}
```
paulproteus: Until the flexbox changes in https://github.com/sandstorm-io/sandstorm/pull/1190 get merged, I don't know that the `priority=` parameter does anything. But OK!

shell/packages/sandstorm-accounts-ui/login_buttons.html
```diff
@@ -1,7 +1,7 @@
 <template name="loginButtons">
   {{!-- Data context is an instance of AccountsUi. --}}
   {{#if isDemoUser}}
-    <a class="{{#if expiringSoon}}expiring-soon{{/if}}">Demo {{demoTimeLeft}} ▾</a>
+    <a>Create Account ▾</a>
```
paulproteus: s/Create Account/Sign up for free/ ? But I can live with this as-is too. (I defer having a strong preference about the wording of this to @neynah .)

shell/packages/sandstorm-accounts-ui/login_buttons.html
```diff
@@ -57,11 +57,11 @@
 </template>
 
 <template name="_loginButtonsLoggedOutDropdown">
-  <h4>Sign in</h4>
-
   {{#if isDemoUser}}
-    <p class="demo-note">Sign in to make your demo data permanent.</p>
+    <h4>Create Account</h4>
+    <p class="demo-note">Your demo data will be transferred to your new account.</p>
```
paulproteus: +1, greatly improved clarity

shell/packages/sandstorm-accounts-ui/login_buttons.html
```diff
@@ -57,11 +57,11 @@
...
   {{else}}
+    <h4>Sign in</h4>
```
paulproteus: I would imagine "Sign in" here makes it sound like you have to already have an account to save your data.

paulproteus:
Hey @hintjens - if you saw something akin to the screenshot above, would that have helped you?

shell/packages/sandstorm-accounts-ui/login_buttons.html
paulproteus: Or

s/Create Account/Create free account/

shell/packages/sandstorm-accounts-ui/login_buttons.html
```diff
@@ -57,11 +57,11 @@
 </template>
 
 <template name="_loginButtonsLoggedOutDropdown">
-  <h4>Sign in</h4>
-
   {{#if isDemoUser}}
-    <p class="demo-note">Sign in to make your demo data permanent.</p>
+    <h4>Create Account</h4>
```
paulproteus: In the screenshot, the 'a' seems to show up in lowercase. In this code, it's uppercase.

All I really hope is that this string is consistent with what's in the top-right.

shell/packages/sandstorm-accounts-ui/login_buttons.html
kentonv: I actually have come to the realization that I vastly prefer "create account" over "sign up".

"Sign up" sounds like I'm committing to something, or paying for something. That's scary, and it's why I always resisted saying "sign up" rather than "sign in" in the past. "Create account" sounds much more passive: it's just creating a bucket to put my data in.

shell/packages/sandstorm-accounts-ui/login_buttons.html
kentonv: Right, this is the branch that displays for people who aren't demo users (and may very well have an account already).

paulproteus: But >99% of humanity doesn't have an account yet. But I am OK with leaving this as "Sign in" for now.

shell/client/shell.html
kentonv: It does! Priority determines the ordering of the topbar items in the HTML, which implies determining the ordering in the top bar. I was seeing a bug where the ordering was non-deterministic. I think this bug predated my changes. Anyway, assigning priorities fixes it.

paulproteus: Gotcha

shell/packages/sandstorm-accounts-ui/login_buttons.html
paulproteus: That's pretty sensible. +0 at least. (I'd prefer "Create free account" but we can do an experiment to fine-tune that, it's no showstopper)

shell/packages/sandstorm-ui-topbar/topbar.html
```diff
@@ -51,6 +51,17 @@
           <button class="close-button" title="Close {{title}}"></button>
         </li>
         {{/each}}
+
+        {{#with accountExpires}}
+          <li><div class="demo-notice">
+            <p>This demo expires in: <span class="countdown {{#if urgent}}urgent{{/if}}">{{countdown}}</span></p>
+
+            <p>To make your data permanent, create an account:
+              <button class="sign-in">Create account »</button></p>
```
paulproteus: Ah hah, here's the lowercase one.

It'd be cool to be consistent about "Create Account" (capital A) vs. "Create account", unless there's a reason not to.

shell/packages/sandstorm-ui-topbar/topbar.html
```diff
@@ -51,6 +51,17 @@
...
+
+            <p class="self-host"><a href="https://sandstorm.io/install" target="_blank">Or, get Sandstorm on your own server (it's open source) »</a></p>
```
paulproteus: Add a trailing `/` to avoid a redirect (not that it matters much) to https://sandstorm.io/install/

paulproteus:
Code review feedback summary for @kentonv:

- I am +1 on you shipping a dot release that makes this live immediately. The hundreds of appdemo daily users are likely going to be happy with this.

- I assume you're asking me for UI review & string review, not JS code review. The JS code looks fine but I haven't read it 100% carefully; you can ask me to do that if you want.

- If that works for you, you can self-merge; no need for me to click the button. Many many thanks for pinging me on this review.

Other less-urgent things:

- Most important: fix capitalization.

- Less important: everything else.

Questions for hintjen hypothetically:

- I'm curious if we asked hintjen -- 'If you had been asked to sign up, but there were a "Continue as guest" button, would you have clicked it and continued as guest?'

Other even less-urgent things:

- It's kind of sad there's so many calls to action on the demo home page. But I guess the guided tour might help with that. (If you want to work on that before I get a chance to, I'd be perfectly happy with that fwiw!)

shell/packages/sandstorm-accounts-ui/login_buttons.html
kentonv: I actually fixed this after taking the screenshot but before pushing. :)

kentonv: Wait I'm confused. I fixed one case of inconsistent capitalization but you found another one. Will do.

paulproteus: :D

paulproteus:
FWIW, for appdemo users, I think they might be confused about what "sandstorm" is when they clicked to try "etherpad". But I think this is livable.

shell/packages/sandstorm-ui-topbar/topbar.html
```diff
@@ -51,6 +51,17 @@
           <button class="close-button" title="Close {{title}}"></button>
         </li>
         {{/each}}
+
+        {{#with accountExpires}}
+          <li><div class="demo-notice">
+            <p>This demo expires in: <span class="countdown {{#if urgent}}urgent{{/if}}">{{countdown}}</span></p>
+
+            <p>To make your data permanent, create an account:
+              <button class="sign-in">Create Account »</button></p>
```
zarvox: Actually, standard style within Sandstorm is to capitalize the first word of the action and no others - see our other "Sign in", "Share access", "Send an invite", "Get shareable link", and so on.

shell/packages/sandstorm-accounts-ui/login_buttons.html
zarvox: https://github.com/sandstorm-io/sandstorm/wiki/Shell-Design-Guidelines from Nena says "Capitalization: - always mixed caps with the first letter capitalized"

hintjens:
Thanks for listening to me. The trick is IMO to remove synthetic concepts like "demo mode" and "timer" and make things work as people expect. Here is how I'd expect it to work:

* I can create content anonymously, and the content is public and temporary, where "temporary" means a day or two. Enough to be useful, not enough to clog up your systems. 
* The URL I'm working in is immediately shareable (this is what I'd assumed, and tried, and failed with). The whole extra dialog of getting a URL with specific access rights was for nothing. 
* Warn me, e.g. in a header, that the content is public and short-lived. Don't make it too specific, no-one cares about a timer except your implementation.
* Offer me then the option of making this content private and/or permanent, via an action on the page somewhere. When I do this, if I'm not logged in, it asks me my email address, and then a password. And then verification etc.

This makes the platform immediately useful as a shared space with no identity and no persistence. No demo mode, no timers, no upfront registration, no reason to say "no."

PR #1236, opened 2015-12-04
Fix setPath/identity switcher interaction

zarvox:
GrainView's `reset()` did not clear all the fields it needed to clear,
so other derived data were incorrect.  This includes having bad values
for the iframe's src in transitional states.


dwrensha:
Looks good to me.

PR #1239, opened 2015-12-04
Make identity provider icons darker and a bit larger, for readability.

dwrensha:
This makes the icons the same shade of gray that they are in @neynah's mock-ups.

before:
before

after:
after


kentonv:
FWIW we should be moving to using the icon font, in which case you can choose the color in CSS. But no time for that right this moment.

PR #1246, opened 2015-12-07
Update Users.lastActive documentation to match current reality: only accounts have this field.

dwrensha:
It might make sense to have this field on identities too, but we would first need to update the stats code in several places.


kentonv:
The `splitAccountUsersAndIdentityUsers()` actually does copy `lastActive` over to the identities. As a side-effect the user stats are now double-counting.

dwrensha:
Eek. Yeah, we're now double-counting any user that was around before the migration.

PR #1247, opened 2015-12-07
Stop relying on function that got refactored-out, and thereby stop crashing during sharing

paulproteus:


paulproteus:
@kentonv ready for your review/merge.

I suppose the fact that this was broken suggests that there should be tests that cover it.

paulproteus:
Thanks for speedy merge.

It is too bad that I didn't catch this when the refactor happened. I
noticed that the refactor was deleting this function but forgot to actually
re-check my code against that refactor before resubmitting for merge.

PR #1249, opened 2015-12-07
sandcats https: Do date arithmetic on the right operands

paulproteus:
This fixes a bug that logs this message:

```
Will switch certs, probably to /var/sandcats/https/{{hostname}}/{{filename}}, at time Invalid Date via real-time SNI re-keying system.
```

which has the side-effect that real-time SNI re-keying didn't work.

Details:

- I was adding `undefined` to an integer and getting `NaN`, then
- Adding `NaN` to another number and getting `NaN`.
- As a result, I had an unused helper value, `twentyMinutesInMilliseconds`.
- Eventually I converted some of these things to `new Date()`s and got an `Invalid Date`.

Testing done:

- Added this patch to `rose.sandcats.io` (my self-install), then
- Manually fiddled with time on my self-install to Sat Dec 5, then
- Caused it to log this message: `Will switch certs, probably to /var/sandcats/https/rose.sandcats.io/1449306826631 , at time Sat Dec 05 2015 04:33:57 GMT+0000 (UTC) via real-time SNI re-keying system.` then
- manually fiddled with time to be Sun Dec 6, then
- made a HTTPS request, and got this log message `Re-keying resulting in a new certificate. Good.`, then
- tested in a browser that it using the new certificate, and indeed it is.

### DIFF
--- a/meteor-bundle-main.js
+++ b/meteor-bundle-main.js
@@ -285,7 +285,7 @@ function monkeypatchHttpAndHttps() {
 
       if (useThisCertificate.keyFilename != referenceCertificate.keyFilename) {
         var twentyMinutesInMilliseconds = 1000 * 60 * 20;
-        var twentyMinutesFromNow = now + twentyMinutesFromNow;
+        var twentyMinutesFromNow = now + twentyMinutesInMilliseconds;
         var notAfterMinusTwentyMinutes = result.notAfter - twentyMinutesInMilliseconds;
 
         // But never attempt to re-key in the past...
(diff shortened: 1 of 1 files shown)


zarvox:
I was about to comment, but then you pushed the second commit fixing the thing I was going to comment on.  :D

paulproteus:
Yeah. You might enjoy my process here:

- Read the code carefully
- Identify this bug
- Make a commit
- Go SSH into my self-install and manually make the "same" change on my self-install against `/opt/sandstorm/latest/main.js` but
- accidentally I made a _different_ change when manually editing the file, then
- manually test per the above, and
- believe the fix I committed was right, even though I had accidentally tested a different thing than what I committed.

paulproteus:
Re-tested with `sudo wget https://raw.githubusercontent.com/sandstorm-io/sandstorm/fix-sandcats-rekeying/meteor-bundle-main.js -O /opt/sandstorm/latest/sandstorm-main.js`


@kentonv ready for your review and/or merge.

PR #1251, opened 2015-12-08
Fix stats to not count identities as separate users.

kentonv:


paulproteus:
Yay.

dwrensha:
I think we should also change the logic where we [update `DeleteStats`](https://github.com/sandstorm-io/sandstorm/blob/09c03628b27e86957b41a7dfc483a0e891da9d96/shell/shared/demo.js#L67-L79) --- we should only count account users there, too. Of course, right now such a change would only affect demo users that are active during the migration.

dwrensha:
I've addressed my concern about `DeleteStats` in #1258.

PR #1252, opened 2015-12-08
Suggest sign-in when opening a sharing link anonymously.

kentonv:
We auto-open the login popup with options for "sign in" (goes to the normal login button list) and "stay anonymous" (closes the popup).

Currently looks like screenshot below. I suspect @neynah can come up with something better, but I think this is good enough to push now.

screenshot from 2015-12-07 21-01-20

### DIFF
--- a/shell/client/_account.scss
+++ b/shell/client/_account.scss
@@ -72,6 +72,19 @@ body>.popup.account>.frame {
   }
 }
 
+.login-suggestion {
+  >button {
+    @extend %button-base;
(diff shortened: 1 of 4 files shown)


paulproteus:
I, for one, am very excited by this changeset!

neynah:
This is great! In general, @kentonv comes up with pretty awesome UI.

dwrensha:
Looks good to me.

PR #1258, opened 2015-12-08
Update lastActive for identity users too.

dwrensha:
Now that we're chasing down any place that this would affect stats, there's no reason not to do this.


dwrensha:
Ugh, git error with libsodium on Jenkins.

dwrensha:
I'm guessing that the Jenkins box still has some dirty state from before #1173. @jparyani, can you please fix it?

jparyani:
Garply, retest this please

PR #1259, opened 2015-12-08
Fix bug where an expired demo user could see a grain just by refreshing the page.

dwrensha:
Problem: [this selector](https://github.com/sandstorm-io/sandstorm/blob/09c03628b27e86957b41a7dfc483a0e891da9d96/shell/shared/grain.js#L1419) matches the `main-content` element.

Solution: use a different class, as we did with `first-sign-in-main-content`.


jparyani:
Garply, retest this please

PR #1260, opened 2015-12-08
Don't clear the admin token after setting just one setting.

kentonv:
The behavior in which the first setting change succeeds but subsequent changes fail is not intuitive to users. The intent of the admin token is to give the user time-limited access to the admin interface -- but not to limit it to only one change.

In the past, the problem was masked by the fact that the first setting change was usually to enable a login service. The user would then log in and, by virtue of being first user, get promoted to admin, so never use an admin token again.

However, the "first user becomes admin" thing has changed. Now, the user must exchange the admin token to receive permanent admin rights. With the token being good for only one change, this means that two admin tokens are required while setting up a server: one to enable login, and another to actually become admin. This is poor UX.

This commit simply removes the logic that clears the admin token after a successful change.

### DIFF
--- a/shell/shared/admin.js
+++ b/shell/shared/admin.js
@@ -855,10 +855,6 @@ if (Meteor.isServer) {
 
       ServiceConfiguration.configurations.upsert({service: options.service}, options);
     },
-    clearAdminToken: function(token) {
-      check(token, String);
-      clearAdminToken(token);
-    },
     clearResumeTokensForService: function (token, serviceName) {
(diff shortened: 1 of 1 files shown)


jparyani:
The "clearAdminToken" meteor method should also be removed since it's not needed anymore (or we could put it to re-use and add a "Clear admin token" button in the UI).

kentonv:
Removed the method.

jparyani:
LGTM

PR #1261, opened 2015-12-08
Work around libsodium's force-push

zarvox:
libsodium's "stable" branch received a force-push recently, which breaks our current approach to `make update-deps`.

This change fetches the stable branch from the remote and checks out the FETCH_HEAD, rather than attempting to pull and fast-forward.

It does cause git to make a kinda scary warning about having a detached head, which is a little sad.  If we don't mind clobbering any data in deps/libsodium we could instead `git reset --hard FETCH_HEAD` but that might be a user's dev folder if they do the ekam symlink trick for deps, so I figured this was preferable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -193,11 +193,18 @@ deps/node-capnp:
 
 update-deps:
 	@$(call color,updating all dependencies)
-	@$(foreach DEP,capnproto ekam libseccomp libsodium node-capnp, \
+	@$(foreach DEP,capnproto ekam libseccomp node-capnp, \
 	    cd deps/$(DEP) && \
 	    echo "pulling $(DEP)..." && \
 	    git pull $(REMOTE_$(DEP)) `git symbolic-ref --short HEAD` && \
 	    cd ../..;)
+	# Argh, libsodium did a force push of their stable branch,
+	# necessitating an explicit checkout.
+	@cd deps/libsodium && \
+	    echo "pulling deps/libsodium..." && \
+	    git fetch $(REMOTE_libsodium) stable && \
+	    git checkout FETCH_HEAD && \
+	    cd ../../
 
 # ====================================================================
 # Ekam bootstrap and C++ binaries


dwrensha:
Does `git checkout FETCH_HEAD` accomplish something different from `git checkout origin/stable`? I guess the point is that the remote might not actually be named "origin"?

zarvox:
Yeah, with this we specify the remote explicitly, rather than relying on origin being the URL we stated.

This is desirable because remotes have changed in the past (libseccomp moved from SourceForge to GitHub), and really we always want to be using whatever's specified in the Makefile.

dwrensha:
I'm somewhat uncomfortable with leaving the repo in a headless state.
Would it make sense to do `git pull --ff-only`, and if that fails delete the current branch and then repull it?
Approximately:

```
BRANCH=`git symbolic-ref --short HEAD`
git pull --ff-only $(REMOTE_$(DEP)) $BRANCH
if $? != 0 then
  git branch -D $BRANCH
  git pull $(REMOTE_$(DEP)) $BRANCH
```

kentonv:
I agree with @dwrensha although that sounds destructive if the user has made any local changes. Can we detect the presence of local changes and bail out if there are any?

zarvox:
You can theoretically bail if there's uncommitted local changes, but if you have a local, committed change that's not a fast-forward of the remote tree, then that looks essentially the same as "remote force-pushed a change", so I'm not sure this is a supportable request.

The difficulty here is a result of unclear ownership of that folder - we can clobber the data with impunity (and should!) if the Sandstorm repo was responsible for setting up and maintaining that folder, but if the user set it up and might have their own work there, then we don't want to mess with it.

Perhaps the best thing to do would be to abandon this commit and instead add a suggestion of what a user could do to recover the tree if the "git pull --ff-only" fails.  Another option: check if `deps/` or `libsodium/` is a symlink (an indication of intended user-ownership) and refuse to take a destructive action if so?

Anything else that could help distinguish who owns the deps repo?

kentonv:
Hmm, I think possibly you could check before the fetch whether `origin/stable` matches `stable`. Theoretically you could even record the commit of `origin/stable` for use in a subsequent `rebase --onto` after the fetch. But maybe that's over-engineering?

We do want to make sure that our continuous build doesn't break every time this happens. Another way to accomplish that might be to introduce a new make target called `clobber-deps` or something which happily destroys any local changes.

zarvox:
Actually, I like that last suggestion a lot - add `make clobber-deps` which does `git fetch $EXPLICIT_REMOTE $EXPLICIT BRANCH && git reset --hard FETCH_HEAD`, and use that in our continuous integration, and if `make update-deps` as it exists today fails, suggest `clobber-deps` as an option.

People who have local changes are likely capable of handling their tree state (read: occasionally doing a `git reset --hard $SOME_APPROPRIATE_BRANCH` when upstream does dumb things) without our scripts.  And everyone else who doesn't care can let `clobber-deps` handle the repos.

PR #1262, opened 2015-12-09
Add a test for grain backup and restore

zarvox:
Dealing with downloads and uploads in selenium is...fraught.

You can't interact with any of the native platform dialogs, so you have to
configure Firefox to auto-download files of a particular MIME type to a fixed
location via a profile which you also have to configure on a per-test basis.

You can't really know when a download is completed, since there's no
programmatic notification of completion available.  We watch the hardcoded
download folder for changes, and assume that since our backup is small, it will
complete within 5 seconds of clicking the button, and within 1 second of the
file being initially `open()`'ed.

You can't set the value of a `<input type="file">` element in Firefox if the
element isn't visible.  The shell makes them invisible by default and supports
interaction by triggering `input.click()` when you click a parent button.  So to
make this work in testing, we patch the style of the `<input>`.

Finally, we need to refactor the shell to allow the tests to call the
appropriate methods to trigger form submission, since we normally activate the
<input> and wait for its value to change (but we can't here, because Selenium
can't interact with the native platform UI like the file-picker).

A new test app ( https://github.com/zarvox/sandstorm-test-python ) was written
to support this test and others.


zarvox:
Garply, retest this please

zarvox:
@dwrensha I'm seeing a test failure from the previous test run ( https://build.sandstorm.io/job/Sandstorm%20PRs/1339/ ) that looks like the grain incognito interstitial failed with a completely blank screen.  Any ideas?

tests/tests/backup-restore.js
```diff
@@ -0,0 +1,195 @@
...
+    .click('li.navitem-grain.current button.close-button')
+    .url(browser.launch_url + "/apps")
+    .waitForElementVisible('button.restore-button', short_wait)
+    .execute(function () {
+      // Firefox/Selenium oddity: invisible inputs cannot have their values set
+      var input = document.querySelector('button.restore-button input[type=file]');
+      input.style["display"] = "inline";
+    }, [])
+    .perform(function (client, done) {
+      // Have to defer referencing fileDownloaded until after the previous steps have run, so this
```
dwrensha: Do you mean `downloadPath` rather than `fileDownloaded`?

jparyani:
The testing part of this PR looks fine to me. @kentonv, can you review the shell.js changes and merge?

tests/tests/backup-restore.js
zarvox: Touché.

dwrensha:
The shell.js changes look good to me!

shell/shared/shell.js
```diff
@@ -1002,41 +1002,47 @@ var startUpload = function (file, endpoint, onComplete) {
...
       }
     });
   });
+};
+
+promptRestoreBackup = function(input) {
+  promptForFile(input, restoreBackup);
+};
+
+// This function is global so tests can call it
```
dwrensha: style nitpick: comment should go inside function

shell/shared/shell.js
```diff
@@ -1002,41 +1002,47 @@ var startUpload = function (file, endpoint, onComplete) {
...
-        }
-      });
-    });
-  });
-}
-
-promptUploadApp = function (input) {
-  promptForFile(input, function (file) {
-    Meteor.call("newUploadToken", function (err, token) {
+// This function is global so tests can call it
```
dwrensha: same style nitpick as below

dwrensha:
(sorry about LGTM'ing before nitpicking)

zarvox:
@dwrensha thanks for the feedback :)

PR #1264, opened 2015-12-09
On initial signup, if billing is enabled, show a billing plan selector.

kentonv:
This lets the user know that billing is a thing. The user can of course choose "free", but perhaps some will want to choose a plan immediately, and others will want to keep in mind that they have options.

The strategy is similar to the profile confirmation page, in that the plan selector overrides other routes. However, there is an exception: if the user is trying to view a shared grain (not one they own), then we do not interrupt them, since this would be weird for people who are only here to interact with their friend's grains.

There is a risk that this change will scare some users away too early. We should consider running a controlled experiment.

screenshot from 2015-12-08 22-01-20

Note how the prompt does not show any plan as "current", even though the free plan is technically current.

This PR is the Sandstorm side of the change. Most of the meat is in Blackrock.

### DIFF
--- a/shell/shared/shell.js
+++ b/shell/shared/shell.js
@@ -701,6 +701,9 @@ if (Meteor.isClient) {
       // Don't show if not logged in.
       if (!user) return;
 
+      // Don't show if not in the experiment.
+      if (!user.experiments || user.experiments.firstTimeBillingPrompt !== "test") return;
+
       // Don't show if the user has selected a plan already.
       if (user.plan && !Session.get("firstTimeBillingPromptOpen")) return;
 
--- a/shell/shared/shell.js
+++ b/shell/shared/shell.js
@@ -689,6 +689,51 @@ if (Meteor.isClient) {
     },
     accountSettingsUi: function () {
       return makeAccountSettingsUi();
+    },
+    firstTimeBillingPromptState: function () {
+      // Should we show the first-time billing plan selector?
+
+      // Don't show if billing is not enabled.
+      if (!window.BlackrockPayments) return;
+
+      var user = Meteor.user();
+
+      // Don't show if not logged in.
+      if (!user) return;
+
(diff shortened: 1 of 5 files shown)


kentonv:
I just changed the second FAQ to be different for this prompt than it is when you run out of space:

> Which plan should I choose?
>
> If you're not sure which plan to choose, start with the free plan. You can always upgrade later. Also remember than Sandstorm is Open Source, so you can install your own server »

shell/shared/shell.js
```diff
@@ -689,6 +689,51 @@ if (Meteor.isClient) {
...
+      // Don't show if the user has selected a plan already.
+      if (user.plan && !Session.get("firstTimeBillingPromptOpen")) return;
+
+      // Only show to account users (not identities).
+      if (!user.loginIdentities) return;
+
+      // Don't show to demo users.
+      if (user.expires) return;
+
+      // Don't show when viewing another user's grain. We don't want to scare people away from
```
jparyani: I believe we redirect users to "/grain" automatically (see https://github.com/sandstorm-io/sandstorm/blob/94db7a3037acad63925a39071545b2c9b8d272b0/shell/client/grainview.js#L470), so we **will** be displaying this to users who login with a sharing link.

jparyani:
Other than my comment, it looks good to me.

shell/shared/shell.js
kentonv: For grain routes, the code below checks whether the active grain is owned by the current user, and does not display the prompt if it isn't.

kentonv:
I've turned this into an experiment. Half of new users will be selected for the experiment; the other half will go into the control group. After a while we can use direct mongo queries to compare:
* Number of accounts who created a (non-demo) grain (evaluates whether users are bothered by the prompt).
* Number of accounts who upgraded to a non-free plan (evaluates the prompt's effectiveness).

PR #1267, opened 2015-12-09
Use normal syntax for template helper

paulproteus:
@jadeqwang reported a bug with the following steps to reproduce:

- Have a paid plan, and

- Go to /referrals on Oasis, and

- See text about how you could earn up to 2GB of storage, when that text really should be telling you that you can earn up to 30GB of storage

I took a shortcut here in the template helper, which was not to wrap it in `function()`, but perhaps that was wrong. So this commit fixes that. I say "perhaps that was wrong" because it seems to work fine both ways for me on localhost. But maybe if I don't make it a function, it doesn't get reactively re-run, but I can't trigger the reactivity on localhost because latency is too low.

@jparyani I had emailed you about this before; everything I know is here, so maybe if we discuss, let's discuss here.

Note that I _can_ reproduce the problem Jade reports using `testrock.sandstorm.io` so that's (approximately) good. It means I should be able to use testrock to confirm if this fixes the bug.

@kentonv I don't suppose there's a way for me to personally push to testrock...?

### DIFF
--- a/shell/shared/shell.js
+++ b/shell/shared/shell.js
@@ -481,7 +481,9 @@ if (Meteor.isClient) {
   });
 
   Template.referrals.helpers({
-    isPaid: (Meteor.user() && Meteor.user().plan !== "free"),
+    isPaid: function() {
+      return (Meteor.user() && Meteor.user().plan && Meteor.user().plan !== "free");
+    },
     notYetCompleteReferralNames: function() {
       return ReferralInfo.find({completed: false});
     },


paulproteus:
Honestly there was no good reason to push the code with this "shortcut" (no `function() { ... }`) in it. I wrote it that way because I was curious if it would work that way, and it seemed to, and then I forgot to clean it up in the progress of getting it merged.

dwrensha:
> But maybe if I don't make it a function, it doesn't get reactively re-run, but I can't trigger the reactivity on localhost because latency is too low.

This sounds like a very plausible explanation to me.

kentonv:
Amusingly, had the code worked without the `function()` wrapping, then we would have been promising paid-level referral bonuses to free users, because the code treated plan === undefined as being a paid plan.

So I'm glad the lack of function() means we're under-promising rather than over-promising. :)

kentonv:
But yes, the reason for the problem is that without the `function()` wrapper, this expression is being evaluated at initial page load. It's not just non-reactive, it's entirely static.

Issue #1255, opened 2015-12-08
Adding an email address: "Error linking identity: Error: Cannot link an identity that can already log into another account [403]"


ocdtrekkie:
This is a confusing error, because I am trying to add an email address (to set as a primary notification address), not link an account. There probably should not be a limit on linking email addresses for notification that are used elsewhere.

ocdtrekkie:
I also tried to use the passwordless email login to log into any possible "account" created with this address, and it asked me to confirm my profile, as if I didn't have one, and there's no way to delete the account. So it seems impossible for me to attach my inbox (at) jacobweisz.com email address to my Sandstorm account on this server.

dwrensha:
Thanks for reporting this. We need to figure out how to make this less confusing.

Adding an email address is the same thing as linking an email identity. The error means that at some point in the past you must have logged in with that email address. The "confirm identity" thing is not surprising, because we previously did not ask for confirmation for guest users, but now we do.

We intend to implement account deletion at some point. Until then, one workaround for freeing up your preferred email identity might be:
  1. Attach a new throwaway email identity to your email's account. (e.g. I might use david+throwaway@sandstorm.io)
  2. Give the new email identity the ability to login.
  3. Unlink your preferred email identity.
  4. Now you can link your preferred email to your original account.

dwrensha:
It seems like it might be a common for a user to have a bunch of accounts under different identities, but only one account that has any grains or installed apps. It would probably be easy for us to do something less confusing in that case than displaying "Cannot link an identity that can already log into another account"

PR #1269, opened 2015-12-09
Automatically delete unused accounts when doing so would free up an identity that a user wants to link.

dwrensha:
This prevents the error seen in #1255.

Note that we now need to explicitly check that we aren't linking an identity that's already linked to the current account --- otherwise, a fresh account that tries to re-link its only identity gets pushed back to the login interstitial with no helpful error reported.

### DIFF
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -1454,7 +1454,7 @@ if (Meteor.isServer) {
     });
   };
 
-  SandstormDb.prototype.deleteUnusedAccount = function(identityId) {
+  SandstormDb.prototype.deleteUnusedAccount = function(backend, identityId) {
     // If there is an *unused* account that has `identityId` as a login identity, deletes it.
 
     check(identityId, String);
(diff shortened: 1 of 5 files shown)


shell/packages/sandstorm-db/db.js
```diff
@@ -1454,6 +1454,23 @@ if (Meteor.isServer) {
...
+    // If there is an *unused* account that has `identityId` as a login identity, deletes it.
+
+    check(identityId, String);
+    var account = Meteor.users.findOne({"loginIdentities.id": identityId});
+    if (account &&
+        account.loginIdentities.length == 1 &&
+        account.nonloginIdentities.length == 0 &&
+        !Grains.findOne({userId: account._id}) &&
+        !ApiTokens.findOne({accountId: account._id}) &&
+        (!account.plan || account.plan === "free") &&
```
kentonv: Also check that account.payments is not present -- otherwise they might have credit that they're losing.

shell/packages/sandstorm-db/db.js
```diff
@@ -1454,6 +1454,23 @@ if (Meteor.isServer) {
...
+        !Notifications.findOne({userId: account._id}) &&
+        !Contacts.findOne({ownerId: account._id})) {
+      Meteor.users.remove({_id: account._id});
```
kentonv: I think you need to call `backend.deleteUser()` too.

dwrensha: Done.

shell/packages/sandstorm-db/db.js
dwrensha: Done.

PR #1275, opened 2015-12-11
Enable specifying a background for the logged-out home page.

kentonv:
If specified, the URL will be displayed in an iframe behind the prompt. This allows every server to customize their home page.

I removed the old "splashDialog" setting as it had become kind of ugly and now seems unnecessary. Plus, it technically allowed the admin to XSS users. While the legitimate admin almost certainly has this power anyway, limiting the power of the admin panel makes hijackings less damaging.


jparyani:
As discussed, a warning to users that this is subject to change would be nice. Also, validating the URL (with `matchWildcardHost`) might be nice, but this an advanced option so I guess it's ok without.

Otherwise, LGTM.

PR #1278, opened 2015-12-12
Make the share-via-email message more recipient-friendly

zarvox:
* Include the grain title at the start of the subject line
* Include the grain title in the button
  * Adjust padding and min-width on button.  The button should grow in width to
    accommodate longer grain titles.
* Change the button to be Sandstorm purple.
* Set the From: header to include the sharing user's display name and
  if available, the first verified email for that identity.
* Avoid jargon like "Shared Grain" in favor of "collaborate" and using
  the sharer's title for the grain.

### DIFF
--- a/shell/packages/sandstorm-email/email.js
+++ b/shell/packages/sandstorm-email/email.js
@@ -148,6 +148,7 @@ SandstormEmail.send = function (options) {
     to: options.to,
     cc: options.cc,
     bcc: options.bcc,
+    envelope: options.envelope,
     replyTo: options.replyTo,
     subject: options.subject,
     text: options.text,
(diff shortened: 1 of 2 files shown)


kentonv:
> Set the From: header to include the sharing user's display name and if available, the first verified email for that identity.

Please use the sender's selected primary email rather than the first verified email. I believe there is a helper function for this.

zarvox:
We discussed this in person, but I'm leaving myself a note for the record lest I forget:

We should not use the sender's primary service-related email for sending this email address by default - we should only send from a verified email address associated with the sharing identity.

Instead, we should add a "Send as:" combobox email to the sharing form, populated with the sharing identities' verified email addresses, with the first address in that array selected by default.

kentonv:
Would love to see this finished soon! :)

PR #1279, opened 2015-12-12
Add postMessage API tests

zarvox:
The test app now renders a template that includes the token and host as JSON,
for easier use in the tests.

### DIFF
--- a/tests/tests/postmessage-api.js
+++ b/tests/tests/postmessage-api.js
@@ -1,0 +1,148 @@
+// Sandstorm - Personal Cloud Sandbox
+// Copyright (c) 2015 Sandstorm Development Group, Inc. and contributors
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


tests/tests/postmessage-api.js
```diff
@@ -0,0 +1,148 @@
...
+
+module.exports['Test setPath'] = function (browser) {
+  var expectedUrl = browser.launch_url + '/grain/' + grainId + '/#setpath';
+  browser
+    .frame('grain-frame')
+      .click('#setPath')
+      .pause(very_short_wait)
+    .frameParent()
+    .assert.urlEquals(expectedUrl);
+    // Should link-sharing links be expected to also hold the current path?
```
zarvox: Curious what others think about this - I was considering adding some code to verify the behavior of generated link-sharing links, but then I realized that I can't really say which behavior is less confusing.

kentonv: I think probably the current path should be included when generating a link, but perhaps if the path is non-empty then we should have some fine-print that says: "Recipients will get access to the whole grain, not just /current/path."

PR #1282, opened 2015-12-13
App index improvements

kentonv:
1. Fix a bug introduced by the "experimental" index change: Detailed metadata from non-approved version of packages was taking precedence over approved packages of the same app. Details for experimental packages are now placed under `/experimental` rather than `/apps`.
2. Reject app updates that fail to bump `appVersion`.

I've actually already updated the master index with these changes because we were displaying bogus metadata on app details pages.


jparyani:
LGTM

PR #1285, opened 2015-12-14
DenormalizedGrainMetadata has an 'icon' field, not an 'appIcon' field.

dwrensha:
Grains that have been shared since #1161 have broken app icons. This patch fixes the problem and deletes some now-unused code.

It would not be too difficult to also write a migration that fixes up any tokens that have been created in the mean time.

Related question: I seem to remember that we had some discussions at one point where we decided that the recipients of shares should not get to see the appId of grains that are shared with them. Do we still think that would be desirable?

### DIFF
--- a/shell/packages/sandstorm-permissions/permissions.js
+++ b/shell/packages/sandstorm-permissions/permissions.js
@@ -470,9 +470,10 @@ SandstormPermissions.createNewApiToken = function (db, provider, grainId, petnam
 
     if (pkg && pkg.manifest && pkg.manifest.metadata && pkg.manifest.metadata.icons) {
       var icons = pkg.manifest.metadata.icons;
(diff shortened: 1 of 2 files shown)


kentonv:
Right, the idea is that if someone builds a customized version of an open source app with their own keys, they should be able to make it indistinguishable from the "real" version as far as anyone else can see.

It could be that this is a silly property that no one actually needs but I'm not comfortable saying that for sure yet. :)

dwrensha:
> they should be able to make it indistinguishable from the "real" version as far as anyone else can see.

OK. I believe this is currently the case as long as the app has an icon. If the app doesn't have an icon, then we store the app ID in the denormalized grain metadata and we use it to generate an identicon.

kentonv:
Correct. The plan is that if we see a need we'll add a way for an app to specify its "identicon ID" which would override appId for this purpose.

PR #1286, opened 2015-12-14
Check minApiVersion and maxApiVersion validity

zarvox:
`spk` should refuse to produce a package that claims compatibility with an API
version of which it is unaware.  It should also refuse to pack an .spk with
maxApiVersion < minApiVersion.

### DIFF
--- a/src/sandstorm/spk.c++
+++ b/src/sandstorm/spk.c++
@@ -63,6 +63,11 @@ typedef kj::byte byte;
 static const uint64_t APP_SIZE_LIMIT = 1ull << 30;
 // For now, we will refuse to unpack an app over 1 GB (decompressed size).
 
+static const uint32_t MAX_DEFINED_APIVERSION = 0;
+// The maximum API version that has been defined, as of this source code's compilation.  We should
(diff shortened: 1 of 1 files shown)


src/sandstorm/spk.c++
```diff
@@ -63,6 +63,14 @@ typedef kj::byte byte;
 static const uint64_t APP_SIZE_LIMIT = 1ull << 30;
 // For now, we will refuse to unpack an app over 1 GB (decompressed size).
 
+static const uint32_t MAX_DEFINED_APIVERSION = 0;
+// The maximum API version that has been defined, as of this source code's compilation.
+// We should outright refuse to pack an app claiming a newer minApiVersion than this, because we
+// can't possibly know what the constraints are on that API.
+//
+// We should additionally warn if the user attempts to pack an app claiming a newer maxApiVersion
```
kentonv: warn -> error

zarvox: Fixed.  Sorry about the force-push; I noticed the bad comment myself and thought I'd sneak in a correction in before anyone noticed, but I was wrong.

PR #1287, opened 2015-12-14
Add a clobber-deps command to the Makefile for CI usage

zarvox:
A CI system using this can not be affected when upstreams make non-fast-forward
pushes to branches that we rely on.

Closes #1261.


jparyani:
LGTM, but I'll let @kentonv merge. I'll update the build server to use this new make target when that happens.

PR #1289, opened 2015-12-14
Test code cleanup

zarvox:
Attempt to remove the need for arbitrary `pause()` invocations.  Ideally, we
should wait for things to reach a desired state, with a timeout for that
happening, so we don't sit idly for the rest of the timeout period.

This change makes login cleanly async, which should save a couple seconds on
each test that performs login.


zarvox:
Garply, retest this please

PR #1304, opened 2015-12-17
Make sharing form a bit more robust if user never completes contact

jparyani:
Also fail with a reasonable error if the contact input is empty.


kentonv:
I'm pushing this for now to fix the bug but the form really shouldn't force the user to "start over" (clearing the whole form) when there's a problem...

PR #1307, opened 2015-12-19
Extract installer-tests tool into separate repo

paulproteus:
This commit:

- Removes run_tests.py, and adds a dependency on a separate project,
  stodgy-tester.

- Adjusts the *.t test files accordingly.

- Adjusts the Makefile and prepare-for-tests.sh accordingly.


paulproteus:
I'm going to self-merge this as it 100% affects `installer-tests/` and the `Makefile` rule that relates to `installer-tests`.

PR #1308, opened 2015-12-19
Fix trivial typo: hyphen vs underscore

paulproteus:
This appears in the Makefile and how it invokes `stodgy-tester`.


paulproteus:
Self-merging because 100% related to installer-tests.

PR #1316, opened 2015-12-27
install.sh: Add support for domain reservation token

paulproteus:
- Includes a new test for this.


paulproteus:
I've run the `install.sh` test suite locally, and all the tests pass.

I'll leave this not-yet-merged so that there's time to think about any implementation details.

paulproteus:
I'm going to self-merge this now, even though the tests fail, since the `installer-tests/` pass.

PR #1317, opened 2015-12-28
twiddle with some symlinks to get libsodium to build

dwrensha:
This makes the directory structure more faithful to the original repository, so that [this include](https://github.com/jedisct1/libsodium/blob/f9d982480ba68895dd90f29cf072bc3e533f09de/src/libsodium/crypto_sign/ed25519/ref10/open.c#L10) has a chance of working.


paulproteus:
Yay!!!​

PR #1320, opened 2015-12-29
(Do not merge) Fix #1318; verify w/ Jenkins that it actually works

### LINKED ISSUE #1318: Edge case when setting blank title or path
With the new `postMessage` API grains can expose meta-data to the Sandstorm shell. I encounter the following edge case:

```js
window.parent.postMessage({setTitle: ''}, '*');
```

Instead of setting a blank tittle, this output the following message in the console:

    postMessage from app not understood: [object Object]

(Setting a blank title isn’t the best idea in the world, but that just happen to me in Wekan due to a load order problem, ie the `DocHead.getTitle()` is called for the first time before the title is set and then re-called again a few milliseconds later due to its reactivity). Anyway I think this edge case should be handled correctly.

This bug is due to JS automatic type conversion [here](https://github.com/sandstorm-io/sandstorm/blob/39c898e30f5f836429c68b7476609beb39f8dbcf/shell/shared/grain.js#L1251). Similar calls wouldn't work either:

```js
window.parent.postMessage({setPath: ''}, '*');
```

paulproteus:


paulproteus:
I will resubmit once @dwrensha 's' fix lands. This shows that my test and my implementation match up.

PR #1331, opened 2016-01-05
Update Sandcats administration docs

paulproteus:
Improve detail. Hopefully this helps one confused user.


paulproteus:
Self-merging as this is 100% docs.

PR #1334, opened 2016-01-06
Update release-certificate for Asheesh

paulproteus:


paulproteus:
Self-merging since this is 100% my release certificate and @kentonv has said it's OK to do that.

PR #1337, opened 2016-01-06
Add docs on docs

paulproteus:


paulproteus:

I did just test this, fwiw.

PR #1344, opened 2016-01-08
Add intro.js-based "guided tour" to teach grain creation

paulproteus:
This paves the way for other "guided tour"-like functionality if we need it.

Note:

- This adjusts /apps/:appId to wait on globalSubs, like most other routes.

- This adds intro.js and therefore adds it to shell.html.


paulproteus:
@kentonv, this is ready for your review/merge.

A question:

- @kentonv is this a sensible way of "vendoring" external JS/CSS files? I realize it's extremely ad-hoc. If you prefer, I can create a Meteor package to contain these dependencies.

- @neynah if you prefer something other than 0.7 for opacity, you and I can experiment and I can submit a follow-up commit that makes that happen.

Screenshots:

working

*warning* - sometimes the highlight box ends up highlighting the wrong area. Here is what that looks like.

not-working-great

The crux of it is that I can't find out when the browser has finished loading the fonts/images for this page. I've added a workaround that, in addition to the way it works now, adds a call to `setTimeout()` so that 2 seconds later, `intro.js` to re-render itself. Pretty sad IMHO. It does result in the box moving to the right place, though, and it looks a little weird but not terrible when that happens.

But anyway, after 2 seconds, it does self-heal.

If there's a better way to do that, I'm all ears. I spent a while tonight trying to make `MutationObserver` work, to no avail.

zarvox:
I would prefer it if you would put intro.js into a meteor package - it's already pretty well-scoped and exports a single symbol.

paulproteus:
I will slightly revise this and make it into a package that lives under `sandstorm/shell/packages/`. Revision coming in a moment.

paulproteus:
Done moving it into a package. Thanks to @zarvox for the suggestion!

zarvox:
Great, thanks!

paulproteus:
@kentonv I'm done with this now and the build bots now like it!

shell/shared/grain.js
```diff
@@ -1644,9 +1644,8 @@ Router.map(function () {
     path: "/apps/:appId",
     template: "appDetails",
     waitOn: function () {
-      return [
-        Meteor.subscribe("appIndex", this.params.appId),
-      ];
+      return globalSubs.concat(
```
kentonv: Does this not lead to duplicate subscriptions which are expensive to dedup server-side?

shell/client/shell.html
```diff
@@ -1389,6 +1389,7 @@ <h2 class="fine-print">The fine print</h2>
         <li><a href="https://github.com/justmoon/node-bignum">node-bignum by Stefan Thomas</a> (MIT license)</li>
         <li><a href="https://github.com/drudru/ansi_up">ansi_up library by Dru Nelson</a> (MIT license)</li>
         <li><a href="https://github.com/stewartlord/identicon.js">identicon.js by Stewart Lord</a> (BSD license)</li>
+        <li><a href="https://github.com/usablica/intro.js">intro.js by Afshin Mehrabani</a> (MIT license)</li>
```
kentonv: Also edit LICENSE.

shell/shared/grain.js
kentonv: Hmm no I guess not, since you're not doing Meteor.subscribe() again, you're just passing in the existing subscription objects.

shell/client/shell.html
kentonv: I'll do it.

shell/packages/sandstorm-ui-app-details/app-details.html
```diff
@@ -52,7 +52,7 @@
   --}}
   <div class="app-details-widget">
     {{debug}}
-    <div class="app-icon" style="background-image: url('{{appIconSrc}}');"></div>
+    <div class="app-icon" style="background-image: url('{{appIconSrc}}'); height: 128; width: 128;"></div>
```
kentonv: This width and height you added are invalid because they don't specify any units, so the browser ignores them. I removed them in 9a9b058.

paulproteus: Thanks for catching that.

shell/client/shell.html
paulproteus: Thanks for handling that! I'll aim to remember in a future pull request.

paulproteus:
Thanks @kentonv for advice/cleanup and @zarvox for advice.

paulproteus:
Oh and thanks @kentonv for the merge!

PR #1349, opened 2016-01-09
Fix mobile double completing contacts on click

jparyani:
It turns out we only need to monitor `mousedown` events. `click` events
don't trigger correctly for these elements on desktop anyways.


paulproteus:
Wa-hoo!​

PR #1351, opened 2016-01-09
Make timeout longer while waiting for appdemo page to load

jparyani:


jparyani:
Self merging because it's just testing changes.

PR #1352, opened 2016-01-09
Remove incorrectly-used ARIA roles

ndarilek:


garply:

zarvox:
Ahhh, this looks like it's caused by us reusing the same template for the children of the menubar dropdown, heard as "your display name button" (where I think the roles are probably appropriate?) and in the main content of the root page (where it should not be described as a menu).

Does that sound right to you?

If so, I might prefer to whip up an alternate patch that threads through whether this instance of `loginButtonsList` is in a menu or not, so we can keep the roles around in the menubar dropdown, but drop them from the place where they don't belong.  Or if you want to do that, I'd be happy to give feedback/iterate with you!

zarvox:
Garply, ok to test

ndarilek:
Hmm, haven't actually looked at the markup, but none of your menus seem 
to get presented as menus under the NVDA, Orca or TalkBack screen readers.

menu/menuitem is really fairly finicky. You pretty much need <element 
role="menu"><child role="menuitem"/>...</element>. If any child isn't a 
menuitem, breakage. If new elements are added to the one containing the 
menu role, breakage. So for instance, if Notifications is a menuitem and 
clicking it pops up the notifications list under the element with the 
menu role, screen readers will have a hard time with that. It's best 
when you have a <nav/>, <ul/> or other item that will only contain menu 
items and never ever anything else. It's hard to get right, and 
something I struggle with implementing in my own apps.

Might be best to remove menu/menuitem roles entirely for now, then add 
them back slightly in carefully-controlled doses. I used them heavily in 
my initial patch because none of the controls were tabbable/interactive, 
but now it seems that all of your controls are fine.

It might be nice to add the roles back someday, but now they're causing 
issues under Android/Firefox where I can't access the form to input my 
email. I removed them here because that was the instance I 
encountered--didn't realize they were used more heavily elsewhere.

zarvox:
All right, +1 from me on dropping the roles then, and thanks for the explanation.  I'll let @kentonv merge.

PR #1354, opened 2016-01-10
Tell app authors how to switch IFRAME

paulproteus:
Thanks to @simonv3 for showing up to IRC to ask this question.

Ready for review/merge by @kentonv 

Curious what @zarvox thinks about this, and what @ndarilek thinks of it - specifically Nolan, would this have saved you time when you ran into the same issue before?

Note that this will log a message once per time someone opens a grain within the same browser tab, so they could end up with the message logged N times, once per grain they open in the same browser tab. I think that is OK.

### DIFF
--- a/shell/shared/grain.js
+++ b/shell/shared/grain.js
@@ -1218,7 +1218,6 @@ if (Meteor.isClient) {
       if (grain.sessionId()) {
         // TODO(soon):  Investigate what happens in background tabs.  Maybe arrange to re-open the
         //   app if it dies while in the background.
-        console.log("keepalive: ", new Date());
         Meteor.call("keepSessionAlive", grain.sessionId(), function (error, result) {
           // Sessions will automatically resume if possible, otherwise they will refresh.
         });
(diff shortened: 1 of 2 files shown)


paulproteus:
Garply, retest this please

ndarilek:
Looks useful. If there's concern about that being spammy, maybe you 
could only emit the message if the user is signed in with a dev account? 
No strong feelings either way, just a thought in response to your point 
about multiple messages.

kentonv:
Are you sure this only prints every time you open a new grain and not every time you switch grain tabs?

Could we make it print just once per session instead? You could do it in a Meteor.startup(). (Maybe it's time to retire the keepalive log spam so it doesn't scroll this away.)

paulproteus:
Kenton, I think your ideas sounds well. I'm happy to revise this
accordingly.

paulproteus:
As an update: @kentonv it does indeed print every time I switch grain tabs! I think `Meteor.startup()` is a smarter way to go about this.

paulproteus:
@kentonv now ready for your re-review/merge if you like. The message got longer, but I'm OK with that, at least. But you might or might not be!

paulproteus:
Yippee!

PR #1364, opened 2016-01-12
offerTemplate $GRAIN_SLUG replacement

mnutt:
This adds the ability to use `$GRAIN_SLUG` in an offer template to generate a useful representation of the name of the grain.

My use case for this is Owncloud Client authentication. While Sandstorm ignores the username field of basic auth, Owncloud Client displays it when listing remote servers, so setting it to the name of the grain is the only real way to distinguish remote servers in Owncloud Client.

image

Unfortunately not anything can go in the `username` part of a URL, so it sanitizes things like `Michael's Grain` down to `michaels-grain`.

### DIFF
--- a/shell/shared/grain.js
+++ b/shell/shared/grain.js
@@ -1814,8 +1814,18 @@ if (Meteor.isClient) {
           // template namespace under key id2.
           const key = "offerTemplate" + id2;
           const host = globalDb.makeApiHost(tokenId);
+          var grain = getActiveGrain(globalGrains.get());
+          var grainTitle = grain.title();
+          // grainTitleSlug is the grain title with url-unsafe characters replaced
+          var grainTitleSlug = grainTitle.toLowerCase().trim();
+          grainTitleSlug = grainTitleSlug.replace(/\s+/g, '-')
+                                         .replace(/[^\w\-]+/g, '')
+                                         .replace(/\-\-+/g, '-')
+                                         .replace(/^-+/, '')
+                                         .replace(/-+$/, '');
           const renderedTemplate = template.replace(/\$API_TOKEN/g, tokenId)
-                                           .replace(/\$API_HOST/g, host);
+                                           .replace(/\$API_HOST/g, host)
+                                           .replace(/\$GRAIN_TITLE_SLUG/g, grainTitleSlug);
           sessionStorage.setItem(key, JSON.stringify({
               token: tokenId,
               renderedTemplate: renderedTemplate,
--- a/shell/shared/grain.js
+++ b/shell/shared/grain.js
@@ -1814,15 +1814,15 @@ if (Meteor.isClient) {
           // template namespace under key id2.
           const key = "offerTemplate" + id2;
           const host = globalDb.makeApiHost(tokenId);
-          var grain = getActiveGrain(globalGrains.get());
-          var grainTitle = grain.title();
+          const grain = getActiveGrain(globalGrains.get());
+          const grainTitle = grain.title();
           // grainTitleSlug is the grain title with url-unsafe characters replaced
-          var grainTitleSlug = grainTitle.toLowerCase().trim();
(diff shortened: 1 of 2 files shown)


garply:

paulproteus:
Garply, ok to test

paulproteus:
@zarvox I wonder if you would be interested in being the one to review this.

Also @mnutt presumably this should have some docs that come with it. I'd be willing to write those docs in `docs/` perhaps in the file that generates this doc: https://docs.sandstorm.io/en/latest/developing/http-apis/#creating-an-offer-template

Would you be interested in writing those docs? If so, great. If not, I can volunteer to write them.

paulproteus:
Garply, ok to test

paulproteus:
I think Garply ignores messages by email, interestingly.

paulproteus:
Also @zarvox it seems to me we should move the offer template generation code, at some point, into a separate npm module and/or into a separate Meteor package, with its own unit tests. Doing that doesn't have to block this, but curious what you would want to see for that.

paulproteus:
(naturally @kentonv will have an opinion too, but I'm especially curious about @zarvox 's )

shell/shared/grain.js
```diff
@@ -1314,8 +1314,12 @@ if (Meteor.isClient) {
             // Store apitoken id1 and template in session storage in the offer
             // template namespace under key id2.
             var key = "offerTemplate" + id2;
+            var grain = getActiveGrain(globalGrains.get());
+            var grainTitle = grain.title();
+            var slug = grainTitle.toLowerCase().trim().replace(/[^\w ]+/g,'').replace(/ +/g,'-');
```
kentonv: Nitpick: It looks like a '-' in the title will actually get removed even though the final output can contain new '-'s, which seems odd. Maybe instead it makes sense to replace any span of non-\w characters with a single '-'.

kentonv:
I like this change but I agree I'd like @zarvox's opinion.

shell/shared/grain.js
zarvox: I think I'd prefer to instead just add `-` to the list of characters that don't get stripped - it'd be weird to convert `'` or `"` into `-`, especially if it's like `Drew's grain` -> `Drew-s-grain`.  

Maybe collapse any instances of `/-+/` into a single `-` as a followup?

shell/shared/grain.js
```diff
@@ -1314,8 +1314,12 @@ if (Meteor.isClient) {
...
             var renderedTemplate = template.replace(/\$API_TOKEN/g, tokenId)
-                                           .replace(/\$API_HOST/g, makeWildcardHost("api"));
+                                           .replace(/\$API_HOST/g, makeWildcardHost("api"))
+                                           .replace(/\$GRAIN_SLUG/g, slug);
```
zarvox: I'd prefer this be called `GRAIN_TITLE_SLUG` rather than `GRAIN_SLUG` to make it slightly more precise.

zarvox:
:+1: on this conceptually - this seems like a nice piece of polish!  Small changes requested. :)

And then either @mnutt or @paulproteus can write docs about this.

paulproteus:
I'm totally happy to own writing docs for this, if that speeds it up. If
I'm to write the docs, then I would prefer to do so after it gets merged,
just to avoid creating inter-person dependencies for this pull request.

mnutt:
Agreed about GRAIN_TITLE_SLUG. Also, I wonder if some replacement code like https://gist.github.com/mathewbyrne/1280286 would work better.

garply:

mnutt:
(I'll fix up the conflicts, change to GRAIN_TITLE_SLUG, and update the regex logic)

mnutt:
Updated to change to $GRAIN_SLUG_TITLE, add better character removal, and some (very light) doc updates.

zarvox:
:shipit:

PR #1365, opened 2016-01-12
ES2015 and JSCS style conformance

zarvox:
Depends on #1361.

ES2015 has some great features and syntax.  Meteor supports it.  Let's modernize our codebase and go all in.

This patch series adds a JSCS config file which mostly follows the airbnb style guide, and then migrates a significant portion of the shell (`lib/`, `client/`, and `server/`) to comply with the style guidelines and to make use of newly-available ES2015 features.  It started as a proof-of-concept for using ES2015 in the shell, and then I got carried away, as can be seen in the diffstat.

Once the whole shell can pass the JSCS checker, I intend to add a pre-commit hook which verifies that there are no deviations from the style rules and maybe fixes things it knows it can.

In working on this, I tried to audit for functions that were left in the global scope and either bring them into local scope if they were unused in other files or to document what else was relying on the functions living in global scope.


paulproteus:
I'm +0.9, by which I mean I don't plan to aggressively review this but in
general am very enthusiastic about these sorts of changes.

In general I am excited to have more ways that robots can check my code.

kentonv:
Test failure looks legit.

I'm excited about this! In retrospect it might have been a good idea to separate basically-automated changes (like changing functions to arrows) from deeper refactorings, but oh well.

zarvox:
Test failure is legit, and I think it will be fixed by the commit I just added.

Amusingly, it's actually the arrow function change that introduced the regression (at least the one in the install flow).  Arrow functions have neither a rebindable `this`, nor do they receive `arguments` in the normal way.  I misunderstood their properties when I redid `server/installer.js` the first time.

Takeaway: don't use arrow functions where you do fancy stuff with `arguments` or need to rebind `this`.  Which, slightly sadly, means they can't be used in most of the places where we interact with Blaze, since it needs to rebind `this`.

kentonv:
Garply, retest this please.

zarvox:
Garply, retest this please.

zarvox:
@kentonv said I could merge this

PR #1369, opened 2016-01-13
Add redirect hack for developing/security-practices/

paulproteus:


paulproteus:
Self-merge due to 100% docs

PR #1371, opened 2016-01-13
Add more security non-events.

kentonv:
I think I covered just about every app we have that is big enough to have CVEs.

A couple notable exceptions:

* MediaWiki's CVEs reflect a Wikipedia-oriented threat model: privilege escalation is not so important when everything is world-writable, but resource consumption (e.g. failure to throttle file uploads) is. We actually don't defend against that sort of thing, but OTOH we don't claim Sandstorm would be a good place to host Wikipedia.
* Gitlab had several CVEs in 2013 but is conspicuously lacking anything more recent, at least according to cvedetails.com. (Gitweb's CVEs are even older.)


paulproteus:
The writing here is very clear. Thanks for writing it.

I'll merge & push now. Some suggestions.

- For publicity purposes, does it make sense to have a counter somewhere? "Sandstorm protected users from 46 CVEs" or something like that, at the top of the document?

Thanks so much. Also, I need to figure out how I can get mkdocs to turn `--` into an em dash.

paulproteus:
Also, I think I'm +1 on people self-merging documentation changes if they have push access to the repo. I'm fine fixing things in follow-up commits or doing post-merge review for docs. I don't think we're gaining much from me standing in the middle here. @kentonv does that work for you in the future?

kentonv:
SGTM

kentonv:
Though one problem is I don't really know how to push documentation changes live, so I currently still rely on you for that. Maybe add that to the readme?

paulproteus:
You don't know how because it relies on a GitWeb Pages grain that only I
have a key for! Oh but I can share it with you... one sec.

PR #1372, opened 2016-01-13
Remove token from admin URL once it's not needed.

### LINKED ISSUE #409: Admin token URL should remove the token from the URL if you log in
**What I did**

While doing emergency server operations, I went to a URL like:

* http://pintle2.sandcats.io:6080/admin/19bc20df04838fdc03101d898be075cc02de66f2

which allowed me to change my login configuration, yay, so I did so.

* Then I logged in.

**What I saw**

* After logging in, I was still at my admin token URL e.g. http://pintle2.sandcats.io:6080/admin/19bc20df04838fdc03101d898be075cc02de66f2

**What I expected**

* Since the admin token is sort of meaningless for people who are already logged in, I would expect to find myself at http://pintle2.sandcats.io:6080/admin/ instead, without the token in the URL.

dwrensha:
Fixes #409.

### DIFF
--- a/shell/shared/admin.js
+++ b/shell/shared/admin.js
@@ -51,8 +51,13 @@ var adminRoute = RouteController.extend({
     });
     var user = Meteor.user();
     if (user && user.loginIdentities) {
-      if (this.params._token && (!user.signupKey || !user.isAdmin)) {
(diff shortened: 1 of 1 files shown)


paulproteus:
Garply, retest this please

Issue #1355, opened 2016-01-10
Misconfigured Google login leads to confusing redirect to admin settings; needs better explanation


paulproteus:
Steps to reproduce:

- Install Sandstorm as a self-hoster
- Enable Google login
- Log in
- Install Etherpad
- Create an Etherpad grain
- Create a sharing link
- Visit that sharing link in an incognito window
- Click on "Sign in" in the dropdown
- Click on the "Sign in with Google" choice

Expected behavior:

- See a login screen

Actual behavior:

- Find oneself at https://rose.sandcats.io/admin/settings which shows a message, "You are not logged in as admin and there isn't a valid token specified"

![anim](https://cloud.githubusercontent.com/assets/25457/12221348/341407ac-b74c-11e5-805d-ead8e0b1a270.gif)

paulproteus:
Bizarrely, I can't reproduce this on oasis.sandstorm.io.

Sample sharing links that do repro the problem:

- https://rose.sandcats.io/shared/QtmEUkQGcgV2e3b0LSy6Qcq2-CrUqg8OYO8222cWtRd

- https://xamar.sandcats.io/shared/Bqa9dftNbc1Ni06D-SgBdkFuM_iky8VHAlTw0Rk1lzN (this is how I noticed the bug, thanks Pablo Jamar)

Sample sharing link that does not repro the problem:

- https://oasis.sandstorm.io/shared/MBt18sGiLbNG7UcexoHFvTbtcYEI8Uja_pzIVCMaVu0

paulproteus:
Actually this seems to be a showstopper bug in login entirely. Steps to reproduce:

- Visit https://xamar.sandcats.io/

- Click "with Google"

Expected behavior:

- Be prompted for Google Account

Actual behavior:

- Find oneself at https://xamar.sandcats.io/admin/settings

dwrensha:
I would guess that [this redirect](https://github.com/sandstorm-io/sandstorm/blob/3a847a42721305daf7931adf1007c0d5d4360e2e/shell/packages/sandstorm-accounts-ui/login_buttons.js#L99-L101) is getting tripped, which ought to only happen in the case of misconfiguration. I haven't been able to reproduce locally, so I'm not sure exactly what kind of misconfiguration that could be.

kentonv:
[...]

PR #1373, opened 2016-01-13
Display an error message on catching ServiceConfiguation.ConfigError.

dwrensha:
This addresses https://github.com/sandstorm-io/sandstorm/issues/1355

### DIFF
--- a/shell/packages/sandstorm-accounts-ui/login_buttons.js
+++ b/shell/packages/sandstorm-accounts-ui/login_buttons.js
@@ -70,8 +70,9 @@ var loginResultCallback = function (serviceName, err) {
     // do nothing
   } else if (err instanceof ServiceConfiguration.ConfigError) {
     loginButtonsSession.errorMessage(
-        "The " + serviceName + " login service has a configuration problem (" + err.message + ")." +
-        "\n\n\n Please contact your server administrator. ");
+      "Configuration problem: " + err.message + ". Please visit the Admin Settings page within " +
+      "Sandstorm, or ask your administrator to do so. You may need an admin token. Read more by " +
+      "clicking Troubleshooting below.");
   } else {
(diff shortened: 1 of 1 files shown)


paulproteus:
FWIW I'm working on a patch that adds a simplistic UI to tell the administrator what to do. So I gently propose you wait for me to finish that and see if it supercedes this.

Sorry to not mention I was going to do that earlier this morning.

paulproteus:
Suggested string change:

from:

- The Google login service has a configuration problem (Service not configured). Please contact your server administrator.

to: 

- Configuration problem: Service not configured. Please visit the Admin Settings page within Sandstorm, or ask your administrator to do so. To learn more, click Troubleshooting below.

paulproteus:
Rationale: This makes sure the admin knows what to do, not merely that some action is required.

paulproteus:
Adjusted the above to say "You may need an admin token" since people will probably need that, and this process will be confusing.

dwrensha:
Since we now have a more complete strategy (#1374) for addressing #1355, I should independently justify this patch.

  1. I can't think of a situation where redirecting to /admin/settings is better than just reporting the error.
  2. The `var serviceName = instance.data.data.displayName;` line fixes a dereference of an otherwise-undefined variable.

paulproteus:
+1 on those grounds

kentonv:
I agree that this change is still needed: changing BASE_URL isn't the only way to end up in this situation. At least one user who reported this problem got there by deleting his Google account to which the OAuth credentials were attached.

I also agree with @paulproteus's wording change. In most cases the user reading the message will actually be the admin (of a personal server), so we should tell them what to do.

dwrensha:
OK, I've updated to Asheesh's wording.

PR #1376, opened 2016-01-14
Be specific about alpha.sandstorm.io

paulproteus:


paulproteus:
100% docs; self-merging

PR #1377, opened 2016-01-14
Add configurable server title and return address for notification emails

kentonv:
For now this is only used on login emails, but I also plan to use it for invoice emails in Blackrock.

Note that I intentionally did NOT apply this to sharing emails because we plan to make those be "from" the sharing user soon.


paulproteus:
Garply, retest this please

jparyani:
LGTM

PR #1379, opened 2016-01-14
Fix references to WebSocketReceiver.queue.

dwrensha:
Looks like this broke in one of the big ES6 changes: https://github.com/sandstorm-io/sandstorm/commit/eed6e5a2dab392f59d610a9ec8f3669e38a4be72 (which GitHub apparently can't render in-browser)

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -2108,8 +2108,8 @@ WebSocketReceiver = class WebSocketReceiver {
   }
 
   go() {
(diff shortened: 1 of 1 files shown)


dwrensha:
Without this fix, BrowserQuest totally fails.

I guess we don't have any unit tests that actually depend on websockets to grains actually working. @zarvox: would make sense to add a websocket test to https://github.com/zarvox/sandstorm-test-python ?

PR #1384, opened 2016-01-15
Makes contact input sized more flexibly

### LINKED ISSUE #1299: When sharing a grain through mail address the address is  truncated.
In "Send an Invite"  mail address get partly hidden  from under the  "G"  in "Get shareable link."  to the right.

jparyani:
Fixes #1299 

I'm not sure this is actually any clearer than  using a table, since I'd never heard of "block formatting contexts" before doing this. I'm also not sure if there's some easier way to do this with display: flex. Feel free to suggest alternatives.

### DIFF
--- a/shell/client/_topbar.scss
+++ b/shell/client/_topbar.scss
@@ -904,6 +904,7 @@ body>.popup {
         border: 0px;
         outline: 0px;
         width: 100%;
+        box-sizing: border-box;
       }
       &.active {
         outline: $focus-outline-color solid 1px;
--- a/shell/client/_topbar.scss
+++ b/shell/client/_topbar.scss
@@ -903,7 +903,7 @@ body>.popup {
       input.emails {
         border: 0px;
         outline: 0px;
-        width: 200px;
+        width: 100%;
       }
       &.active {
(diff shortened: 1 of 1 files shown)


kentonv:
Hmm, just width: 100% doesn't do the trick?

jparyani:
Setting width=100% will always put it on a new line, which looks a little janky when there's only 1 contact completed. I'm ok with doing that if you prefer to keep the css cleaner.

kentonv:
When I edit the `<input>` tag's style manually (via debug tools on Oasis) and add `width: 100%; box-sizing: border-box;` I seem to get the desired behavior. I don't see it go to the next line. Note that we're just trying to get the input box to match the width of the surrounding `<div>` which is creating the border.

Or is the goal to make the "can edit" drop-down appear on the same line as it used to? (Currently I *always* see it on a separate line.)

jparyani:
Ok, after discussing with Kenton, it's been changed to set the input's width to 100%.

kentonv:
I think you also need `box-sizing: border-box` otherwise it now runs over the right border.

jparyani:
Done

PR #1385, opened 2016-01-15
use /shared/ links in share-by-identity emails

dwrensha:
This allows us to display a helpful error message if the intended recipient is not logged in when they click the link:

access denied

### DIFF
--- a/shell/client/grainview.js
+++ b/shell/client/grainview.js
@@ -1,7 +1,7 @@
 let counter = 0;
 
 GrainView = class GrainView {
-  constructor(grainId, path, token, parentElement) {
+  constructor(grainId, path, tokenInfo, parentElement) {
     // `path` starts with a slash and includes the query and fragment.
     //
     // Owned grains:
@@ -22,7 +22,8 @@ GrainView = class GrainView {
     this._grainId = grainId;
     this._originalPath = path;
     this._path = path;
-    this._token = token;
+    this._tokenInfo = tokenInfo;
+    this._token = tokenInfo && tokenInfo._id;
     this._parentElement = parentElement;
     this._status = 'closed';
     this._dep = new Tracker.Dependency();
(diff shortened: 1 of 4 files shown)


paulproteus:
FWIW, what's needed to make it fully helpful is a clear statement of action
the user should take when they see this.

So perhaps add a message saying:

Visit your <a href="/account">account settings</a> to add that identity to
your account.

paulproteus:
Along those lines, I would avoid stating "Access Denied" since that sounds
like a fatal that a user cannot fix.

dwrensha:
Ok, now I suggest two possible actions:
actions


Also, here is what the logged-out case looks like:
logged out

dwrensha:
"Access Denied" is a concise and precise description of what's happening here. Do you have an alternative suggestion?

paulproteus:
Maybe:

- Sign in to access

- Sign in with another account

for the two cases? This avoids the problem (which IMHO is a problem) that
"Access Denied" sounds like fixing the problem would be hard.

neynah:
What about something like:

Logged in case:
Access to this grain is restricted to this identity
[identity]
To gain access, link this identity to your account or sign into the appropriate account.

Logged out case:
Access to this grain is restricted to this identity

(I'm also not sure how important it is to state "this grain was shared to the following identity")

dwrensha:
Garply, retest this please

dwrensha:
Okay, I've added an updated that uses @neynah's proposed wording, except instead of "access to this grain" I have "access through this URL", because it's possible that the user can access the grain through other means.

It maybe makes sense to redirect to the grain if the current user has access through other means. Or maybe in that case there should be a link such as "open this grain as <identity>, which already has access". However, it seems best to defer such extra functionality to a later pull request, if we decide that we want it.

paulproteus:
+1 to David's changed wording

I am also +1 redirecting to grain in case the user has access through other
means, but +1 on that being a separate pull request.

kentonv:
Garply, retest this please

dwrensha:
Garply, retest this please

Issue #1374, opened 2016-01-13
When BASE_URL changes, deconfigure OAuth providers more fully


paulproteus:
The plan, per @dwrensha , as I understand it:

1. disable all oauth services when their config is cleared (when BASE_URL changes)
2. reject empty-string configs in the /admin/settings UI
3. Show a pseudo-notification in /admin/settings explaining why this happened

paulproteus:
Refs: https://github.com/sandstorm-io/sandstorm/issues/1355 https://github.com/sandstorm-io/sandstorm/pull/1373

This probably supercedes the need for https://github.com/sandstorm-io/sandstorm/pull/1373

dwrensha:
Done in #1387.

paulproteus:
I'm so very happy about this!

PR #1387, opened 2016-01-15
Implement the plan described in #1374

dwrensha:
Disables oauth services on BASE_URL change, but does not delete the config. Displays red boxes in the admin area, explaining why the services were disabled.

Looks like this:
admin

### DIFF
--- a/shell/client/_admin.scss
+++ b/shell/client/_admin.scss
@@ -49,6 +49,9 @@
   .fade-out {
     opacity: .5;
   }
+  .reset-reason {
+    margin: 10px;
+  }
 }
 
 .settings-tabs {
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -839,14 +839,30 @@ limitations under the License.
   <form id="admin-settings-form">
     <p>
       <label><input type="checkbox" class="oauth-checkbox" name="googleLogin"
-             value="value" checked={{googleEnabled}} data-servicename="google"> Enable Google Login
-       (<a class="configure-oauth" data-servicename="google" href="">configure</a>)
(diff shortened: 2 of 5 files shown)


paulproteus:
A question, fwiw: will this "fix" (actively go ahead and disable)
misconfigured Sandstorm installs, where someone changed BASE_URL and it
half-disabled their Google auth?

I'd prefer to see a change where it does actively fully disable someone's
Google auth, rather than keep it half-enabled as it is now. Which is to say
I'm hoping that an auto-update causes the Google login button to go away
from https://xamar.sandcats.io/ and https://rose.sandcats.io/ to force
those admins to look into why logging in isn't working.

That doesn't have to prevent this from merging, but fwiw, that's a thing I
want.

kentonv:
I like it.

We should probably tweak the language to make clear that the user will need to modify the configuration at the Google or Github side. Probably we should explicitly tell them "You will need to update the OAuth redirect URL in the {Google Deveolper Console,Github application settings}".

I think that shoving a description string into the database probably isn't the right approach, especially when you consider localization (which we should do someday). Instead, we should define a specific set of reason IDs, or maybe the field should simply be `resetDueToBaseUrlChange` and we can add other flags later.

Please be sure to update the docs in db.js for the schema change.

@paulproteus People who modify BASE_URL after this change will have the "google" and "github" buttons disappear from the login drop-down, yes. I suppose we could retroactively "fix" existing servers using a migration that checks for the situation where the setting is enabled but the configuration is gone.

dwrensha:
OK. Updated to not store the description string in the database. Also, now clicking on the checkbox after an automatic reset will pop up the configuration form.

paulproteus:
: D​

PR #1396, opened 2016-01-20
Call browser.end() at the end of each unit test.

dwrensha:


dwrensha:
This ensures that nightwatch calls [DELETE session](https://code.google.com/p/selenium/wiki/JsonWireProtocol#DELETE_/session/:sessionId) after tests. Otherwise, as far as I can tell, Selenium never cleans up the open browser windows. (For example, if you watch the Selenium console output, after this change there are a bunch of lines like "Executing: [delete session: 7c309f57-0ad7-4988-9241-39e5fd0b0d18])" that didn't exist before. Or try running Selenium in non-headless mode!)

Also, this changes the appdemo test so that the user who installs the app does not create a grain. Perhaps that simplification will make the test somewhat more robust...

dwrensha:
Ah, now I'm noticing lines like "Session b413fa68-b2d7-4105-a41d-69c8cf52af37 deleted due to client timeout", so I guess that not calling `end()` is not disastrous. I still think the changes in this pull request are a good idea.

jparyani:
Should there be a .end() called in `tests/apps/powerbox.js`? Otherwise, LGTM

dwrensha:
> Should there be a .end() called in tests/apps/powerbox.js?

Added.

Current test failure:
```
TEST FAILURE: 1 assertions failed, 327 passed (4m 49s)
 ✖ apps/roundcube
   - Sending Mail
     Failed [equal]: ("Failed to start listening for SMTP server Error: listen EADDRINUSE" == "") - Expected "" but got: "Failed to start listening for SMTP server Error: listen EADDRINUSE"
```

jparyani:
Garply, retest this please

jparyani:
I think that error is spurious. If you want to track down the root cause for it as well, that'd be awesome. Feel free to do it in a separate PR, if you choose to work on it.

jparyani:
LGTM and all tests pass. Click merge whenever you're ready.

PR #1398, opened 2016-01-20
Initial version of the inline powerbox

jparyani:
It's still a bit rough around the edges, but the basic functionality
works.


tests/apps/powerbox.js
```diff
@@ -129,7 +131,70 @@ module.exports["Test Powerbox with failing requirements"] = function (browser) {
...
+        cancelable: true,
+      });
+      document.querySelector(".inline-powerbox").dispatchEvent(ev);
+    })
+    .frame("grain-frame")
+    .waitForElementVisible("#request-result", short_wait)
+    .assert.containsText("#request-result", "successfully fetched page");
+};
+
+module.exports["Install Faling Inline Powerbox"] = function (browser) {
```
dwrensha: s/Faling/Failing/

tests/apps/powerbox.js
```diff
@@ -129,7 +131,70 @@ module.exports["Test Powerbox with failing requirements"] = function (browser) {
...
+  browser
+    .init()
+    .installApp("http://sandstorm.io/apps/jparyani/inline-powebox-test-1.spk", "b58a280b3ca4dc72c8fc4c7b41d3e03d", "8stkfx4ez54109qzmzjtthaq105nf7f4sqfdzzp00g22p1r3uxg0")
+    .assert.containsText("#grainTitle", "Untitled InlinePowerboxTest");
+};
+
+module.exports["Test Faling Inline Powerbox"] = function (browser) {
```
dwrensha: s/Faling/Failing/

tests/apps/powerbox.js
```diff
@@ -129,7 +131,70 @@ module.exports["Test Powerbox with failing requirements"] = function (browser) {
...
             });
     })
+};
+
+// This tests the basic functionality of the inline powerbox.
+// Source: https://github.com/jparyani/sandstorm-test-app/tree/inline-powerbox
+module.exports["Install Inline Powerbox"] = function (browser) {
+  browser
+    .init()
+    .installApp("http://sandstorm.io/apps/jparyani/inline-powebox-test-1.spk", "b58a280b3ca4dc72c8fc4c7b41d3e03d", "8stkfx4ez54109qzmzjtthaq105nf7f4sqfdzzp00g22p1r3uxg0")
```
dwrensha: s/powebox/powerbox/ and update the name of the spk on the server

shell/server/core.js
```diff
@@ -226,6 +226,38 @@ Meteor.methods({
...
+      throw new Meteor.Error(403, "This identity doesn't have view permissions to the grain.");
+    }
+    const requirements = []; // We don't actually want the user's permission check as a requirement.
+    const grainOwner = {grain: {
+      grainId: grainId,
+      introducerIdentity: identityId,
+      saveLabel: url + " websession",
+    }};
+    const sturdyRef = waitPromise(saveFrontendRef(
+      {externalWebSession: {url: url}}, grainOwner, requirements)).sturdyRef;
```
dwrensha: Please update the schema in db.js for this new variant of `ApiToken.frontendRef`.

shell/shared/grain.js
```diff
@@ -1358,6 +1432,19 @@ if (Meteor.isClient) {
             return "remove";
           }
         });
+      } else if (event.data.inlinePowerbox) {
+        var inlinePowerbox = event.data.inlinePowerbox;
+        check(inlinePowerbox, Object);
+        var rpcId = inlinePowerbox.rpcId;
```
dwrensha: It would be good to check the type of `rpcId` here. Have we actually settled what type it is allowed to be? Your test app appears to use `Number` for this method. However, the `renderTemplate` method, above, requires that `rpcId` be a string.

zarvox: I'm inclined to say that we should either define `rpcId` to be `String` or we should simply let it be `Any`.  `rpcId` exists solely so that the app can give a unique identifier to the shell with a request so it can tell which request a `postMessage()`'d reply refers to, if there are multiple requests away at once.

Given that, rpcId might as well be `Any` - it's an opaque blob, the equivalent of a void* userdata in a struct.

ocdtrekkie:
What is this PR in comparison to Sandstorm today, @zenhack? Should this be closed?

zenhack:
It looks like this is a slightly different feature/design than what sandstorm currently has; if I'm reading the code correctly it shows an input just below the bottom of the grain frame, instead of a modal dialog floating over it. What I don't know is what the intent was here -- were folks expecting to have both UIs for some reason, or is this just obsolete? Is there a design doc somewhere?

@kentonv, any thoughts?

kentonv:
The "inline powerbox" refers to the ["autocomplete style" powerbox UI described in the roadmap](https://github.com/sandstorm-io/sandstorm/tree/master/roadmap/platform/powerbox#todofeature-autocomplete-style).

I do think this style of powerbox is important in the long term. Many kinds of everyday interactions -- such as selecting recipients of an e-mail -- become cumbersome with a full powerbox, but could be made to feel almost normal using the inline powerbox, without sacrificing security.

However, implementing the inline powerbox is pretty hairy, since it requires communication between the app and the shell to capture keystrokes and decide where to render the overlay. I think we should focus on lower-hanging fruit for the time being.

PR #1402, opened 2016-01-21
get the backgrounding test to pass on selenium 2.49

dwrensha:
Simple alternative to #1401.

### DIFF
--- a/shell/client/_topbar.scss
+++ b/shell/client/_topbar.scss
@@ -119,6 +119,7 @@ body>.topbar {
       background-size: 24px 24px;
       color: transparent;  // hide text
       width: 32px;
+      height: 32px;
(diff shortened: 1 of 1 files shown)


shell/client/_topbar.scss
```diff
@@ -119,6 +119,7 @@ body>.topbar {
       background-size: 24px 24px;
       color: transparent;  // hide text
       width: 32px;
+      height: 32px;
```
zarvox: prefer `$topbar-height-desktop`

dwrensha: Many other places in this file have `height: 32px`, including `body>.topbar` itself. I propose that we tidy such things up in a separate pull request.

PR #1404, opened 2016-01-21
Send all users (old and new), a notification about the referral program

jparyani:
This only effects servers with Meteor.settings.public.quotaEnabled set
to true (ie. only Oasis).

### DIFF
--- a/shell/packages/sandstorm-db/migrations.js
+++ b/shell/packages/sandstorm-db/migrations.js
@@ -411,6 +411,14 @@ function initServerTitleAndReturnAddress() {
   Settings.insert({_id: "returnAddress", value: "no-reply@" + hostname});
 }
 
+function sendReferralNotifications() {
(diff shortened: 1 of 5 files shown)


shell/packages/sandstorm-db/migrations.js
```diff
@@ -411,6 +411,14 @@ function initServerTitleAndReturnAddress() {
   Settings.insert({_id: "returnAddress", value: "no-reply@" + hostname});
 }
 
+function sendReferralNotifications() {
+  if (Meteor.settings.public.quotaEnabled) {
+    Meteor.users.find({loginIdentities: {$exists: 1}}, {fields: {_id: 1}}).forEach(function (user) {
+      sendReferralProgramNotification(user._id);
```
dwrensha: should demo users get this notification?

PR #1406, opened 2016-01-22
install.sh: Create error codes for install.sh failures

paulproteus:
NOTE that this changes the "fail" function to have two arguments.

Also:

- installer-tests: Adjust failure string to be more lenient.

- Capitalize the first word in a sentence.

- Include half-written code for sending errors to an Oasis grain, if
  users opt in. This is disabled by default at the moment.

- Avoid race condition in permissions for sandcats logs by creating
  the file with touch and then setting its permissions.

### DIFF
--- a/installer-tests/fail-on-32-bit-system.t
+++ b/installer-tests/fail-on-32-bit-system.t
@@ -2,7 +2,7 @@ Title: Can't install Sandstorm on old kernels
 Vagrant-Box: debian-7.8-32-nocm
 
 $[run]CURL_USER_AGENT=testing /vagrant/install.sh
-$[slow]Sorry, the Sandstorm server currently only runs on x86_64 machines.
-*** INSTALLATION FAILED ***
-Report bugs at: http://github.com/sandstorm-io/sandstorm
+$[slow]*** INSTALLATION FAILED ***
+Sorry, the Sandstorm server currently only runs on x86_64 machines.
+bugs at: http://github.com/sandstorm-io/sandstorm
 $[exitcode]1
--- a/install.sh
+++ b/install.sh
@@ -62,9 +62,45 @@ error() {
 }
 
 fail() {
+  local error_code="$1"
+  shift
+  if [ "${SHOW_FAILURE_MSG:-yes}" = "yes" ] ; then
+    echo "*** INSTALLATION FAILED ***" >&2
+    echo ""
+  fi
   error "$@"
-  echo "*** INSTALLATION FAILED ***" >&2
-  echo "Report bugs at: http://github.com/sandstorm-io/sandstorm" >&2
+  echo "" >&2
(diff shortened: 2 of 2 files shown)


paulproteus:
Eek, something has gone horribly wrong. Closing for now.

paulproteus:
This test was flakey for me locally:

- full-server-install-with-domain-reservation-token.t

Distressingly it seemed to hang before printing this line successfully:

> Successfully auto-configured HTTPS

...apparently this is an actual problem with GlobalSign. I've emailed them.

paulproteus:
And the tests pass on my machine. It's like there's a race condition where Garply isn't using the `install.sh` from the latest commit. Hmm.

paulproteus:
Garply, retest this please

kentonv:
I wonder if it's really worth asking the user for permission to send the error back, considering that the installer script is already obviously in communication with the server. I'm having trouble imagining who would actually object.

@ocdtrekkie what do you think?

ocdtrekkie:
I prefer keeping the option. Your communication with the server prior to is to download content (i.e., the user considers this data flow one way). Windows long set a comfortable precedent of asking before sending error reports.

You may note that with Windows 10, the vast majority of the telemetry they send is innately innocuous, but people are still very unhappy it is done without their consent. (I suppose one could always comment out part of the install.sh, mind you.)

ocdtrekkie:
As a general rule, I am usually comfortable sending data/feedback provided it is optional. If it is not optional, my first instinct is to wonder why it isn't.

installer-tests/fail-on-32-bit-system.t
```diff
@@ -2,7 +2,7 @@ Title: Can't install Sandstorm on old kernels
 Vagrant-Box: debian-7.8-32-nocm
 
 $[run]CURL_USER_AGENT=testing /vagrant/install.sh
-$[slow]Sorry, the Sandstorm server currently only runs on x86_64 machines.
-*** INSTALLATION FAILED ***
-Report bugs at: http://github.com/sandstorm-io/sandstorm
+$[slow]*** INSTALLATION FAILED ***
+Sorry, the Sandstorm server currently only runs on x86_64 machines.
+bugs at: http://github.com/sandstorm-io/sandstorm
```
kentonv: Did you mean to remove the word "Report"?

kentonv: Oh I guess this is some sort of substring match?

paulproteus: Yeah - this way it's a more lenient check - this test system tests for
substring presence, not strict equivalence.

The start of the string changed, so rather than hard-code the new start of
the string, I just stopped asserting its presence.

paulproteus: Yes (-:

jparyani:
Garply, retest this please

PR #1408, opened 2016-01-22
Introduce 'user.planBonus' for aggregating additional quota bonuses.

kentonv:
The Oasis payments module will be implementing new kinds of bonuses which will be aggregated into this field, such as a bonus if the user signs up for the mailing list, and bonuses from the indiegogo campaign.

Also eliminate a redundant query from referral processing.

### DIFF
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -82,7 +82,9 @@ if (Meteor.isServer && process.env.LOG_MONGO_QUERIES) {
 //                the invite was sent to.
 //   hasCompletedSignup: True if this account has confirmed its profile and agreed to this server's
 //                       terms of service.
-//   plan: _id of an entry in the Plans table which determines the user's qutoa.
+//   plan: _id of an entry in the Plans table which determines the user's quota.
+//   planBonus: {storage, compute, grains} indicating bonus amounts to add to the user's plan,
+//              filled in by the payments module. (Missing fields should be treated as zeros.)
 //   storageUsage: Number of bytes this user is currently storing.
 //   payments: Object defined by payments module, if loaded.
 //   dailySentMailCount: Number of emails sent by this user today; used to limit spam.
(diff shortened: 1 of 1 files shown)


shell/packages/sandstorm-db/db.js
```diff
@@ -82,7 +82,9 @@ if (Meteor.isServer && process.env.LOG_MONGO_QUERIES) {
 //                the invite was sent to.
 //   hasCompletedSignup: True if this account has confirmed its profile and agreed to this server's
 //                       terms of service.
-//   plan: _id of an entry in the Plans table which determines the user's qutoa.
+//   plan: _id of an entry in the Plans table which determines the user's quota.
+//   planBonus: {storage, compute, grains} indicating bonus amounts to add to the user's plan,
+//              filled in by the payments module. (Missing fields should be treated as zeros.)
```
paulproteus: Suggested small comment rewrite:

```
planBonus: {storage, compute, grains} bonus amounts to add to the user's plan. The payments module writes data here; we merely read it. Missing fields should be treated as zeroes. Does not yet include referral bonus, which is calculated separately.
```

shell/packages/sandstorm-db/db.js
```diff
@@ -590,7 +592,7 @@ isSignedUpOrDemo = function () {
   return false;
 }
 
-var calculateReferralBonus = function(accountId, plan) {
+var calculateReferralBonus = function(user, plan) {
```
paulproteus: Since you're now passing the actual `user` in here (although I'd rather call it `account`), you could stop passing in the `plan` parameter and pull it out of the user.

paulproteus:
+1 as-is

There are some suggestions as comments, which would be nice, but aren't necessarily essential. I did honestly get confused by the fact that this repo never writes to `planBonus` so I'd love to see the comment get changed.

shell/packages/sandstorm-db/db.js
kentonv: It appears to me that getUserQuota() has been calling this with the incorrect second parameter all along -- it passes in a plan object, but this function expects a string name. Why does this work?

kentonv: I think the answer is that we're currently enforcing the wrong quota, but displaying the right one.

kentonv:
Applied review comments and fixed apparent quota enforcement bug in 7a58764

PR #1409, opened 2016-01-22
Add docs on splash URL

paulproteus:
Also mention this from the "administrator's guide".


paulproteus:

PR #1411, opened 2016-01-22
Make dev login element a button when not expanded.

dwrensha:
I noticed that it was difficult to navigate to dev login using just the keyboard, probably because the dev login element in the login buttons list is a form, not a button. This makes it easier.


jparyani:
Garply, retest this please

Issue #1388, opened 2016-01-18
Wekan, Mediagoblin installations fail with "Package hash did not match"


rlafuente:
All parts of the Sandstorm installation went fine, and it's actually _fun_ to set up. Thanks everyone for making the process so smooth!

I did run into a problem on my self-hosted instance: certain apps throw an error when trying to install -- `Package hash did not match`. This happens with Wekan and Mediagoblin. All Ether* apps, Draw.io, Gitweb and a few others I tried install fine.

Relevant portion from `sandstorm.log`:

```
Downloading app: https://app-index.sandstorm.io/packages/7c76eaa536d4a28742e965ebccc1dc89
Failed to install app: Error: Package hash did not match.
    at AppInstaller.<anonymous> (server/installer.js:492:1)
    at IncomingMessage.<anonymous> (server/installer.js:365:1)
    at IncomingMessage.emit (events.js:117:20)
    at _stream_readable.js:944:16
    at process._tickCallback (node.js:448:13)
Downloading app: https://app-index.sandstorm.io/packages/35f82e4bf3ef0683558336494a3dc9d6
App ready: 35f82e4bf3ef0683558336494a3dc9d6
fetching keybase 5FDCAA736BE1402EDA0107A4F157DCE30ECD34E4
fetching keybase 5FDCAA736BE1402EDA0107A4F157DCE30ECD34E4
Downloading app: https://app-index.sandstorm.io/packages/54240bdb86c60e44b2bf0437561e236f
Failed to install app: Error: Package hash did not match.
    at AppInstaller.<anonymous> (server/installer.js:492:1)
    at IncomingMessage.<anonymous> (server/installer.js:365:1)
    at IncomingMessage.emit (events.js:117:20)
    at _stream_readable.js:944:16
    at process._tickCallback (node.js:448:13)
```

paulproteus:
Interesting! You're not the only person who's reported an issue like this.
I'd love to get to the bottom of it.

Please run:

wget
https://app-index.sandstorm.io/packages/7c76eaa536d4a28742e965ebccc1dc89
spk verify 7c76eaa536d4a28742e965ebccc1dc89


and see if that seems to print some JSON as output. If it does, then I
wonder what's going on!

rlafuente:
Okay, unexpected result that might yield a clue: the wget process gets repeatedly terminated between 20-30% wth the error: `Read error at byte 40255184/119446716 (The TLS connection was non-properly terminated.). Retrying.`

[...]

PR #1412, opened 2016-01-22
Fix autofocus in login dropdown and account dropdown.

dwrensha:
Fixes some fallout from https://github.com/sandstorm-io/sandstorm/commit/1887ba24dcf79d9e07dd75143d97b1c1748c9b65.

Currently, I see errors like `TypeError: Cannot read property 'focus' of null` when I expand the login menu.


dwrensha:
The recent build failures here and elsewhere have the error message "Failed to install app. Package hash did not match," as in #1388. The cases that fail are the ones that need to download from the Alpha server.

paulproteus:
...is this about CloudFlare, I wonder...?​

jparyani:
Garply, retest this please

PR #1415, opened 2016-01-23
installer-tests: Update order of FAILED vs. details

paulproteus:
Also hard-code the full "You can report bugs..." string.


paulproteus:
FWIW: I've run the two tests that I modified, and they now pass.

I'm going to self-merge since this is 100% installer-tests.

PR #1416, opened 2016-01-23
installer-tests: Adjust no-root.t for new strings

paulproteus:


paulproteus:

PR #1417, opened 2016-01-23
Allow "upgrade" when appVersion is equal.

kentonv:
Show the 'upgrade grains' button not only when grains with smaller version numbers exist, but also grains of the same version number but different package.

This fixes difficulties @ndarilek described.


paulproteus:
I'm not @zarvox but I had no idea we could satisfy @ndarilek 's desires with such a small change.

zarvox:
:+1:

PR #1422, opened 2016-01-24
Docs tooling cleanups

paulproteus:


paulproteus:
Self-merging because 100% in `docs/`

PR #1423, opened 2016-01-25
Ignore notifications in SandstormDb.deleteUnusedAccount().

dwrensha:
@phildini reported being unable to link an identity from a unused account on Oasis. The problem was that the account had received the referrals announcement notification, so Sandstorm did not consider it to be unused.

This patch changes the logic so that Sandstorm would still have considered that account to be unused. Deleting an account that has some notifications but nothing else should not be dangerous.


phildini:
I'm @phildini and I approve of this PR. :smile:

PR #1431, opened 2016-01-26
Add identity-linking test.

dwrensha:
This test is designed to prevent a reappearance of the bug fixed in https://github.com/sandstorm-io/sandstorm/commit/6e7887db839749e96b376387a3adf16d979ddd0e.

To make the implementation of the test nice, I also added a text input to the dev login form, so you can log in as any dev user. It looks like the following image:

with dev account

Note that I use `display: flex`.

Finally, this pull request gets rid of some introjs-related test flakiness, so that the tests pass more reliably on the latest Selenium.

### DIFF
--- a/shell/packages/sandstorm-accounts-ui/login_buttons.html
+++ b/shell/packages/sandstorm-accounts-ui/login_buttons.html
@@ -142,11 +142,9 @@
       <li><button type="button" class="login-dev-account" data-name="Dave Dev User">
           Dave
       </button></li>
-      <li>
-        <div class="flex-box">
-          <input name="name" class="custom-name" type="text" placeholder="Eve" required>
-          <button class="login-dev-account-custom" type="submit">Custom</button>
-        </div>
+      <li class="custom-name">
+        <input name="name" type="text" placeholder="Eve" required>
+        <button type="submit">Custom</button>
       </li>
     </ul>
   </form>
--- a/shell/packages/sandstorm-accounts-ui/login_buttons.html
+++ b/shell/packages/sandstorm-accounts-ui/login_buttons.html
@@ -127,7 +127,7 @@
 <template name="devLoginForm">
   {{#if expanded}}
   <form class="dev login expanded">
-    with a Dev account ▴
+    with a Dev account <button type="button" title="unexpand" class="unexpand"> ▴ </button>
     <ul class="dev-identities">
       <li><button type="button" class="login-dev-account" data-name="Alice Dev Admin"
                   data-is-admin="true">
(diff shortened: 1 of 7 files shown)


dwrensha:
I've added a commit that switches the "Test role sharing" test from using Etherpad to using GitWeb. This should be faster and more robust.

The problem I observed with Etherpad is that when an Etherpad grain is loading Firefox appears to simply swallow clicks sometimes. As in, the user clicks to navigate the sharing menu, but one of the clicks doesn't make it through. Selenium reports such clicks as success. I was actually able to observe such dropped clicks outside of Selenium. The dropped clicks seem to be correlated with spikes in client cpu usage when the grain is loading. It's also notable that the javascript console prints a lot of warnings/errors at such times, including one about a synchronous XMLHttpRequest.

zarvox:
big :+1: for digging into and fixing flaky tests.

shell/packages/sandstorm-accounts-ui/login_buttons.html
```diff
@@ -142,9 +142,12 @@
       <li><button type="button" class="login-dev-account" data-name="Dave Dev User">
           Dave
       </button></li>
-      <li><button type="button" class="login-dev-account" data-name="Eve Dev User">
-          Eve
-      </button></li>
+      <li>
+        <div class="flex-box">
```
kentonv: This class name specifies implementation detail rather than semantics.

Also, do you actually need a `<div>` here, or could you make the parent `<li>` be the flexbox?

I would give it `class="custom-name"`, then not assign classes to the inner elements but instead nest their selectors inside `.custom-name`.

shell/packages/sandstorm-accounts-ui/login_buttons.html
```diff
@@ -142,9 +142,12 @@
...
+          <input name="name" class="custom-name" type="text" placeholder="Eve" required>
```
kentonv: Doesn't marking this required mean that you can't submit the form by clicking the other buttons, e.g. "Bob"?

zarvox: Yes, it prevents you from submitting the form, but the other buttons don't trigger form submission anyway, they fire click events.

shell/packages/sandstorm-accounts-ui/login_buttons.html
dwrensha: Good idea! Done.

shell/packages/sandstorm-accounts-ui/login_buttons.html
dwrensha: Right, the other buttons have click handlers. It would perhaps be nicer for them not to be part of the form element, but then we would have to rejigger the styling, which currently benefits from closely following the pattern of the the email login form.

Issue #693, opened 2015-08-17
HTTP Proxy for outbound requests


JonTheNiceGuy:
Further to the email discussion with @kentonv:

> Currently there are several places in Sandstorm's code where we use Node's "http" and "https" modules to make outgoing HTTP requests.
> 
> It looks like the following files use these modules:
>
> shell/server/sandcats.js: To talk to the Sandcats server for dynamic DNS updates, if you use sandcats.
> shell/server/installer.js: To download packages.
> shell/server/hack-session.js: To implement the httpGet() API method, which e.g. TinyTinyRSS uses to fetch RSS feeds.
> shell/server/drivers/external-ui-view.js: To implement a more general mechanism for apps to make outgoing HTTP requests (not yet used by any apps).
>
> And this file uses Meteor's `HTTP` API:
>
> shell/packages/sandstorm-db/user.js: To fetch profile pictures from Google and Github.
>
> Meteor's HTTP is a wrapper around the `request` npm package, which in turn wrap's node's HTTP client. It looks like `request` explicitly supports proxies (and Meteor's HTTP supports passing arbitrary options to `request`):
>
> https://www.npmjs.com/package/request#proxies
>
> It's also possible to get Node's raw HTTP client to use proxies per this StackOverflow question:
>
> http://stackoverflow.com/questions/3862813/how-can-i-use-an-http-proxy-with-node-js-http-client
>
> So we need to make sure that all the places where we make HTTP requests use the proxy.
>
> I think the right way to do this would be to introduce a new Sandstorm-internal HTTP API that either wrap's Meteor's `HTTP` or the npm `request` package (probably the latter) and allows us to force all requests globally through a porxy, configurable through the admin interface. Then we should update all the above code to use it.
>
> Additionally, run-bundle.c++ shells out to `curl` to fetch Sandstorm updates. I believe you can make `curl` use a proxy by setting an environment variable, so if you set that variable in the initscript that calls `sandstorm start` it should apply. (Maybe it makes sense for the Sandstorm shell to pick up the http_proxy environment variable automatically as well, rather than require setting it in the admin settings?)

A-Picogna:
Hi,

I also have a problem with proxy, there is a way to add it to sandstorm for the moment ? to force the plate-forme going through it ?

paulproteus:
Hi @A-Picogna ,

It's your lucky week! I'm testing this right now and one of my goals for the week is to make sure that this works.

[...]

PR #1433, opened 2016-01-27
(Mock-up, do not merge) Ensure that we can install apps when using a HTTP proxy

paulproteus:
This is not ready for prime-time, but it does work.

- Question: What _is_ a good way to get access to `request` from the node world? Should I make a package wrapping it, vaguely similar to what I did for `introjs`? Should I try to `Meteor.require("../../something/request/")` instead and "steal" request from the Meteor packages that bundle it? (Namely `HTTP` probably.)"

- I'll take any & all feedback, even if it's obvious, but feel free to alternatively ignore this until I submit it for final review.

This pull request:

- Copies the `http_proxy` and `https_proxy` (lowercase!) environment variables at Sandstorm start time

- Adds a `meteorhacks:npm`-based package to depend on `request` from the `npm` world. (See above on the fact that this is hacky.)

- Adjusts the SPK download flow to use `request`, which does respect `http_proxy`, unlike the `http` and `https` node modules which we used to use.

- Keybase info fetching seems to "just work" due to Meteor's built-in support for HTTP proxies.

- @kentonv you remarked in the app download code that Sandstorm would hopefully not need a CA bundle. But we are now using HTTPS as our only trust mechanism for Keybase, so presumably it does need a CA bundle.

Future work:

- I need to test other flows.

- Clean up this patch to be more reasonable - avoid `meteorhacks:npm` and make sure I'm being reasonable with memory (do I need to `free()` things?) in `run-bundle.c++` patch.


paulproteus:
see also https://github.com/sandstorm-io/sandstorm/issues/693

kentonv:
Regarding CA bundles, I think for some reason I assumed we weren't shipping one since I hadn't, you know, done any particular work to make sure that we were. But the fact that the keybase requests work implies that we are, in fact, shipping a CA bundle. So I guess node just takes care of that. I think it's probably fine, then, to check certificates even in cases where we know it's redundant (package downloads, where we already have a verified hash).

paulproteus:
Closing this WIP pull request in favor of #1438 that was successfully merged.

PR #1438, opened 2016-01-28
Add support for using a HTTP proxy for outbound connections (just for app install for now)

### LINKED ISSUE #693: HTTP Proxy for outbound requests

paulproteus:
This adjusts `installer.js` to use the `request` API, rather than the node
HTTP/HTTPS API. It also adjusts `run-bundle.c++` to pass this variable through,
and adjusts the docs to explain how to use this feature.

Refs #693

Does not fully close #693 yet because a few other places need a similar update.

### DIFF
--- a/shell/server/installer.js
+++ b/shell/server/installer.js
@@ -310,92 +310,74 @@ AppInstaller = class AppInstaller {
     };
 
     let protocol;
-    if (url.protocol === 'http:') {
-      protocol = Http;
-    } else if (url.protocol === 'https:') {
+    if (url.protocol === 'http:' || url.protocol === 'https:') {
+      protocol = Request.defaults({
+        maxRedirects: 1,
       // Since we will verify the download against a hash anyway, we don't need to verify the server's
       // certificate. In fact, the only reason we support HTTPS at all here is because some servers
       // refuse to serve over HTTP (which is, in general, a good thing). Skipping the certificate check
       // here is helpful in that it means we don't have to worry about having a reasonable list of trusted
       // CAs available to Sandstorm.
-      options.rejectUnauthorized = false;
-      protocol = Https;
+        strictSSL: false
+      });
     } else {
       throw new Error('Protocol not supported: ' + url.protocol);
     }
 
     // TODO(security):  It could arguably be a security problem that it's possible to probe the
     //   server's local network (behind any firewalls) by presenting URLs here.
-    const request = protocol.get(options, this.wrapCallback((response) => {
-      if (response.statusCode === 301 ||
-          response.statusCode === 302 ||
-          response.statusCode === 303 ||
-          response.statusCode === 307 ||
-          response.statusCode === 308) {
-        // jscs:disable requireDotNotation
-        // Got redirect. Follow it.
-        this.url = Url.resolve(this.url, response.headers['location']);
-        this.doDownloadTo(out);
-        return;
+    let bytesExpected = undefined;
+    let bytesReceived = 0;
+    const hasher = Crypto.createHash('sha256');
+    let done = false;
+    const updateDownloadProgress = _.throttle(this.wrapCallback(() => {
+      if (!done) {
+        if (bytesExpected) {
+          this.updateProgress('download', bytesReceived / bytesExpected);
+        } else {
+          this.updateProgress('download', bytesReceived);
+        }
       }
+    }), 500);
 
-      if (response.statusCode !== 200) {
-        throw new Error('Download failed with HTTP status code: ' + response.statusCode);
-      }
+    const request = protocol.get(this.url);
 
-      let bytesExpected = undefined;
-      let bytesReceived = 0;
-
+    request.on('response', this.wrapCallback((response) => {
       if ('content-length' in response.headers) {
         bytesExpected = parseInt(response.headers['content-length']);
       }
-
-      let done = false;
-      const hasher = Crypto.createHash('sha256');
-
-      const updateDownloadProgress = _.throttle(this.wrapCallback(() => {
-        if (!done) {
-          if (bytesExpected) {
-            this.updateProgress('download', bytesReceived / bytesExpected);
-          } else {
-            this.updateProgress('download', bytesReceived);
-          }
-        }
-      }), 500);
-
-      response.on('data', this.wrapCallback((chunk) => {
-        hasher.update(chunk);
-        out.write(chunk);
-        bytesReceived += chunk.length;
-        updateDownloadProgress();
-      }));
-
-      response.on('end', this.wrapCallback(() => {
-        out.done();
-
-        if (hasher.digest('hex').slice(0, 32) !== this.packageId) {
-          throw new Error('Package hash did not match.');
-        }
-
-        done = true;
-        delete this.downloadRequest;
-
-        this.updateProgress('unpack');
-        out.saveAs(this.packageId).then(this.wrapCallback((info) => {
-          this.appId = info.appId;
-          this.authorPgpKeyFingerprint = info.authorPgpKeyFingerprint;
-          this.done(info.manifest);
-        }), this.wrapCallback((err) => {
-          throw err;
-        }));
-      }));
-
-      response.on('error', this.wrapCallback((err) => { throw err; }));
     }));
 
-    this.downloadRequest = request;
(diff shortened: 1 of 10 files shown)


shell/server/installer.js
```diff
@@ -310,92 +310,74 @@ AppInstaller = class AppInstaller {
     };
 
     let protocol;
-    if (url.protocol === 'http:') {
-      protocol = Http;
-    } else if (url.protocol === 'https:') {
+    if (url.protocol === 'http:' || url.protocol === 'https:') {
+      protocol = Request.defaults({
+        maxRedirects: 1,
```
kentonv: Why only one? Note that we have historically had app URLs that were redirects, and double-redirects are not that uncommon.

shell/server/installer.js
```diff
@@ -310,92 +310,74 @@ AppInstaller = class AppInstaller {
...
       // CAs available to Sandstorm.
-      options.rejectUnauthorized = false;
-      protocol = Https;
+        strictSSL: false
+      });
     } else {
       throw new Error('Protocol not supported: ' + url.protocol);
     }
 
     // TODO(security):  It could arguably be a security problem that it's possible to probe the
```
kentonv: Hey I think you're fixing this TODO! (Networks with significant private internals should always specify a proxy.)

src/sandstorm/run-bundle.c++
```diff
@@ -1162,12 +1162,23 @@ private:
 
     // The environment inherited from the host is probably no good for us. E.g. an oddball
     // locale setting can crash Mongo because we don't have the appropriate locale files available.
+    //
+    // That said, there are a few environment variables that we do re-export.
+    char * http_proxy, * https_proxy;
```
kentonv: Generally in C++ it's recommended to associate the `*` with the type, i.e. `char* x` not `char *x`. This of course leads to confusion when declaring multiple variables at once as you are doing here, and generally the answer to that is: don't do that. :) That is, declare them on separate lines.

src/sandstorm/run-bundle.c++
```diff
@@ -1162,12 +1162,23 @@ private:
...
+    KJ_SYSCALL(http_proxy = getenv("https_proxy"));
```
kentonv: `s/https/http/`

paulproteus: +1

src/sandstorm/run-bundle.c++
paulproteus: +1

shell/server/installer.js
paulproteus: How about 5? : D

I just want to pick a constant, any constant.

src/sandstorm/run-bundle.c++
kentonv: Also please use camelCase, not under_scores.

shell/server/installer.js
paulproteus: Clearly you've never worked at Eventbrite!

Which is to say - I think there should be some other mechanism for avoiding "server side request forgery". And/or we should document the heck out of this.

src/sandstorm/run-bundle.c++
kentonv: `KJ_SYSCALL` only works for functions that return `int`. Also it's only meant for calls that can fail, but getenv() cannot fail (it returns null when there's no match, but that's not a failure).

src/sandstorm/run-bundle.c++
paulproteus: Oh, I was mirroring the environment variable name. +1, I can use camelCase for this.

src/sandstorm/run-bundle.c++
```diff
@@ -1162,12 +1162,23 @@ private:
...
+    KJ_SYSCALL(https_proxy = getenv("https_proxy"));
     KJ_SYSCALL(clearenv());
```
kentonv: I believe that this `clearenv()` call frees the memory that your `http_proxy` and `https_proxy` pointers are pointing at. You will need to copy the strings, preferably using `kj::str()` -- but of course only if they aren't null.

paulproteus: "it seems to work fine" (but I share your assessment from the man page)

+1 to `kj::str()` to copy

shell/server/installer.js
kentonv: Apparently Chrome and Firefox use a limit of 20 by default.

src/sandstorm/run-bundle.c++
paulproteus: Cool, I will try without `KJ_SYSCALL`.

shell/server/installer.js
paulproteus: Fixed in latest push

paulproteus:
@kentonv , how do you suggest I express the following?

```
    const char * keepTheseVars[] = {"http_proxy", "https_proxy"};
    auto numberOfVarsToKeep = 2; // FIXME: How am I supposed to express this & the previous line?
    for (int i = 0; i < numberOfVarsToKeep; i++) {
```

Rationale for doing it this way: I had typos in `https_proxy` vs. `http_proxy` earlier, and so I want to do this in a less repeating-myself fashion to avoid future typos and increase clarity. But I don't know how to make a literal string array in KJ/C++.

Other than the above style question, I believe this addresses all review feedback.

kentonv:
`kj::size(keepTheseVars)` will evaluate to `2`.

Also, in C++11 you can write:

    for (char* var: keepTheseVars) {

paulproteus:
Thanks @kentonv 

I've pushed a style fix for that, and additionally I've manually re-tested that app installs work, for both http and https app URLs. Ready for your re-review/merge.

src/sandstorm/run-bundle.c++
paulproteus: Fixed by avoiding having so many temp variables in the first place.

src/sandstorm/run-bundle.c++
paulproteus: Fixed by not using `KJ_SYSCALL` on getenv, merely on setenv, which can error.

paulproteus:
Pushed a fix to my "style fix" which makes it actually compile. I will let Jenkins test this; anyway, I did also re-manually-test this locally.

kentonv:
I added some OCD style tweaks directly in github. I'll wait for Jenkins to verify I didn't typo anything then merge.

PR #1442, opened 2016-01-29
Add / extend Cap'n Proto interfaces for key drivers.

kentonv:
Note that while the outgoing-HTTP and email drivers may be implemented as pseudo-drivers, I think that the web publishing driver may have no such luck, as it needs to store significant amounts of data and we will want to charge that against the user's quota. Plus we don't have a great place to store the data if not in a grain.

@zarvox @jparyani

### DIFF
--- a/src/sandstorm/api-session.capnp
+++ b/src/sandstorm/api-session.capnp
@@ -30,7 +30,103 @@ interface ApiSession @0xc879e379c625cdc7 extends(WebSession.WebSession) {
     # it.  However, for those that benefit from it (like analytics), clients can opt into passing
     # their IP on to the backend by adding an "X-Sandstorm-Passthrough: address" header to their
     # request.  This would be a privacy leak for WebSession, since the grain can give the client
-    # scripts which would send the header, but ApiSession requires a user action, so it's safe here.
+    # scripts which would send the header, but ApiSession requires a user action, so it's safe
+    # here.
     remoteAddress @0 :IpAddress;
+  }
+
+  struct PowerboxTag {
+    # A tag used for PowerboxDescriptors describing HTTP APIs. Such descriptors should typically
+    # indicate `ApiSession` as the requested interface and use HttpApiTag to specify what API is
+    # expected.
+    #
+    # Usually, this tag is used to request APIs implemented by internet services external to
+    # Sandstorm. However, a Sandstorm grain may advertise itself as supporting a compatible API
+    # to the Powerbox, and if so the Powerbox will allow the user to select that grain for such
+    # requests. Note that publishing an `ApiSession` has nothing to do with `UiView` or Sandstorm's
+    # normal HTTP API support.
+    #
+    # Requests of this type trigger the HTTP driver, such that the powerbox UI will display a
+    # special flow for connecting to external web services. The Powerbox will also surface other
+    # grains that advertise their ability to handle these requests under the usual matching rules.
+    # This struct is carefully designed such that the usual Powerbox query rules make sense for
+    # performing such queries.
+    #
+    # If you make a Powerbox request containing multiple PowerboxDescriptors describing HTTP APIs,
+    # the HTTP driver will offer the user a choice among the options. This can make sense for, say,
+    # allowing the user to choose from multiple popular servers implementing the same protocol, or
+    # to choose whether or not to grant the app optional permissions (OAuth scopes). The HTTP
+    # driver will return a capability with a descriptor matching one of the requested descriptors
+    # which best-describes the user's choice.
+
+    canonicalUrl @0 :Text;
+    # The standard URL at which this service is found. Use this especially for traditional SaaS
+    # services. For example, you might request "https://api.github.com" to request access to the
+    # GitHub API.
+    #
+    # If you include a path in the URL, requests will automatically be prefixed with that path. You
+    # should usually include enough path components to identity a specific product and version, but
+    # usually not specific collection types within that product. For example, you would request the
+    # Google Calendar API version 2 as "https://apidata.googleusercontent.com/caldav/v2". However,
+    # you would not normally request "https://api.github.com/users" because "GitHub users" is not
+    # considered a separate API but rather one part of the overall GitHub API. Note that
+    # `canonicalUrl` should never end with a '/', because request paths to `ApiSession` are
+    # requried to start with a '/', so this would result in two consecutive '/'s which is usually
+    # wrong.
+    #
+    # The HTTP driver will present `canonicalUrl` as a strong suggestion to the user. However, the
+    # user is always allowed to substitute a different URL instead, causing requests to be
+    # redirected to some other service. This is the user's choice. In cases where the app does not
+    # trust the user, the app will need to defend itself against the possibility that the user
+    # connects it to a malicious server.
+    #
+    # The Powerbox will also offer matches published by other grains using the usual query matching
+    # rules. That is, a grain may advertise that it can handle queries for HTTP APIs with a
+    # particular `canonicalUrl`, indicating that the grain offers ApiSession capabilities
+    # implementing a compatible protocol.
+
+    struct OAuthScope {
+      name @0 :Text;
+    }
+
+    oauthScopes @1 :List(OAuthScope);
+    # List of OAuth scopes required by the requesting app. OAuth APIs usually publish a list
+    # of "scope" names representing permissions that can be requested. For example, see:
+    #
+    #     https://developer.github.com/v3/oauth/#scopes
+    #
+    # When this list is present (even if empty), it indicates that the requested API requires
+    # OAuth-based authentication. The HTTP driver will guide the user through connecting their
+    # Sandstorm account to the remote service and requesting the appropriate permissions.
+    #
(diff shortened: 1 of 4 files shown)


src/sandstorm/web-publishing.capnp
```diff
@@ -0,0 +1,110 @@
...
+  # under that URL. Note that the web driver assigns *no* special meaning to the character '/', so
+  # you should use a suffix ending in '/' if you want a site that represents a subdirectory.
+
+  # TODO(someday): Allow registering dynamaic handlers.
+
+  # TODO(someday): Allow transactionally changing a bunch of things at once.
+}
+
+interface WebResource {
+  # Represents an HTTP resource. Although StaticResource is an interface, an implementation must
```
dwrensha: I guess you renamed this from `StaticResource` to `WebResource` at some point?

src/sandstorm/email.capnp
```diff
@@ -27,8 +27,17 @@ using Util = import "util.capnp";
 struct EmailAddress {
   # Email addresses are (usually) of the form "Full Name <example@example.org>".
   # This struct separates the display name and address into distinct fields.
+
   address @0 :Text;
   name @1 :Text;
+
+  # TODO(someday): We could add a field `token` which is a capability representing "email captial",
```
dwrensha: s/captial/capital/

src/sandstorm/web-publishing.capnp
kentonv: Yikes, WebResource is an older version of my thoughts which I meant to delete entirely before pushing, but apparently forgot. Removed.

src/sandstorm/email.capnp
kentonv: fixed

src/sandstorm/web-publishing.capnp
```diff
@@ -0,0 +1,81 @@
...
+# limitations under the License.
+
+@0xd5d3e63bd0a552b6;
+
+$import "/capnp/c++.capnp".namespace("sandstorm");
+
+using Util = import "util.capnp";
+using Handle = Util.Handle;
+
+interface WebSite {
```
zarvox: We should specify an explicit stable interface ID, if we want people to powerbox request it.

src/sandstorm/web-publishing.capnp
```diff
@@ -0,0 +1,81 @@
...
+  # until enough data has actually been uploaded for them to respond. If the stream is dropped
+  # without calling `done()`, the blob will be left in a broken state in which some methods will
+  # throw exceptions.
+
+  getSubsite @4 (suffix :Text) -> (site :WebSite);
+  # Append `suffix` to the URL and return a `WebSite` capability that can only modify resources
+  # under that URL. Note that the web driver assigns *no* special meaning to the character '/', so
+  # you should use a suffix ending in '/' if you want a site that represents a subdirectory.
+
+  # TODO(someday): Allow registering dynamaic handlers.
```
zarvox: s/dynamaic/dynamic/

src/sandstorm/web-publishing.capnp
```diff
@@ -0,0 +1,81 @@
...
+
+    body :union {
+      bytes @3 :Data;
+      # Use when content is smaller than 1MB.
+
+      blob @4 :Util.Blob;
+      # Use when content is larger than 1MB. Currently, the blob must be created via uploadBlob().
+    }
+  }
+
```
zarvox: Is there any way to enumerate the currently-known entities?  This API appears to require that you know a priori the exact full path of anything that you want to examine or modify, which may make removing content challenging.

src/sandstorm/web-publishing.capnp
```diff
@@ -0,0 +1,81 @@
...
+    mimeType @0 :Text;  # Content-Type header.
+    language @1 :Text;  # Content-Language header (optional).
+    encoding @2 :Text;  # Content-Encoding header (optional).
+
+    body :union {
+      bytes @3 :Data;
+      # Use when content is smaller than 1MB.
+
+      blob @4 :Util.Blob;
+      # Use when content is larger than 1MB. Currently, the blob must be created via uploadBlob().
```
dwrensha: I suppose the idea here is that someday this blob could alternatively come from an object-storage grain?

src/sandstorm/web-publishing.capnp
```diff
@@ -0,0 +1,81 @@
...
+    }
+  }
+
+  getEntities @1 (path :Text) -> (entities :Util.Assignable(List(Entity)));
+  # Get the list of static HTTP entities that make up the resource at the given path. Often there
+  # is only one entity, but if you provide multiple entities, one will be chosen based on "Accept"
+  # headers provided by the client. Setting the entity list empty effectively deletes the resource;
+  # the client will receive a 404 error, which can be customizedb using `getNotFoundEntities()`.
```
dwrensha: s/customizedb/customized/

src/sandstorm/api-session.capnp
```diff
@@ -30,7 +30,103 @@ interface ApiSession @0xc879e379c625cdc7 extends(WebSession.WebSession) {
...
     # request.  This would be a privacy leak for WebSession, since the grain can give the client
-    # scripts which would send the header, but ApiSession requires a user action, so it's safe here.
+    # scripts which would send the header, but ApiSession requires a user action, so it's safe
+    # here.
     remoteAddress @0 :IpAddress;
   }
+
+  struct PowerboxTag {
+    # A tag used for PowerboxDescriptors describing HTTP APIs. Such descriptors should typically
+    # indicate `ApiSession` as the requested interface and use HttpApiTag to specify what API is
```
dwrensha: Should this `HttpApiTag` be defined somewhere?

src/sandstorm/api-session.capnp
```diff
@@ -30,7 +30,103 @@ interface ApiSession @0xc879e379c625cdc7 extends(WebSession.WebSession) {
...
+    # expected.
```
dwrensha: So the ID of `PowerboxTag` is meant to be used in `PowerboxDescriptor.Tag.id`, with a value of type `PowerboxTag` in `PowerboxDescriptor.Tag.value`?

src/sandstorm/api-session.capnp
dwrensha: This makes more sense if I s/HttpApiTag/PowerboxTag/.

src/sandstorm/api-session.capnp
```diff
@@ -30,7 +30,103 @@ interface ApiSession @0xc879e379c625cdc7 extends(WebSession.WebSession) {
...
+    # driver will return a capability with a descriptor matching one of the requested descriptors
+    # which best-describes the user's choice.
+
+    canonicalUrl @0 :Text;
+    # The standard URL at which this service is found. Use this especially for traditional SaaS
+    # services. For example, you might request "https://api.github.com" to request access to the
+    # GitHub API.
+    #
+    # If you include a path in the URL, requests will automatically be prefixed with that path. You
+    # should usually include enough path components to identity a specific product and version, but
```
dwrensha: s/identity/identify/

dwrensha:
I don't see anything in web-publishing.capnp about 300-level redirect responses. Do we want to support redirects, and how would they fit in with what you've designed here?

src/sandstorm/web-publishing.capnp
```diff
@@ -0,0 +1,81 @@
...
+interface WebSite {
+  # Represents an HTTP URL to which a web site can be published.
+  #
+  # Make a powerbox request for `WebSite` when you wish to publish a web site accessible over HTTP
+  # without going through the Sandstorm UI.
+
+  getUrl @0 () -> (path :Text);
+  # Get the URL of this resource.
+
+  struct Entity {
```
zarvox: This interface should probably grow support for permitting cross-origin queries, so we can remove the hacks at https://github.com/sandstorm-io/sandstorm/blob/cf984260cca26a1d5b7abad82ee8fed73e5ff2d9/shell/server/pre-meteor.js#L86..L97

src/sandstorm/web-publishing.capnp
kentonv: Yes. Or, raw Blackrock storage.

src/sandstorm/api-session.capnp
kentonv: Yeah, I meant PowerboxTag. Went through some renames.

src/sandstorm/api-session.capnp
kentonv: No, the type ID would be ApiSession's ID. The docs for PowerboxDescriptor specify that when Tag.id is an interface ID, Tag.value's type is specified by the interface's dogs. ApiSession specifies that ApiSession.PowerboxTag shall be used as the tag type.

src/sandstorm/web-publishing.capnp
kentonv: Actually, I'd argue that this is best controlled not by the app or even the user, but by the server admin. If the system is on the public internet, then it's safe and appropriate to set `access-control-allow-origin: *` on all published sites. If the system is behind a firewall (only visible on some corp network), then `access-control-allow-origin` should specify only sites that are allowed to reach into the firewall. The app doesn't really have any reason to be expressing an opinion on the matter.

PR #1443, opened 2016-01-29
Request Access button

dwrensha:
When a user visits a grain they are unauthorized to access, displays a button that requests access from the grain owner. The button looks like this:

requestaccess0

After the button is clicked, the grain owner gets an email that looks like this:

requestaccess

The "Open Sharing Menu" button is a link to `/share/:grainId/:identityId`, which redirects to the grain with the sharing menu popped up and prepopulated with the requester's identity.

The `replyTo` field in the email is the address of the access-requester.

### DIFF
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -202,7 +202,9 @@ limitations under the License.
          <p>Sending your request...</p>
       {{/if}}
       {{#if success}}
-         <p>Your request has been sent.</p>
+        <p>
+          Your request has been sent. You will receive an email if your request gets approved.
+        </p>
       {{/if}}
       {{#with error}}
         <p>Your request failed because: {{.}}</p>
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -192,7 +192,7 @@ limitations under the License.
 
 <template name="requestAccess">
   <div class="grain-interstitial">
-  <p>You currently do not have permission to access this grain.</p>
+  <p>You do not have permission to access this grain.</p>
   {{#if currentUser}}
     {{#with status}}
       {{#if showButton}}
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -141,7 +141,11 @@ limitations under the License.
   {{#with unpackedGrainState}}
     <div class="grain-container {{#if active}}active-grain{{else}}inactive-grain{{/if}}">
     {{#if error}}
-      <pre>{{error}}</pre>
+      {{#if unauthorized}}
+        {{> requestAccess}}
+      {{else}}
+        <pre>{{error}}</pre>
+      {{/if}}
     {{else}}
     {{#if appOrigin}}
       {{!-- Selenium requires iframes to have an id in order to select them. id="grain-frame" is only
@@ -152,14 +156,14 @@ limitations under the License.
       {{/if}}
     {{else}}
     {{#if interstitial.chooseIdentity}}
-      <div class="interstitial-button-box">
+      <div class="grain-interstitial">
         <p>With which of your identities would you like to open this grain?</p>
         {{> identityPicker identityPickerData}}
         <button class="incognito-button" data-token="{{token}}">Open in Incognito Mode</button>
       </div>
     {{else}}
       {{#if interstitial.directShare}}
-        <div class="direct-share-denied">
+        <div class="grain-interstitial">
           <p>
             Access through this URL is restricted to this identity:
           </p>
@@ -184,6 +188,30 @@ limitations under the License.
     {{/if}}
     </div>
   {{/with}}
+</template>
+
+<template name="requestAccess">
+  <div class="grain-interstitial">
+  <p>You currently do not have permission to access this grain.</p>
+  {{#if currentUser}}
+    {{#with status}}
+      {{#if showButton}}
+        <button class="request-access">Request Access</button>
+      {{/if}}
+      {{#if waiting}}
+         <p>Sending your request...</p>
+      {{/if}}
+      {{#if success}}
+         <p>Your request has been sent.</p>
+      {{/if}}
+      {{#with error}}
+        <p>Your request failed because: {{.}}</p>
+      {{/with}}
+    {{/with}}
+  {{else}}
+    Please sign in to request access.
+  {{/if}}
+  </div>
 </template>
 
 <template name="invalidToken">
@@ -486,7 +514,7 @@ limitations under the License.
   <form class="email-invite">
     <p>
     <img class="inline-icon" src="/people-m.svg" role="presentation">
-    {{> contactInputBox contacts=contacts}}
+    {{> contactInputBox contacts=contacts preselectedIdentityId=preselectedIdentityId}}
     {{> selectRole viewInfo=viewInfo}}
     </p>
     <div>
@@ -655,6 +683,16 @@ limitations under the License.
   {{#if showPowerboxOffer}}
     {{>sandstormTopbarItem name="offer" topbar=globalTopbar template="grainPowerboxOffer"
         startOpen=true popupTemplate="grainPowerboxOfferPopup" data=powerboxOfferData}}
+  {{/if}}
+</template>
+
+<template name="share">
+  {{#if currentUser}}
+    {{#with grainNotFound}}
+      No grain found with ID {{.}}. Maybe you need to sign in to a different account?
+    {{/with}}
+  {{else}}
+    You must be signed in to grant access to a grain.
   {{/if}}
 </template>
 
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -210,7 +210,7 @@ limitations under the License.
       {{/if}}
       {{#if success}}
         <p>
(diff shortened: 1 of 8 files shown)


shell/shared/grain.js
```diff
@@ -386,6 +389,57 @@ Meteor.methods({
...
+      globalDb.addContact(grainOwner._id, identityId);
+
+      const fromEmail = "Sandstorm server <no-reply@" + HOSTNAME + ">";
+      const senderEmails = SandstormDb.getVerifiedEmails(identity);
+      const replyTo = senderEmails.length > 0 ? senderEmails[0] : fromEmail;
+
+      // TODO(soon): In the HTML version, we should display an identity card.
+      let identityNote = "";
+      if (identity.profile.service === "google") {
+        identityNote = " (" + identity.privateIntrinsicName + " on Google)";
```
kentonv: Is privateIntrinsicName a display name or an email address?

I think it needs to be an email address, since that's what we can verify. In that case I'd remove the "on Google" part; that's not really relevant.

shell/shared/grain.js
```diff
@@ -386,6 +389,57 @@ Meteor.methods({
...
+      } else if (identity.profile.service === "github") {
+        identityNote = " (" + identity.profile.intrinsicName + " on GitHub)";
+      } else if (contact.profile.service === "email") {
+        identityNote = " (" + identity.profile.intrinsicName + ")";
+      }
+
+      const message = identity.profile.name + identityNote +
```
kentonv: I wonder if it's a phishing risk to include the requester's display name. We do include the verified email/username immediately thereafter which means we provide enough information to recognize impostors but I bet a lot of people won't look closely enough. Arguably we should remove the display name and only show the email address / github username which we can verify.

OTOH, in order to request access the user already has to have obtained the grain link which is itself a barrier to malicious users sending a request at all. Maybe at this point the usability benefit outweighs the phishing concern?

Thoughts? @zarvox interested in your take as well.

shell/shared/grain.js
```diff
@@ -386,6 +389,57 @@ Meteor.methods({
...
+      const email = _.findWhere(SandstormDb.getUserEmails(grainOwner), {primary: true});
+      if (!email) {
+        throw new Meteor.Error("no email", "Grain owner has no email address.");
+      }
+      const emailAddress = email.email;
+
+      const identity = globalDb.getIdentity(identityId);
+      globalDb.addContact(grainOwner._id, identityId);
+
+      const fromEmail = "Sandstorm server <no-reply@" + HOSTNAME + ">";
```
kentonv: If you want this to come from the server, use `db.getServerTitle()` and `db.getReturnAddress()`.

But note that you could actually make the "from" be the sender's address as long as the envelope sender is the Sandstorm server. See e.g. how we send mail from apps.

shell/shared/grain.js
```diff
@@ -832,6 +886,33 @@ if (Meteor.isClient) {
...
+  Template.requestAccess.onCreated(function () {
+    this._status = new ReactiveVar({showButton: true});
+  });
+
+  Template.requestAccess.events({
+    "click button.request-access": function (event, instance) {
+      let identityId = Accounts.getCurrentIdentityId();
+      let grainId = getActiveGrain(globalGrains.get()).grainId();
+      Meteor.call("requestAccess", getOrigin(), grainId, identityId, function(error, result) {
+        console.log("result", result);
```
kentonv: debug logging?

kentonv:
Is there anything stopping me from spamming someone with access requests? Maybe we should have a limit of one request per document.

Also perhaps there should be some explanation that the button sends an email to the owner.

shell/shared/grain.js
dwrensha: Done.

dwrensha:
I've added an `accessRequests` field to the `Grains` table to enforce an anti-spam limit of one request per identity per grain. The field includes a timestamp, so if we decide we want the limit to expire at some point, we could do that too.

shell/shared/grain.js
dwrensha: I was going to edit the `From` field in the email to also include the display name, but this comment dissuaded me.

I like the message text as-is, but I'd be receptive to thoughts from @zarvox.

shell/shared/grain.js
dwrensha: Removed.

kentonv:
It looks like in the case that the user has already requested access once, then reloads the page, they'll be able to click "request access" again, but will get an exception. It would be nicer to change the UI so that it indicates that access was already requested.

shell/shared/grain.js
```diff
@@ -386,6 +389,78 @@ Meteor.methods({
...
+      const url = origin + "/share/" + grainId + "/" + identityId;
+
+      let html = message + "<br><br>" +
+          emailLinkWithInlineStyle(url, "Open Sharing Menu");
+
+      SandstormEmail.send({
+        to: emailAddress,
+        envelopeFrom: envelopeFrom,
+        from: fromEmail,
+        subject: "Request for access to " + grain.title,
```
zarvox: Prefer ordering this grain title first, so that the most relevant information is at the beginning of the subject.  This way, on mobile phones where the subject may be truncated, the context is clearer.

`subject: grain.title + " - Request for access,`

shell/shared/grain.js
```diff
@@ -386,6 +389,78 @@ Meteor.methods({
...
+      if (identity.profile.service === "google") {
+        identityNote = " (" + identity.privateIntrinsicName + ")";
+      } else if (identity.profile.service === "github") {
+        identityNote = " (" + identity.profile.intrinsicName + " on GitHub)";
+      } else if (identity.profile.service === "email") {
+        identityNote = " (" + identity.profile.intrinsicName + ")";
+      }
+
+      const message = identity.profile.name + identityNote +
+            " has requested access to your grain: " + grain.title + ".";
```
zarvox: Prefer present tense here: "is requesting access" rather than "has requested access".

zarvox:
I think I'm in favor of:

* Include display name and intrinsic name in the email body
* One of the following, ranked from "favorite" to "least favorite":
  * Let the user select which of <current identity's verified email addresses> they want to request access as
  * Show which email address and identity the request will be sent from on the page before the user clicks the button.
  * Send the request as the server admin.
* Put the grain title at the beginning of the email subject.

I'm fine with punting on the email selection UI until later, but I do think it should be clear which identity and email address you're possibly revealing to the grain owner, since that's easy to lose track of.

shell/client/shell.html
```diff
@@ -186,6 +190,30 @@
...
+  {{#if currentUser}}
+    {{#with status}}
+      {{#if showButton}}
+        <button class="request-access">Request Access</button>
+      {{/if}}
+      {{#if waiting}}
+         <p>Sending your request...</p>
+      {{/if}}
+      {{#if success}}
+         <p>Your request has been sent.</p>
```
zarvox: Consider setting expectations about how the user will know if/when they have been granted access to the grain - will they receive an email?  Will it appear silently in their grain list?

shell/shared/grain.js
dwrensha: Done.

shell/shared/grain.js
dwrensha: Done.

shell/shared/grain.js
dwrensha: Done.

dwrensha:
Now clicking the Request Access button displays an identity chooser and a warning that you will be revealing your identity.

dwrensha:
I ditched my old approach of once-per-grain-per-requester spam control. My latest set of changes adds a new field to the Users table that limits each account to ten access-requests per day.

dwrensha:
Ready for re-review.

kentonv:
Merge conflicts. :/

kentonv:
I tested this out and it seems pretty great. It would be neat if the "request access" screen would reactively get replaced with the grain as soon as access is granted, but I suppose that's hard?

shell/client/shell.html
```diff
@@ -186,6 +190,39 @@
...
+        <p>Requesting access will reveal your identity to the grain owner.</p>
+        <p> {{chooseIdentityText}} </p>
+        {{> identityPicker identityPickerData}}
+      {{/if}}
+      {{#if waiting}}
+         <p>Sending your request...</p>
+      {{/if}}
+      {{#if success}}
+        <p>
+          Your request has been sent. You will receive an email if your request gets approved.
```
kentonv: "gets approved" sounds ever so slightly weird to me. I feel like "is approved" would sound less weird. This may just be me.

dwrensha: Heh, I wonder whether I'm slipping into Pittsburghese... a request "needs approved" until it "gets approved".

dwrensha:
OK, I've updated the wording to be "is approved", and I've made the Request Access page automatically load the grain once access is granted.

kentonv:
@dwrensha Can you fix the js-style issues? Sorry. `make stylecheck` runs the checks locally.

dwrensha:
rebased and linted

PR #1445, opened 2016-01-29
Use server title in document title.

### LINKED ISSUE #945: Allow setting an instance title
Lately I find myself moving between 3 separate Sandstorm instances:

 * Oasis
 * An instance run by my company
 * My personal instance

It would be nice if each individual Sandstorm instance could have its own title. At the moment they're all just "Sandstorm" so I have to look to my addressbar to figure out which one I'm working with.

kentonv:
Fixes #945

### DIFF
--- a/shell/client/grainview.js
+++ b/shell/client/grainview.js
@@ -243,18 +243,18 @@ GrainView = class GrainView {
   frameTitle() {
     this._dep.depend();
     if (this._frameTitle !== undefined) {
-      return this._frameTitle;
+      return this._frameTitle + ' · ' + globalDb.getServerTitle();
     }
 
     const appTitle = this.appTitle();
     const grainTitle = this.title();
     // Actually set the values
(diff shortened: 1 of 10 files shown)


zarvox:
I note that `serverTitle` might be null or undefined if not explicitly set by the admin - consider making that value "Sandstorm" instead, so we don't wind up stripping that from all existing users?

If you intend for the field to be totally optional, then when `!serverTitle`, we probably don't need the cdot separator either, so in that case consider extracting the appending to some function which appends the center dot and the serverTitle, if one is given.

If you consider both of these things insufficiently important, whatever :shipit: . :)

kentonv:
`serverTitle` will not be null or undefined because a migration always initializes it. :)

I broke a test. Updating...

zarvox:
Ahhh, touché.  Ship it once tests pass, then. :)

PR #1446, opened 2016-01-29
SSL docs changes dwrensha and more

paulproteus:
Combines @dwrensha 's #1426 with turning `ssl.md` into a topic guide for easier reading.


paulproteus:

PR #1448, opened 2016-02-01
Make sure to stop all subscriptions and observers when a grainview is destroyed.

dwrensha:
Currently, if you open a grain and then close it from the navbar, its GrainView can linger on due to a still-alive `_sessionObserver` observer. That observer actually gets triggered every three minutes when `gcSessions()` happens. At that point, the GrainView calls `openSession()`, causing the session to be renewed and the grain's `lastUsed` time to be updated. The confusing result: the "Last Activity" column in the grain list continues to update for grains that have been closed!

### DIFF
--- a/shell/client/grainview.js
+++ b/shell/client/grainview.js
@@ -61,15 +61,10 @@ GrainView = class GrainView {
     this._hostId = undefined;
     this._sessionId = null;
     this._sessionSalt = null;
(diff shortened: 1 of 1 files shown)


paulproteus:
If this really makes my grains stop mysteriously being "active" while I'm
not looking at them, that will be so great!

PR #1451, opened 2016-02-01
Fix vestigial references to userId that should be identityId

### LINKED ISSUE #1449: Shared with me grains don't display proper icons
Instead, it seems that the generic identicon displays rather than an App's icon. @zarvox has more details.

zarvox:
When ApiTokens migrated from using `userId` to using `identityId`, we appear to
have missed a couple of places we used the old value.

As usual, this could have been caught by a typechecker and a strongly-typed
interface.

Fixes #1449.

### DIFF
--- a/shell/client/grainview.js
+++ b/shell/client/grainview.js
@@ -588,7 +588,7 @@ GrainView = class GrainView {
       // Case 2
       const apiToken = ApiTokens.findOne({
         grainId: this._grainId,
-        'owner.user.userId': Meteor.userId(),
+        'owner.user.identityId': this.identityId(),
       }, {
         sort: {created: 1},
       });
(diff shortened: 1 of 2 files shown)


kentonv:
I guess our sharing stats have been broken for a while now?

zarvox:
It would appear that way, yes.

paulproteus:
Can I humbly request the addition of either a test or some typechecker fun?
If not, that's fine, but I thought I should at least ask.

zarvox:
I guess the test would look like:

1. Install some app
2. Create a grain
3. Make sharing link
4. Log in as second user
5. Redeem sharing link as second user
6. Verify navbar icon correctness...somehow.  It's not clear how, since we really just have the URL, which is a randomly generated static asset ID, which isn't based on content.  I guess we could extract the URL, fetch it again from the testcase, and compare to a known hash, or something?

I'm +1 on having a test for the observed regression (verify that app icons work for shared grains (and maybe grains in general)), but writing that test is going to take substantially longer than writing the two-line fix, so I opted to just do the fix, and see if anyone else was excited about writing tests later.

PR #1454, opened 2016-02-02
Add DB methods to find identities and accounts by verified email address

kentonv:


dwrensha:
Cool, we can also use this on first login to detect duplicate accounts.

How often do you think this method be used? It might be worthwhile to index the email fields. Currently "services.email.email" is indexed but the other two aren't.

Otherwise, looks good to me.

kentonv:
I added indexes and also moved all our ensureIndex() calls into db.js, where I think they should live going forward. This way db.js defines the whole database schema -- and tells us what we need to do if we wanted to move to a different database.

We should go through and add indexes to all the collections soon. We're still doing linear queries for, like, everything. >_> (But Oasis would be the first to have problems with this, and so far it hasn't...)

dwrensha:
I'm not sure about pulling the `ensureIndex()` out of the sandstorm-accounts-oauth package. It actually looks like we might be able to delete the sandstorm-accounts-oauth package and go back to using the plain accounts-oauth package that it was forked from. For reference, sandstorm-accounts-oauth was added in https://github.com/sandstorm-io/sandstorm/pull/303. The recent changes in https://github.com/sandstorm-io/sandstorm/pull/1342 get rid of most of the differences between the packages, and the `deregisterService()` method has been upstreamed as `unregisterService()`: https://github.com/meteor/meteor/pull/4014.

kentonv:
Eh, I think having a comprehensive list of indexes seems more useful than avoiding repetition given that the function is idempotent anyway.

kentonv:
That is to say, it's fine if we keep the ensureIndex()es here *and* switch back to the Meteor OAuth package that calls esureIndex() itself, since only the first call will have an effect.

dwrensha:
> Eh, I think having a comprehensive list of indexes seems more useful than avoiding repetition given that the function is idempotent anyway.

Fair enough.

Looks good to me.

PR #1459, opened 2016-02-03
Picker powerbox: SealedUiView support

zarvox:
This implements the request and offer sides of the picker powerbox for the SealedUiView interface, which can be used to implement grain sharing.

A test app exercising this functionality can be found at https://alpha-hlngxit86q1mrs2iplnx.sandstorm.io/test-4.spk .  The source lives at https://github.com/zarvox/sandstorm-test-python .

### DIFF
--- a/shell/packages/sandstorm-permissions/permissions.js
+++ b/shell/packages/sandstorm-permissions/permissions.js
@@ -417,7 +417,11 @@ SandstormPermissions.createNewApiToken = function (db, provider, grainId, petnam
   check(owner, Match.OneOf({webkey: {forSharing: Boolean,
                                      expiresIfUnusedDuration: Match.Optional(Number)}},
                            {user: {identityId: String,
-                                   title: String}}));
+                                   title: String}},
+                           {grain: {grainId: String,
+                                    saveLabel: Match.ObjectIncluding({defaultText: String}),
+                                    introducerIdentity: String,
+                                    sealed: Boolean}}));
 
   var grain = db.getGrain(grainId);
   if (!grain) {
@@ -462,18 +466,7 @@ SandstormPermissions.createNewApiToken = function (db, provider, grainId, petnam
       apiToken.expiresIfUnused = new Date(Date.now() + owner.webkey.expiresIfUnusedDuration);
     }
   } else if (owner.user) {
-    var pkg = db.collections.packages.findOne(grain.packageId);
-    var appTitle = (pkg && pkg.manifest && pkg.manifest.appTitle) || { defaultText: ""};
-    var grainInfo = {appTitle: appTitle};
-
-    if (pkg && pkg.manifest && pkg.manifest.metadata && pkg.manifest.metadata.icons) {
-      var icons = pkg.manifest.metadata.icons;
-      grainInfo.icon = icons.grain || icons.appGrid;
-    }
-    // Only provide an app ID if we have no icon asset to provide and need to offer an identicon.
-    if (!grainInfo.icon && pkg) {
-      grainInfo.appId = pkg.appId;
-    }
+    var grainInfo = db.getDenormalizedGrainInfo(grainId);
     apiToken.owner = {
       user: {
         identityId: owner.user.identityId,
@@ -481,6 +474,15 @@ SandstormPermissions.createNewApiToken = function (db, provider, grainId, petnam
         // lastUsed: ??
         denormalizedGrainMetadata: grainInfo,
       }
+    };
+  } else if (owner.grain) {
+    apiToken.owner = {
+      grain: {
+        grainId: owner.grain.grainId,
+        saveLabel: owner.grain.saveLabel,
+        introducerIdentity: owner.grain.introducerIdentity,
+        sealed: owner.grain.sealed,
+      },
     };
   }
 
--- a/shell/packages/sandstorm-permissions/permissions.js
+++ b/shell/packages/sandstorm-permissions/permissions.js
@@ -480,7 +480,6 @@ SandstormPermissions.createNewApiToken = function (db, provider, grainId, petnam
         grainId: owner.grain.grainId,
         saveLabel: owner.grain.saveLabel,
         introducerIdentity: owner.grain.introducerIdentity,
-        sealed: owner.grain.sealed,
       },
     };
   }
--- a/shell/packages/sandstorm-permissions/permissions.js
+++ b/shell/packages/sandstorm-permissions/permissions.js
@@ -124,8 +124,8 @@ function collectEdges(db, vertex) {
   // trivial and its `roleAssignments` field is an empty list. If `terminalEdge` is not present,
   // then there is no such link.
   //
-  // TODO(someday): Once UiView tokens can have membrane requirements, we'll need to account for
-  // them in this computation.
+  // TODO(now): UiView tokens can have membrane requirements; we need to account for them in this
+  // computation.
   check(vertex,
         Match.OneOf({token: Match.ObjectIncluding({_id: String, grainId: String})},
                     {grain: Match.ObjectIncluding(
--- a/shell/packages/sandstorm-permissions/permissions.js
+++ b/shell/packages/sandstorm-permissions/permissions.js
@@ -420,8 +420,7 @@ SandstormPermissions.createNewApiToken = function (db, provider, grainId, petnam
                                    title: String}},
                            {grain: {grainId: String,
                                     saveLabel: Match.ObjectIncluding({defaultText: String}),
-                                    introducerIdentity: String,
-                                    sealed: Boolean}}));
+                                    introducerIdentity: String,}}));
 
   var grain = db.getGrain(grainId);
   if (!grain) {
--- a/shell/packages/sandstorm-ui-powerbox/powerbox-client.js
+++ b/shell/packages/sandstorm-ui-powerbox/powerbox-client.js
@@ -319,22 +319,22 @@ Template.powerboxRequest.onRendered(function () {
 Template.powerboxRequest.helpers({
   grains: function () {
     const ref = Template.instance().data.get();
-    return ref.filteredCardData();
+    return ref && ref.filteredCardData() || [];
   },
 
   selectedProvider: function () {
     const ref = Template.instance().data.get();
-    return ref._selectedProvider.get();
+    return ref && ref._selectedProvider && ref._selectedProvider.get();
   },
 
   selectedProviderTemplate: function () {
     const ref = Template.instance().data.get();
-    return ref._selectedProvider.get().templateName;
+    return ref && ref._selectedProvider && ref._selectedProvider.get().templateName;
   },
 
   selectedProviderTemplateData: function () {
     const ref = Template.instance().data.get();
-    return ref._selectedProvider.get().templateData();
+    return ref && ref._selectedProvider && ref._selectedProvider.get().templateData();
   },
 
   requestedInterfaceIsImplementedByFrontendRef: function () {
@@ -352,7 +352,7 @@ Template.powerboxRequest.helpers({
     // empty list, as requests for webkeys.  Later, we"ll want to transition the test apps to
     // specify interfaces, and implement frontendrefs to support those interfaces.
     const ref = Template.instance().data.get();
-    return !ref._requestInfo.query;
+    return !(ref && ref._requestInfo && ref._requestInfo.query);
   },
 
   webkeyError: function () {
--- a/shell/packages/sandstorm-ui-powerbox/powerbox-client.js
+++ b/shell/packages/sandstorm-ui-powerbox/powerbox-client.js
@@ -61,9 +61,9 @@ SandstormPowerboxRequest = class SandstormPowerboxRequest {
     return false;
   }
 
-  completeSealedUiView(roleAssignment) {
+  completeUiView(roleAssignment) {
     const fulfillingProvider = this._selectedProvider.get();
-    if (fulfillingProvider.type === "frontendref-sealeduiview") {
+    if (fulfillingProvider.type === "frontendref-uiview") {
       const fulfillingGrainId = fulfillingProvider.grainId;
       const fulfillingApiToken = fulfillingProvider.apiToken; // Possibly irrelevant?
       const fulfillingGrainTitle = fulfillingProvider.title;
@@ -74,7 +74,6 @@ SandstormPowerboxRequest = class SandstormPowerboxRequest {
           grainId: this._requestInfo.grainId,
           saveLabel: saveLabel,
           introducerIdentity: this._requestInfo.identityId,
-          sealed: this.requestedInterfaceMatchesTag({ id: "10145426045091321204" }), // SealedUiView
         },
       };
       const provider = {
@@ -98,7 +97,7 @@ SandstormPowerboxRequest = class SandstormPowerboxRequest {
               token: apiToken,
               descriptor: {
                 tags: [
-                  { id: "10145426045091321204" }, // SealedUiView
+                  { id: "15831515641881813735" }, // UiView
                 ],
                 quality: "acceptable",
               },
@@ -115,10 +114,10 @@ SandstormPowerboxRequest = class SandstormPowerboxRequest {
 
   selectGrain(grainCard) {
     this._selectedProvider.set({
-      type: "frontendref-sealeduiview",
+      type: "frontendref-uiview",
       grainId: grainCard.grainId,
       title: grainCard.title,
-      templateName: "powerboxProviderSealedUiView",
+      templateName: "powerboxProviderUiView",
       templateData: () => {
         const grain = this._db.collections.grains.findOne(grainCard.grainId);
         const viewInfo = grain.cachedViewInfo;
@@ -131,7 +130,7 @@ SandstormPowerboxRequest = class SandstormPowerboxRequest {
           lastUsed: grainCard.lastUsed,
           viewInfo: viewInfo,
           onComplete: (roleAssignment) => {
-            this.completeSealedUiView(roleAssignment);
+            this.completeUiView(roleAssignment);
           },
         };
       },
@@ -147,11 +146,11 @@ SandstormPowerboxRequest = class SandstormPowerboxRequest {
         const viewInfo = result;
         this.annotateViewInfo(viewInfo);
         this._selectedProvider.set({
-          type: "frontendref-sealeduiview",
+          type: "frontendref-uiview",
           grainId: apiTokenCard.grainId,
           apiToken: apiTokenCard._id,
           title: apiTokenCard.title,
-          templateName: "powerboxProviderSealedUiView",
+          templateName: "powerboxProviderUiView",
           templateData: () => {
             return {
               _id: apiTokenCard._id,
@@ -161,7 +160,7 @@ SandstormPowerboxRequest = class SandstormPowerboxRequest {
               lastUsed: apiTokenCard.lastUsed,
               viewInfo: viewInfo,
               onComplete: (roleAssignment) => {
-                this.completeSealedUiView(roleAssignment);
+                this.completeUiView(roleAssignment);
               }
             }
           }
@@ -344,7 +343,7 @@ Template.powerboxRequest.helpers({
       // TODO: support additional frontendref types
       //ref.requestedInterfaceMatchesTag({id: "12214421258504904768"}) || // IpNetwork
       //ref.requestedInterfaceMatchesTag({id: "16369547182874744570"}) || // IpInterface
-      ref.requestedInterfaceMatchesTag({ id: "10145426045091321204" }) // SealedUiView
+      ref.requestedInterfaceMatchesTag({ id: "15831515641881813735" }) // UiView
     );
   },
 
@@ -405,7 +404,7 @@ Template.powerboxRequest.events({
   },
 });
 
-Template.powerboxProviderSealedUiView.events({
+Template.powerboxProviderUiView.events({
   "click .connect-button": function (event) {
     event.preventDefault();
     const ref = Template.instance().data;
--- a/shell/packages/sandstorm-ui-powerbox/powerbox-client.js
+++ b/shell/packages/sandstorm-ui-powerbox/powerbox-client.js
@@ -32,6 +32,18 @@ SandstormPowerboxRequest = class SandstormPowerboxRequest {
     }
   }
 
+  hexValueForWellKnownTagId(decimalString) {
+    // Since there's not a convenient way to convert between the decimal and hex representations of
+    // 64-bit integers in Javascript due to Javascript's numeric type only having 53 bits of
+    // precision, we hardcode the integer and hex values of the interface IDs implemented by
+    // frontendrefs so we can match them in the browser immediately.
+    return {
+      "15831515641881813735": "0xdbb4d798ea67e2e7", // UiView
+      "12214421258504904768": "0xa982576b7a2a2040", // IpNetwork
+      "16369547182874744570": "0xe32c506ee93ed6fa", // IpInterface
+    }[decimalString];
+  }
+
   requestedInterfaceMatchesTag(target) {
     // This whole function should probably be migrated to use the powerbox interface matching code,
     // once that exists.
@@ -51,8 +63,13 @@ SandstormPowerboxRequest = class SandstormPowerboxRequest {
           const tag = tags[j];
           // TODO: implement the more precise request matching algorithm in grain.capnp
           // which also considers tag values
-          if (tag.id && tag.id === target.id) {
-            return true;
+          if (tag.id) {
+            // Since Cap'n Proto 64-bit integer types are canonically represented as strings, but
+            // node-capnp also accepts string of any base, we special-case matching the hex value of
+            // a few well-known tag ids.
+            if (tag.id === target.id || tag.id === this.hexValueForWellKnownTagId(target.id)) {
+              return true;
+            }
           }
         }
       }
--- a/shell/packages/sandstorm-ui-powerbox/powerbox-client.js
+++ b/shell/packages/sandstorm-ui-powerbox/powerbox-client.js
@@ -1,0 +1,429 @@
+SandstormPowerboxRequest = class SandstormPowerboxRequest {
+  constructor(db, requestInfo) {
+    check(requestInfo, Match.ObjectIncluding({
+      source: Match.Any,
+      origin: Match.Any,
+      rpcId: Match.Any,
+      // For reasons I don't understand, Match.Optional does not work here.
+      query: Match.OneOf(undefined, [Object]),
+      saveLabel: Match.Optional(Match.Any),
+      grainId: String,
+      identityId: String,
+      onCompleted: Function,
+    }));
+
+    this._db = db;
+    this._requestInfo = requestInfo;
+    this._completed = false;
+    this._error = new ReactiveVar(undefined);
+
+    // State to track UI status.
+    this._filter = new ReactiveVar("");
+    this._selectedProvider = new ReactiveVar(undefined);
+  }
+
+  finalize() {
+    if (!this._completed) {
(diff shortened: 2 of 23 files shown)


zarvox:
Note: this patchset does not implement enforcement of membraneRequirements for UiViews.  That will come in a separate patchset.

zarvox:
Rebased out the commits that do `box-sizing: border-box` everywhere, since apparently there are more places that could break that I need to test.

shell/packages/sandstorm-ui-powerbox/powerbox.html
```diff
@@ -0,0 +1,80 @@
...
+  {{#if showWebkeyInput}}
+    <form id="powerbox-request-form">
+      <label>Please input an api token: </label><input name="token" type="text" id="powerbox-request-input"><br>
+      <button class="submit">Import token</button>
+    </form>
+    {{#if webkeyError}}
+    <div class="error">{{webkeyError}}</div>
+    {{/if}}
+  {{else}}
+    <h4>Connect this grain with:</h4>
```
kentonv: I feel like this wording may be confusing in common use cases. E.g. when sharing through Rocket.Chat, users wouldn't normally think of what they're doing as "connecting". Maybe "attaching" makes sense, or maybe this should just say "Choose a grain:"? I think it's well-enough implied from the context that the currently-open app is making the request and will get access -- similarly, the browser's file-open dialog doesn't explicitly say that you're granting the web page access to the file.

shell/packages/sandstorm-ui-powerbox/powerbox-client.js
```diff
@@ -0,0 +1,429 @@
...
+    return (
+      // TODO: support additional frontendref types
+      //ref.requestedInterfaceMatchesTag({id: "12214421258504904768"}) || // IpNetwork
+      //ref.requestedInterfaceMatchesTag({id: "16369547182874744570"}) || // IpInterface
+      ref.requestedInterfaceMatchesTag({ id: "10145426045091321204" }) // SealedUiView
+    );
+  },
+
+  showWebkeyInput: function () {
+    // Transitional feature: treat requests that specified no query patterns to match, not even the
```
kentonv: I feel like there should always be some way to input a webkey, possibly hidden behind a click on "advanced" or some such.

kentonv:
@dwrensha can you review the permissions / sharing graph stuff?

shell/packages/sandstorm-ui-powerbox/powerbox.html
zarvox: Yeah, I agree the copy here is tricky.  "Attaching" might make sense in the context of Rocket.Chat, but is probably less suitable for e.g. a folder of grains.  "Connect" is reasonably good for things like static publishing, or remote hosts to connect to (e.g. RSS feeds) but not super great in general.  I do like that it hints at the persistent nature of these connections, though, which isn't necessarily obvious from the get-go.

"Choose a grain:" is better for this specific case, but "grain" might be confusing when frontendrefs enter the mix, like if the app is requesting an IpNetwork or IpInterface, which may be implemented by either the frontend or by apps.  We could have a small dictionary for the various Sandstorm-recognized types, but then non-Sandstorm-aware interfaces wind up disadvantaged, or we have .  Maybe the request should specify an overriding noun phrase, but then we have to change the request/offer interfaces?  Maybe something else?

The text I originally wrote tries to dodge the "what noun should we use?" question entirely.  I want to have some sort of verb here indicating what the user should do next, and ideally the noun would be one of the items in the list.  "Grain" is almost right, except when it's not.

Maybe this could just say "Select one:" and rely on the context provided by the app making the request?  I think I like this option the best, though it's also the least specific/helpful.

@neynah if you have any ideas here, I'm all ears

shell/packages/sandstorm-ui-powerbox/powerbox-client.js
zarvox: Sure, we can incorporate that in a future revision.

shell/packages/sandstorm-ui-powerbox/powerbox.html
kentonv: FWIW, I think we can reasonably have different language when selecting SealedUiView vs. other capability types. "Connect" could make more sense for general capabilities, maybe.

"Select one:" also seems reasonable.

paulproteus: FWIW, I like "Select one:" or "Choose:"​

src/sandstorm/grain.capnp
```diff
@@ -739,7 +752,17 @@ struct RoleDef {
   # "default", then such sharing actions will be treated as having an empty permissions set (the
   # user can open the grain, but the grain is told that the user has no permissions).
   #
+  # This field should *not* be used to specify a default role hint for sharing!  For that, use
+  # `defaultSelection` immediately below.
+  #
   # See also `ViewSharingLink.RoleAssignment.none`, below.
+
+  defaultSelection @6 :Bool = false;
```
dwrensha: I'm happy that this field will now exist. Did you mean to actually use it in this pull request? I don't see any other uses of the the string "defaultSelection" in the diff.

src/sandstorm/sealed.capnp
```diff
@@ -0,0 +1,29 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+@0xfe5c7cde99284f21;
+
+$import "/capnp/c++.capnp".namespace("sandstorm");
+
+using SystemPersistent = import "supervisor.capnp".SystemPersistent;
+
+interface SealedUiView @0x8ccbcb3c38aa1574 {}
```
dwrensha: I'm struggling to understand why we need to define two new interfaces for `UiView`. Could there be some more explanatory text about that here? The `Persistent` interface already has a `sealFor` parameter. How is that different form the sealing that is happening here?

src/sandstorm/grain.capnp
```diff
@@ -633,7 +633,20 @@ interface SessionContext {
...
   #
-  # (eg. window.parent.postMessage({powerboxRequest: {rpcId: myRpcId, query: [{}]}}, "*")
+  # e.g.:
+  #   window.parent.postMessage({
+  #     powerboxRequest: {
+  #       rpcId: myRpcId,
+  #       query: [
+  #         {
+  #           tags: [
+  #             { id: "10145426045091321204" },
```
dwrensha: Ah, man, I wish I could use the hexadecimal `id` here. Is there a convenient way to do that?

shell/server/core.js
```diff
@@ -147,6 +148,46 @@ class NotificationHandle {
...
+        if (this._token && this._token.owner && this._token.owner.grain &&
+            this._token.owner.grain.saveLabel && this._token.owner.grain.saveLabel.defaultText) {
+          owner.user.title = this._token.owner.grain.saveLabel.defaultText;
+        }
+
+        // Denormalize this grain's app metadata into owner.user.
+        owner.user.denormalizedGrainMetadata = globalDb.getDenormalizedGrainInfo(grainId);
+      }
+
+      const ret = makeChildTokenInternal(this._token._id, owner, this._membraneRequirements, grainId);
```
dwrensha: See [the documentation for the `parentToken` field](https://github.com/sandstorm-io/sandstorm/blob/618193034f4fcd5d45e0f720b783ab61a68878f7/shell/packages/sandstorm-db/db.js#L365-L37) --- in addition to the `grainId` field, we should make sure that the child inherits the `identityId` and `accountId` fields of the parent.

Maybe the first argument to `makeChildTokenInternal` should be the actual entry in `ApiTokens` representing the parent?

dwrensha:
When the test app makes a powerbox request, I seem to be unable to deny it. I can click the "x" to minimize the popup, but the "P" button remains in the topbar until I fulfill the request or the app makes a new request.

src/sandstorm/grain.capnp
zarvox: I'm trying to follow the node-capnp encoding rules as closely as possible, here, so that one day this could become more strongly typed without breaking compatibility.  `id` here is really "a capnproto uint64", but because JS can't represent a 64-bit numeric type, we fake it with strings.

For some reason, I though that node-capnp only accepted base-10 strings as 64-bit numerics, but looking at `node-capnp/src/node-capnp/capnp.cc:1224` it looks like we actually accept anything that `strtoull(3)` accepts.  So I guess I can switch everything to hex ids after all!

In which case I should probably do so everywhere.

src/sandstorm/grain.capnp
zarvox: Oh, whoops, I probably meant to use it to prepopuplate which roleAssignment was selected when picking a roleAssignment for the SealedUiView powerbox provider.

src/sandstorm/sealed.capnp
zarvox: Ahhh, yes.  I'll add a comment explaining how this is only unsealable into a UiView by a user, not by a grain or other entity.

The goal was that we be able to share a reference to grain A without granting the intermediary grain B that holds the reference (currently a SealedUiView capability) the ability to do literally anything that UiView allows, like open new sessions for grain A impersonating other users.

Perhaps overloading the term `sealed` here was a poor choice.  If you have a better name for this, I'm definitely open to changing it.

src/sandstorm/grain.capnp
dwrensha: When you do that, make sure to fall back to `default` if there is no `defaultSelected`.

dwrensha:
Aside from the `identityId`/`accountId` thing, this looks good to me from a permissions perspective.

I'll need to update the new direct-share-revocation code so that it doesn't call these grain-owned strudyrefs "links", but the revocation should work as-is.

zarvox:
The "unable to deny" thing is that way by design, so in case the user clicks or taps outside of the popup when it first appears, they can get back to it without losing state.  While it's true that right now there's very little state to lose, this could become a more painful failure mode once we open a grain with a `newRequestSession()` and you have to take more steps within that grain to select the capability that grain will be offering.

I'm realizing we'd have to revise how we manage popup item state and iframes to make that sort of thing work in the first place, though, so probably it makes sense to switch to the behavior you initially expected for now, and if we find that people make painful errors, then we can revisit the design then.

shell/server/core.js
zarvox: Hmmm.  This is where things get kinda weird.

The saved child token kind-of-isn't-really-a-SealedUiView.  It's a full UiView, which should have a membrane requirement that it can't be restored unless the SealedUiView that led to its creation is still valid.

A SealedUiView isn't actually a UiView; it's an opaque handle associated with a UiView that the Sandstorm shell will give you or redeem into a proper UiView if you offer it to a client with the (invisible and implicit) permission "is a human user".

Given that: is using `parentToken` still the right thing to do here, if in practice the relation doesn't match what `parentToken` is documented as meaning?  Is there an alternative expression that would be more consistent?

dwrensha: > A SealedUiView isn't actually a UiView; it's an opaque handle associated with a UiView that the Sandstorm shell will give you or redeem into a proper UiView if you offer it to a client with the (invisible and implicit) permission "is a human user".

This is the kind of text that would be nice to have in `sealed.capnp`!

dwrensha: Sounds to me like you want it to add a `tokenValid` membrane requirement and make it not a child token.

src/sandstorm/grain.capnp
zarvox: I thought the point of having a separate field was so that we could unbind the two.

Actually, hmm.  Now I find myself wishing that `default` and `defaultSelection` were integers on `ViewInfo`, rather than booleans on `RoleDef`, given that it only makes sense for one role to be default for upgrading the old permissions model, and for only one role to be default in selection UI.

I think I may instead back this piece of the change out for now, and maybe do it as a separate piece later.

shell/server/core.js
zarvox: So, after some discussion with @kentonv and @jparyani, reversing what I said above: 

A SealedUiView *is* actually a UiView.  UiViews are defined to have an additional (implicit) permission, `isHuman`, which human users have, and which is required to make any method calls to the default UiView implementation.  SealedUiView is a child UiView token that has that permission removed.

This means that only `restore()` needs to compute the implicit `mayBeHuman` permission, which is currently always false, since humans do not use the `restore()` function.  This also means that we do not currently require a `sealed` Boolean at all, since we've instead defined it in terms of other properties of the ApiToken - in particular, it's `!!owner.grain`.

This will necessitate some code changes, so I'll do those and get back to you.

src/sandstorm/grain.capnp
zarvox: So, making this change has unveiled a second problem, which is that to do tag id matching, the `uint64_t`'s representation needs to be canonicalized, and it's awkward to canonicalize bignum-representing-strings in the browser without a bignum library or something like that.

My options now appear to be:

* require base-10 strings everywhere, since that's what node-capnp will produce by default when I refer to `Grain.UiView.typeId` on the server.  Requires apps to use the base 10 id representation.  What this patchset, as-is, implements.
* require base-16 strings everywhere, since that's what `capnp compile -o capnp` gives you.  Requires manual conversion of base10 strings to base16 strings on the server.  Requires apps use base16 strings only.
* Require both apps and sandstorm to convert strings into uint8 arrays or something like that.  This makes the client-side comparison easy and correct, but means weird conversions that don't mix cleanly with the other APIs on either side.
* Implement/use a pure-JS bignum library in the browser to parse and compare the two string values.  This does not put requirements on what the app code sends.

None of these options seem particularly great.  Do you have a preference for any of these?

In all cases, we still need to make a decision on what the shell should `postMessage()` back to the app in the powerboxDescriptor when the user grants a token - should it be a base10 string, a base16 string, or something else?  Regardless of what we choose, the app's backend is going to have to parse it into a 64-bit uint.

dwrensha: I think that in any case the `descriptor` field in the `postMessage()` call should be a valid node-capnp value. 

If we don't accept base-16 strings, I think we should at least provide some javascript that apps can use to make the conversion themselves.

dwrensha: > I think that in any case the descriptor field in the postMessage() call should be a valid node-capnp value.

I suppose we could always extend node-capnp so that it can handle uint8 arrays or buffers or whatever.

zarvox: Yes, I agree that it should be a valid node-capnp value, but the issue is that there are multiple things which node-capnp will accept, but when doing an implicit capnproto-struct-to-JS-object, there can be only one which it will emit in the thing that it generates to `postMessage()` to the client.  Whatever that is should be the canonical JSON encoding of a capnproto 64-bit int.

Since node-capnp is not an option in frontend javascript code (which will be the immediate consumer of this postMessage() that we send back to the app with the sturdyref and the descriptor at issue), we can't assume that apps will have that available to do the reverse conversion.  So I think we ought to specify what that encoding will be, so that apps can rely on *something* without having to pull in node-capnp on the backend.  And today, for better or worse, that canonical encoding, as implemented by the capnproto JSON encoder/decoder, is a decimal string.  So what we `postMessage()` back to apps should likewise be a decimal string.  At which point for simplicity on the app side, we should probably just use decimal strings in the requests, even if we also patch the frontend to accept hex strings.  This way, if app authors treat capnp ids as opaque identifiers, they can rely on string equality for equality testing.

I'll make the frontend able to accept either the decimal or hex representation of just the frontendref-specific interfaces, but after thinking about all this, I intend to continue using the decimal representation in the documentation and my example app, since that's more canonically the JSON representation of a capnproto 64-bit int.

zarvox:
Here's a bunch of changes that I think address everything we talked about.  Take another pass over it when you have the chance.

zarvox:
FYI, an updated test app which requests the appropriate UiView interface now exists at https://alpha-hlngxit86q1mrs2iplnx.sandstorm.io/test-5.spk

src/sandstorm/grain.capnp
```diff
@@ -401,6 +401,14 @@ interface UiView {
     # two UiViews are really the same view with different permissions applied, and can then combine
     # them in the UI as appropriate.
     #
+    # Implementation-wise, in addition to the permissions enumerated here, there is an additional
+    # "is human" pseudopermission provided by the Sandstorm platform, which is invisible to apps,
```
dwrensha: is it a "pseudopermission" or a "pseudoprivilege"?

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -482,6 +474,15 @@ SandstormPermissions.createNewApiToken = function (db, provider, grainId, petnam
         denormalizedGrainMetadata: grainInfo,
       }
     };
+  } else if (owner.grain) {
+    apiToken.owner = {
+      grain: {
+        grainId: owner.grain.grainId,
+        saveLabel: owner.grain.saveLabel,
+        introducerIdentity: owner.grain.introducerIdentity,
+        sealed: owner.grain.sealed,
```
dwrensha: stray `sealed` that should be removed?

zarvox: whoops, must have missed that.  Fixed in c8e6270ee9734f1b353eae0591f468e15c196354

src/sandstorm/grain.capnp
zarvox: "pseudoprivilege" sounds better, but I want to invoke the concept of `Permission`s like the ones declared by the `ViewInfo`, so I'd better use "pseudopermission" everywhere.

src/sandstorm/grain.capnp
```diff
@@ -401,6 +401,14 @@ interface UiView {
...
+    # but affects how UiView capabilities may be used.  In particular, all users are said to possess
+    # the "is human" pseudopermission, whereas all other ApiTokenOwners do not, and by default, the
+    # privilege is not passed on.  All method calls on UiView require that the caller have the "is
```
dwrensha: "is not passed on" to whom? Child tokens?

zarvox: Well, child tokens, or their owners, in the same way that existing tokens may have a `roleAssignment` which passes on permissions to their children.  Is there another way of thinking about it that I'm missing/forgetting?

shell/server/core.js
```diff
@@ -147,6 +148,64 @@ class NotificationHandle {
   }
 }
 
+class PersistentUiViewImpl {
+  constructor(token, sturdyRef, membraneRequirements) {
+    this._token = token;
+    this._sturdyRef = sturdyRef;
+    // We need this token so we can later pass it to SandstormPermissions.createNewApiToken, which
+    // requires the sturdyRef of the token it's producing a child for, not just the hashed token id.
```
dwrensha: I don't think there would be any problem with allowing "createNewApiToken()` to accept a `{hashedParentToken: String}` provider. Indeed, we'll need that ability if we ever implement the "Asheesh button" for generating child tokens from the Who Has Access view.

It's interesting that `rawParentToken` works for the present use case, though!

zarvox: Yeah, that sounds useful!

One thing to keep aware of is that because tokens can chain, we want to make sure we're creating new child tokens as children of the token passed to restore(), not some parent token of that token that gives the restore()'d token any authorization.  Indeed, `this._token` is actually possibly a parent token with `owner: grain`, as looked up by `restoreInternal()` - see the

```} else if (token.parentToken) {```

 branch.

File: shell/server/core.js [hunk 11]
```diff
@@ -332,13 +393,31 @@ restoreInternal = (tokenId, ownerPattern, requirements, parentToken) => {
...
     return waitPromise(globalBackend.useGrain(token.grainId, (supervisor) => {
       return supervisor.restore(token.objectId, requirements, parentToken);
     }));
   } else if (token.parentToken) {
+    // A token which chains to some parent token.  Restore the parent token (possibly recursively),
+    // checking requirements on the way up.
     return restoreInternal(token.parentToken, Match.Any, requirements, parentToken);
+  }
+
+  // Note that hereafter, `token` should be, by process of elimination, a UiView token.
```
dwrensha: Maybe we should at least validate that the `grainId` field is present?

shell/server/hack-session.js
```diff
@@ -43,24 +43,40 @@ SessionContextImpl = class SessionContextImpl {
...
+            // The following fields will be overwritten by PersistentUiView.save(), so no need to
+            // pass them in:
+            //lastUsed: new Date(),
+            //title: "", // This will be replaced by the token's title
+            //denormalizedGrainMetadata: {}, // This will look up the package for the grain referenced.
+          },
+        };
+      }
+
+      const save = castedCap.save(apiTokenOwner);
```
dwrensha: I realize that this isn't part of your change here, but I just noticed this gap between `save()` and the addition of the membrane requirements. Ideally they would happen as a single atomic operation, right? Actually, it looks like what happens now is if the membrane requirement fails, then the token gets persisted without any membrane requirements, because this method exists early. Am I missing something here? It seems like, at the very least, the `save()` should go directly before the `ApiTokens.update()`. cc @jparyani

zarvox: Good catch.  I asked Jason in person and he agreed with your recommendation.  So 2601f461ef7aea8b5ac138a580397abed31e63b7 does the simple thing for now.

shell/server/core.js
zarvox: added in 1de2465714e324833b37dd7ed4671b6ab495d1e9

zarvox:
BTW, big thanks for giving this a thorough review.  I appreciate your attention to detail!

shell/packages/sandstorm-ui-powerbox/powerbox-client.js
```diff
@@ -0,0 +1,445 @@
...
+    if (!this._completed) {
+      // postMessage back to the origin frame that the request was cancelled.
+      this._requestInfo.source.postMessage({
+        rpcId: this._requestInfo.rpcId,
+        error: this._error.get() || "User cancelled request",
+      }, this._requestInfo.origin);
+    }
+  }
+
+  hexValueForWellKnownTagId(decimalString) {
```
dwrensha: I wrote you a function that should eliminate the need for this map.

```javascript
function hexadecimalToDecimal(n) {
  if (typeof n != "string") {
    throw new Error("Expected string.");
  }
  if (n.length != 18 || !n.startsWith("0x")) {
    throw new Error("Expected '0x' followed by 16 hexadecimal digits");
  }
  var upper32 = parseInt(n.substring(0, 10));
  var lower32 = parseInt("0x" + n.substring(10, 18));

  var result = "";
  for (var exponent = 0; exponent < 22; exponent += 1 ) {
    var w = Math.floor(upper32 / 10);
    var x = upper32 % 10;

    var lowerPlusRemainder = (x * Math.pow(2, 32)) + lower32;
    var y = Math.floor(lowerPlusRemainder / 10);
    var z = lowerPlusRemainder % 10;

    result = z.toString() + result;

    upper32 = w;
    lower32 = y;

    if (upper32 == 0 && lower32 == 0) {
       break;
    }
  }
  return result;
}
```

zarvox: **tips hat**

Integrated with gentle modification (older browsers may lack `String.prototype.startsWith`) in 9e433d2a4f6d1878810ea58e4ed5c0c5ec98b93a

shell/packages/sandstorm-ui-powerbox/powerbox-client.js
```diff
@@ -0,0 +1,479 @@
...
+        error: this._error.get() || "User cancelled request",
+      }, this._requestInfo.origin);
+    }
+  }
+
+  hex64ToDecimal(n) {
+    if (typeof n != "string") {
+        throw new Error("Expected string.");
+    }
+    if (n.length != 18 || n.lastIndexOf("0x", 0) !== 0) {
```
dwrensha: Heh, now I'm reviewing my own code...

Should also validate that all characters are hexadecimal digits. And should verify that the `parseInt()` calls succeed.

zarvox: Added a regex match in 18cae37f897f6b68c9b739a43bd438a9c8096cfd

The `parseInt()` cannot fail if the regex match succeeds.

zarvox:
@dwrensha Should I expect more comments on this, or is this in a state suitable for merging? :)

dwrensha:
The test app gives me a 500 error if I do the following:

  1. Create a new grain.
  2. Click the "Request UiView" and choose some other grain.
  3. Create a share link.
  4. Open the share link anonymously.
  5. Click the "Offer" button from the anonymous view.

What is supposed to happen in this case?

zarvox:
I'm not sure if anonymous users are supposed to be able to own/transmit non-`owner:webkey` capabilities - I would by default expect not.

Maybe the supervisor should require in offer() that the session context we're offering this capability to is not anonymous?

dwrensha:
For what it's worth, when I took these actions I was expecting that the anonymous user would get a view for the other grain.

zarvox:
That would break some assumptions we make about anonymous users only being able to have one grain open at a time.  I'm not sure how you'd compute that the membraneRequirement on that token would be satisfied, given that you only have a webkey for the grain that's making the offer.  Maybe `openSessionFromApiToken needs to send all the sturdyRefs currently held by the browser window, so the backend can see that it's allowed to open that grain?

Supporting anonymous delegated access is going to be a nontrivial undertaking.

dwrensha:
> @dwrensha Should I expect more comments on this, or is this in a state suitable for merging?

I don't expect to have further comments. I'm okay with punting on the anonymous-user-offer question, as long as we have in mind a plausible way forward there.

dwrensha:
> That would break some assumptions we make about anonymous users only being able to have one grain open at a time.

We probably ought to not make such assumptions.

zarvox:
>  I'm okay with punting on the anonymous-user-offer question, as long as we have in mind a plausible way forward there.

Looks like this is officially our current policy in https://github.com/sandstorm-io/sandstorm/blob/master/shell/server/hack-session.js#L51 .  I suppose I could stand to make the test app handle `offer()` throwing more gracefully, but that's less of a priority at the moment.

Some more thoughts:

There are some other usability issues that arise if we try to support multiple grains in anonymous windows, especially for not-logged-in-at-all users:

* where do the webkeys that the user is using in their current session get stored?  Today, they live in the URL bar, so when the page refreshes, you don't lose access, but that only works for one grain at a time.  We'd potentially have to persist these secrets in browser storage somewhere (though I guess this is not terribly different from our existing "on refresh you only get one grain in your sidebar" problem).

* where do you draw the line on what anonymous users can do?  We already disallow them from creating new sharing links, or in general, doing anything that would mutate the ApiTokens table.  I think there are confinement reasons why this is necessary.

I guess being able to open a grain with a token from another grain might be okay, but I'm not sure that's super helpful given that the next thing they try might also fail, and be something we can't let succeed without breaking confinement.

But it would certainly be cool to make anonymous access exactly as powerful as the sum of the tokens you have collected!

zarvox:
@kentonv I await your look-over and merge!

shell/server/core.js
```diff
@@ -332,13 +393,34 @@ restoreInternal = (tokenId, ownerPattern, requirements, parentToken) => {
...
+  if (!token.grainId) {
+    throw new Meteor.Error(500, "Expected token " + token._id + " to have a grainId");
+  }
+
+  if (token.owner && token.owner.grain) {
```
kentonv: I don't think this logic should be based on owner type. restore() should always return a "sealed" UiView, unless the token is explicitly marked as allowing direct access (a marking we have not yet defined). Human users opening the view in the sandstorm UI go through a completely different path that doesn't involve restore(), so we don't need to account for them here. So I would remove this if() and its whole else branch.

zarvox: fixed in fe10afa0556f6ae16f57bee3e82fc6de3a38b5dd

zarvox:
:confetti_ball:

paulproteus:
Wowee!

PR #1460, opened 2016-02-03
tests: deal with more guided tour failure possibilities

zarvox:
Extract `disableGuidedTour` into a helper command, since we want to use it in a
couple places.  Use `installApp` in more places.


dwrensha:
:thumbsup:

PR #1461, opened 2016-02-03
Fix bug where dev mode actions all launched the last action

zarvox:
Variable hoisting can be weird.


paulproteus:
BTW @erikmaarten - I think you were running into this a few weeks ago.

The issue only occurred in dev mode.

erikmaarten:
So it was a bug after all. Thanks for notifying me Asheesh!

PR #1462, opened 2016-02-03
docs: Clarify self-signed SSL certificate docs

paulproteus:


paulproteus:
Self-merging now because this is 100% docs.

Thanks to @benthemeek for asking this question and being very patient with the answer!

Issue #1218, opened 2015-12-01
Support PATCH method outside of API


ndarilek:
Loomio is transitioning from a single app to an API back-end with an HTML/JS front-end. At this time, however, I don't think they've decoupled the two. They're using PATCH in some places, and these requests are failing because the HTTP server for grain instances doesn't appear to implement PATCH. I'd rather not change their underlying API by switching methods, nor do I want to route everything through the API host to get PATCH support.

Any chance that the grain HTTP implementation can be made to support PATCH requests?

kentonv:
I agree Sandstorm should be made to support PATCH. @mnutt I don't suppose you'd be interested in working on this, since you know exactly where it goes? :)

ndarilek:
I actually don't know where it goes, it was someone else on the list and 
I haven't touched C++ since college. :( Sorry.

ndarilek:
Er, never mind, just saw that the last sentence wasn't addressed to me. :)

Still very interested in this though. I may be able to hack around it 
temporarily--I think Rails used to have a hack that makes the method 
whatever value was passed as some query parameter or other but I don't 
recall the specifics.

rdbartlett:
Rich from Loomio here :)

We're getting an increasing volume of requests for a secure version of Loomio, especially from activists working in really dangerous parts of the world. Sandstorm seems like the perfect solution. 

If there's anything we can do to support someone to resolve this PATCH issue, please let me know. It would really make a big impact for people doing amazing work in the world.

:heart:

paulproteus:
Wow, that's great! Thanks for chiming in, @rdbartlett . I'll see what we can do.

rdbartlett:
Thanks for the prompt injection of enthusiasm @paulproteus :dancer: 

Just cross-referencing here to the relevant conversation happening in Loomio: https://www.loomio.org/d/yw0D8Sc6/help-porting-loomio-to-sandstorm

paulproteus:
Given that #1465 is resolved, I'm going to mark this as closed. Thanks to all who requested it, worked on it, reviewed it, nudged us, and more.

PR #1465, opened 2016-02-04
add PATCH http method support

mnutt:
From sandstorm's perspective, PATCH and PUT are functionally equivalent. PATCH RFC: https://tools.ietf.org/html/rfc5789

Addresses https://github.com/sandstorm-io/sandstorm/issues/1218.

### DIFF
--- a/src/sandstorm/web-session.capnp
+++ b/src/sandstorm/web-session.capnp
@@ -258,6 +258,7 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
       requestUriTooLong     @8 $httpStatus(id = 414, title = "Request-URI Too Long");
       unsupportedMediaType  @9 $httpStatus(id = 415, title = "Unsupported Media Type");
       imATeapot            @10 $httpStatus(id = 418, title = "I'm a teapot");
+      unprocessableEntity  @12 $httpStatus(id = 422, title = "Unprocessable Entity");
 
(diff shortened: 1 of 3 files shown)


garply:

src/sandstorm/web-session.capnp
```diff
@@ -172,6 +173,14 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
     encoding @2 :Text;  # Content-Encoding header (optional).
   }
 
+  struct PatchContent {
+    # TODO(apibump): Remove this and replace it with `PostContent` (renamed to `Content`).
```
kentonv: Since this is a newly-introduced struct, can you go ahead and use PostContent? It turns out it was a mistake to have separate structs for each method, so we're planning to merge everything into PostContent (and rename it). Some of the WebDAV methods are already using PostContent.

zarvox:
I think we also need to add "422 Unprocessable Request" to `ClientErrorCode` in `web-session.capnp`?  http://tools.ietf.org/html/rfc5789#section-2.2

mnutt:
Thanks for the feedback. Updated both.

kentonv:
Thanks for doing this!

paulproteus:
Yippee! cc: @ndarilek

ndarilek:
Yes, I'm the one who pointed them here. :P

PR #1468, opened 2016-02-05
Allow revocation of direct shares in the Who Has Access view.

dwrensha:
Baby steps towards a more fully-featured Who Has Access view.

This change adds a revoke button for each time the current user appears in the "added by" column. It looks like:

whohasaccess

Clicking the button revokes the share. If the share went through a sharing link, a confirmation dialog is displayed that explains who else might also be affected by the revocation.

### DIFF
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -368,7 +368,7 @@ limitations under the License.
             <span class="token-petname" data-token-id="{{_id}}">
               {{petname}} ({{dateString created}})
             </span>
-            <button class="revoke-token" title="revoke" data-token-id="{{_id}}">revoke</button>
+            <button class="revoke-token" title="Revoke" data-token-id="{{_id}}">revoke</button>
           </li>
         {{/if}}
       {{/each}}
@@ -405,7 +405,7 @@ limitations under the License.
                   <li>
                     {{displayName identityId}}
                     {{#if isCurrentIdentity identityId}}
-                      <button class="revoke-access" title="revoke" data-recipient="{{../recipient}}">
+                      <button class="revoke-access" title="Revoke" data-recipient="{{../recipient}}">
                         revoke
                       </button>
                     {{/if}}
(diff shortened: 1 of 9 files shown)


shell/client/shell.html
```diff
@@ -401,8 +401,15 @@ <h4 class="who-has-access">Who has access</h4>
...
             <td> added by
               <ul>
-                {{#each shares}}
-                  <li>{{displayName identityId}}</li>
+                {{#each dedupedShares}}
+                  <li>
+                    {{displayName identityId}}
+                    {{#if isCurrentIdentity identityId}}
+                      <button class="revoke-access" title="revoke" data-recipient="{{../recipient}}">
+                        revoke
```
kentonv: Nitpick: We usually capitalize button titles/labels, I think.

shell/shared/grain.js
```diff
@@ -985,7 +971,116 @@ if (Meteor.isClient) {
...
+      const transitiveShares = instance.transitiveShares.get();
+      const tokensById = instance.downstreamTokensById.get();
+      const recipientShares = _.findWhere(transitiveShares, {recipient: recipient});
+      const currentIdentityId = Accounts.getCurrentIdentityId();
+      const recipientTokens = _.where(recipientShares.allShares, {identityId: currentIdentityId});
+
+      // Two cases:
+      // 1. All of the tokens are direct shares. Easy. Just revoke them.
+      // 2. Some of the links are child tokens. For each, we walk up the chain of parents to the
+      //    root. Then we walk downwards to see whether an other identities would be immediately
```
kentonv: an -> any

dwrensha:
Comments have been addressed.

kentonv:
Looks like you updated the "title" attributes but not the button body text? Not that the body text is visible anyway, but usually we capitalize that too...

dwrensha:
Ah, OK, got those too now.

PR #1470, opened 2016-02-05
sandstorm-accounts-oauth -> accounts-oauth

dwrensha:
As I pointed out in [this comment](https://github.com/sandstorm-io/sandstorm/pull/1454#issuecomment-178768431), we no longer need to maintain this forked package. Let's drop it!


paulproteus:
Quoth @zarvox:

Code deleted is code tested
​

PR #1478, opened 2016-02-06
Remove the `sandstorm reset-oauth` command.

dwrensha:
The `sandstorm reset-oauth` command was useful back when we had no admin settings page and when changing the `BASE_URL` didn't not automatically clear the oauth config. Nowadays, however, I think its existence is a net negative.

@mrdomino reported invoking the command, expecting it to log out all currently logged-in users, but instead getting into a state with broken config and then not noticing until a rather long time later that nobody could log in. Had he visited the admin/settings page instead of invoking `sandstorm reset-oauth`, this confusion would likely have been averted. (Note that our admin/settings page has both "reconfigure" and "log out all users of this provider" options.) The situation was made worse by the fact that `db.meteor_accounts_loginServiceConfiguration.remove({})` leaves the database in a weird state. It would be better if the reset did something like we do [here](https://github.com/sandstorm-io/sandstorm/blob/4e679e43fa29148fb9024f7b1fbe9dd87c60d4f7/shell/server/startup.js#L30-L36), but that would be somewhat nontrivial to do from run-bundle.c++; I think it would at least require us to enumerate the oauth login providers there, which would be sad.

### DIFF
--- a/src/sandstorm/run-bundle.c++
+++ b/src/sandstorm/run-bundle.c++
@@ -481,15 +481,6 @@ public:
               return alternateMain->getMain();
             },
             "Manipulate spk files.")
-        .addSubCommand("reset-oauth",
-            [this]() {
-              return kj::MainBuilder(context, VERSION,
-                      "Resets the OAuth configuration of Meteor by deleting the configuration "
-                      "that is stored in Mongo.")
-                  .callAfterParsing(KJ_BIND_METHOD(*this, resetOauth))
-                  .build();
-            },
-            "Reset OAuth configuration.")
         .addSubCommand("continue",
             [this]() {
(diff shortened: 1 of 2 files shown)


kentonv:
Agreed.

paulproteus:
+1 it had not occurred to me to outright _remove_ this command. Bravo.

paulproteus:
Doing `git grep reset-oauth` shows one hit in the docs, so I'll file a PR to remove/update that.

paulproteus:
Docs updated. Props to y'all. FWIW in the future consider doing `git grep reset-oauth` or similar when removing admin commands, but I can live with doing these sorts of things for now.

PR #1479, opened 2016-02-06
Remove remaining reset-oauth mention in docs

paulproteus:
#1478 removes the command, so let's just stop mentioning it.

The admin UI will step the user through any further steps they need.

Thanks to @mrdomino for using this command which we realized was sort of useless so we deleted it, and to @dwrensha and @kentonv for deleting it.


paulproteus:

PR #1481, opened 2016-02-07
Renames @asheesh gpg key to sig filetype for consistency.

patcon:
Was copy-pasting from docs for GPG verification, but trusting asheesh's key instead. The diff in filename caused confusion that made this take about 15 minutes more :)


kentonv:
Note that Asheesh is actually @paulproteus. :)

patcon:
Thanks, but found that via `keybase id paulproteus@github` ;) I am finding keybase incredibly useful already! Yay!

paulproteus:
+1

Thanks @patcon for finding this! I'll merge this.

PR #1487, opened 2016-02-09
On admin screen, detect misconfigured WILDCARD_HOST & show warning

paulproteus:
This is some fairly minimal changes for showing information to the admin about a misconfigured `WILDCARD_HOST`.

Screenshot of what admin looks like when this warning is visible:

wildcard-host-warning

Review request from:

- @neynah and/or @zarvox - please sanity-check my proposed text

Review & merge requested from:

- @kentonv

Note for Kenton: This adds a new subdomain pattern, `selftest*`.

Note that this _only_ affects `/admin/*` routes. That means that people can still get a forever-spinning grain spinner.

From my perspective, @jadeqwang and I want to limit how much time I spend on this particular bit of self-testing, so I'd like to declare making the `grain-frame` provide warning text as out-of-scope for this pull request. IMHO the most important thing is that people see this when they are doing initial setup.

Testing done:

- Set my `BASE_URL` and `WILDCARD_HOST` to point at `mouse-potato.com` which is a domain that resolves to 127.0.0.1 Verified that I saw the message.

- Fixed my `WILDCARD_HOST` and loaded the page up via `local.sandstorm.io`.

Rationale for new subdomain type:

- When I was manually testing this in Firefox, Firefox would cache the CORS result from a broken domain - and the CORS result from a broken domain is "no, CORS is not OK". Once I realized that, I realized that I needed each self-test to be on its own domain. So I created a new subdomain type.

### DIFF
--- a/shell/server/pre-meteor.js
+++ b/shell/server/pre-meteor.js
@@ -211,7 +211,7 @@ function checkMagic(buf, magic) {
 }
 
 function serveSelfTest(req, res) {
-  inMeteor(() => {
+  try {
     if (req.method === 'GET' &&
         req.url === '/') {
       const content = new Buffer("Self-test OK.");
@@ -226,7 +226,9 @@ function serveSelfTest(req, res) {
       res.end('Bad request to self-test subdomain.');
       return;
     }
-  });
+  } catch (err) {
+    console.log("WARNING: An error occurred while serving self-test; proceeding anyway:", err);
+  }
 }
 
 function serveStaticAsset(req, res) {
--- a/shell/server/pre-meteor.js
+++ b/shell/server/pre-meteor.js
@@ -210,7 +210,24 @@ function checkMagic(buf, magic) {
   return true;
 }
 
-function serveStaticAsset(req, res) {
+function serveSelfTest(req, res) {
+  inMeteor(() => {
+    if (req.method === 'GET' &&
+        req.url === '/') {
+      return serveStaticAsset(req, res, {
+        content: new Buffer("Self-test OK.")
+      });
+    } else if (req.method === 'OPTIONS') {
+      return serveStaticAsset(req, res);
+    } else {
+      res.writeHead(400, { 'Content-Type': 'text/plain' });
+      res.end('Bad request to self-test subdomain.');
+      return;
+    }
+  });
+}
+
+function serveStaticAsset(req, res, overrideAsset) {
   inMeteor(() => {
     if (req.method === 'GET') {
       // jscs:disable validateQuoteMarks
@@ -230,8 +247,13 @@ function serveStaticAsset(req, res) {
         return;
       }
 
-      const url = Url.parse(req.url);
-      const asset = globalDb.getStaticAsset(url.pathname.slice(1));
+      let asset;
+      if (overrideAsset) {
+        asset = overrideAsset;
+      } else {
+        const url = Url.parse(req.url);
+        asset = globalDb.getStaticAsset(url.pathname.slice(1));
+      }
 
       if (asset) {
         const headers = {
@@ -468,6 +490,12 @@ Meteor.startup(() => {
       if (id === 'static') {
         // Static assets domain.
         serveStaticAsset(req, res);
+        return;
+      }
+
+      if (id.match(/^selftest/)) {
+        // Self test domain pattern.
+        serveSelfTest(req, res);
         return;
       }
 
--- a/shell/server/pre-meteor.js
+++ b/shell/server/pre-meteor.js
@@ -214,11 +214,13 @@ function serveSelfTest(req, res) {
   inMeteor(() => {
     if (req.method === 'GET' &&
         req.url === '/') {
-      return serveStaticAsset(req, res, {
-        content: new Buffer("Self-test OK.")
+      const content = new Buffer("Self-test OK.");
+      res.writeHead(200, {
+        'Content-Type': 'text/plain',
+        'Content-Length': content.length,
+        'Access-Control-Allow-Origin': process.env.ROOT_URL,
       });
-    } else if (req.method === 'OPTIONS') {
-      return serveStaticAsset(req, res);
+      res.end(content);
     } else {
       res.writeHead(400, { 'Content-Type': 'text/plain' });
       res.end('Bad request to self-test subdomain.');
@@ -227,7 +229,7 @@ function serveSelfTest(req, res) {
   });
 }
 
-function serveStaticAsset(req, res, overrideAsset) {
+function serveStaticAsset(req, res) {
   inMeteor(() => {
     if (req.method === 'GET') {
       // jscs:disable validateQuoteMarks
@@ -247,13 +249,8 @@ function serveStaticAsset(req, res, overrideAsset) {
         return;
       }
 
-      let asset;
-      if (overrideAsset) {
-        asset = overrideAsset;
-      } else {
-        const url = Url.parse(req.url);
-        asset = globalDb.getStaticAsset(url.pathname.slice(1));
-      }
+      const url = Url.parse(req.url);
+      const asset = globalDb.getStaticAsset(url.pathname.slice(1));
 
       if (asset) {
         const headers = {
--- a/shell/server/pre-meteor.js
+++ b/shell/server/pre-meteor.js
@@ -490,8 +490,9 @@ Meteor.startup(() => {
         return;
       }
 
-      if (id.match(/^selftest/)) {
-        // Self test domain pattern.
+      if (id.match(/^selftest-/)) {
+        // Self test domain pattern. Starts w/ hyphen to avoid ambiguity with grain session/static
+        // publishing wildcard hosts.
         serveSelfTest(req, res);
         return;
       }
--- a/shell/shared/admin.js
+++ b/shell/shared/admin.js
@@ -46,6 +46,29 @@ var adminRoute = RouteController.extend({
   },
 
   action: function () {
+    // Test the WILDCARD_HOST for sanity.
+    Deps.nonreactive(function() {
+      if (Session.get("alreadyTestedWildcardHost")) {
+        return;
+      }
+      let req = new XMLHttpRequest();
+      req.timeout = 2000;
+      const fail = () => {
+        Session.set("alreadyTestedWildcardHost", true);
+        Session.set("wildcardHostWorks", false);
+      };
+      const succeed = () => {
+        Session.set("alreadyTestedWildcardHost", true);
+        Session.set("wildcardHostWorks", true);
+      };
+      req.addEventListener("load", succeed);
+      req.addEventListener("error", fail);
+      req.addEventListener("abort", fail);
+      req.open("GET", "//" + makeWildcardHost("selftest" + Random.hexString(20)));
+      req.overrideMimeType('text/plain');
+      req.send();
+    });
+
     var state = this.state;
     Meteor.call("getSmtpUrl", this.params._token, function(error, result){
       state.set("smtpUrl", result);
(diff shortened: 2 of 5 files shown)


paulproteus:
p.s. `mouse-potato.com` I love you so.

paulproteus:
Note for reviewers:

the link goes to https://docs.sandstorm.io/en/latest/administering/faq/#why-do-i-see-an-error-when-i-try-to-launch-an-app-even-when-the-sandstorm-interface-works-fine which is a directly link to the particular relevant FAQ entry

shell/server/pre-meteor.js
```diff
@@ -210,7 +210,24 @@ function checkMagic(buf, magic) {
   return true;
 }
 
-function serveStaticAsset(req, res) {
+function serveSelfTest(req, res) {
+  inMeteor(() => {
+    if (req.method === 'GET' &&
+        req.url === '/') {
+      return serveStaticAsset(req, res, {
```
zarvox: I dislike this overriding on `serveStaticAsset`.  They are not conceptually related, and I think this is the wrong level of abstraction to reuse code at.  I'd rather you just

```
res.writeHead(200, {"Content-Type": "text/plain"});
res.end("Self-test OK.");
```

for `GET`, extract a method for handling the 'OPTIONS' request (if the proper handling matches the one in `serveStaticAsset` exactly), and explicitly 405 anything else.

neynah:
Wow that's scary!

Regarding copy:
- do we say Sandstorm install or Sandstorm Server? (difference is unclear to me)
- perhaps: "Once you fix the issue, reload this page." My assumption here is that there's no actual "test" other than refreshing to see if the issue is resolved?
- we could remove the documentation sentence and instead say "See documentation." or "Learn more." I think we can justify shortening it because it's clear the documentation or learn more is regarding WILDCARD_HOSTS due to the context.
- I'm unsure about "we want to help you succeed" as we don't say this anywhere else (as far as I know). And if we don't, then it makes this warning dialogue seem oddly personal. Maybe instead: "If you're still having problems, e-mail us at support@sandstorm.io." This also suggests that they should at least attempt to fix the problem themselves first.

shell/server/pre-meteor.js
paulproteus: I can live with that. FWIW I overloaded `staticServeAsset` because I don't
love the idea of writing copy-pasta for OPTIONS. I could extract _that_
into a function.

zarvox: Note that you don't actually care about getting the same etag behavior as the static asset domain, since these are single-use host/filename pairs.  You only have one content type, you don't want to accept POSTs, and since you don't allow arbitrary user content and serve up text-plain, you don't need to worry about the rest of the CSP headers.

Maybe you need the Access-Control-Allow-Origin header, and the Content-Length header too.

paulproteus:
> do we say Sandstorm install or Sandstorm Server? (difference is unclear to me)

There's no real difference. I don't think we're consistent.

> perhaps: "Once you fix the issue, reload this page." My assumption here is that there's no actual "test" other than refreshing to see if the issue is resolved?

Yeah, the page does a diagnostic when it loads.

> we could remove the documentation sentence and instead say "See documentation." or "Learn more." I think we can justify shortening it because it's clear the documentation or learn more is regarding WILDCARD_HOSTS due to the context.

Interesting; will do that.

> I'm unsure about "we want to help you succeed" as we don't say this anywhere else (as far as I know). And if we don't, then it makes this warning dialogue seem oddly personal. Maybe instead: "If you're still having problems, e-mail us at support@sandstorm.io." This also suggests that they should at least attempt to fix the problem themselves first.

+1 I just want to make sure they're not totally put off from using Sandstorm after seeing this

shell/server/pre-meteor.js
paulproteus: Agreed, I suppose, although personally my heart grows warmer with everything I block via CSP.

neynah:
Oh! Another thing worth mentioning, perhaps we should state what is impacted by this issue? I know some people (myself included) ignore warnings especially if we can continue doing the thing we're doing.

shell/shared/admin.js
```diff
@@ -46,6 +46,29 @@ var adminRoute = RouteController.extend({
   },
 
   action: function () {
+    // Test the WILDCARD_HOST for sanity.
+    Deps.nonreactive(function() {
```
dwrensha: s/Deps/Tracker/ (see https://github.com/meteor/meteor/blob/devel/packages/deps/package.js)

paulproteus: +1, on it, many thanks

shell/shared/admin.js
```diff
@@ -46,6 +46,29 @@ var adminRoute = RouteController.extend({
...
+      if (Session.get("alreadyTestedWildcardHost")) {
+        return;
+      }
+      let req = new XMLHttpRequest();
```
kentonv: Meteor's `HTTP` package has a much nicer interface. Consider using it rather than using XHR directly?

shell/server/pre-meteor.js
kentonv: I agree this should not use the `static` subdomain, since all kinds of special-purpose behavior might apply to it in the future. I'd prefer using an independent subdomain like `self-test`. You don't need to handle `OPTIONS` -- for a `GET` request, setting `Access-Control-Allow-Origin: *` in the response is all you need.

paulproteus: +1 to all this

paulproteus: FWIW I went with `selftest` not `self-test` because `matchWildcardHost` doesn't permit hyphens, and I can live without a hyphen.

shell/shared/admin.js
paulproteus: Sensible, will try that

kentonv:
* Note that the FAQ link has an auto-generated ID which would change with any slight change to the question heading. Should we perhaps trampoline through a stable URL that will redirect to the appropriate place, so that if needed we can update the link server-side rather than have to push a new Sandstorm release?

* I wonder if changing WILDCARD_HOST and restarting the server will automatically cause clients to refresh. I think it will because Meteor.settings will have changed. So maybe we don't even need to instruct people to refresh manually?

paulproteus:
> Note that the FAQ link has an auto-generated ID which would change with any slight change to the question heading. Should we perhaps trampoline through a stable URL that will redirect to the appropriate place, so that if needed we can update the link server-side rather than have to push a new Sandstorm release?

I link to that from enough places that I know I'm going to have this problem in the future.

For now, my plan is (in the long run) to have docs releases correspond to Sandstorm releases, and then this link can link to the heading in the version of Sandstorm the user is running. So I'm -0 on doing  anything smart about this right now.

> I wonder if changing WILDCARD_HOST and restarting the server will automatically cause clients to refresh. I think it will because Meteor.settings will have changed. So maybe we don't even need to instruct people to refresh manually?

Possible. I can test it.

shell/server/pre-meteor.js
kentonv: Oh I see you were using a separate subdomain all along, so my comment was silly. But reusing the code to handle OPTIONS still seems unnecessary since GETs don't require preflights...

shell/shared/admin.js
paulproteus: Using that now.

paulproteus:
New warning text:

wildcard-host-warning-retr

paulproteus:
New text, in context:

wildcard-host-warning-retry-2

paulproteus:
wildcard-host-warning-retry-3

Adjusted text to say "Until you fix it", which makes it more ominous and also emphasizes that action is required.

paulproteus:
OK @kentonv I believe this addresses all feedback.

Note: @zarvox I chose to do 400 for all requests that aren't either `GET` to `/`, or `OPTIONS`. I could have conditional logic to `400` or `404` on `GET` of paths that aren't `/`, and additionally `405` (bad method) on non-`GET` methods, but in the interest of implementation simplicity I would prefer not to. If you would prefer I go all-in on 400 vs 404 vs 405, then I can do that.

paulproteus:
@dwrensha thanks again for the note about `Tracker` vs `Deps`. It's a real pleasure to work with people who provide quick feedback to help my code be more reasonable.

Same to @kentonv and @zarvox for code feedback, and @neynah for string feedback.

paulproteus:
Oh, also, I narrowed `Access-Control-Allow-Origin` to just the shell's origin.

paulproteus:
@kentonv this is ready for your re-review & merge, if you're willing!

shell/server/pre-meteor.js
```diff
@@ -210,6 +210,25 @@ function checkMagic(buf, magic) {
   return true;
 }
 
+function serveSelfTest(req, res) {
+  inMeteor(() => {
```
kentonv: This inMeteor() is unnecessary since you aren't using any meteor APIs in the code within.

shell/server/pre-meteor.js
```diff
@@ -210,6 +210,25 @@ function checkMagic(buf, magic) {
...
+    if (req.method === 'GET' &&
+        req.url === '/') {
+      const content = new Buffer("Self-test OK.");
+      res.writeHead(200, {
+        'Content-Type': 'text/plain',
+        'Content-Length': content.length,
+        'Access-Control-Allow-Origin': process.env.ROOT_URL,
```
kentonv: You could make this `*`, which would avoid the need to change it later when we support multiple `ROOT_URL`s. I don't see any danger in allowing all origins.

shell/server/pre-meteor.js
```diff
@@ -471,6 +490,12 @@ Meteor.startup(() => {
         return;
       }
 
+      if (id.match(/^selftest/)) {
```
kentonv: Technically, a valid grain host could begin with "selftest". Note that "static" is not a valid grain host because grain host IDs are always longer than that, but for "selftest" you're only matching a prefix. I think it would make sense to permit hyphens in wildcard IDs so that you can make this "selftest-*", which is never a valid grain host.

kentonv: (We may want to use this sort of pattern for other things in the future, e.g. API endpoints being `api-*`.)

kentonv: FWIW I think making hyphens be allowed is a one-char change to a regex in db.js.

shell/server/pre-meteor.js
paulproteus: But it does mean that exceptions won't bubble-up to the top and destroy the entire process, which otherwise they would do, I believe.

If I'm mistaken, let me know, but I'm gently in favor of keeping it given the above.

shell/server/pre-meteor.js
paulproteus: In theory allowing all origins is an information leak that this entity is running a Sandstorm at their domain. If we don't need `*` yet, then I'm happier to limit this to just what is needed. I can allow `*` if you very much prefer, though.

shell/server/pre-meteor.js
paulproteus: I always use `Random.hexString(20)` in the "client" code for this, so I'm +1 on me changing this to be a different length, too.

I'd prefer the length-based way to differentiate the selftest hosts, but I make it use a hyphen if you prefer that.

paulproteus: OK, I'll support hyphens! Done in a follow-up commit.

shell/server/pre-meteor.js
paulproteus: Upon re-reading this comment, I see that `inMeteor()` is something that we wrote, not something that comes with Meteor. Since `try/except` is all I need, I'll switch to that.

paulproteus:
@kentonv in your last review, you had these requests.

- Support `Access-Control-Allow-Origin: *`. However, I think it's OK to only allow the shell's endpoint. I opened https://github.com/sandstorm-io/sandstorm/pull/1490 if you prefer to merge it that way.

- Stop using `inMeteor()`: Done. The only thing I needed was try/catch.

- Use a hyphen to clarify `selftest-*` wildcards from other wildcards. Done.

Ready for your fresh review/merge. I personally prefer this branch over #1490 but either way works for me.

kentonv:
Hmm, apparently we need to update the kernel headers on our build machine.

kentonv:
Garply, retest this please.

PR #1492, opened 2016-02-09
docs: Fork theme; add separate sandstorm.io link to nav

### LINKED ISSUE #1405: [Docs] Navigation back to website is inefficient
For the past 4 user tests I've conducted on our old and new website there is a persisting annoyance for users navigating from Docs back to sandstorm.io. Currently, the only way to navigate back to the website would be to hit the back button (or type in the URL). This seems fine at first but if you start clicking on navigation within docs, then you have to click the back button multiple times to get back to the website.

I have 2 options in mind with a preference for option A.

A. Add a navigation button/link on docs to go to sandstorm.io

B. Have docs open in a new tab

paulproteus:
This commit:

- Forks base.html from the mkdocs readthedocs theme.

- Copies in the Javascript unmodified, since I was having problems with the JS
  not executing locally otherwise.

- Adds a README.md to this directory.

- Uses extra_css.

- Adjusts mkdocs.yml to point at the customizations.

- Uses flexbox to add a "sandstorm.io" link so people can get to
  the sandstorm.io home.

- Adjusts mkdocs.yml to rename this project to just "Docs" rather tha
  "Sandstorm" or "Sandstorm Docs" so that the navigation link in the top
  is clearer.

- Tells mkdocs to ignore the env/ directory, which contains a
  virtualenv.

Close #1405

### DIFF
--- a/docs/extra.css
+++ b/docs/extra.css
@@ -1,0 +1,41 @@
+.wy-side-nav-search>.home-and-project {
+    display: flex;
+    text-align: left;
+}
+
+.home-and-project>a.docs-home {
+    /* Push anything else in the flexbox as far right as possible. */
+    flex: 1 0;
+}
+
+
+.wy-side-nav-search>.home-and-project>a {
+    /* Fix vertical alignment problem */
+    line-height: initial;
+    /* Use default A styling for .wy-side-nav-search from theme.css */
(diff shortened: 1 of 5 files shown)


paulproteus:

PR #1493, opened 2016-02-09
docs: Remove vendor'd js; move virtualenv up one level

paulproteus:
vendoring the JS created a hilarious problem: the 'toggleClass()' call
would be run *twice* per click, making it look like it was called 0 times.

Having the virtualenv be in the docs/ directory made mkdocs confused, so
it's best to just not do that, I guess. (cc: @dwrensha )


paulproteus:
Also cc: @kentonv - you'll probably want to `rm -rf docs/env` and re-set-up your docs environment per the new README.

paulproteus:

PR #1491, opened 2016-02-09
Patching away issue #1437

synchrone:
Issue #1437 prevents CalDAV\CardDAV applications (e.g infCloud) from PUTting their objects onto backend. This patch makes it possible to ignore If-None-Match: * header value, since it does not make anything different for sandstorm.

PR #1494, opened 2016-02-10
Support If-None-Match: *.

### LINKED ISSUE #1437: If-None-Match: * should be accepted (rather than turn the request into a 400)
@synchrone noticed on #sandstorm earlier today that we crash in `proxy.js` if a request comes in with `if-none-match: *`.

@dwrensha noticed we do handle `if-match: *` but don't yet handle `if-none-match: *`.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html indicates that `*` is an acceptable value.

kentonv:
Fixes #1437

Closes #1491

Note that https://github.com/kentonv/node-capnp/commit/10e269d298189236881d038413950b0302610e9a is also needed for `*` to work correctly (for both `If-Match` and `If-None-Match`).

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -1928,6 +1928,10 @@ const parsePreconditionHeader = (request) => {
   }
 
   if (request.headers['if-none-match']) {
+    if (request.headers['if-none-match'].trim() === '*') {
+      return { doesntExist: null };
+    }
+
     const noneMatches = parseETagList(request.headers['if-none-match']);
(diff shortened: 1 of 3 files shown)


jparyani:
LGTM

paulproteus:
After reading this patch and saying out loud, "This is so great," I then spent about 5 minutes explaining to someone how awesome this patch is. (Thank you @kentonv for a high-clarity fix.)

PR #1495, opened 2016-02-10
Apparently now libseccomp defines the new syscall numbers.

dwrensha:
This undoes #1488. Apparently the latest libseccomp now defines the values? I'm curious whether this will pass in Jenkins...

### DIFF
--- a/src/sandstorm/supervisor.c++
+++ b/src/sandstorm/supervisor.c++
@@ -1112,9 +1112,6 @@ void SupervisorMain::setupSeccomp() {
   // New syscalls that don't seem useful to Sandstorm apps therefore we will disallow them.
   // TODO(cleanup): Can we somehow specify "disallow all calls greater than N" to preemptively
   //   disable things until we've reviewed them?
-#ifndef __NR_userfaultfd
-#define __NR_userfaultfd 323
-#endif
   CHECK_SECCOMP(seccomp_rule_add(ctx, SCMP_ACT_ERRNO(ENOSYS), SCMP_SYS(userfaultfd), 0));
 
   // TOOD(someday): See if we can get away with turning off mincore, madvise, sysinfo etc.


dwrensha:
Fascinating. Apparently `__NR_seccomp` and `__NR_bpf` are still not defined.

dwrensha:
It looks like [these libseccomp commits](https://github.com/seccomp/libseccomp/compare/7f3ae6e6a1...099f4214ce) got pushed to master yesterday.  They include https://github.com/seccomp/libseccomp/commit/d2ca11b7cdddbba3782b1e306ceacf19e898faee, which defines `__NR_userfaultfd`.

Unfortunately, it gets defined to `-10200`, which, as far I understand, is not the correct value (and I verified that this is indeed what the the Sandstorm supervisor ends up seeing for the value). I guess that this is just meant to be a dummy value?

I'm also confused about why such a definition would exist for the `userfaultfd` syscall but not for the `seccomp` and `bpf` syscalls.

Does anyone have a better understanding of what's going on here? @amluto?

dwrensha:
This seems to indicate that we should move our `#ifndef __NR_userfaultfd` so that it happens before we `#include <seccomp.h>`.

amluto:
Yuck.  They should not be injecting incorrect definitions like that.  I'll
complain at them.

At some point we should drop libseccomp.  It's problematic for a few
reasons.

kentonv:
Yeah uh IMO this is a security flaw in libseccomp -- presumably this means that if you build the supervisor against old headers, you silently end up with a sandbox that doesn't filter the newer syscalls. That's really bad!

@amluto I'm always happy to indulge in more NIH. ;)

kentonv:
Hmm actually I suspect what they're doing is making up for old kernel headers by falling back to internal tables. Presumably the weird negative value gets mapped through a table inside the library to get the correct value for the host platform. We should probably test to find out.

It's still weird that they'd override the `__NR_*` constants which are used by more than just libseccomp. This could mean that someone who uses the `syscall()` function ends up making a totally bogus syscall.

amluto:
I filed:

https://github.com/seccomp/libseccomp/issues/27

dwrensha:
Closing due to https://github.com/sandstorm-io/sandstorm/pull/1672

PR #1502, opened 2016-02-10
Remove some obsolete styling.

dwrensha:


dwrensha:
These styles applied to the old dev-login page, which no longer exists.

PR #1506, opened 2016-02-11
install.sh: Clarified (sub)domain recovery

elimisteve:
It asked for domain, but if you type the domain (rather than the subdomain)
you get an error.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1536,7 +1536,7 @@ sandcats_provide_help() {
   echo "* Recover access to a domain you once registered with sandcats"
   echo ""
   echo "* Just press enter to go to the previous question."
-  DESIRED_SANDCATS_NAME=$(prompt "What Sandcats domain do you want to recover?" "none")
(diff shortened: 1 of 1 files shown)


elimisteve:
Impetus for this:

```
What Sandcats domain do you want to recover? [none] elimisteve.sandcats.io
OK. We will send a recovery token to the email address on file. Type no to abort.
OK to continue? [yes] 
{"error":{"rawHostname":{"Invalid 
format":true,"maxLength":true,"message":"*Invalid Input*"}}}
```

But it worked when I just typed the subdomain, hence the suggested word change.

paulproteus:
Thanks @elimisteve ! I'm going to merge this right now.

I'm also going to add behavior in `install.sh` where, if it notices this easy-to-make mistake, it tells you that it wants just the subdomain part.

elimisteve:
Perfect, thanks @paulproteus!

PR #1511, opened 2016-02-11
Add docs on xip.io

### LINKED ISSUE #1508: docs: Need to explain how to use an internal IP, such as xip.io (or adjust sandcats protocol to allow internal IP addresses)
In #1500 the nice person wants Sandstorm running on a private IP address. Unfortunately our docs provide no advice for this.

We should fix that.

paulproteus:
Thanks @sstephenson for creating & maintaining this service and for making it open source!

Close #1508


paulproteus:
Going to self-merge because this is 100% docs. Happy to hear any feedback that people have.

PR #1516, opened 2016-02-12
Use `box-sizing: border-box;` everywhere

zarvox:
This changeset switches to border-box box-sizing for everything by default.  This is widely considered best practice: http://www.paulirish.com/2012/box-sizing-border-box-ftw/

I did my best to audit our codebase for places where box-sizing would affect layout -- namely, elements with both a `width` or `max-width` and a `margin` or `border`.  To discover sites where layout would be affected, I manually added the rules in the last commit in Chrome to a tab open to oasis.sandstorm.io (to ensure that I discovered issues that only exist when e.g. payments are enabled), and toggled the rule on the `<html>` tag.  Where things changed appearance, I adjusted element sizes.

While there may be some subtle changes in layout that I have missed, I believe they should be generally inconsequential.

Things I specifically tested:

* Account page
* Notifications
* Referrals page
* About page
* App list page
* Grain list page
* App details page

If we discover a situation where we require `content-box` layout, we can specify `box-sizing: content-box;` on that element, which will give us the default browser behavior.

### DIFF
--- a/shell/client/_about.scss
+++ b/shell/client/_about.scss
@@ -88,12 +88,14 @@
         img {
           padding: 0 16px;
           display: block;
-          max-width: 122px;
+          box-sizing: border-box;
+          max-width: 154px;
           max-height: 121px;
         }
         .uniregistry-image {
           @extend %pseudo-img-tag;
-          width: 122px;
+          width: 154px;
+          box-sizing: border-box;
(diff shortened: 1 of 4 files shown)


paulproteus:
I, for one, welcome our new border-box overlords.​

PR #1518, opened 2016-02-12
Add tooltips to more topbar elements

### LINKED ISSUE #1505: Hover titles missing for some top nav items
Discovered to be inconvenient in user testing. I propose to be consistent and add hover text to the following top bar items (that are missing hover titles):
- share access
- webkey
- notifications

uifix-hovertext

zarvox:
Fixes #1505.

### DIFF
--- a/shell/client/_topbar.scss
+++ b/shell/client/_topbar.scss
@@ -411,9 +411,6 @@ body>.topbar {
     >.account, >.login {
       >button.show-popup {
         @extend %hamburger-menu-item;
-        >a.expiring-soon {
-          color: #e82538;
-        }
       }
 
       @media #{$desktop} {
@@ -421,14 +418,12 @@ body>.topbar {
         display: block;
 
         >button.show-popup {
-          >a {
-            padding: 0 8px;
-            display: block;
-          }
+          padding: 0 8px;
+          display: block;
         }
       }
       @media #{$mobile} {
-        >a {
+        >button.show-popup {
           font-weight: normal;
         }
       }
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -197,10 +197,12 @@ limitations under the License.
 </template>
 
 <template name="notifications">
+  <button class="show-popup" title="Notifications">
   {{#if notificationCount}}
     <span class="count">{{notificationCount}}</span>
   {{/if}}
   Notifications
+  </button>
 </template>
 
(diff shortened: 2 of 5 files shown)


shell/client/_topbar.scss
```diff
@@ -411,9 +411,6 @@ body>.topbar {
     >.account, >.login {
       >button.show-popup {
         @extend %hamburger-menu-item;
-        >a.expiring-soon {
```
zarvox: There appears to be no `expiring-soon` in any of our HTML templates, so this is dead and I'm removing it.

paulproteus: +1 to removing stale nonsense​

Extra +1 if you can autodetect it.

kentonv:
Somewhat sad to denormalize the button HTML into all the places. Would it work to make the title be a template parameter instead?

zarvox:
Hmmm, maybe.  I suppose we should probably be setting event handlers on the buttons themselves, rather than on the `<li>`s that contain them.  On the flipside, it is also sad that we wind up basically making our own pseudotemplate language to express a bunch of things that HTML already has rules for, and which winds up complicating how we write the stylesheets.

It's not even super clear that the topbar should own these buttons - what happens if we want a thing in the topbar that isn't a button, or has different margin/padding requirements?  How might I set a minimum-size for the powerbox buttons, or add a class to flash them briefly when the user triggers an action that makes a new apiToken?  You'd have to have a selector that captures a fair bit of context of the document, rather than being able to write a class that you stick on the thing that has the different needs.  And it's harder to tell, at a glance, what a particular topbar item is going to look like in the DOM.

I wrote a fairly long thing about why I'm pretty convinced that having the root element of a logical unit owned and rendered by its parent is wrong and leads to more fragile and confusing code and styles, but then it seemed like a ridiculous rant to a perfectly reasonable question.  I'll probably post the long form with a practical example comparing the approaches as a gist some time, but in short: the same reasons that make `box-sizing: border-box` better for managing element size make components-owning-their-roots better for styles and visualization reuse - it's easier to make changes without having to modify things unrelated to the change at hand, and is a better impedance match for what the programmer/designer wants to express.

If you'd still prefer the title be a template parameter instead, let me know and I'll do that real quick, since I want to see this UX issue fixed, but I think the principle behind that request makes templates and styles more confusing and less reusable on average.

PR #1519, opened 2016-02-12
Initial definition of feature keys.

kentonv:
A feature key unlocks "Sandstorm for Work" features. Most commercial software would use the term "license key" or "license file". However, a sandstorm feature key does not correspond to a license. The license is Apache 2.0, and everyone receives the license even if they don't have a feature key.

### DIFF
--- a/src/sandstorm/feature-key.capnp
+++ b/src/sandstorm/feature-key.capnp
@@ -63,7 +63,7 @@ struct FeatureKey {
 
   isElasticBilling @5 :Bool;
   # If true, then userLimit is a ceiling, but the customer will actually be billed based on
-  # monthly active user count at the end of each month.
+  # observed active user count at the end of each billing cycle.
 
   isTrial @6 :Bool;
   # Is this a trial key? For display purposes only.
--- a/src/sandstorm/feature-key.capnp
+++ b/src/sandstorm/feature-key.capnp
@@ -1,0 +1,83 @@
+# Sandstorm - Personal Cloud Sandbox
+# Copyright (c) 2016 Sandstorm Development Group, Inc. and contributors
+# All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+@0xd24581e9cd6a6772;
+
+using PublicSigningKey = import "update-tool.capnp".PublicSigningKey;
+
+$import "/capnp/c++.capnp".namespace("sandstorm");
+
+struct FeatureKey {
+  # A FeatureKey describes a customer of Sandstorm for Work and the features for which
+  # they have paid. This "key" is actually more of a certificate: it is a signed statement.
+  # However, it is also a key in that it alone unlocks the features, and should be treated as
+  # a secret.
+  #
+  # A complete key is formed by:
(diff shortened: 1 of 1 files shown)


src/sandstorm/feature-key.capnp
```diff
@@ -0,0 +1,83 @@
...
+  # However, it is also a key in that it alone unlocks the features, and should be treated as
+  # a secret.
+  #
+  # A complete key is formed by:
+  # 1) Serialize a FeatureKey in Cap'n Proto packed format. (Note that canonicalization
+  #    is not necessary, since we don't use detached signatures.)
+  # 2) Sign it using libsodium's crypto_box_easy().
+  # 3) Base64 the whole thing, for easy copy/paste.
+  # 4) Optionally insert newlines or other whitespace for readability.
+  # 5) Optionally add lines beginning with '-', which will be ighnored. (Commonly, lines like
```
zarvox: s/ighnored/ignored/

src/sandstorm/feature-key.capnp
```diff
@@ -0,0 +1,83 @@
...
+    contactEmail @3 :Text;      # email address of contact
+  }
+
+  customer @1 :Customer;
+
+  issued @2 :UInt64;
+  expires @3 :UInt64;
+  # Unix timestamp (seconds) when this key was issued and expires.
+  #
+  # Use UINT64_MAX for `expires` to mean (effectively) "never expires".
```
zarvox: Is there a reason to prefer a magic number over a union here (or in general)?

src/sandstorm/feature-key.capnp
```diff
@@ -0,0 +1,83 @@
...
+  #
+  # Tip: Although 64-bit, these values can safely be represented as numbers in Javascript.
+
+  userLimit @4 :UInt32;
+  # How many users are allowed? Use UINT32_MAX to mean (effectively) "unlimited".
+
+  isElasticBilling @5 :Bool;
+  # If true, then userLimit is a ceiling, but the customer will actually be billed based on
+  # monthly active user count at the end of each month.
```
zarvox: Just an observation: this needn't be bound to a monthly billing cycle.  It could be yearly, or quarterly, or whatever makes the most sense (which is almost certainly monthly).

zarvox:
Fix typo and merge at will!

src/sandstorm/feature-key.capnp
kentonv: It's not a magic number, it's a date so impossibly far in the future that it is effectively infinite. :) That is to say, I would not expect to see any code explicitly checking for said number.

kentonv: To expand on that a bit: I like this approach when it makes sense because it requires no special code to implement thus is harder to get wrong. Otherwise, we would need to explicitly test non-expiring certificates in addition to expiring ones.

zarvox: If we ship such a thing, it would be pretty unprofessional/confusing to put something in the UI that says that that customer's feature key expiry date is "15:30:08 UTC on Sunday, 4 December 292,277,026,596". :)  Instead, I'd special case it so that it says that there is no expiry date.  At which point I am checking for the magic number UINT64_MAX after all.

Maybe it makes the enforcement check less branchy, but when you consider that you need to show the user something that matches their expectations, the simplification is not as helpful as you might think.

Also, we won't be able to safely `parseInt()` the value in JS if you actually set it to UINT64_MAX.  Which maybe pokes a hole in the "harder to get wrong" claim.

In practice, I'm not strongly opinionated on this and you can ship it as-is; I just wanted to understand your reasoning, since it was a case where it felt like any work we save on the capnproto side we wind up paying for later.

kentonv: Hmm, you make a good point that we still need to special-case these for display purposes.

I don't think the `parseInt()` issue is a real problem, though: it would be perfectly fine test for "later than the year 1,000,000 AD" rather than for equality. :)

I guess with my approach, we only need special cases for display (which, if missed, is ugly and confusing but doesn't harm functionality), whereas with a union we need special cases in core logic as well as in display.

At this point I don't have a strong opinion either. Maybe if starting over I'd use a union but I think I'm going to merge as-is.

PR #1520, opened 2016-02-12
JSCS compliance for everything not in packages

zarvox:
This time, with double quotes preferred, more ES6 support, and a pile of rebased manual fixes.

### DIFF
--- a/shell/.jscsrc
+++ b/shell/.jscsrc
@@ -1,7 +1,9 @@
 {
   "preset": "airbnb",
+  "esnext": true,
+  "verbose": true,
   "fileExtensions": [".js"],
-  "validateQuoteMarks": {"mark": "'", "escape": true},
+  "validateQuoteMarks": {"mark": "\"", "escape": true},
   "disallowVar": true,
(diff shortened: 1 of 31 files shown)


zarvox:
Test failures are reproducible locally, so I'll look into figuring out why.

zarvox:
Okay, looks like it's unsafe to convert 

`var namedFunction = function () { ...`

into

`const namedFunction = function () { ...`

if namedFunction is declared within a block e.g. `if (Meteor.isClient) {` but then used outside of that block.  Which is something we shouldn't be doing at all, but which we definitely do in `shared/admin.js` and `shared/shell.js`.  I'll look for similar problems elsewhere.

zarvox:
I audited the codebase for other potentially dangerous blocks by running

`ag '\ +const.*= function'`

and

`ag '^\ +const.*=.*(.*) => {'`

and checking the output.

Any remaining blocks that match those patterns, I manually verified are safe to leave as-is, currently.

paulproteus:
yay!

Issue #1521, opened 2016-02-12
Add CalDAV\CardDAV user-agents to HTTP Basic auth whitelist


synchrone:
[Radicale app](https://github.com/synchrone/sandstorm-radicale) needs to provide access to clients for CardDAV\CalDAV background synchronization with phones and desktop applications.

Some of them are gathered in a [corresponding issue](https://github.com/synchrone/sandstorm-radicale/issues/3).

Pull request will follow

synchrone:
Discussion at https://github.com/sandstorm-io/sandstorm/pull/1522

PR #1522, opened 2016-02-12
Adding CalDAV\CardDAV client user-agents. Also curl for debugging.

synchrone:
Relates to issue #1521

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -138,6 +138,13 @@ BASIC_AUTH_USER_AGENTS = [
   'Mozilla\\/5\\.0 \\(iOS\\) ownCloud-iOS\\/',
   'Mozilla\\/5\\.0 \\(Android\\) ownCloud-android\\/',
   'litmus\\/',
+  'curl',
+  '.*iOS\\/.+dataaccessd', //also for iOS UA check compatibility mode employed by some clients
+  'CalDAV-Sync',
+  'CardDAV-Sync',
+  'MSFT-', //windows phones, and other MS clients
+  '.+Thunderbird\\/.+Lightning\\/',
+  'Jakarta Commons-HttpClient', //generic java clients who didn't bother to override default UA
 ];
 BASIC_AUTH_USER_AGENTS_REGEX = new RegExp('^(' + BASIC_AUTH_USER_AGENTS.join('|') + ')', '');
 


garply:

kentonv:
So I'm starting to think this is not the right approach after all. I guess I hadn't considered the case of a standard protocol used by lots of client apps, rather than a specific client app.

So now I'm thinking maybe we don't need to have a whitelist after all if we assign unique API hosts. There is still an XSRF risk if basic auth is used from a browser context, but as long as we can find a way to block basic-authed requests from browsers "most of the time", maybe that's enough that apps won't make the mistake of suggesting basic-auth-from-browser to users. Maybe then we can say that apps which _do_ make this mistake are responsible for the consequences of exposing themselves.

I'll try to work on this in the next week.

synchrone:
@kentonv have you considered denying based on the presence of Origin header? It seems to be 1) not sent by api consumers, 2) is secure from JS modifications, 3) sent by modern browsers in almost any case.

kentonv:
@synchrone Actually the origin header is sent on XHRs but not (IIRC) on cross-origin `<form>` posts. We actually explicitly want to allow XHRs, so that other web sites can access APIs. What we don't want is for the response to be rendered in a browser window, but XHRs don't do that.

garply:

zarvox:
@kentonv https://wiki.mozilla.org/Security/Origin suggests that the Origin header *is* sent on cross-origin `<form>` POSTs.

zarvox:
But in testing in Firefox, this does not appear to be the case.  Darn.

kentonv:
@zarvox Yeah I would _love_ it if they would send `Origin` for cross-site form posts. Then we could make our CSRF protection be 100% (even in the presence of DNS request sniffing). But I don't think anyone does that currently.

synchrone:
@kentonv from our conversation with @paulproteus on IRC, I recall the idea was to detect all the browser requests and forbid them, for the cases when the grain app misuses GET for modifying data and a malicious  can destroy your data.

I feel like there's a solid threat model for API endpoint missing. Maybe there is a need to define potential attackers and scenarios. Otherwise it seems to me the core dev views are not aligned.

kentonv:
@synchrone With yesterday's release (`sudo sandstorm update` if you don't have it already) we now assign each API token to a unique host and we now allow HTTP basic auth to all clients. I believe that makes this PR no longer needed. Can you check that Radicale works with these clients now? I'm eager to get it on the app market!

> I recall the idea was to detect all the browser requests and forbid them, for the cases when the grain app misuses GET for modifying data and a malicious ￼ can destroy your data.

No, we aim to prevent browsers from directly rendering content (and executing script) served from an API, but we explicitly do want to allow XHRs (which are not rendered). Abstractly speaking, the API host has different security properties than app GUI hosts. We have designed GUI hosts to be "sandboxed" in various ways, both to defend against external attacks and to prevent the app itself from launching attacks. Some of the things we do for GUI hosts don't work for API hosts. For example, GUI hosts are short-lived, which together with their unguessable nature helps to mitigate XSRF attacks. API hosts fundamentally can't be short-lived since you don't want to have to reconfigure your client for every session.

In the past, all APIs were served off a single host, with only the authorization token determining which grain's API was served. This had the nice security property of making it obvious that there was no way to access an API host without having the token, since the token literally identified the API. This is a fundamental property of capability-based security. However, this theoretical niceness had practical problems, in particular that we could not allow browsers to use HTTP basic auth when contacting this host, because browsers will cache HTTP auth credentials, which would then make it easy for any site in the browser to access the API in question.

synchrone:
Thank you for a sane solution!

PR #1524, opened 2016-02-12
Add the Origin header in sandstorm-http-bridge.

dwrensha:
This adds an Origin header to every request that goes through `WebSession`.

See the [Origin RFC](http://tools.ietf.org/html/rfc6454#section-7.3), the [WebSocket RFC](https://tools.ietf.org/html/rfc6455#section-10.2), and [this discussion on sandstorm-dev](https://groups.google.com/forum/#!topic/sandstorm-dev/MaFW3v3Refw).

### DIFF
--- a/src/sandstorm/sandstorm-http-bridge.c++
+++ b/src/sandstorm/sandstorm-http-bridge.c++
@@ -1213,9 +1213,12 @@ private:
     }
     lines.add(kj::str("X-Sandstorm-Permissions: ", permissions));
     if (basePath.size() > 0) {
+      auto host = extractHostFromUrl(basePath);
+      auto protocol = extractProtocolFromUrl(basePath);
       lines.add(kj::str("X-Sandstorm-Base-Path: ", basePath));
-      lines.add(kj::str("Host: ", extractHostFromUrl(basePath)));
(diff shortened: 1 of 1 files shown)


kentonv:
Hmm, this isn't quite right in the case where the WebSocket arrived over the API endpoint, in which case it won't necessarily be same-origin. In those cases, the socket is authorized via an authorization header, which prevents XSRF, but probably lying to the app and claiming the request is same-origin isn't a great idea. Really we want the app to understand that it is trusting Sandstorm to authenticate WebSockets, ignoring the origin header.

paulproteus:
in which case maybe sandstorm-http-bridge should generate the right magic
Origin header?​

kentonv:
One option might be to have a bogus origin (e.g. https://sandstorm-api) that is always used for API requests? But you need to distinguish API requests, which I guess you could do based on the session type being ApiSession, though I think Sandstorm falls back to WebSession if the app claims not to support ApiSession.

dwrensha:
Closing, because the implications of this change are unclear, and the person who had been hitting problems with the Origin header found a workaround: https://groups.google.com/d/msg/sandstorm-dev/MaFW3v3Refw/W9LVzcu6AAAJ

PR #1529, opened 2016-02-13
Add install & uninstall docs

paulproteus:
People sometimes want to know how to uninstall Sandstorm. There isn't a script, but there is now a complete guide.

I also took the liberty of improving the backups docs at the same time.

I also took the liberty of explaining what the install script actually does. I think this would have addressed @glyph's questions when he asked me things on IRC. I'll go check with him after pushing this.

### DIFF
--- a/docs/administering/backups.md
+++ b/docs/administering/backups.md
@@ -32,13 +32,16 @@ snapshots you can make a quick backup by running:
 
     cp -a /opt/sandstorm $HOME/sandstorm-snapshot-from-$(date -I)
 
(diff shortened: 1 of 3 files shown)


paulproteus:
I also took the liberty of adjusting the "Installation" of vagrant-spk in the table of contents to explain that it's not about installing Sandstorm itself.

paulproteus:
Self-merging because it's 100% docs.

If other people want to review, I'd be glad, but it's OK if not, too!

PR #1530, opened 2016-02-13
Tell people mongod failure could be due to low disk free

paulproteus:
Having a better log message could have helped @sanderant debug this without assistance, I believe.

We found ~350MB triggered the error case, but 500MB free resulted in success.

Requesting review/merge from @kentonv

### DIFF
--- a/docs/administering/faq.md
+++ b/docs/administering/faq.md
@@ -313,3 +313,23 @@ not by a member of the Sandstorm team. If you want to run your own wildcard DNS 
 xip.io inside your own organization, you can do so by [downloading
 xipd](https://github.com/sstephenson/xipd) which Sam generously licenses as open source software.
 You can also set up your own `*.sandstorm.example.com` subdomain within your organization's domain.
+
+## mongod failed to start. What's going on?
+
+If your Sandstorm server isn't working, and you find this text in `/opt/sandstorm/var/log/sandstorm.log`:
+
+```
+**mongod failed to start. Initial exit code: 100, bailing out now.
+```
+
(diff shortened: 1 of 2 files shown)


kentonv:
Note that this kind of hint, when it turns out to be wrong, leads to confusion. Would it make sense to link to some troubleshooting docs instead, where we could list multiple possible reasons for the problem?

paulproteus:
I don't think it's ever really been wrong before, and it wasted about half a day of my life personally. I am willing to write a quick troubleshooting doc note, though.

This commit adds it, and links to it by search query, so that if we move it around, we'll still be in good shape. @kentonv requesting your review/merge.

kentonv:
We've definitely seen Mongo fail for other reasons in the past (locale issues, unclean shutdown, etc. -- things we've fixed, but it has happened).

It occurs to me that we should really note somewhere that Mongo's true error message can be found in `/opt/sandstorm/var/log/mongo.log`. Might even be worth mentioning directly in the run-bundle.c++ error message, in addition to the docs.

Feel free to make whatever English text changes you deem appropriate and then click merge.

paulproteus:
@kentonv thanks for your feedback; you are totally right that we've seen other Mongo issues in the past.

I've adjusted the docs text to de-emphasize disk space, remarking it's just one option, and to point people to the Mongo log. I've also added a note about mongo.log to the `run-bundle.c++` error message; mega +1 on pointing people to the Mongo log file.

I'll let Jenkins test these, then self-merge. Any further feedback on the text is welcome, from you or others.

PR #1532, opened 2016-02-14
JSCS compliance for `shell/packages/*`

zarvox:
With this branch, we can now run:

```
jscs shell/
```

and get an exit code of 0 a couple seconds later.

I specifically skipped changing `shell/packages/sandstorm-permissions` (beyond adding a comment to disable JSCS on the files in the module) since I know @dwrensha is actively working on that bit and didn't want to make massive conflicting changes.  If there are other modules people are actively working on that they'd like me to leave alone for now, comment here and I'll rebase them out.

@jparyani How do you think we should automate keeping the shell in compliance as people make future changes?  A couple ideas:

* an automated test that runs against PRs and complains about style checker violations (ideally, orthogonal to the existing functional test suite, so that style errors don't prevent functional tests from running/demonstrating success)
* a pre-commit git hook that runs jscs on `shell/`, if `jscs` is on your `$PATH`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,6 +147,9 @@ test: sandstorm-$(BUILD)-fast.tar.xz
 installer-test:
 	(cd installer-tests && bash prepare-for-tests.sh && PYTHONUNBUFFERED=yes TERM=xterm SLOW_TEXT_TIMEOUT=120 ~/.local/bin/stodgy-tester --plugin stodgy_tester.plugins.sandstorm_installer_tests --on-vm-start=uninstall_sandstorm --rsync)
 
+stylecheck:
+	cd shell && jscs .
+
 # ====================================================================
 # Dependencies
 
--- a/Makefile
+++ b/Makefile
@@ -148,6 +148,7 @@ installer-test:
 	(cd installer-tests && bash prepare-for-tests.sh && PYTHONUNBUFFERED=yes TERM=xterm SLOW_TEXT_TIMEOUT=120 ~/.local/bin/stodgy-tester --plugin stodgy_tester.plugins.sandstorm_installer_tests --on-vm-start=uninstall_sandstorm --rsync)
 
 stylecheck:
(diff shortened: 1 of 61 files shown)


paulproteus:
I for one would be in favor of this being part of the PR review process via
a new Jenkins task run by the Garply build bot stuff.

It could be a separate check, per e.g.
https://github.com/blog/1935-see-results-from-all-pull-request-status-checks
, so that it's clear when something passes/fails style vs. passes/fails
functionality tests. So I like that idea you had.

Secondarily, "a pre-commit git hook that runs jscs on shell/, if jscs is on
your $PATH" is nice. One has to enable pre-commit git hooks on each clone,
and I probably would end up removing it locally to avoid slowing down my
"wip" commits. In which case, I'd love a "make stylecheck" thing I could
run before submitting.

zarvox:
@paulproteus Added a "make stylecheck" command for you.  I'm going to still rely on you to have `jscs` installed on your `PATH` somewhere.  If we don't like that, we could write an npm `package.json` which specifies a dependency on jscs at a particular version, and then run that one from `node_modules/.bin/jscs` (`npm run` I think will invoke the same?).

+1 on a second, orthogonal Jenkins task that just runs `make stylecheck`.

paulproteus:
I can live with relying on me to have jscs on my path if there's somewhere
documented on how to install it. I'm OK with the documentation being vague
and not in the "main" docs, so long as the docs are discoverable somehow.

For example, maybe make stylecheck could do:


stylecheck:
        which jscs > /dev/null 2>&1 || ( echo "You need jscs installed.
Consider installing with e.g. npm -g install jscs "; exit 1)
        jscs --options shell/* --whatever-drew-needs

zarvox:
That is a fine idea, which I have implemented in 40af7dbd318eac4318298d17d17253b77b6d5ad8

PR #1536, opened 2016-02-17
JSCS: don't warn about line length

zarvox:
We have a bunch of places that are over 100 characters, and I don't think they
are particularly harmful to readability.  I'd welcome over-time reworking
toward a 100-character line limit, but I don't think we need to be religious
about line lengths in general.

This rule was added to the airbnb preset recently, which is why my previous
changesets did not deal with this warning.


zarvox:
Self-merging for triviality and because I probably own the JSCS machinery anyway.

PR #1537, opened 2016-02-17
Add a Tiny Tiny RSS vulnerability to non-events.

kentonv:


paulproteus:
Typo: Sansdtorm => Sandstorm

I'll fix that in a follow-up commit since communication incurs latency.

PR #1543, opened 2016-02-18
Implement app ID replacement table.

kentonv:
This allows developers to recover from a lost or stolen app ID. The idea is that we compile directly into Sandstorm a mapping of lost keys and new keys that are replacing them. Apps signed with the new key will be treated by Sandstorm and all tools as if they are signed with the old key. Old keys can also optionally be revoked while specifying an exact list of known-good packages that had been signed with said keys. If the old key was merely lost but not believed to have leaked (e.g. all copies were destroyed), then revocation is not necessary.

@JamborJan in particular needs this for Paperwork.

This change is cleanly split into two commits which I recommend reading separately.

Also: Yay unit tests.

### DIFF
--- a/src/sandstorm/appid-replacements.c++
+++ b/src/sandstorm/appid-replacements.c++
@@ -1,0 +1,77 @@
+// Sandstorm - Personal Cloud Sandbox
+// Copyright (c) 2016 Sandstorm Development Group, Inc. and contributors
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "appid-replacements.h"
+#include "id-to-text.h"
+
+namespace sandstorm {
+
+void applyAppidReplacements(
+    kj::ArrayPtr<kj::byte> appId, kj::ArrayPtr<const kj::byte> packageId,
+    capnp::List<spk::AppIdReplacement>::Reader replacements) {
+  // Given an input app ID that was just verified to have signed the given package ID,
+  // check appid-replacements.capnp to see if the app ID is revoked (throws exception) or the
+  // package should be treated as some other app (replaces appId).
+
+  // The logic here is slightly weird because the replacement list is organized into events --
+  // which makes it easier for people modifying it to understand what to do -- rather than into
+  // rules -- which could be more directly processed. Each event introduces one or two rules:
+  // a revocation of an original key (except for grandfathered whitelist) and a mapping of a
+  // replacement key to and original key.
+
+  KJ_REQUIRE(appId.size() == APP_ID_BYTE_SIZE);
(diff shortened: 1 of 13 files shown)


kentonv:
Added commits to:

1. Not break installs -- Sandstorm actually pipes in the SPK data and my code had inadvertently made spk expect a file.
2. Make `spk publish` sign requests using the replacement key, and the app index expect this. This also changes things so that the pkgdef can (and should) go on listing the app's original ID, and spk will automatically use the new key as appropriate.

src/sandstorm/appid-replacements.capnp
```diff
@@ -0,0 +1,84 @@
...
+  # ---- example entry ----
+
+  (original = "vjvekechd398fn1t1kn1dgdnmaekqq9jkjv3zsgzymc4z913ref0",
+   replacement = "wq95qmutckc0yfmecv0ky96cqxgp156up8sv81yxvmery58q87jh",
+   revokeExceptPackageIds = ["b5bb9d8014a0f9b1d61e21e796d78dcc"]),
+  # This is the ID for My Cool Example App by Kenton Varda. The old app key was compromised when
+  # it was accidentally committed to the app's public git repo. Only one version ("b5bb...") had
+  # ever been published.
+
+  # ---- end exmaple; real entries follow ----
```
zarvox: s/exmaple/example/

src/sandstorm/appid-replacements.capnp
```diff
@@ -0,0 +1,84 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+@0xa53cae3f717a1676;
+# This file is used to recover from lost or stolen app keys.
+#
+# If you are an app developer and you have lost access to your key or suspect it may have been
```
zarvox: This probably needs to also make it into the docs, and be adapted for `vagrant-spk`, even if that's just saying `vagrant-spk ssh` first and specifying `--keyring /host-dot-sandstorm/sandstorm-keyring`

cc @paulproteus

src/sandstorm/appid-replacements.c++
```diff
@@ -0,0 +1,96 @@
...
+
+  // First check if this app ID is revoked.
+  for (auto item: replacements) {
+    if (item.getOriginal() == appidStr) {
+      // The app ID matches the `original` of this entry. Check if it was revoked.
+      if (item.hasRevokeExceptPackageIds()) {
+        // This app ID is revoked, except for specific package IDs...
+        for (auto allowed: item.getRevokeExceptPackageIds()) {
+          if (pkgidStr == allowed) {
+            goto checkReplacements;  // Need to break outer loop...
```
zarvox: This is a totally reasonable usage of `goto`, but man, I haven't seen one of these in ages! :D

src/sandstorm/appid-replacements.capnp
```diff
@@ -0,0 +1,84 @@
...
+# compromised, do the following:
+# 1. Generate a new key using `spk keygen`. The new key ID will be printed.
```
zarvox: I wrote https://github.com/sandstorm-io/vagrant-spk/pull/134 so you can now replace the `spk` in each of these with `vagrant-spk` and people will have a good time.

zarvox:
Address comments (spelling nitpick, s/spk/vagrant-spk/ in that one comment block) and merge at will!  (The js-style failure is from an unrelated commit, and is already fixed in master.)

PR #1546, opened 2016-02-19
Add new CSS for "Get free storage" section on /account

paulproteus:
This only affects Oasis.

It also only creates new CSS, so I believe it is totally safe to merge. *

*. Never believe engineers when they say something is totally safe. And yet I do think this is totally safe!

### DIFF
--- a/shell/client/_account.scss
+++ b/shell/client/_account.scss
@@ -885,6 +885,38 @@ div.billing {
       font-size: 90%;
     }
 
+    div.free-stuff-section-header {
+      padding-bottom: 0.6em;
+    }
+
+    div.free-stuff {
+      /* "Get free storage" area of /account */
+      display: flex;
+      align-items: center;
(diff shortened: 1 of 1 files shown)


paulproteus:
Responded to feedback from @kentonv. Currently blocked on a libseccomp build issue. Gotta say I don't love it when that happens. Maybe we can pin libseccomp within a release, and then right after doing a release, update it to the latest "stable", so that it only breaks master right after a release.

shell/client/_account.scss
```diff
@@ -880,11 +884,43 @@ div.billing {
...
+        font-family: Arial;
+        font-size: 16px;
+        font-weight: normal;
+        width: 125px;
+        flex-shrink: 0;
+      }
+    }
+
+
     .subscribe-prompt {
```
kentonv: subscribe-prompt is still here

PR #1547, opened 2016-02-19
Place each API token on a separate random subdomain.

kentonv:
Although we already do a lot to prevent API content from being consumed in any context where same origin policy matters, using separate subdomains provides defense-in-depth.

More importantly, though, on token-specific subdomains, we can relax the restriction on the use of HTTP basic auth. This will allow many clients not explicitly designed for Sandstorm to connect to Sandstorm-based apps. Note that an app which encourages users to access it through basic auth *in a browser* may be shooting itself in the foot by creating an XSRF problem: browsers will usually cache the credentials which may allow other web sites to address requests to the hostname without knowing the token. API hosts are unguessable like grain UI hosts, but unlike grain UI hosts, API hosts are long-lived, so there is somewhat more risk. Still, this risk seems overall small and the benefit of being able to use a wide range of client apps is large.

@synchrone This should make #1522 unnecessary -- all clients will be able to use basic auth now.

@mnutt With Davros, the box for the URL now has a horizontal scroll for me since the hostname is 32 chars longer than before. IIRC the reason you needed to include a username and password on this host was because the client starts by making an unauthenticated OPTIONS request? I wonder if we can now make these requests return the right thing.


mnutt:
Owncloud client has an authentication process whereby in the first step it discovers the server via an unauthenticated `/status.php` request before the UI asks for a username and password. I get around this by including the username and password in the URL and Owncloud Desktop is fine with this approach while Owncloud iOS and Android disallow this. (they implement their own http connection handling and refuse the connection if the server sends an authentication challenge, even though they have the username/password)

I'd love to solve this problem since it prevents the iOS and Android clients from working entirely, but I'm not sure there's a solution on the Sandstorm side that would allow an unauthenticated `/status.php` so I think we're stuck with it for now. From a UX perspective, I think a combination of the copy-to-clipboard button and some sort of visual indication that the text is overflowing the input field may be sufficient.

kentonv:
I realized that this change actually breaks Davros entirely because it displays multiple offer templates, one for the host and one for the password, and now the password will not work for the given host.

Possible solutions:

1. Try to automatically synchronize offer templates displayed by the same grain at the same time.
2. Require grains to opt-in to the new behavior, and also give them an explicit API to make offer templates align.

I'm going to try option 1 first.

@mnutt Could we hard-code Sandstorm to respond to an unauthenticated `status.php` request in a way that causes the ownCloud clients to continue? I'm thinking something that we'd apply to all API endpoints as a hack -- perhaps only for known ownCloud user agents.

paulproteus:
To be pedantic, _I_ realized it breaks Davros, I think. : P

kentonv:
@paulproteus Indeed, sorry!

paulproteus:
(-: np at all​ :-)

mnutt:
@kentonv hard-coding to return an unauthenticated `/status.php` would solve a ton of problems and clear up a huge source of headaches for me. I just hesitated to suggest it because of how ugly a solution it is. Hopefully other apps don't also use `/status.php` and need it authenticated for some reason.

kentonv:
@mnutt Can you be more specific about what the response to /status.php should be? I think we are already returning a 401 Unauthorized code with www-authenticate header in the case that the client is on the basic auth whitelist.

kentonv:
@mnutt Hmm in retrospect I think what you're saying is that `status.php` has to actually return correct information about the server even when not authenticated. That makes it difficult to hard-code the response because it's potentially different for different apps. We definitely cannot boot up the app if we haven't received a valid key -- some day that key will be required to decrypt the grain's storage.

I suppose we could develop an API in which the app can specify a single small resource to be served statically, without authentication. That would be a separate PR, of course.

kentonv:
@zarvox I've added a commit to memoize the results of newApiToken() for a short time, which solves the problem with Davros.

I also added a commit which changes the hash computation for hostnames so that it is infeasible to determine which ApiToken corresponds to some hostname without knowing the secret token.

I'm hoping to get this change in for the build that I'll push this weekend, either Saturday or Sunday night.

paulproteus:
My personal remark on @mnutt + Davros: Since the box for the URL has horizontal scroll, if you do the username & password in that URL, I would be happy if the box were scrolled all the way to the _right_, so I can see my happy domain name and feel comfortable wanting to copy and paste that thing.

shell/shared/grain.js
```diff
@@ -1562,31 +1570,55 @@ if (Meteor.isClient) {
...
+          const renderedTemplate = template.replace(/\$API_TOKEN/g, tokenId)
+                                           .replace(/\$API_HOST/g, host);
+          sessionStorage.setItem(key, JSON.stringify({
+              token: tokenId,
+              renderedTemplate: renderedTemplate,
+              expires: Date.now() + selfDestructDuration,
+            })
+          );
+          sessionStorage.setItem(key + "-host", host);
+
```
zarvox: Any reason you didn't just add a `host` key to the object above?

zarvox:
Just the one curiosity, which needn't block merge. :+1:

shell/shared/grain.js
kentonv: Because I'm very silly and did not read carefully! Fixed.

PR #1548, opened 2016-02-19
Emphasize how to manually re-run setup.sh

paulproteus:
This hopefully minimizes confusion when people modify `setup.sh` but don't realize they need to re-run `provision` or `destroy && up`.


paulproteus:
I'm going to self-merge this since it's 100% docs.

Feedback welcome, especially from @zarvox.

See also https://github.com/sandstorm-io/vagrant-spk/pull/133

PR #1551, opened 2016-02-19
Persist the user's shrink-navbar choice into localStorage

### LINKED ISSUE #1525: Make hiding of vertical left-column panel (with grains list) sticky
It's awesome that the left-column panel with the grains list can be hidden with the left-arrow icon.

It would be awesomer if, once hidden, it stayed hidden across all apps until unhidden OR there was some other way to make it default to hidden.

I bring this up because I have a friend who uses his monitor in a 'portrait' configuration, and the left bar is causing him to feel grief about the amount of valuable horizontal screen real estate it's wasting.

I know the usecase is uncommon, but it seems like a fix wouldn't be too hard!

Thanks!

jparyani:
Fixes #1525

### DIFF
--- a/shell/lib/db-deprecated.js
+++ b/shell/lib/db-deprecated.js
@@ -74,6 +74,7 @@ if (Meteor.isServer) {
 if (Meteor.isClient) {
   Session.setDefault("shrink-navbar", false);
   globalGrains = new ReactiveVar([]);
+  Session.set("shrink-navbar", window.localStorage.getItem("shrink-navbar") === "true");
   globalTopbar = new SandstormTopbar(globalDb,
     {
       get() {
@@ -91,6 +92,7 @@ if (Meteor.isClient) {
       },
 
       set(value) {
+        window.localStorage.setItem("shrink-navbar", value);
         Session.set("shrink-navbar", value);
       },
     });


shell/lib/db-deprecated.js
```diff
@@ -74,6 +74,7 @@ if (Meteor.isServer) {
 if (Meteor.isClient) {
   Session.setDefault("shrink-navbar", false);
   globalGrains = new ReactiveVar([]);
+  Session.set("shrink-navbar", window.localStorage.getItem("shrink-navbar") === "true");
```
paulproteus: Will this be reactive? It seems like it wouldn't be, which would mean that different browser tabs would get out of sync on this, which is a little sad, but maybe that's life.

jparyani: No, and I guess it's theoretically possible to listen to `storage` events across your other tabs, but I'd say that's outside the scope of this quick fix. Another solution is to persist it into mongo as a user preference and get reactivity by means of that, but the discussion on #1525 seemed to decide that localStorage was best.

paulproteus: +1 to this getting merged

One day I might make it reactive to other tabs, but I think it's not a
showstopper.

PR #1553, opened 2016-02-20
Remove the SandstormBackend.runningGrains table.

### LINKED ISSUE #1550: "Grain no longer in use; shutting down" can sometimes happen even when the grain is in use
@mrdomino has been observing that sometimes Groove Basin grains spontaneously shut down even when they are in use.

To reproduce the problem:
 1. Open a private Groove Basin grain.
 2. Refresh your browser to get a new session.
 3. Start playing some music.
 4. Wait approximately 4.5 minutes.
 5. Notice that the stream has stopped and the grain log has a message reading "Grain no longer in use; shutting down"

The problem is that `gcSessions()` calls `supervisor.close()` on supervisor reference that is shared between multiple proxy objects, so later calls to `keepAlive()` fail with "capability closed".

Proposed fix in an upcoming pull request...

dwrensha:
Fixes #1550.

The `runningGrains` table allows us to avoid calling `Backend.startGrain()` in many cases. Because we currently call `waitPromise()` when we call `Backend.startGrain()`, that means that we avoid a round trip in those cases. However, the table is problematic in that can lead to separate `Proxy` objects holding the same reference to a supervisor. When one of those proxies closes, it calls `supervisor.close()`, invalidating the reference for everybody. Moreover, as far as I can tell, we don't actually need to wait for `startGrain()` to finish, and there's no harm in calling `startGrain()` on grains that are already started.

Narrower fixes to #1550 are possible, but this looks to me like a good opportunity to get rid of some cruft.

### DIFF
--- a/shell/packages/sandstorm-backend/sandstorm-backend.js
+++ b/shell/packages/sandstorm-backend/sandstorm-backend.js
@@ -51,7 +51,6 @@ SANDSTORM_ALTHOME = Meteor.settings && Meteor.settings.home;
 
 SandstormBackend = function (db, backendCap) {
   this._db = db;
-  this.runningGrains = {};
   this._backendCap = backendCap;
 };
 
@@ -66,7 +65,6 @@ SandstormBackend.prototype.deleteUser = function (userId) {
 SandstormBackend.prototype.shutdownGrain = function (grainId, ownerId, keepSessions) {
   if (!keepSessions) {
     Sessions.remove({ grainId: grainId });
-    delete this.runningGrains[grainId];
   }
 
   const grain = this._backendCap.getGrain(ownerId, grainId).supervisor;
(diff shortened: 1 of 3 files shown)


kentonv:
We still cache proxies by API token, right? So this won't cause startGrain() to be called for every request to the same API token? This is somewhat important because startGrain() (as well as getGrain()) on Oasis currently will incur two round trips to the storage service, the first of which obtains the user's entire grain list (which we'll fix eventually...), even if the grain is already running.

dwrensha:
Yeah, along with `proxiesByHostId`, we still have `proxiesByApiToken`.

When a grain is brought up to service a API requests, it will be shut down after 1.5 minutes unless there are also web sessions open and sending keepalives. We only call `startGrain()` when we initially start the grain and whenever we notice that it needs to be restarted, so at most once every 90 seconds.

Reading this code has just reminded me that we don't ever actually clean up the `proxiesByApiToken` table, except for when a token changes its permissions. That's a bug for a separate pull request, though.

dwrensha:
I'm still thinking about this.

With this change, basically every call to `openSession()` will incur two calls to `Backend.startGrain()`. Maybe we can do better...

kentonv:
Two calls to `startGrain()` in service of the same operation definitely sounds wrong...

dwrensha:
To be clear, the first call to `startGrain()` would be due to [this optimization](https://github.com/sandstorm-io/sandstorm/blob/929f2c7ecc7b6c38fe9089f8fa45dd6a52cca609/shell/packages/sandstorm-backend/sandstorm-backend.js#L294) and the second one would happen when the first request comes in. But yeah, if we're worried about the cost of `startGrain()`, this seems bad.

I'm open to further discussion, but I think I now prefer the immediate fix in #1557.

kentonv:
Hmm. Well, it's not a _huge_ concern, and the memory leak that is `runningGrains` is a concern in itself.

Could we modify `openSession` (in `proxy.js`) to pre-populate the `proxiesByHostId` table?

Otherwise I would probably say that Blackrock should temporarily cache results from startGrain() to fix the problem.

dwrensha:
> Could we modify openSession (in proxy.js) to pre-populate the proxiesByHostId table?

Done.

dwrensha:
The integration tests passed locally for me.

dwrensha:
Garply, retest this please

kentonv:
It looks like Garply is out of disk space.

PR #1554, opened 2016-02-20
Fix styling on sidebar shrink button and add :hover css

### LINKED ISSUE #1266: Sidebar arrow highlight on hover
This is a solution to @zarvox's suggestion that users may mistakenly click on the close grain button. 

- whole row highlights on hover (indicates action)
- row height increases to match grain height (harder to miss)
- highlight is somewhat subtle
newsidebar-hover

### LINKED ISSUE #1513: Sidebar collapse button is too small
Currently, the sidebar collapse button is exceedingly discreet and easy to miss (I did for a couple of weeks).

I went to the [original issue](https://github.com/sandstorm-io/sandstorm/issues/872) where the button was suggested, and it looks great on the screenshots with the custom "<", better than the arrow actually. But on the actual front-end it's just the character `<`, which doesn't look half as good as the larger glyph.

Replacing the plaintext `<` character with an SVG icon that looks closer to the original design intent would solve this.

jparyani:
Fixes #1513
Fixes #1266

### DIFF
--- a/shell/client/_geometry.scss
+++ b/shell/client/_geometry.scss
@@ -13,6 +13,6 @@ $topbar-height-desktop: 32px;
 $topbar-height-mobile: 48px;
 $navbar-width-desktop: 200px;
 $navbar-width-desktop-shrunk: 48px;
-$navbar-shrink-button-height: 24px;
+$navbar-shrink-button-height: 32px;
 
 $topbar-popup-padding: 10px;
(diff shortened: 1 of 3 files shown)


dwrensha:
The new glyphs are apparently not present on my system (OSX v10.11.3). The icon now looks like this in Firefox:

firefox

and like this in Chrome:

chrome

kentonv:
On my system, the new glyph looks like an up ad down arrow side-by-side.

Reverting.

PR #1556, opened 2016-02-20
Switch the default Vagrantfile to Debian

### LINKED ISSUE #263: make fails on ubuntu 14.04 in Vagrant

### LINKED ISSUE #544: "Vagrant up" should install linux deps automatically inside the new VM
I expect that when I first vagrant ssh into the new VM, I should already be able to have git, build-essential, and g++ installed in there. `./run-dev.sh` expects the latter 2 dependencies.

paulproteus:
Rationale:

- Close #263 for now because the build works on Debian, at least in theory

- Fewer things for us to support, since we already use this base box in vagrant-spk.

- Document how to install dependencies, allowing me to close #544 (by also
  overriding SERVER_USER)

- While we're at it, link to some docs.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -5,11 +5,8 @@
 VAGRANTFILE_API_VERSION = "2"
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
-  # We base ourselves off the trusty (Ubuntu 14.04) base box.
-  config.vm.box = "trusty64"
-
-  # The url from which to fetch that base box.
-  config.vm.box_url = "https://cloud-images.ubuntu.com/vagrant/trusty/current/trusty-server-cloudimg-amd64-vagrant-disk1.box"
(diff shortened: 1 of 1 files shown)


paulproteus:
Self-merging as this is 100% Vagrantfile.

PR #1557, opened 2016-02-20
Don't call supervisor.close() in resetConnection().

dwrensha:
Conservative fix for #1550.

Note that we currently do not have a satisfactory story for how we clean up `SandstormBackend.runningGrains`. One approach might be to add a `Handle` argument to `Backend.startGrain()`, so that the frontend would get notified when a supervisor died and could then delete the appropriate entry in `runningGrains`.


jparyani:
I think the problem is also due to the `close` call in Proxy.close (see https://github.com/sandstorm-io/sandstorm/blob/c2d98a153d89c35725901db85e668b81ad87b567/shell/server/proxy.js#L1093).

dwrensha:
@jparyani: Cood catch.

kentonv:
I think this is obsolete since I merged #1553?

PR #1572, opened 2016-02-22
Send notification emails to admins when an app lands in the app index.

kentonv:
This is actually already live.

### DIFF
--- a/shell/server/pre-meteor.js
+++ b/shell/server/pre-meteor.js
@@ -84,8 +84,9 @@ function wwwHandlerForGrain(grainId) {
     };
 
     if (path === "apps/index.json" ||
-        path === "apps/index-experimental.json" ||
-        path.match(/apps\/[a-z0-9]{52}[.]json/)) {
+        path.match(/apps\/[a-z0-9]{52}[.]json/) ||
+        path === "experimental/index.json" ||
+        path.match(/experimental\/[a-z0-9]{52}[.]json/)) {
       // TODO(cleanup): Extra special terrible hack: The app index needs to serve these JSON files
       //   cross-origin. We could almost just make all web sites allow cross-origin since generally
       //   web publishing is meant to publish public content. There is one case where this is
--- a/src/sandstorm/app-index/app-index.c++
+++ b/src/sandstorm/app-index/app-index.c++
@@ -321,6 +321,7 @@ public:
       } else if (path.startsWith("reject/")) {
         indexer.reject(path.slice(strlen("reject/")),
             kj::str(params.getContent().getContent().asChars()));
+        indexer.updateIndex();  // remove from experimental
         context.getResults().initNoContent();
       } else if (path.startsWith("unapprove/")) {
         indexer.unapprove(path.slice(strlen("unapprove/")));
--- a/src/sandstorm/app-index/app-index.c++
+++ b/src/sandstorm/app-index/app-index.c++
@@ -67,12 +67,13 @@ template <typename Context>
 void handleError(Context& context, kj::Exception&& e) {
   KJ_LOG(ERROR, e);
   auto error = context.getResults().initServerError();
-  error.setDescriptionHtml(kj::str("Error: ", htmlEscape(e.getDescription())));
+  error.setDescriptionHtml(kj::str("Error: ", htmlEscape(e.getDescription()), "\n"));
 }
 
 class SubmissionSession final: public ApiSession::Server {
 public:
-  explicit SubmissionSession(Indexer& indexer): indexer(indexer) {}
+  explicit SubmissionSession(Indexer& indexer, HackSessionContext::Client session)
+      : indexer(indexer), session(kj::mv(session)) {}
 
   kj::Promise<void> post(PostContext context) override {
     return kj::evalNow([&]() -> kj::Promise<void> {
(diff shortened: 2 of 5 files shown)


paulproteus:
Never mind! I see you already cut my work out for me!

paulproteus:
I see I have my work cut out for me!

src/sandstorm/app-index/app-index.c++
```diff
@@ -163,7 +164,38 @@ public:
...
+        if (!req.isSetState() ||
+            !response.getRoot<SubmissionStatus>().isPending()) {
+          return kj::READY_NOW;
+        }
+
+        // Send notification email to app index reviewers.
+        auto appTitle = indexer.getAppTitle(packageId);
+        auto notificationText = kj::str(
+            "An app package is pending review in the app index.\n\n"
+            "https://alpha.sandstorm.io/grain/NujwEZfut8oZoSdcrFzy9p/\n\n"
```
paulproteus: I guess it's nice if this grain link is configurable, or something.

kentonv: Yeah but not worth the time, meh.

kentonv:
Added additional changes to improve the experimental index.

src/sandstorm/app-index/app-index.c++
```diff
@@ -67,12 +67,13 @@ template <typename Context>
 void handleError(Context& context, kj::Exception&& e) {
```
paulproteus: Random note: I'm not sure I've seen `&&` in a function signature in C++ before.

shell/server/pre-meteor.js
```diff
@@ -84,8 +84,9 @@ function wwwHandlerForGrain(grainId) {
     };
 
     if (path === "apps/index.json" ||
-        path === "apps/index-experimental.json" ||
-        path.match(/apps\/[a-z0-9]{52}[.]json/)) {
+        path.match(/apps\/[a-z0-9]{52}[.]json/) ||
+        path === "experimental/index.json" ||
+        path.match(/experimental\/[a-z0-9]{52}[.]json/)) {
```
paulproteus: +1, naturally, this is sad, but if today is still "not the day" to have CORS indication in app manifest, or even as a server-wide admin setting, then OK.

paulproteus:
```
         // TODO(security): Verify request's webkey hash. Need to know our own webkey, somehow.
 
         auto packageId = packageIdString(req.getPackageId());
```

Curious why we would need that. Not a show-stopper, just curious.

paulproteus:
Also, presumably, it's safe to tell an app the _hash_ of the current webkey being used to access it, in theory, as a possible future Sandstorm API addition? If the app needs that for some reason, though I'm not sure I understand the reason here yet.

paulproteus:
```

         auto outBytes = kj::heapArray<byte>(
             status.totalSize().wordCount * sizeof(capnp::word) + 128);
```

128 here - I guess that's so that we have enough space in the buffer for the NUL byte that comes before the data? Why not ` + 1` in that case? Or I guess it's so that we can round up to the nearest something?

paulproteus:
Also w/r/t curl being hard to detect errors in, I use this in the installer:

```
curl -w '%(http_code)'
```

which works if you can save curl's HTTP response output to a file, or to /dev/null.

src/sandstorm/app-index/app-index.c++
paulproteus: I _seem_ to be getting duplicate emails in some cases, but I don't consider that a showstopper, but thought I'd mention it.

src/sandstorm/app-index/app-index.c++
```diff
@@ -163,7 +164,38 @@ public:
...
+        // Send notification email to app index reviewers.
+        auto appTitle = indexer.getAppTitle(packageId);
+        auto notificationText = kj::str(
+            "An app package is pending review in the app index.\n\n"
+            "https://alpha.sandstorm.io/grain/NujwEZfut8oZoSdcrFzy9p/\n\n"
+            "title: ", appTitle, "\n"
+            "packageId: ", packageIdString(req.getPackageId()), "\n"
+            "requested state: ", req.getSetState().getNewState(), "\n");
+
+        return session.getPublicIdRequest().send()
```
paulproteus: I guess this email sending has no error handling, but that is probably OK.

src/sandstorm/app-index/app-index.c++
```diff
@@ -288,6 +321,7 @@ public:
       } else if (path.startsWith("reject/")) {
         indexer.reject(path.slice(strlen("reject/")),
             kj::str(params.getContent().getContent().asChars()));
```
paulproteus: No CSRF protection except the Sandstorm built-in one. You're bold! I like that! This is not an objection.

paulproteus:
I don't really understand this loop yet:

```
         auto appId = appIdString(info.getAppId());
         auto iter = appMap.find(appId);
         if (iter == appMap.end() || info.getVersion() >= iter->second.version) {
```

Not a showstopper, just a remark.

src/sandstorm/app-index/indexer.c++
```diff
@@ -389,7 +394,7 @@ void Indexer::updateIndexInternal(kj::StringPtr outputFilename, kj::StringPtr ou
     kj::FdOutputStream(file.getFd()).write(text.begin(), text.size());
     file.finalize(kj::str(outputDir, "/", appEntry.first, ".json"));
 
-    if (approvedApps) {
+    if (!experimental) {
```
paulproteus: Not sure I understand this block, but not necessarily a showstopper if it actually works.

paulproteus: In particular, it's a bit weird to use `/var/apps` within this. Is there an analogous thing for experimental apps?

paulproteus: I guess the idea is that we unconditionally write the app SPK file symlink to `/var/apps`, but since we only need to do it once, we skip that work `if (experimental)`?

paulproteus:
OK, I'm +0.5 on this. I want to ask some questions about this code, but those aren't showstoppers, so I'll click merge.

paulproteus:
CORRECTION: I'm not supposed to click merge, @kentonv is, so @kentonv you may do so now.

src/sandstorm/app-index/app-index.c++
kentonv: I haven't seen any dupes but I think there was a case of multiple versions of an app being submitted in rapid succession.

src/sandstorm/app-index/app-index.c++
kentonv: It's one of the greatest new features in C++11! Abstractly speaking, it means that ownership is being transferred without a copy.

src/sandstorm/app-index/indexer.c++
kentonv: This symlink is written to keep track of the latest published package version so that we can reject duplicates. It doesn't apply to experimental because if we reject an experimental app for trivial reasons and it is resubmitted, it is allowed to have the same version number.

src/sandstorm/app-index/app-index.c++
kentonv: If it fails an exception is thrown. The submitter ends up seeing an error. You could argue that we shouldn't display an error to the submitter given that their app was successfully submitted and only the notification email failed but I do not see a better place for the error report to go that we're actually likely to see, and I'd hate for the submitter to falsely believe that their submission will be looked at when in fact we haven't received any notification and so won't be looking.

kentonv:
I guess you reviewed more than just what was changing...

> Curious why we would need that. Not a show-stopper, just curious.

Prevent a submission made to one app index from being replayed against another app index. Not a huge threat.

> 128 here - I guess that's so that we have enough space in the buffer for the NUL byte that comes before the data?

That and the Cap'n Proto header and root pointer. Technically I think +17 would have worked but I didn't feel like being precise. Note that an exception will be thrown if somehow the buffer overflows.

PR #1574, opened 2016-02-23
Fix a bunch of icons on IE, mostly in the sharing UI.

### LINKED ISSUE #1560: Minor Style: Down arrow displays wrong in Edge
I don't remember what particular quirk of formatting causes this, but it was fixed on most of the other SVGs in Sandstorm. This probably affects IE11 too, but I didn't check.

image

kentonv:
I went ahead and switched over to font icons since we need to do this anyway and they display consistently in IE (amazingly).

This also fixes some minor mobile styling regressions caused by the box-sizing change.

@jparyani I found that the webfont-icon mixin wasn't very reusable. Initially I tried to add parameters to customize but then I realized I was customizing everything so there was no point. We may want to remove this.

Fixes #1560

### DIFF
--- a/shell/client/_app-details.scss
+++ b/shell/client/_app-details.scss
@@ -149,13 +149,14 @@
             @extend %unstyled-button;
             padding-left: 8px;
             padding-right: 24px;
-            background-repeat: no-repeat;
-            background-position: right center;
-            background-size: 24px 24px;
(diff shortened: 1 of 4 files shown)


jparyani:
LGTM

I've run into the same issue with webfont-icon as well. It probably should be renamed to topbar-icon.

PR #1588, opened 2016-02-25
docs: Add sniproxy guide (share TLS port w/other server).

pjamar:
Added sniproxy tutorial. Graph in the doc is pointing to a personal server. Asheesh should move that to a safer/more resilient option. Thanks!

### DIFF
--- a/docs/administering/sniproxy.md
+++ b/docs/administering/sniproxy.md
@@ -1,0 +1,145 @@
+# sniproxy for sharing an TLS port 443 with Sandstorm
+
+**The purpose of this tutorial is to set up sniproxy so it’s possible to use regular https:// URLs for your main web server, like nginx or Apache, as well as Sandstorm, which has its own HTTPS handling. To make that work, they will need to share port 443.**
+
+## Introduction
+
+The purpose of this tutorial is to set up *sniproxy* so it's possible to use *Sandstorm* verified TLS encryption while coexisting with another web server that also uses TLS and make share a port for it (443).
+
+The main reason is to allow other users to connect with your *Sandstorm* instance in the standard HTTPS port (443) and keep using that port also for any other web apps.
+
+This action is going to change the BASE_URL of Sandstorm, presumably, so you'll trigger the erasing of OAuth configuration. To fix that, people will need to follow the "How do I log in, if there's a problem with logging in via the web?" section of https://docs.sandstorm.io/en/latest/administering/faq/
+
+I assume the server is running Debian or one of its derivatives (e.g. Ubuntu). Note that there will be some downtime in this process so you might want to do it when your server is not very busy.
+
+## Install sniproxy
+
+If you're lucky the package *sniproxy* might be present on your GNU/Linux distro; otherwise, you'll have to install it yourself. In my case (Ubuntu 10.04 I had to do it manually). Follow the instructions on https://github.com/dlundquist/sniproxy for your operating system and install it.
+
+In my case, on Ubuntu 10.04, I had to do it manually:
+```bash
+# Install required packages
+sudo apt-get install autotools-dev cdbs debhelper dh-autoreconf dpkg-dev gettext libev-dev libpcre3-dev libudns-dev pkg-config fakeroot git
+
+# Clone sniproxy repo from Github
+git clone https://github.com/dlundquist/sniproxy.git
+
+# Compile and create the package
+cd sniproxy
+./autogen.sh && dpkg-buildpackage
+
+# Install the package
+sudo dpkg -i ../sniproxy_*_*.deb
+```
+
+## Setting it up
+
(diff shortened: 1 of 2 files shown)


garply:

paulproteus:
Thanks for filing this, @pjamar ! I'll review/merge in the next day or so.

paulproteus:
When viewing this at https://github.com/pjamar/sandstorm/blob/docs_add_sniproxy/docs/administering/sniproxy.md I notice some issues; here are suggested changes.

- "an TLS port" => "a TLS port"

- Other docs don't bold the first paragraph, so I propose not bolding the first paragraph.

- In this text: "section of https://docs.sandstorm.io/en/latest/administering/faq/" you provide the full URL as a clickable link. Other sections of the docs use a markdown link; I suggest using that here also. For one example on how to do that, look at https://github.com/pjamar/sandstorm/blob/docs_add_sniproxy/docs/administering/faq.md , specifically the "Why does Sandstorm require a wildcard host?" section, for the internal link.

- s/make share a port for it (443)/share port 443 with it/

- s/The main reason is to allow other users/The main reason is to allow users/

- s/people will need to follow the/you will need to follow the/

- s/If you're lucky the package/If you're lucky, the package/

- s/as long as you're being consistent in replacing my choices in the/as long as you're consistent in replacing the ports in the/

- s/sniproxy configurations/sniproxy configuration files/

- s/We'll disable HTTP proxy on/We'll disable the HTTP proxy feature in/

- Before this code block: `sudo nano /etc/sniproxy.conf`, consider writing a one-sentence paragraph such as: You will need to edit the sniproxy configuration file by running something like:

- s/We'll have to ensure sniproxy starts when rebooted/We'll have to ensure sniproxy starts when the system is rebooted/

- s/For that we'll need to ensure is enabled/For that we'll need to ensure it is enabled/

- In this text: "start on boot up::" you only need one `:` character. Remove one of them.

- s/TWe/We/

- s/so that it listens on port that/so that it listens on the port that/

- s/Since that is the standard/Since that is the standard port for HTTPS/

- s/We'll assume you have a example.sandcats.io as your Sandstorm address in the configuration example/We'll assume you have a example.sandcats.io as your Sandstorm address in the configuration example; you can change the example to use a different domain if you need to/

- s/I'm using nginx and I could change all the configuration files using sed to replace 443 for 7443. What I did is:/If you are using nginx, you can change all the configuration files using sed to replace 443 with 7443. To do that, run this command:/

- s/Supposing you use nginx this could be:/For example, if you are using nginx, you can run:/

- s/Also you most probably have/You probably have/

- s/Run the following and follow screen instructions:/Run the following and follow the instructions that it prints out:/

- The other docs don't italicize Sandstorm, so IMHO you shouldn't here either.

- The other docs don't say "I..." so IMHO you shouldn't here either. s/I assume the server is running Debian or one of its derivatives (e.g. Ubuntu)./This document assumes the server is running Debian or one of its derivatives (e.g. Ubuntu). If your server is not, you might need to adjust some steps./

- The other docs generally use backticks to indicate code-ish terms, so in this sentence: "If you're lucky the package sniproxy" you should (IMHO) write `sniproxy` (backticks) rather than italicize it.

- In this sentence: "Follow the instructions on https://github.com/dlundquist/sniproxy" two suggestions. (1) "the instructions at" not "the instructions on". (2) Explain what the link is, such as: "Follow the instructions at [the sniproxy homepage](https://github.com/dlundquist/sniproxy)"

- In the "Install sniproxy" sentence you write this sentence twice: "In my case (Ubuntu 10.04 I had to do it manually)." Remove the first one IMHO.

- s/and make him decide where/and allow it to decide where/

- s/which switched from listening on port 443 to port 7443/which we will switch from listening on port 443 to port 7443/ - just a small change, inserting the word "we", makes that sentence valid. You should say "will switch" not "switched' because in this doc, the reader has not yet made the change.

General feedback:

- The diagram is great! Move the diagram higher up in the document. I suggest putting the diagram right after the intro paragraph. You could alternatively blend the "Setting it up" section into the "Introduction" section.

- The document as a whole is great. I know I gave you a lot of small suggested fixes above, but it really really is.

- Please add a navigation link to `mkdocs.yml` so that people can find this doc via the navigation on the left. I propose you use this title: "Sharing the HTTPS port"

- Please update `ssl.md` so that people can find this document that way, too.

If you want me to do any of the above, I can do it if you say so. If you're willing to, that's even easier for me! Let me know what you think.

paulproteus:
Also: if you can export the image as an SVG file somehow, that's great, and if you can't, that's life!

It's nice if you can:

- Change the `alt text` to something like `Diagram of a request flowing through sniproxy`

- Use a white background, even if you can't export it as SVG

paulproteus:
@ocdtrekkie This is assigned to me, but I'm going to deassign it because I'm not sure I'm going to be addressing it any time soon.

ocdtrekkie:
We need to do some more work on ssl.md this week in light of #3299 but I think we should probably revisit this, maybe making the commits to fix this with Asheesh's review notes, and get it merged in, since this is still something people might want to do.

The currently linked Sandstorm server where this guide existed is a dead link, but the content lives on in this PR.

ocdtrekkie:
I am going to close this in favor of #3317 which includes this commit. Thanks @pjamar for the contributon!

PR #1590, opened 2016-02-25
Destroy all existing GrainViews on login.

### LINKED ISSUE #1581: "Request Access" doesn't go away after login if I already have access
1. Log out.
2. Visit a grain URL you own. The page instructs you to log in to request access.
3. Log in.

Expected behavior: Since I own the grain, it loads normally.

Actual behavior: I am told I do not have access and offered a "request access" button. Reloading the page, or closing the grain tab and re-opening, allows me to get access.

dwrensha:
Fixes #1581 and adds an integration test for it.

Destroying all grainviews on login used to be problematic in the case of linking an email identity, but after the identity-provider refactor of last month, I don't think it's an issue.

### DIFF
--- a/tests/tests/grain.js
+++ b/tests/tests/grain.js
@@ -199,7 +199,37 @@ module.exports["Sign in at grain URL"] = function (browser) {
             .waitForElementPresent("#publish", medium_wait)
             .assert.containsText("#publish", "Publish")
             .frame(null)
-            .end();
+            // Now try it with a /shared/ path.
+            .click('.topbar .share > .show-popup')
+            .waitForElementVisible('#shareable-link-tab-header', short_wait)
+            .click('#shareable-link-tab-header')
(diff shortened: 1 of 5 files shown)


tests/tests/grain.js
```diff
@@ -177,6 +177,63 @@ module.exports["Test grain not found"] = function (browser) {
     .end()
 }
 
+module.exports["Sign in at grain URL"] = function (browser) {
+  browser
```
paulproteus: <3 for this test

dwrensha:
This strategy of using `Accounts.onLogin()` apparently does not work well when a server admin does `sandstorm restart`. I'm currently investigating alternatives...

kentonv:
What about when the server has't restarted but the WebSocket gets randomly disconnected? I bet that also triggers a new login when reconnecting...

PR #1595, opened 2016-02-26
Admin feature keys

zarvox:


kentonv:
Should we feature-flag this so that people don't see it until it's actually ready? Or do you think this is ready to ship with LDAP?

zarvox:
This patchset now works correctly with admin tokens.

I believe this piece is ready to ship with LDAP, and does not need to be feature flagged.

PR #1597, opened 2016-02-26
Subscribe to "requestingAccess" immediately on displaying the Request Access page.

dwrensha:
This fixes a bug in #1590 where users would see blank pages after a `sandstorm restart`.

This also fixes a few places in #1585 where I forgot to project out the `email` field.

### DIFF
--- a/shell/server/hack-session.js
+++ b/shell/server/hack-session.js
@@ -242,7 +242,7 @@ HackSessionContextImpl = class HackSessionContextImpl extends SessionContextImpl
     const identity = globalDb.getIdentity(grain.identityId);
 
     const emails = SandstormDb.getVerifiedEmails(identity);
-    const email = _.findWhere(emails, { primary: true }) || identity.unverifiedEmail;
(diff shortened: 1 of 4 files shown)


dwrensha:
Oops, I had forgotten about the case where `getVerifiedEmails()` returns an empty list. Fixed now.

I also removed the logic that favored the Sandstorm account primary email over the identity provider primary email. (Consider the case where I have a private email that Github knows about and that I want Sandstorm to use to notify me about service announcements, but that I don't want to give out to my Sandstorm contacts.)

kentonv:
> Consider the case where I have a private email that Github knows about and that I want Sandstorm to use to notify me about service announcements, but that I don't want to give out to my Sandstorm contacts.

That seems far-fetched to me. And in any case, we don't actually reveal email addresses to contacts, do we? I prefer the old behavior...

dwrensha:
Meh, okay. Old behavior restored.

> we don't actually reveal email addresses to contacts, do we?

We do not reveal email addresses to all contacts. But certain interactions, like requesting access to a grain, do reveal email addresses.

PR #1601, opened 2016-02-27
Remove / from BASE_URL for a CORS header

### LINKED ISSUE #1599: If BASE_URL ends in /, then /admin/settings shows a spurious WILDCARD_HOST warning
Steps to reproduce:

- Visit (e.g.) https://rose.sandcats.io/admin/settings and see that there is no big red warning text

- Edit `sandstorm.conf` so `BASE_URL=...` now _gains_ a `/` at the end of it

- Restart Sandstorm

Expected behavior:

- There is no big red warning text

Actual behavior:

- Big red warning text about WILDCARD_HOST

red-warning

paulproteus:
Close #1599

### DIFF
--- a/shell/server/pre-meteor.js
+++ b/shell/server/pre-meteor.js
@@ -222,6 +222,7 @@ function serveSelfTest(req, res) {
       if (rootUrlAsOrigin.slice(-1) === "/") {
         rootUrlAsOrigin = rootUrlAsOrigin.slice(0, -1);
       }
+
       res.writeHead(200, {
         "Content-Type": "text/plain",
         "Content-Length": content.length,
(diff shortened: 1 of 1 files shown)


paulproteus:
As a side note:

If I had followed @kentonv's advice to make this allowed to be accessed from all origins (`*`) then this bug would never have happened. But I do like the fail-closed nature of the way I wrote it! Hmm!

paulproteus:
Testing done:

- Set rose.sandcats.io's BASE_URL to end in a `/`

- Reproduce the bug

- Install this code (built via `make fast`)

- Reload `/admin/settings`

- Notice the problem has gone away

- Adjust `WILDCARD_HOST` to be `*.zrose.sandcats.io`

- Reload `/admin/settings`

- Notice that I (correctly) see a big red warning.

- Fix the problem.

- Reload `/admin/settings`

- See the red warning go away.

paulproteus:
<3 style checker

kentonv:
I kind of suspect there are a lot more bugs that will happen if the root URL ends in a '/', so we should consider canonicalizing it at the source (in run-bundle).

paulproteus:
+1 to that. I can see about doing that in the fullness of time. I filed a
bug.

PR #1602, opened 2016-02-27
Support adding unauthenticated options headers and resources to API hosts

kentonv:
This supports both:
* Setting the `DAV` header to return for unauthenticated OPTIONS requests.
* Defining small static resources which can be `GET`ed unauthenticated, e.g. `/status.php`.

Example app code: https://github.com/kentonv/apihost-testapp/blob/master/client/index.html

Also in this PR, I extended web-session.capnp to pass through DAV extension identifiers in OPTIONS requests.

I would like to get this out in this weekend's build so that we can fix up Radicale next week. If @zarvox can't review in time, I'll probably just merge as-is.

@mnutt @synchrone

### DIFF
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -391,6 +391,8 @@ ApiTokens = new Mongo.Collection("apiTokens");
 //              be considered revoked, and all live refs and sturdy refs obtained transitively
 //              through it must also become revoked. Each item is the JSON serialization of the
 //              `MembraneRequirement` structure defined in `supervisor.capnp`.
+//   hasApiHost: If true, there is an entry in ApiHosts for this token, which will need to be
+//              cleaned up when the token is.
 //
 // It is important to note that a token's owner and provider are independent from each other. To
 // illustrate, here is an approximate definition of ApiToken in pseudo Cap'n Proto schema language:
@@ -971,6 +973,20 @@ _.extend(SandstormDb.prototype, {
 
 if (Meteor.isServer) {
   SandstormDb.prototype.getWildcardOrigin = getWildcardOrigin;
+
+  const Crypto = Npm.require("crypto");
+  SandstormDb.prototype.removeApiTokens = function (query) {
+    // Remove all API tokens matching the query, making sure to clean up ApiHosts as well.
+
+    this.collections.apiTokens.find(query).forEach(function (token) {
+      // Clean up ApiHosts for webkey tokens.
+      if (token.hasApiHost) {
+        var hash2 = Crypto.createHash("sha256").update(token._id).digest("base64");
+        ApiHosts.remove({hash2: hash2});
+      }
+    });
+    this.collections.apiTokens.remove(query);
+  }
 }
 
 // =======================================================================================
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -430,6 +430,29 @@ ApiTokens = new Mongo.Collection("apiTokens");
 //   ...
 // }
 
+ApiHosts = new Mongo.Collection("apiHosts");
+// Allows defining some limited static behavior for an API host when accessed unauthenticated. This
+// mainly exists to allow backwards-compatibility with client applications that expect to be able
+// to probe an API host without authentication to determine capabilities such as DAV protocols
+// supported, before authenticating to perform real requests. An app can specify these properties
+// when creating an offerTemplate.
+//
+// Each contains:
+//   _id:          apiHostIdHashForToken() of the corresponding API token.
+//   hash2:        hash(hash(token)), aka hash(ApiToken._id). Used to allow ApiHosts to be cleaned
+//                 up when ApiTokens are deleted.
+//   options:      Specifies how to respond to unauthenticated OPTIONS requests on this host.
+//                 This is an object containing fields:
+//     dav:        List of strings specifying DAV header `compliance-class`es, e.g. "1" or
+//                 "calendar-access". https://tools.ietf.org/html/rfc4918#section-10.1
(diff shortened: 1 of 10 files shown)


tests/tests/apihost.js
```diff
@@ -0,0 +1,94 @@
...
+var medium_wait = utils.medium_wait;
+var very_long_wait = utils.very_long_wait;
+
+module.exports = {
+};
+
+module.exports['Install and launch test app'] = function (browser) {
+  browser
+    .init()
+    .installApp("https://alpha-qkhxczi7kki1x49pfakw.sandstorm.io/apihost-testapp.spk", "e148b304d9642e8c98ef8ea2df2f72ca", "w304h9n5rjx1pzfa8e4guheue5mq3dkwv63aajy1rscupw6e38mh")
```
dwrensha: Could you add a comment linking to the source code of this app?

shell/packages/sandstorm-db/db.js
```diff
@@ -938,6 +974,21 @@ _.extend(SandstormDb.prototype, {
 
 if (Meteor.isServer) {
   SandstormDb.prototype.getWildcardOrigin = getWildcardOrigin;
+
+  const Crypto = Npm.require("crypto");
+  SandstormDb.prototype.removeApiTokens = function (query) {
```
zarvox: I wish there were a way to neuter `ApiTokens.remove()`/`db.collections.apiTokens.remove()`, so it'd be impossible for new code to get this wrong.  Perhaps we can monkeypatch:

```js
ApiTokens.removeUnchecked = ApiTokens.remove;
ApiTokens.remove = function () { throw new Error("Use db.removeApiTokens") }
```

and then use `ApiTokens.removeUnchecked()` in this function.  Thoughts?

shell/packages/sandstorm-db/db.js
```diff
@@ -430,6 +432,29 @@ ApiTokens = new Mongo.Collection("apiTokens");
 //   ...
 // }
 
+ApiHosts = new Mongo.Collection("apiHosts");
```
zarvox: This is a separate collection so that clients (which basically never care about this collection) don't have to load the data associated with the associated ApiToken?

shell/packages/sandstorm-db/db.js
```diff
@@ -430,6 +432,29 @@ ApiTokens = new Mongo.Collection("apiTokens");
...
+// Allows defining some limited static behavior for an API host when accessed unauthenticated. This
+// mainly exists to allow backwards-compatibility with client applications that expect to be able
+// to probe an API host without authentication to determine capabilities such as DAV protocols
+// supported, before authenticating to perform real requests. An app can specify these properties
+// when creating an offerTemplate.
+//
+// Each contains:
+//   _id:          apiHostIdHashForToken() of the corresponding API token.
+//   hash2:        hash(hash(token)), aka hash(ApiToken._id). Used to allow ApiHosts to be cleaned
```
zarvox: Is there any benefit to this being `hash(hash(token))` instead of just `hash(token)`?  It could just as easily be the `ApiToken._id`, which would make it possible to look up other information like "does the grain that created this still exist?" when servicing a request.

It's kinda weird that the static resources will be stored and served forever, even if the grain that caused them to be created has since been deleted.

shell/packages/sandstorm-db/db.js
```diff
@@ -391,6 +391,8 @@ ApiTokens = new Mongo.Collection("apiTokens");
 //              be considered revoked, and all live refs and sturdy refs obtained transitively
 //              through it must also become revoked. Each item is the JSON serialization of the
 //              `MembraneRequirement` structure defined in `supervisor.capnp`.
+//   hasApiHost: If true, there is an entry in ApiHosts for this token, which will need to be
```
zarvox: It seemed a little odd at first glance to have this information duplicated between the two tables, but I guess it saves a roundtrip in the token cleanup codepath, so that seems probably worthwhile.

shell/packages/sandstorm-db/db.js
kentonv: > Is there any benefit to this being hash(hash(token)) instead of just hash(token)?

I have this weird desire to make sure that it is impossible to map from an API hostname back to the ApiToken. It's possible I'm being silly.

> It's kinda weird that the static resources will be stored and served forever, even if the grain that caused them to be created has since been deleted.

I think deleting the grain also deletes all ApiTokens and therefore all ApiHosts?

zarvox: Oh, I see.  It removes all of the ApiTokens associated with that grain that are either owned by that user or by `{ webkey: null }`.

I thought all ApiTokens were left lying around because if you share a grain, the ApiToken owned by the person you shared with definitely is still around after you delete the grain, but I see that's not the case in general. :)

shell/packages/sandstorm-db/db.js
kentonv: Ehh.... In theory I like that, but in practice I worry that it will have some unintended consequences, and I'm not excited about thoroughly testing it right now. Also note that some of our old migration use ApiTokens.remove, and I didn't update them -- I _could_ update them but then I'd need to test them...

tests/tests/apihost.js
kentonv: Done -- and extended the test to cover the DAV header, which I had forgotten about.

kentonv:
Going to merge now and build a release but happy to address further comments in later changes.

mnutt:
Great! I'm going to test it out on some WebDAV clients tonight.

kentonv:
@mnutt @synchrone At the last moment before release I decided that the new parameter should really be called `unauthenticated` rather than `static`. `static` is a keyword in es6 (although context-dependent, so it can be used as a name too) and `unauthenticated` more clearly warns the developer that what they write here had better not be secret.

PR #1613, opened 2016-02-28
Group membership check in readConfig (fixes #711)

### LINKED ISSUE #711: Running sandstorm admin-token as non-privileged user crashes
Should probably offer a user-readable error "Run as root user or under sudo"

```
$ sandstorm admin-token
*** Uncaught exception ***
sandstorm/util.c++:47: failed: open(name.cStr(), flags, mode): Permission denied; name = ../var/sandstorm/adminToken
stack: 0x541e08 0x541d89 0x541d7a 0x5143a1 0x515025
```

mrdomino:
Testing: tried running `sandstorm admin-token` as root with a SERVER_USER, and as non-root with a matching uid, and with a different uid that was a supplementary member of SERVER_USER's group (passes the check in all cases), and with a different uid that was not a member (fails the check).

### DIFF
--- a/src/sandstorm/run-bundle.c++
+++ b/src/sandstorm/run-bundle.c++
@@ -234,8 +234,10 @@ kj::Array<uint> parsePorts(kj::Maybe<uint> httpsPort, kj::StringPtr portList) {
   return kj::mv(result);
 }
 
-kj::Maybe<UserIds> getUserIds(kj::StringPtr name) {
+kj::Maybe<UserIds> getUserIds(kj::Maybe<kj::StringPtr> name) {
   // We can't use getpwnam() in a statically-linked binary, so we shell out to id(1).  lol.
+  // This returns the UserIds for the passed name, or the user this process is running as if nullptr
+  // is passed.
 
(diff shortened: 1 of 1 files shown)


garply:

mrdomino:
Updated: I now default config.uids to the current user's uid/gid. Testing, in addition to the above: verified that `sandstorm admin-token` with no SERVER_USER passes for a non-root user.

kentonv:
Why do all this work instead of calling `access()`?

mrdomino:
I was concerned about possible cases like calling sandstorm without a var/sandstorm at some point -- maybe a hypothetical sandstorm install or something. But just calling access definitely feels better to me than running through gids. So if we can rely on ../var/sandstorm existing and having correct permissions, then I'll happily resubmit with that approach.

mrdomino:
I went and did it anyway: https://github.com/sandstorm-io/sandstorm/pull/1617. Yeah, way simpler.

kentonv:
Closing this since I like the access() approach better. Thanks!

PR #1617, opened 2016-02-29
Fail early if we can't access var/sandstorm

### LINKED ISSUE #711: Running sandstorm admin-token as non-privileged user crashes

mrdomino:
Fixes #711 via the `access()` method.

### DIFF
--- a/src/sandstorm/run-bundle.c++
+++ b/src/sandstorm/run-bundle.c++
@@ -821,6 +821,7 @@ public:
 
   kj::MainBuilder::Validity adminToken() {
     changeToInstallDir();
+    checkAccess();
 
     // Get 20 random bytes for token.
     kj::byte bytes[20];
@@ -924,6 +925,19 @@ private:
   void changeToInstallDir() {
     KJ_SYSCALL(chdir(getInstallDir().cStr()));
     changedDir = true;
+  }
+
+  void checkAccess() {
+    KJ_ASSERT(changedDir);
+    if (access("../var/sandstorm", W_OK) == -1) {
+      if (errno == EACCES) {
+        KJ_FAIL_REQUIRE(
+            "Sandstorm was not run with appropriate privileges; rerun as root or the user for "
+            "which it was installed.");
+      } else {
+        KJ_FAIL_SYSCALL("access", errno);
+      }
+    }
   }
 
   void checkOwnedByRoot(kj::StringPtr path, kj::StringPtr title) {


garply:

src/sandstorm/run-bundle.c++
```diff
@@ -647,13 +647,13 @@ public:
     // Make sure the pidfile is close-on-exec.
     KJ_SYSCALL(fcntl(pidfile, F_SETFD, FD_CLOEXEC));
 
-    changeToInstallDir();
+    changeToInstallDirAndCheckAccess();
```
kentonv: I'm somewhat terrified that there could be some significant number of installations out there that randomly don't have /var/sandstorm writable by the group, but have the appropriate subdirectories writable such that things do actually still work. If we add a check in `continue_` then all those servers will die when they update. Similarly if the check is in `start` then they'll all die on their next reboot. I really don't want people to start fearing that an auto-update might randomly kill their server, so I'd rather not perform the check in these cases.

src/sandstorm/run-bundle.c++
```diff
@@ -742,7 +742,7 @@ public:
   }
 
   kj::MainBuilder::Validity status() {
-    changeToInstallDir();
+    changeToInstallDirAndCheckAccess();
```
kentonv: Arguably `status` shouldn't require this check. It can often complete just fine without any write permissions.

src/sandstorm/run-bundle.c++
```diff
@@ -921,8 +921,15 @@ private:
     return kj::heapString(exeName.slice(0, KJ_ASSERT_NONNULL(exeName.findLast('/'))));
   }
 
-  void changeToInstallDir() {
+  void changeToInstallDirAndCheckAccess() {
     KJ_SYSCALL(chdir(getInstallDir().cStr()));
+    if (access("../var/sandstorm", W_OK) == -1) {
+      if (errno == EACCES) {
+        KJ_FAIL_REQUIRE("Sandstorm must be run as root or a member of SERVER_USER's group.");
```
kentonv: This error is not quite correct:
- If Sandstorm was installed as root, then it must be started as root, regardless of SERVER_USER.
- If Sandstorm was installed as non-root, then it must always be run as the same user who installed it, and I think SERVER_USER is not actually specified in the config.

kentonv: However, `dev` should never run as root, even if Sandstorm was installed as root; it should be run by someone inside the server user's group.

kentonv:
Hmm, this seems to be more complicated than it seemed. Possibly we should focus solely on doing the check in the `admin-token` command for now, since that was what issue #711 complained about?

mrdomino:
Sounds reasonable. I'll resubmit next time I get a chance to work on it.

src/sandstorm/run-bundle.c++
mrdomino: Okay, so special message for the dev command? Is there an easily-detectable way of finding out whether Sandstorm was installed as root? Or should the message just include the full conditional?

mrdomino:
Resubmitted. Sorry it took so long -- unusually busy these days.

kentonv:
Thanks! To answer this question:

> Okay, so special message for the dev command? Is there an easily-detectable way of finding out whether Sandstorm was installed as root? Or should the message just include the full conditional?

There's a `runningAsRoot` member which tells you if the command is running as root. If `readConfig()` has been called, it will have verified if `runningAsRoot` is true then `SERVER_USER` is configured. Arguably if `runningAsRoot` is not true then `readConfig()` should fail if `SERVER_USER` *is* configured, since the presence of `SERVER_USER` implies that Sandstorm should start as root (and should setuid() itself later).

Anyway, not important.

PR #1618, opened 2016-02-29
Ignore child tokens in 'requestingAccess' publish.

dwrensha:
Currently, if a user receives access to a grain through a sharing link, that link gets revoked, and then the user tries to visit the grain from their grain list, the client goes into an infinite retry loop.

This fixes the problem.


kentonv:
Is there a similar problem if the user has received a direct share from Bob, but Bob's own access has been revoked?

dwrensha:
No, because the `requestingAccess` subscription only picks up non-revoked direct shares from the grain owner.

kentonv:
Ah, I see.

PR #1620, opened 2016-03-01
Update permissions computation to account for permissionsHeld membrane requirements.

dwrensha:


shell/packages/sandstorm-db/db.js
```diff
@@ -130,6 +130,8 @@ Meteor.users.ensureIndexOnServer("loginIdentities.id", { unique: 1, sparse: 1 })
 Meteor.users.ensureIndexOnServer("nonloginIdentities.id", { sparse: 1 });
 Meteor.users.ensureIndexOnServer("services.google.id", { unique: 1, sparse: 1 });
 Meteor.users.ensureIndexOnServer("services.github.id", { unique: 1, sparse: 1 });
+Meteor.users.ensureIndexOnServer("apiTokens.grainId", { sparse: 1 });
```
kentonv: Did you mean:

    ApiTokens.ensureIndexOnServer("grainId", { sparse: 1 });

?

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -14,333 +14,773 @@
...
-    var old = this.array[ii];
-    this.array[ii] = this.array[ii] && !other.array[ii];
-    if (old != this.array[ii]) {
-      changed = true;
+  isSubsetOf(other) {
+    check(other, PermissionSet);
+    for (let ii = 0; ii < this.array.length; ++ii) {
+      const mine = this.array[ii];
+      const yours = other.array[ii] || false;
+      if (mine != yours) {
```
kentonv: Shouldn't e.g. [false, true] be considered a "subset" of [true, true]? It seems like the logic here is actually verifying that this PermissionSet is a prefix of `other` rather than a "subset". It would help to add some doc comments to the method explaining the expected behavior.

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -14,333 +14,773 @@
...
+  // Methods for mutating a PermissionSet by combining it with another PermissionSet.
+  // These return a boolean indicating whether the operation had any effect.
+
+  add(other) {
+    check(other, PermissionSet);
+    let changed = false;
+    for (let ii = 0; ii < other.array.length; ++ii) {
+      const old = this.array[ii];
+      this.array[ii] = this.array[ii] || other.array[ii];
+      if (old != this.array[ii]) {
```
kentonv: If `old` is `undefined` but the new value is `false`, then the PermissionSet hasn't really changed, but this will treat it as having changed.

Also, please use `!==` rather than `!=`.

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -14,333 +14,773 @@
...
+    return changed;
+  }
+
+  remove(other) {
+    check(other, PermissionSet);
+    let changed = false;
+    for (let ii = 0; ii < other.array.length && ii < this.array.length; ++ii) {
+      const old = this.array[ii];
+      this.array[ii] = this.array[ii] && !other.array[ii];
+      if (old != this.array[ii]) {
```
kentonv: Same here.

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -14,333 +14,773 @@
...
-    this.array[ii] = this.array[ii] && other.array[ii];
-    if (old != this.array[ii]) {
-      changed = true;
+  intersect(other) {
+    check(other, PermissionSet);
+    let changed = false;
+    for (let ii = 0; ii < this.array.length; ++ii) {
+      const old = this.array[ii];
+      this.array[ii] = this.array[ii] && other.array[ii];
+      if (old != this.array[ii]) {
```
kentonv: And here.

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -14,333 +14,773 @@
...
-          break;
+  hash() {
+    // It is often useful to form sets of clauses. This method returns a hash that is suitable to
+    // be used as a key for such a set.
+    //
+    // TODO(soon): Investigate other approaches. We do not need this hash to be cryptographically
+    //   secure; we just need it to avoid accidental collisions. Maybe we should instead represent
+    //   sets of clauses as ordered maps, or maybe as proper hash maps with bucketing?
+    const hasher = Crypto.createHash("sha256");
+    for (let grainId in this.identityPermissions) {
```
kentonv: Do you need to make sure these are in a canonical order? Note that iterating over keys in a JS object iterates in the order in which the keys were added, not lexical order.

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -14,333 +14,773 @@
...
-      for (var ii = 0; ii < viewInfo.roles.length; ++ii) {
-        var roleDef = viewInfo.roles[ii];
-        if (roleDef.default) {
-          result = new PermissionSet(roleDef.permissions);
-          break;
+  hash() {
+    // It is often useful to form sets of clauses. This method returns a hash that is suitable to
+    // be used as a key for such a set.
+    //
+    // TODO(soon): Investigate other approaches. We do not need this hash to be cryptographically
```
kentonv: Is this a performance concern? `TODO(perf)`, then.

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -14,333 +14,773 @@
...
-    grainId = vertex.token.grainId;
-  } else if (vertex.grain) {
-    grainId = vertex.grain._id;
-  }
-  var grain = db.getGrain(grainId);
-  if (!grain) {
-    return {grainDoesNotExist: true};
-  }
+class MembranedPermissionSet {
+  // A PermissionSet that is contingent upon some membrane requirements. The membrane is repesented
```
kentonv: It took me a bit to understand what you mean by "the membrane is represented as a Clause". What you actually mean is "the requirements are represented as a Clause", right?

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -14,333 +14,773 @@
...
+    this.permissions = permissions;
+    this.membrane = membrane;
+    this.tokensUsed = {}; // Token IDs for the tokens that these permissions depend upon.
   }
 
-  var result = {edgesByRecipient: {}};
+  static fromToken(token, viewInfo) {
+    const permissions = PermissionSet.fromRoleAssignment(token.roleAssignment, viewInfo);
+    const result = new MembranedPermissionSet(permissions, new Clause());
+    result.membrane.addMembraneRequirements(token.requirements);
```
kentonv: Maybe add a comment that the parent token -- which is a kind of requirement in itself -- is handled differently, to reassure people that it's not being forgotten about here.

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -14,333 +14,773 @@
...
+    const clause = new Clause();
+    destinationPermissions[clause.hash()] = new MembranedPermissionSet(permissionSet, clause);
   }
 
-  var openerIdentityId = edges.terminalEdge.sharer;
-  var edgesByRecipient = edges.edgesByRecipient;
-  var owner = "OwningAccount";
+  const permissionsMap = {};
+  // Map<vertexId, Map<membraneHash, MembranedPermissionSet>>.
+  // May from vertex ID to permissions that we've already shown flow from that vertex to our
```
kentonv: s/May/Map/

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -14,333 +14,773 @@
...
+    let incomingEdges = [];
 
-        var newPermissions = roleAssignmentPermissions({allAccess: null}, viewInfo);
-        inEdge.roleAssignments.forEach(function(roleAssignment) {
-          newPermissions.intersect(roleAssignmentPermissions(roleAssignment, viewInfo));
-        });
-        newPermissions.intersect(permissionsMap[recipient]);
+    function tokenToEdge(token) {
+      let sharerId = "i:" + token.identityId;
+      if (token.parentToken) {
```
kentonv: I feel like a ternary operator would be more readable, and would also avoid computing a string concatenation that just gets thrown away in the case that there's a parent token.

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -14,333 +14,773 @@
...
+
+    if (vertexId.slice(0, 2) == "o:") {
+      // Owner. We don't need to do anything.
+      incomingEdges = [];
+    } else if (vertexId.slice(0, 2) == "t:") {
+      const token = context.tokensById[vertexId.slice(2)];
+      if (token) {
+        incomingEdges = [tokenToEdge(token)];
+      }
+    } else if (ownerIdentityIds.indexOf(vertexId.slice(2)) >= 0) {
```
kentonv: I think you should explicitly check here that the vertexId starts with "i:", in case a different kind of vertexId is added later.

kentonv: (the remaining else/if cascade should be gated by this check)

dwrensha:
@kentonv I've addressed your comments. I've also added a new unit test and I've added some crude performance monitoring to `SandstormPermissions.grainPermissions()`.

kentonv:
I have to admit I actually haven't wrapped my head around anything after the class definitions -- that is, from `backpropagateVertex()` on.

I think what I'm going to do is pull this down and go through everything while writing my own comments that help me understand. Then I'll send the commit back to you to verify.

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -14,333 +14,790 @@
...
+          newChosen[hashedClause] = chosen[hashedClause];
         }
-        if (!stackedUsers[sharer]) {
-          userStack.push(sharer);
-          stackedUsers[sharer] = true;
+      }
+
+      if (fullCover(newChosen)) {
+        minimal = false;
+        // TODO(perf): keep track of values we've already explored.
```
kentonv: So, if I'm reading this right (which I might not be!), without implementing this TODO, I think this function is O(n!). *With* implementing this TODO, I think the function is still O(2^n) (a set of size n has 2^n subsets). Either way, this is dangerous. I think this function will need to be rethought.

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -14,333 +14,790 @@
...
+
+  for (const hashedClause in goalClauses) {
+    const goalClause = goalClauses[hashedClause];
+    const tokensUsed = goalClause.tokensUsed;
+    if (goalClause.clause.isEmpty()) {
+      // We have an empty goal. Trivially true.
+      return { yes: { tokensUsed: tokensUsed } };
+    }
+
+    clauses[hashedClause] = { clause: goalClause.clause, tokensUsed: tokensUsed };
```
kentonv: It seems like the object literal here is just reconstructing `goalClause`?

Does `clauses` end up being a shallow-ish copy of `goalClauses`? Maybe make that more explicit?

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -14,333 +14,790 @@
...
+  }
+
+  const grainsFullyQueried = {};
+
+  while (clauseStack.length > 0) {
+    const hashedClause = clauseStack.pop();
+    const clause = new Clause();
+    clause.conjoin(clauses[hashedClause].clause);
+    const tokensUsed = clauses[hashedClause].tokensUsed;
+    const goal = clause.popFirstGoal();
```
kentonv: It seems like some memoization of goals is needed here, otherwise you could easily end up checking the same goal over and over as it appears in different clauses.

kentonv: Also, I think goals could be cyclic. Of course, a goal which is implied by itself should be treated as false (think of it as: all goals are false until shown to be true).

I think this situation gets complicated, though:

A <= B or C
B <= A
C <= true

A naive algorithm trying to prove A might start by following the branch to B, which leads back to A, which loops.

A better but still naive algorithm might follow the branch to B, see that it leads back to A, and so decide that B is false because cycles are false, and memoize this. It may then try to prove C, find that it's true, and therefore decide that A is true. However, B has now been incorrectly memoized as false, when in the end it should have been true.

A better algorithm might have three results: "true", "false", and "cycle", where "cycle" is a false result but is never memoized. This might work? Will not memoizing mean repeating work too often?

I'm sure this problem has been deeply studies, of course, so maybe we should look up the solution.

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -14,333 +14,790 @@
...
+      context.addGrain(db, grainId);
+    }
+
+    grainsFullyQueried[grainId] = true;
+
+    const viewInfo = context.grains[grainId].cachedViewInfo || {};
+    const result = backpropagateVertex(context, goal.vertex, goal.permissions, viewInfo);
+    const newGoals = normalize(result, goal.permissions);
+
+    // For each new clause, conjoin it with the remaining goals in `clause`.
```
kentonv: I think you'll want to split out all the goals to memoize separately, so not try to merge them back into new clauses.

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -14,333 +14,790 @@
...
+      };
+
+      // Optimization: we don't care about permissions that we've already proven the opener has.
+      if (permissionsMap["o:Owner"]) {
+        for (const hashedClause in permissionsMap["o:Owner"]) {
+          const membranedPermissionSet = permissionsMap["o:Owner"][hashedClause];
+          if (hashedClause in sequenced) {
+            const smps = sequenced[hashedClause];
+            smps.permissions.remove(membranedPermissionSet.permissions);
+            if (smps.permissions.array.length == 0) {
```
kentonv: Shouldn't you be checking that all elements in the array are false, rather than that the array is empty? (Maybe add an `isEmpty()` method to `PermissionSet`?)

kentonv:
I added a lot of comments in commit 54b1495 (branch `permissions-commentary`); please merge it and check that my comments are correct.

The commit also adds a bunch of TODO(perf)s. I think some of them may be important enough to implement now.

shell/packages/sandstorm-permissions/permissions.js
dwrensha: I think what it probably does make sense to memoize is the mapping from `goal` to `newGoals` which is basically "apply `backpropagateVertex()` and then `normalize()`". Is that the memoization that you are suggesting here?

dwrensha: `proveClauses()` avoids loops by keeping track of what clauses it has already seen.

I've added a `TODO(perf)` about the memoization and I've added a comment that I hope addresses your concern about looping.

shell/packages/sandstorm-permissions/permissions.js
kentonv: I'm saying that feels weird to me. Instead given `A ∧ B ∧ C` I would start by trying to prove A, B, and C separately. Expanding one item and then forming a new clause seems a lot harder to reason about.

What if there are cycles?

`A ∧ B ∧ C` -> `A ∧ W ∧ B ∧ C` -> `A ∧ W ∧ W ∧ B ∧ C` -- here the two W's would be de-duped of course, and then we're back at `A ∧ W ∧ B ∧ C`, which I guess would already be in the map and therefore we don't have a loop? But I'm having trouble being confident that there aren't other cases that become problematic, since this seems like an unusual way of thinking.

dwrensha:
I'm still thinking about the perf issues, but in the meantime, I've pushed a commit that adds some abstractions that should make it much easier to read and write permissions tests.

shell/packages/sandstorm-permissions/permissions.js
dwrensha: Note that the immediately straightforward way to compute `normalize()` takes can take p^n time, where p is number of permissions and n is the number of ConditionalPermissionSets in the input.

I've updated the function so that it groups together ConditionalPermissionSets that correspond to the same PermissionSet. This way, when we do the "minimal cover" computation, all of the subsets we deal with are unique. I think this makes the computation (modulo the TODO(perf) about remembering states we've visited already) come to O(2^m) where m is the number of unique PermissionSets in the input permissions flow.

Because it can still blow up, I've added an iteration counter that bails after a while. I suspect that one would need to willingly construct pathological examples in order to get this to happen.

dwrensha:
I've gone through all the comments here and in the permissions-commentary commit, and I've addressed everything that I considered urgent.

shell/packages/sandstorm-permissions/permissions.js
dwrensha: > mmediately straightforward way to compute normalize() takes can take p^n time

I think I might mean n^p, rather.

Issue #1383, opened 2016-01-15
Multi-identity: Make it clearer when logging in as a non-login user how to upgrade them to a login user


kentonv:
We had a user who tried to transfer his Google identity from an empty account into his main account to use for login. He almost succeeded, but he missed the checkbox to allow login from that identity. When he subsequently tried to log in using Google, he was confused by the interstitial, and thought that the old empty account was still around and causing the problem.

Perhaps when a user adds a new identity to their account, and that identity is eligible to be a login identity (because it is not attached to any other account, or was attached to an empty account which was then deleted), we should automatically make it a login identity.

Additionally, we should improve the interstitial in the case where the identity is attached to only one account to make it easier to convert it into a login identity. Perhaps in this case, we could have a single button that initiates login as the account's main identity and a checkbox next to it for "Allow me to log in with [identity the user tried first] in the future"?

seagreen:
I'm that user!

> Perhaps when a user adds a new identity to their account, and that identity is eligible to be a login identity (because it is not attached to any other account, or was attached to an empty account which was then deleted), we should automatically make it a login identity.

Though I'm not familiar with all the ins and outs of your login process, this would have definitely taken care of me.

paulproteus:
+1 to "login identity by default"

dwrensha:
It would need to be "login identity by default and then fall back to non-login identity if some other account already has the identity linked".

PR #1623, opened 2016-03-02
by default, give newly-linked identities the ability to login

dwrensha:
Partially addresses #1383.


kentonv:
You mentioned in the meeting that you thought there might be a bug in this if the identity is attached to an empty account. Have you had a chance to check on that?

dwrensha:
The thing I was worried about was not a problem. I've added a commit that adds a test for the empty account case.

PR #1627, opened 2016-03-03
Split most of shared/ into client/ and server/

zarvox:
Code-splitting at build time is better than branchy code at runtime for a
couple of reasons:
    
1. More amenable to static analysis.  If you're defining a function in an
   `if (Meteor.isServer)` block, eslint and friends aren't going to know that the
   function will definitely be available in a second `Meteor.isServer` block.
2. More performant.  *All* code in `shared` has to be loaded by the browser,
   even if a majority of the file happens to live in an if (Meteor.isServer) block.
   Splitting code only includes the appropriate subset in the browser bundle,
   which means less data on the wire and faster load times.  (While it seems like it might have been low-hanging fruit, Meteor's build tooling is not intelligent enough to remove dead branches based on the value of `Meteor.isClient` and `Meteor.isServer`.)
3. More compatible with ES2015 modules.  ES2015 does not allow conditional
   imports or exports, which means that any exports or imports that are conditional on
   Meteor.isClient or Meteor.isServer are not compatible with refactoring into
   standard ES2015 modules.

In practice, the **vast** majority of our code is intended to run on either the client or the server.  Template helpers and event handlers are strictly client-side.  Method implementations and publishes are, for all intents and purposes, strictly server-side.  Most of our methods do not do anything useful when run in simulation, and most interactions are in fact more usable when tristated (default, submitting, completed) than predictively committed.  Indeed, most of our `Meteor.call` calls await the result before taking further action.

Additionally, the more we split code into smaller files and modules, the less likely we are to get merge conflicts when collaborating.

This changeset reduces the size of the client JS bundle by 20k, which, while only about 2.5% of the whole bundle size, is still about 10% of the code previously shipped not in `packages/`, as measured by https://gist.github.com/zarvox/eb20b813be3e18096ee2 .

At a request from @jparyani I've moved all of our SCSS into `shell/client/styles`, so the numerous files there don't make it so hard to browser the directory tree, and moved our vendored-wholesale-into-shell/client libs into `shell/client/vendor`.

In the interest of not disrupting others' workflows with changes that move around large swaths of code, @kentonv please merge other open pull requests before this one, in favor of forcing this one to rebase.


kentonv:
Is there some way we can avoid having multiple files with the same name? When I jump around in my editor, I do it by name, not by path.

paulproteus:
FWIW, in e.g. Emacs, "uniquify" somewhat addresses this problem:
https://stackoverflow.com/questions/5607794/emacs-switching-between-buffers-with-the-same-name-but-in-different-directories

I don't know if your editor has something similar.

zarvox:
Hmmm.  I guess we can repeat the path information into the filename - `client/admin.js` becomes `client/admin-client.js` and so on.

FWIW if you have two tabs open with files with the same basename, Atom will also show the containing folder in the tab for clarity.  I can't speak to Qt Creator's behavior.

zarvox:
Renamed files per requests and rebased.

zarvox:
Rebased again.

PR #1628, opened 2016-03-04
Create command-line tool for generating feature keys.

kentonv:
Also:
- Removed isFeatureKeyValid Meteor setting. Instead, feature keys with "isForTesting = true" will work only when the server is running in testing mode (another Meteor setting). I will generate such testing certificates for everyone to use.
- Fixed key reassembly in feature-key.js to be little-endian.

### DIFF
--- a/shell/server/feature-key.js
+++ b/shell/server/feature-key.js
@@ -34,7 +34,6 @@ const signingKey = bits0
 const isTesting = Meteor.settings && Meteor.settings.public &&
                   Meteor.settings.public.isTesting;
 
-// Export for use in Meteor method in admin.js
 function verifyFeatureKeySignature(buf) {
   // buf is a Buffer containing an feature key with attached signature.
   // This function returns the signed data if the signature is valid,
--- a/shell/server/feature-key.js
+++ b/shell/server/feature-key.js
@@ -17,20 +17,25 @@
 const Capnp = Npm.require("capnp");
 const FeatureKey = Capnp.importSystem("sandstorm/feature-key.capnp").FeatureKey;
 
-// These byte-packing tricks that are so convenient in non-memory-safe languages are a
-// bit of a pain in memory-safe languages.  Argh.  At least I have bignums on the server.
+// We don't currently have a way of reading the raw bytes of a capnp struct in Javascript the
+// way we do in C++. Awkwardly, what we end up with is four 64-bit ints that have been stringified
+// since JS can't actually deal with 64-bit ints. We have to laboriously reconstruct the underlying
+// bytes using Bignum. Don't forget that everything is little-endian, as DJBeesus intended.
 const bits0 = Bignum(FeatureKey.signingKey.key0);
 const bits1 = Bignum(FeatureKey.signingKey.key1);
 const bits2 = Bignum(FeatureKey.signingKey.key2);
 const bits3 = Bignum(FeatureKey.signingKey.key3);
-const signingKey = bits0.shiftLeft(64 * 3)
-    .add(bits1.shiftLeft(64 * 2))
-    .add(bits2.shiftLeft(64))
-    .add(bits3)
-    .toBuffer();
+const signingKey = bits0
+    .add(bits1.shiftLeft(64 * 1))
+    .add(bits2.shiftLeft(64 * 2))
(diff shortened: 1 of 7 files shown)


shell/server/feature-key.js
```diff
@@ -17,20 +17,25 @@
...
+    .add(bits2.shiftLeft(64 * 2))
+    .add(bits3.shiftLeft(64 * 3))
+    .toBuffer({ endian: "little", size: 32 });
+
+const isTesting = Meteor.settings && Meteor.settings.public &&
+                  Meteor.settings.public.isTesting;
 
 // Export for use in Meteor method in admin.js
-verifyFeatureKeySignature = function (buf) {
+function verifyFeatureKeySignature(buf) {
```
zarvox: If this function doesn't need to be exported, let's declare it `const` here so it's clear it's not used elsewhere.

shell/server/feature-key.js
```diff
@@ -17,20 +17,25 @@
...
+const signingKey = bits0
+    .add(bits1.shiftLeft(64 * 1))
+    .add(bits2.shiftLeft(64 * 2))
+    .add(bits3.shiftLeft(64 * 3))
+    .toBuffer({ endian: "little", size: 32 });
+
+const isTesting = Meteor.settings && Meteor.settings.public &&
+                  Meteor.settings.public.isTesting;
 
 // Export for use in Meteor method in admin.js
```
zarvox: You can remove this comment, looks like.

shell/server/feature-key.js
```diff
@@ -17,20 +17,25 @@
...
-const signingKey = bits0.shiftLeft(64 * 3)
-    .add(bits1.shiftLeft(64 * 2))
-    .add(bits2.shiftLeft(64))
-    .add(bits3)
-    .toBuffer();
+const signingKey = bits0
+    .add(bits1.shiftLeft(64 * 1))
+    .add(bits2.shiftLeft(64 * 2))
+    .add(bits3.shiftLeft(64 * 3))
+    .toBuffer({ endian: "little", size: 32 });
```
zarvox: Ahhh, this worked for me because when I generated a fake `key{0,1,2,3}` for testing, I manually serialized my test signing key to capnp as big endian, and then deserialized here as big endian, so it came out the same.

src/sandstorm/feature-key-tool.c++
```diff
@@ -0,0 +1,196 @@
...
+  byte key[crypto_sign_ed25519_SECRETKEYBYTES];
+  byte publicKey[crypto_sign_ed25519_PUBLICKEYBYTES];
+
+  const byte* structToBytes(capnp::AnyStruct::Reader reader, size_t size) {
+    auto data = reader.getDataSection();
+    KJ_REQUIRE(data.size() == size);
+    return data.begin();
+  }
+
+  byte* structToBytes(capnp::AnyStruct::Builder builder, size_t size) {
```
zarvox: Maybe I'm paranoid of interior mutability, but this sounds like any old serialization function.  A name like `getInternalBuffer()` might make it clearer that this lets you modify the bytes out from under the Builder.  Especially when you have an identically named function that returns a `const byte*` above.

But maybe constness or lack thereof makes that clear enough.

zarvox:
Only nitpicks; merge at will.

src/sandstorm/feature-key-tool.c++
kentonv: Agreed. Renamed to `getUnderlyingBytes` (both the reader/const and builder/non-const versions). Also renamed in update-tool.c++ which has similar helpers.

shell/server/feature-key.js
kentonv: Indeed!

shell/server/feature-key.js
kentonv: Actually binding a name with `function` has similar semantics to `var`, so won't be visible in other files. Arguably, perhaps, we should move to a style of declaring all functions as `const f = (...) =>` for the same reasons we've banned `var`, but currently it looks like we still declare many things using `function` so I'll leave it for a later mass cleanup.

zarvox:
:+1: , thanks!

PR #1631, opened 2016-03-04
Make the smtp server also listen on BIND_IP

### LINKED ISSUE #1605: Mail server listening on all interfaces (0.0.0.0)
It looks like the sandstorm mail server always binds to 0.0.0.0, even when BIND_IP='127.0.0.1'

I would have expected the mail server to either:
1) bind to the same IP as the web interface (configured via BIND_IP)
or
2) bind to a separately configured IP and default to 127.0.0.1

jparyani:
Fixes #1605


zarvox:
:shipit:

PR #1635, opened 2016-03-07
(WIP) UI Styleguide

neynah:
@zarvox Available for viewing but not ready to be merged yet.

### DIFF
--- a/shell/client/_widgets.scss
+++ b/shell/client/_widgets.scss
@@ -1,3 +1,194 @@
+.button-primary1 {
+  @extend %button-small;
+  @extend %button-primary;
+}
+
+.button-primary2 {
+  @extend %button-medium;
+  @extend %button-primary;
+}
+
+.button-primary3 {
+  @extend %button-large;
+  @extend %button-primary;
+}
+
+.button-primary4 {
+  @extend %button-xlarge;
+  @extend %button-primary;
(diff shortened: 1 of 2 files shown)


shell/client/_widgets.scss
```diff
@@ -1,3 +1,194 @@
...
+  &:disabled {
+    @extend %button-disabled;
+  }
+}
+
+%button-margin {
+  margin: 10px 5px;
+}
+
+%button-disabled {
```
zarvox: Two comments on this:

* Disabled buttons should probably be lower contrast, to make it clear that they are disabled, especially when you're looking at a form with a single disabled button in isolation. `#777777` for the foreground color and `lighten(#afafaf, 10%)` for the border looked decent to me; other ideas totally welcome.
* Disabled buttons should also override `&:hover` (with the same values) so that disabled buttons don't show their "other" colors on mouseover.

neynah: Thanks for the feedback. Your suggested colours seem fine to me.

kentonv:
Do we intend to merge this eventually? If not, let's close the pull request. If we do want to merge it, it looks like it needs to be updated for merge conflicts.

neynah:
Probably not any time soon. Ill close it for now

PR #1637, opened 2016-03-08
Fix position of popup close button on mobile

zarvox:
The popup frame should be `position: relative`, so that children (like the close
button) will be placed relative to that frame.

### DIFF
--- a/shell/client/_topbar.scss
+++ b/shell/client/_topbar.scss
@@ -784,7 +784,7 @@ body>.popup {
     }
 
     @media #{$mobile} {
-      position: static;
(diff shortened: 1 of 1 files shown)


zarvox:
Weird jenkins failure: https://build.sandstorm.io/job/Sandstorm%20PRs%20-%20build/128/

Roundcube incoming mail test appears to have crashed something.

zarvox:
Garply, retest this please

kentonv:
This change, which explicitly applied to mobile, broke mobile. :(

PR #1639, opened 2016-03-11
Tweak Sandstorm for Work

kentonv:
- Group the organizational management (auto-upgrade) settings together rather than putting them next to each provider, because this is an important concept and needs some explanation. New users won't understand the concept of "auto-upgrade" and could easily miss the checkbox otherwise.
- Move LDAP settings into their own section because cramming them between basic login providers created an unnatural split.
- Make auto-upgrade actually work. It din't work previously because the feature key was not subscribed except in admin routes.
- Add a button to delete the feature key, mainly for testing purposes.

screenshot from 2016-03-10 17-04-18

### DIFF
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -986,6 +986,10 @@ limitations under the License.
 <template name="adminSettings">
   {{setDocumentTitle}}
   <form id="admin-settings-form">
+    <h4>Basic login providers</h4>
+
+    <p>Choose which services users may use to identify themselves. Anyone can log in, but only users you {{#linkTo route="adminInvites" data=getToken}}invite{{/linkTo}} will be able to install apps or create files.</p>
+
     <p>
       <label><input type="checkbox" class="oauth-checkbox" name="googleLogin"
                     value="value" checked={{googleSetting.value}} data-servicename="google">
@@ -993,16 +997,6 @@ limitations under the License.
         (<a class="configure-oauth" data-servicename="google" href="">configure</a>)
         (<a class="reset-login-tokens" data-servicename="google" href="">log out all users</a>)
       </label><br>
-
-      {{#if isFeatureKeyValid}}
-      <label>
-        Google Organization Domain:
-        <input type="text" name="organizationGoogle" value="{{organizationGoogle}}">
-      </label>
-      {{/if}}
-
-      <span class="details">Users within this Google Apps for Work domain will automatically be upgraded to full accounts upon logging in. Leave blank to disable.</span>
-      <br>
       {{#with googleSetting.automaticallyReset}}
         <p class="alert-danger reset-reason">
           Google login was automatically disabled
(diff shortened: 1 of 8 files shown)


shell/client/shell.html
```diff
@@ -1354,8 +1361,12 @@ <h3 id="app-stats">Apps</h3>
     {{else}}
       <div class="feature-key-button-row">
         <button class="button-primary feature-key-upload-button">Upload new feature key...</button>
+        <button class="button-secondary feature-key-delete-button">Delete feature key</button>
```
zarvox: I question the utility of this.  I don't think anyone who isn't us will actually want this, and it's a bit of a footgun to leave in the product.

zarvox: Proposed alternatives:

* add a `testRemoveFeatureKey()` JS function
* `echo "db.featureKey.remove({})" | sudo sandstorm mongo`

zarvox:
+1 for the improved UX.  This is a better conceptual organization than the previous state, for sure. :)

I would appreciate a reasonably-quick review/merge of this -- I'll need to rebase my email settings changes atop this, which have some changes in the same form and event handlers.

shell/client/shell.html
kentonv: It seems wrong to me that we have a setting which, once set, cannot be undone.

It's theoretically not a problem if the feature key strictly adds features, but I'm not sure that's the case. The feature key imposes a user limit, and it's not entirely clear whether that user limit should only apply to the organization (in which case you can deconfigure your organization to avoid the limit) or to all invited users. My current thinking leans towards the latter, in which case adding a feature key in itself could actually cripple a server.

Either way, I think users will be confused if they can't undo the change. They might wonder what will happen to their server when the key expires, etc.

kentonv:
Added a commit which fixes some minor LDAP-related things.

shell/client/shell.html
```diff
@@ -1062,6 +1038,37 @@ <h4 class="share-with-others">Share with others</h4>
...
+      <p>These users will be considered part of your organization. They will automatically be
+        able to log in, install apps, and create grains.</p>
+
+      <p>
+        <label><input type="checkbox" name="isOrganizationLdap" value="value" checked={{isOrganizationLdap}}> Users authenticated via LDAP.</label><br>
+        <label><input type="checkbox" name="isOrganizationGoogle" value="value" checked={{isOrganizationGoogle}}> Users authenticated via Google Apps for Work domain:</label> <input type="text" name="organizationGoogle" value="{{organizationGoogle}}" placeholder="example.com">
+        <label><input type="checkbox" name="isOrganizationEmail" value="value" checked={{isOrganizationEmail}}>Users authenticated via passwordless email login with addresses at domain:</label> <input type="text" name="organizationEmail" value="{{organizationEmail}}" placeholder="example.com"></p>
+    {{else}}
+      <p>To get support for LDAP, SAML, and organization management, enable
+        <a href="{{featuresPath}}">Sandstorm for Work</a>.</p>
```
jparyani: I'd slightly prefer using iron router's linkto helper (https://github.com/iron-meteor/iron-router/blob/devel/Guide.md#linkto), but not a huge deal.

jparyani:
LGTM

shell/client/shell.html
kentonv: That does look nicer! Though I'd still need a "token" helper, and all this code is about to be replaced, so I think I'll just leave it for now.

PR #1644, opened 2016-03-11
Rework email config

zarvox:


paulproteus:
Presumably `install.sh` should also stop referring to `MAIL_URL`

paulproteus:
+0.9 to this (only not +1 because I haven't reviewed in depth)

zarvox:
@paulproteus `install.sh` does not mention `MAIL_URL`.

That said, the docs around configuring email should probably get more thoroughly updated - can I talk you into signing up for that? :D

paulproteus:
@zarvox agreed! I was looking at an out-of-date copy.

shell/client/admin-client.js
```diff
@@ -192,16 +213,21 @@ Template.adminSettings.events({
...
+    if (event.target.emailTokenLogin.checked &&
+            (!event.target.smtpHostname.value || !event.target.smtpPort.value)) {
       handleErrorBound(new Meteor.Error(400,
         "You must configure an SMTP server to use email login."));
       return false;
     }
 
+    // Construct state from email form.
+    const mailConfig = emailConfigFromForm(event.target);
+    console.log(mailConfig);
```
jparyani: Debug logging?

paulproteus:
+0  re: me writing some more/better email docs

If you're breaking something that used to work, I'd like to see you own the docs changes. ("bug")

If you're requesting new docs on something, sure, I can write that. File a bug requesting it. ("feature")

zarvox:
Fair.  I think this qualifies as "breaking something that used to work", so I went ahead and wrote the docs changes. :)

shell/client/admin-client.js
```diff
@@ -760,13 +790,11 @@ Template.adminAdvanced.events({
     const state = Iron.controller().state;
     const token = this.token;
     resetResult(state);
-    state.set("numSettings", 12);
+    state.set("numSettings", 11);
```
paulproteus: Just skimming this review, I have no idea as a reader why a seemingly-magic number of 12 changed to a different seemingly-magic number of 11.

paulproteus: (not a show-stopper)

paulproteus:
+1 to the docs changes

Many thanks @zarvox !

shell/client/admin-client.js
zarvox: The number is the number of settings that we're saving, which is reduced by one because `returnAddress` is no longer present in this form.

I agree it's weird, and I'll try to make saving settings be an atomic action per-page in the new UI, rather than one with 11 independently succeeding-or-failing parts, where possible :P

shell/client/shell.html
```diff
@@ -1078,6 +1070,51 @@ <h4 class="share-with-others">Share with others</h4>
       <p>To get support for LDAP, SAML, and organization management, enable
         <a href="{{featuresPath}}">Sandstorm for Work</a>.</p>
     {{/if}}
+    <div class="legacy-email-config">
+      <h4>SMTP Configuration: <button id="admin-settings-send-toggle">Test</button></h4>
```
kentonv: Nitpick: It feels weird to me that the Sandstorm for Work features would be inserted between non-SFW feature sections rather than appended to the end. Maybe this should stay before LDAP login? I understand the desire to group identity service checkboxes together, but LDAP is already separated from the others by an `<h4>` so I'm not sure there can be said to be any such grouping currently.

shell/client/shell.html
```diff
@@ -1287,8 +1324,6 @@ <h3 id="app-stats">Apps</h3>
   <form id="admin-settings-form">
     <p>Server Title: <input type="text" name="serverTitle" value="{{serverTitle}}"><br>
       <span class="details">The name of this server, for use when communicating with users. Example: "Bob's Sandstorm Server"</span></p>
-    <p>Return Address: <input type="email" name="returnAddress" value="{{returnAddress}}"><br>
```
kentonv: Server Title and Return Address are closely-related: the title is used as the display name in the "from" of emails sent. Today that's actually the only place it's used. Kind of sad for these now to be on separate pages.

kentonv:
Can you post a screenshot of the new UI? (In general I think we should include screenshots in any PR that introduces new UI.)

shell/client/shell.html
zarvox: Mmmm, it's partly about keeping identity service checkboxes together, and partly about keeping identity management and org management (which ties in deeply with identity management) together.  On the whole, I think that whether a feature is enabled by Sandstorm for Work is a less-interesting grouping to users than the things that it interacts with.

What is clear is that email relay configuration doesn't really belong here, so it's good that the new admin UI separates this out.

If anything, perhaps LDAP and SAML don't need their own `<h4>`, or things that are Sandstorm-for-Work-exclusive features could be made to have a different background color.  But I think that's not really important here, and I'm disinterested in spending much more time on a design that I'm aiming to replace in the near term. :P

I intend to leave this one as-is.

shell/client/shell.html
zarvox: It's also used as the last unit in the page `<title>`.  Which is sufficiently non-email-related that I think it doesn't rightly live under email configuration.

Sharing emails should migrate (Soon™) to sending as (From:) the sharer's display name and selected verified email address.  I guess this setting would still be relevant for login and transactional emails, but I think that it applies more broadly than email.

Perhaps the optimal design is to have an *additional* setting for the name of the email sender in the email config, which if empty, defaults to the value of Server Title (and explains all this).

zarvox:
top-of-page

bottom-of-page

test-email-expanded

shell/client/shell.html
kentonv: I disagree with this but the whole UI is going away so whatever.

PR #1646, opened 2016-03-11
Restore last-opened grains on load.

### LINKED ISSUE #1578: misleading error message if no cookies allowed in browser
Cookies are not allowed in my browser by default (Firefow with Cookie Monster Add-On). When opening a sandstorm instance (e.g. https://oasis.sandstorm.io/ ) I get the following misleading message:

> Oops, looks like there's no route on the client or the server for url: "https://oasis.sandstorm.io/." 

Maybe it should just say I have to allow cookies. (If I allow cookies, everything works fine.)

kentonv:
We store the list of open grains in localStorage at shutdown and restore it at startup. We also clear the saved list at startup so that if you open a second window it doesn't open the same grains again. Additionally, we do *not* store anything when closing a window that only has one grain open (or none), as probably in this case the user is relying on browser tabs rather than our sidebar tabs and restoring the last-opened grain would be surprising.

(Also fixed a random exception I've been seeing and fixes #1578.)

### DIFF
--- a/shell/client/grainview.js
+++ b/shell/client/grainview.js
@@ -63,7 +63,7 @@ GrainView = class GrainView {
     // reconstruct the same view later. However, the parentElement and initialPopup arguments
     // can't be included since they are live objects, not data.
 
-    return [ this._grainId, this._path, this._tokenInfo ];
+    return [this._grainId, this._path, this._tokenInfo];
   }
 
   reset(identityId) {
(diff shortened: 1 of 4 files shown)


shell/client/grainview.js
```diff
@@ -748,3 +756,112 @@ const onceConditionIsTrue = (condition, continuation) => {
...
+  if (old) {
+    const oldParsed = JSON.parse(old);
+
+    if (oldParsed.time > Date.now() - 5000) {
+      // Crap. It seems that some other tab was closed in the last 5 seconds that had the same
+      // URL (perhaps a common one like "/apps"). We have no way to distinguish our tab from this
+      // other tab. Rather than arbitrarily clobber one tab's grains list with the other -- which
+      // will likely confuse the user, opening grains in multiple places that weren't previously --
+      // we will have to give up and not restore anything. :(
+      Meteor._localStorage.setItem(JSON.stringify({time: Date.now(), grains: []}));
```
jparyani: This is missing `key`: should be `Meteor._localStorage.setItem(JSON.stringify(key, {time: Date.now(), grains: []}));`

zarvox:
Fix the thing Jason commented on, and then merge!

paulproteus:
By the way - thanks for doing this. It makes using "multigrain" in
Sandstorm (that is, multiple Sandstorm tabs open to different grains) much
more usable.

PR #1650, opened 2016-03-15
install.sh: Switch to type=install_v2

paulproteus:
We use the "type" value when calculating how many people attempted to do a
Sandstorm install. We had to stop using "install" because vagrant-spk happens
to use &type=install during situations that we do not want to categorize as an
attempt by a human to install Sandstorm.


paulproteus:
This is 100% `install.sh` so I will self-merge.

PR #1651, opened 2016-03-15
Update docs to reflect the fact the we no longer depend on ImageMagick.

dwrensha:
See https://github.com/sandstorm-io/sandstorm/commit/f9e5db8a488502bcf10a7d0b5cba7bcab75c6307.


paulproteus:
Yippee, thanks.

PR #1654, opened 2016-03-16
Implement `tabId` to correlate sessions in a grain tab.

kentonv:
In the future, a grain tab may in fact be displaying multiple grains at once, and thus multiple sessions. This can happen when the Picker Powerbox is displaying an embedded picker, or when a grain embeds other grains (e.g. the "Tabbify" proposal). It may also be useful for a single grain to recognize server-side when multiple sequential back-end sessions actually come from the same tab, as occurs e.g. when the user suspends their laptop and then comes back later without refreshing their browser.

In some cases it may be helpful if these grains can recognize that they live in the same tab. Hence we generate a `tabId` which is passed to `newSession()` and uniquely identifies the tab.

The current motivating use case is one where the Powerbox is being used in an authentication interaction: The app wishes the user to "verify" their email address via a powerbox query. The user chooses which address to verify, and a verification capability is returned. We need to make sure the app cannot take this verification capability and pass it off to other users who can then "authenticate" as the address to other apps. To do that, each app needs to be able to check that the verification was actually created by this session's user. We could do this by attaching verification to user IDs, but that would not work in the case that the user is anonymous. Tab IDs are assigned even to anonymous users.

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -309,7 +309,7 @@ Meteor.methods({
                                                      null, null, cachedSalt);
     const result = opened.methodResult;
     const proxy = new Proxy(grainId, this.userId, result.sessionId,
-                            result.hostId, identityId, false, opened.supervisor);
+                            result.hostId, result.tabId, identityId, false, opened.supervisor);
     proxiesByHostId[result.hostId] = proxy;
     return result;
   },
@@ -398,7 +398,7 @@ Meteor.methods({
 
       const result = opened.methodResult;
       const proxy = new Proxy(apiToken.grainId, grain.userId, result.sessionId,
-                              result.hostId, identityId, false);
+                              result.hostId, result.tabId, identityId, false);
       proxy.apiToken = apiToken;
       proxiesByHostId[result.hostId] = proxy;
       return result;
@@ -608,7 +608,8 @@ const getProxyForHostId = (hostId, isAlreadyOpened) => {
         // Note that we don't need to call mayOpenGrain() because the existence of a session
         // implies this check was already performed.
 
-        const proxy = new Proxy(grain._id, grain.userId, session._id, hostId, session.identityId, false);
+        const proxy = new Proxy(grain._id, grain.userId, session._id, hostId, session.tabId,
+                                session.identityId, false);
         if (apiToken) proxy.apiToken = apiToken;
 
         // Only add the proxy to the table if it was not concurrently deleted (which could happen
@@ -880,6 +881,8 @@ function getApiSessionParams(request) {
 getProxyForApiToken = (token, request) => {
   check(token, String);
   const hashedToken = Crypto.createHash("sha256").update(token).digest("base64");
+  const tabId = Crypto.createHash("sha256").update("tab:").update(hashedToken)
+      .digest("hex").slice(0, 32);
   const serializedParams = getApiSessionParams(request);
   const hashedParams = Crypto.createHash("sha256").update(serializedParams).digest("base64");
   return Promise.resolve(undefined).then(() => {
@@ -910,7 +913,7 @@ getProxyForApiToken = (token, request) => {
             identityId = tokenInfo.identityId;
           }
 
-          proxy = new Proxy(tokenInfo.grainId, grain.userId, null, null, identityId, true);
+          proxy = new Proxy(tokenInfo.grainId, grain.userId, null, null, tabId, identityId, true);
           proxy.apiToken = tokenInfo;
           proxy.apiSessionParams = serializedParams;
         }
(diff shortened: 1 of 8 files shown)


dwrensha:
I'm curious about the details of this proposed `EmailVerifier` interface. What would its methods be?

kentonv:
The idea is that you do a powerbox request for a type called `EmailVerification`. The powerbox prompts the user to choose one of their emails. The returned `EmailVerification` capability has no methods, but it can be passed to an `EmailVerifier.verify()` to get the email address. `verify()` also takes a `tabId` as a second parameter, and will only succeed if the `EmailVerification` was originally minted in the specified tab.

    verify(tabId :Data, verification :EmailVerification) -> (address :Text)

(The `EmailVerifier` itself is obtained by the grain's owner doing a powerbox request for `EmailVerifier` during first-time setup. The powerbox lets the owner choose which login providers to trust for verification.)

shell/server/proxy.js
```diff
@@ -880,6 +881,8 @@ function getApiSessionParams(request) {
 getProxyForApiToken = (token, request) => {
   check(token, String);
   const hashedToken = Crypto.createHash("sha256").update(token).digest("base64");
+  const tabId = Crypto.createHash("sha256").update("tab:").update(hashedToken)
```
zarvox: Maybe ApiSessions shouldn't all share the same tabId?

Consider the Piwik case - all of the requests from the tracked page will share the same `tabId`, even though they come from different browsers, different IPs, and will pass the requests on to different `Proxy`s (due to the `apiSessionProxies.get() call below`).

zarvox:
I'd rename the proposed `EmailVerification` to `VerifiableEmail`.

Broader thoughts: it seems like the workflow for setting up grains which will act as public-facing services will require some initial configuration by the owner, like requesting things like an `EmailVerifier` or a domain to be published at or a notification sender.  It *might* be nice if apps didn't necessarily have to build a bunch of UI for requesting these things if they want to treat them as static, global resources.  Parameterized actions might make this a nice workflow, but I guess there's still the problem of "what happens to the grain if the EmailVerifier capability gets revoked and the grain doesn't have UI for requesting a new one" so maybe that doesn't save devs any work.  I guess it could still lead to a neater user experience for the deployer, though.

dwrensha:
Not that this would necessarily be a better design, but would it work to add a `getTabId()` method on `SessionContext`, instead of adding a `tabId` parameter to the `newSession` methods?

dwrensha:
How would the implementation of `EmailVerifier.verify()` work? Does the `EmailVerification` capability have some non-public methods that allow it to be matched with a `tabId`?

kentonv:
> would it work to add a `getTabId()` method on `SessionContext`?

It's kind of unclear whether `SessionContext` is trusted. If someone has an unsealed `UiView` capability (not possible now, but maybe in the future), they can call `newSession()` with arbitrary parameters. We have already decided that Sandstorm will intercept the `newSession()` call and replace parameters as appropriate, but so far there is no specific need for sandstorm to also wrap the `SessionContext`, which is nice. That said, some of the declared-but-not-yet-implemented methods of `SessionContext` do look like they might require trust, but we could very well decide never to implement those.

> How would the implementation of `EmailVerifier.verify()` work?

A capability implementer is able to recognize and "unwrap" its own capabilities returned to it. `EmailVerifier` and `EmailVerification` would both be implemented by the frontend, so yes, it can get at the private internals.

dwrensha:
I wonder whether "tab" is too global of a notion. It seems like we're designing something somewhat analogous to iframes, but instead of providing `window.parent`, we're providing a way to reference the root frame. I worried that this won't nest well. For example, what if the user has a bunch of tabbified views to grains, and multiple of them want to perform actions like email verification that require `tabId` coordination. Might they interfere with each other?

kentonv:
@dwrensha Think of `tabId` as being similar to `userId` except that it can't be tracked across tabs (and therefore can safely identify anonymous users). All nested grains do in fact have the same user, right? How would they interfere in a nested situation? It actually seems useful for nested apps to be able to pass an EmailVerification capability around between themselves and all be able to verify it.

In fact the reason I introduced this rather than use `userId` itself is so that we can hypothetically support verifying email addresses without requiring login.

dwrensha:
Ah, if we plan to constrain nested grains in a tab to all have the same user/identity, then I'm less worried. I had been assuming that nested grains would work basically the same as our tabs do now, with each possibly having a different user/identity.

Anyway, this pull request looks good to me.

zarvox:
Hmm, I'm possibly confused by this constraint about all nested grains having the same user/identity.  What is expected to happen if I have:

* an account with two linked identities: identity1 and identity2
* grain1 owned by (account, identity1)
* grain2 owned by (account, identity2)
* grain2 makes a powerbox request which grain1 can satisfy

Should I see grain1 listed in the powerbox as something which can fulfill grain2's request?  As a user, I would expect the answer to be "yes", but if I'm reading the above correctly, the answer would be "no".

kentonv:
I think in the tabbify use case, we do in fact want all nested grains to be opened under the same identity -- anything else would be very confusing.

For the powerbox case, I think the grain embedded in the powerbox could have a different user ID than the outer grain while still sharing a tab ID. I don't see any obvious problems with that. In fact, it seems useful for the email verification case: you may want to verify an email address that isn't your main identity.

PR #1658, opened 2016-03-17
GrainViewList

### LINKED ISSUE #1625: You can appear to be using a grain anonymously, even if it doesn't allow anonymous access, if you sign out in another tab
Steps to repro: go to https://r.wholezero.org (signed in). Stop streaming. Open a new tab or window. Go to https://mrdomino.sandcats.io. Sign out. Expected: you get signed out and the radio tab goes to "request access", or you stay signed in on that tab because it didn't propagate or something. Actual: you're now not signed in in the sandstorm UI, but the stream is going?!

dwrensha:
This defines a new class `GrainViewList` for the value held by `globalGrains`. The list is automatically cleared whenever `Meteor.userId()` is detected to have changed. When an oauth identity is linked, the grain restoration logic is postponed until the linking is complete, so that the automatic clearing does not destroy any open grains.

Fixes #1625. (@kentonv: I don't fully understand your comment about membranes there. I think it suffices to remove the grain iframe from the DOM, because that should cause any javascript associated with the grain to stop executing.)

### DIFF
--- a/tests/tests/restore-open-grains.js
+++ b/tests/tests/restore-open-grains.js
@@ -89,5 +89,5 @@ module.exports["Test restore open grains"] = function (browser) {
               });
           });
       });
-    })
+    });
 }
--- a/tests/tests/restore-open-grains.js
+++ b/tests/tests/restore-open-grains.js
@@ -1,0 +1,93 @@
+// Sandstorm - Personal Cloud Sandbox
+// Copyright (c) 2016 Sandstorm Development Group, Inc. and contributors
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+var utils = require('../utils'),
+    appDetailsTitleSelector = utils.appDetailsTitleSelector,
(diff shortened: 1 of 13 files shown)


tests/tests/restore-open-grains.js
```diff
@@ -0,0 +1,93 @@
...
+              .click(".navbar-grains>li[data-grainid='" + grainIds[2] + "']>button.close-button")
+              .execute(function (){
+                return document.querySelectorAll(".navbar-grains>li").length;
+              }, [], function (response) {
+                browser.assert.equal(response.value, 0);
+                browser.end();
+              });
+          });
+      });
+    })
```
zarvox: Semicolon here.

zarvox:
Broad :+1: on the refactor; especially the descriptive `getActive()` and `getByOrigin()`.  Thanks!

Since `GrainView`'s constructor now requires a `GrainViewList` instance to be passed in, it might make more sense to make "creating a new GrainView" be done by some factory method `GrainViewList.newGrainView` (with the former GrainView constructor's arguments), which injects `this` as the first argument to `GrainView.constructor()`.

Then, if I'm reading this correctly, `GrainView` itself doesn't need to be exported from the package.  Thoughts?

Either way, LGTM.

dwrensha:
> `GrainView` itself doesn't need to be exported from the package. 

It would still be useful to export it. For example, someone might want to write a function that checks that its parameter has type `GrainView`.

kentonv:
> I don't fully understand your comment about membranes there.

I think I misread the bug as being about revocation rather than logout, in which case a client-side solution would of course be insufficient. I suppose on voluntary logout client-side is fine.

kentonv:
I agree with @zarvox's comment about making a `GrainViewList.newGrainView()` method.

dwrensha:
I've added a `GrainViewlist.addNewGrainView()` method, which also takes care of pushing the new `GrainView` onto the list.

kentonv:
This package seems to still be referring to many globals defined in db-deprecated.js...

PR #1662, opened 2016-03-18
mailConfig.port is a string

dwrensha:
After https://github.com/sandstorm-io/sandstorm/pull/1644, I noticed that I could no longer send email. This fixes the problem.


shell/packages/sandstorm-email/email.js
```diff
@@ -19,7 +19,7 @@ const makePool = function (mailConfig) {
   const pool = simplesmtp.createClientPool(
     mailConfig.port,
```
zarvox: I guess this should be parseInt(mailConfig.port);

Or alternately, the setting should just be stored as a `Number` in the db.

PR #1663, opened 2016-03-18
Check that immediately after demo upgrade, the current identity is the non-demo identity.

dwrensha:
A logged-in user has a "current identity" for each open grain, as well as one for the Sandstorm shell in general. Currently we make sure that after a demo user upgrades, their shell current identity is equal to their non-demo identity.

We do not switch the identity on any other grains. User testing has suggested that it would maybe be less confusing if we did switch those identities, including for any open grain views. However, I'm not entirely convinced that's the correct behavior, as there may be some apps that present a completely different view for different identities, so automatically switching identities for them could make it seem like data has been lost. In any case, switching the identities would not be a completely trivial change. If we decide that's the way to go, we should open an issue for it.


paulproteus:
This does also mean that a Demo Account identity would still show up in the
list of identities, right? If it's unnecessary to keep that identity, I'd
be excited about deleting that identity from the list.

Not a showstopper for this pull request.

PR #1666, opened 2016-03-18
Make error clearer for contact input

### LINKED ISSUE #1657: Improve "Send an Invite" input error message
User tried to share a grain with me by entering in my name and did not succeed. (We hadn't shared any grains between each other before.) They also could have entered my e-mail but they may not have remembered the "Enter email addresses or names" placeholder text, and the input error message did not help them.

Instead of "Invalid text in contact input box: nena nguyen" I lightly suggest: "Enter in a valid name or email address"

It'd be nice if we could communicate that you can only enter names/handles of people you "know" but I can't think of an efficient way to say that.

(Spotted in user testing.)
image
image

jparyani:
Fixes #1657

### DIFF
--- a/shell/client/grain-client.js
+++ b/shell/client/grain-client.js
@@ -1120,7 +1120,7 @@ Template.emailInviteTab.events({
         emails.value = "";
       } else {
         instance.completionState.set(
-          { error: "Invalid text in contact input box: " + emailsValue });
+          { error: "Unknown user \"" + emailsValue + "\". Try entering an e-mail address." });
         return;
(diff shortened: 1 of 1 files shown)


kentonv:
This is still a bit unclear. What is a "valid name"?

How about: `Unknown user "foo". Try entering an e-mail address.`

They'll figure out that auto-complete is a thing when they see it.

jparyani:
Done

neynah:
Great! Thanks for fixing this. :)

PR #1667, opened 2016-03-18
Email port should probably be a Number, not a String

zarvox:
Closes #1662.

### DIFF
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -1094,7 +1094,7 @@ limitations under the License.
       </span>
 
       <label>Relay port:
-        <input type="text" name="smtpPort" value="{{smtp.port}}" pattern="[0-9]+" />
+        <input type="number" name="smtpPort" value="{{smtp.port}}"/>
       </label>
       <span class="details">SMTP (unencrypted or with StartTLS) uses port 25, TLS-encrypted SMTPS
           uses port 465.</span>
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -1094,7 +1094,7 @@ limitations under the License.
       </span>
 
       <label>Relay port:
-        <input type="text" name="smtpPort" value="{{smtp.port}}" />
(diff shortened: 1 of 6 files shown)


dwrensha:
Looks like you need to update [this comment](https://github.com/sandstorm-io/sandstorm/blob/30ee7f3b6aed4da8a4ffc73273047714057f1cd7/shell/packages/sandstorm-email/email.js#L129) too.

zarvox:
Fixed, thanks.

dwrensha:
For the curious: [here](https://github.com/andris9/simplesmtp/blob/v0.3.35/lib/pool.js#L30) is where SimpleSmtp claims to require an Number for `port`.

zarvox:
FWIW, it was working because under the hood, [tls.connect](https://nodejs.org/api/tls.html#tls_tls_connect_port_host_options_callback allows) either String or Number, and simplesmtp wasn't actually enforcing any type correctness.

But since simplesmtp documents expecting a Number, we should stick to that.

shell/client/shell.html
```diff
@@ -1094,7 +1094,7 @@ <h4 class="share-with-others">Share with others</h4>
       </span>
 
       <label>Relay port:
-        <input type="text" name="smtpPort" value="{{smtp.port}}" />
+        <input type="text" name="smtpPort" value="{{smtp.port}}" pattern="[0-9]+" />
```
kentonv: Maybe use `type="range" min="1" max="65535"`? This way the browser can give a better error when the value doesn't match (and I think will even display a numeric keypad on mobile). Alternatively, give the input a `title` which will be displayed on error.

zarvox: `type="range"` produces a slider, not a text input box, so I think that's not what we want.

`type="number"` is mildly odd in that it gives up/down arrows, but it does provide the mobile hint and better validation behavior, so I've gone with that.

PR #1668, opened 2016-03-18
Move feature key explanation into another section of the same page

paulproteus:


paulproteus:
Self-merging, 100% docs.

PR #1669, opened 2016-03-18
Add ldap filter field

jparyani:
This allows users to specify a raw ldap filter that they want to apply.


jparyani:
Added the ability for admins to specify a pre-search bind dn/pass for ldap servers that don't allow unauthenticated searches:
image

PR #1674, opened 2016-03-20
Remove db-deprecated references from grainview code.

dwrensha:


shell/packages/sandstorm-ui-grainview/grainview.js
```diff
@@ -136,7 +158,7 @@ GrainView = class GrainView {
         }
       });
     } else {
-      if (ApiTokens.findOne({
+      if (this.db.collections.apiTokens.findOne({
```
zarvox: s/db/_db/

shell/packages/sandstorm-ui-grainview/grainview-list.js
```diff
@@ -67,6 +85,15 @@ GrainViewList = class GrainViewList {
     });
   };
 
+  setTopbar(topbar) {
+    // A SandstormTopbar wants a reference to its GrainViewList, and a GrainViewList wants a
+    // reference to its SandstormTopbar. To break the cycle, we construct the GrainViewList first
+    // and then require that the contructor to SandstormTopbar call this method. This all happens
+    // before Meteor.startup(). Luckily, a GrainViewList never needs to refer to its topbar before
+    // Meteor.startup().
+    this._topbar = topbar;
```
zarvox: Tracing through, it looks like the only place we use `this._topbar` is passing it to the constructor of `GrainView`, which only uses the reference in its constructor to possibly set an initial popup (either `"login"` if you are using a webkey anonymously or possibly `"share"` if `Session.get("share-grain-${grainId}")` is set.)

I wonder if the job of opening the popup is actually better owned by the route handler, rather than being threaded through `GrainViewList` and `GrainView`, which would break the circular dependency.  Given that our popups are modal, rather than per-grain, I think that might also make more structural sense.

shell/packages/sandstorm-ui-grainview/grainview.js
dwrensha: Fixed.

shell/packages/sandstorm-ui-grainview/grainview-list.js
dwrensha: Good idea! I went ahead and implementing this.

zarvox:
Nice.  :shipit:

PR #1675, opened 2016-03-20
Refactor Picker Powerbox so that query matching happens on the server side.

kentonv:
PowerboxDescriptors are now provided in base64url-serialized-packed Cap'n Proto format, not JSON. For example, a UiView can be requested as "EAZQAQEAABEBF1EEAQH_5-Jn6pjXtNsAAAA".

This will be necessary going forward because PowerboxDescriptor.Tag.value is an AnyPointer, the schema of which Sandstorm may not even know. The query-matching algorithm is designed to operate on Cap'n Proto binary format without the need for a schema, but without a schema it is not possible to convert to/from JSON.

I expect most apps won't need to customize their queries very much, so can probably get away with using hard-coded queries copied from documentation. Copying a whole query string seems not much worse than copying a type ID. I also imagine adding an spk sub-command for generating query strings.

### DIFF
--- a/shell/packages/sandstorm-ui-powerbox/powerbox-client.js
+++ b/shell/packages/sandstorm-ui-powerbox/powerbox-client.js
@@ -92,6 +92,7 @@ SandstormPowerboxRequest = class SandstormPowerboxRequest {
             this._requestInfo.source.postMessage({
               rpcId: this._requestInfo.rpcId,
               token: apiToken,
+              // encoded/packed/base64url of (tags = [(id = 15831515641881813735)])
               descriptor: "EAZQAQEAABEBF1EEAQH_5-Jn6pjXtNsAAAA",
             }, this._requestInfo.origin);
             // Completion event closes popup.
--- a/shell/packages/sandstorm-ui-powerbox/powerbox-client.js
+++ b/shell/packages/sandstorm-ui-powerbox/powerbox-client.js
@@ -134,6 +134,20 @@ SandstormPowerboxRequest = class SandstormPowerboxRequest {
   }
 
   filteredCardData() {
+    // Returns an array of "cards" (options from which the user can pick), sorted in the order in
+    // which they should appear. Each has the fields described in the comments for
+    // Meteor.publish("powerboxOptions") in powerbox-server.js as well as the following fields
+    // added client-side:
+    //
+    // _id: Unique identifier for this card among the results.
+    // title: Human-readable title string.
+    // appTitle (optional): Human-readable title of the app serving this option.
+    // iconSrc (optional): URL of an icon.
+    // lastUsed (optional): Date when this item was last accessed.
+    // callback: Function returning a function to call if this card is chosen. (The double-function
+    //     is necessary because when a function value is named in a Blaze template, Blaze calls
+    //     the function, thinking it is a helper.)
+
     const cards = PowerboxOptions.find({ requestId: this._requestId }).map(cardData => {
       // Use ID as title if we don't find anything better.
       cardData.title = cardData._id;
--- a/shell/packages/sandstorm-ui-powerbox/powerbox-client.js
+++ b/shell/packages/sandstorm-ui-powerbox/powerbox-client.js
@@ -316,11 +316,6 @@ Template.powerboxRequest.helpers({
     return ref && ref._selectedProvider && ref._selectedProvider.get().templateData();
   },
 
-  requestedInterfaceIsImplementedByFrontendRef: function () {
-    const ref = Template.instance().data.get();
-    return true;
-  },
-
   showWebkeyInput: function () {
     // Transitional feature: treat requests that specified no query patterns to match, not even the
     // empty list, as requests for webkeys.  Later, we"ll want to transition the test apps to
--- a/shell/packages/sandstorm-ui-powerbox/powerbox-client.js
+++ b/shell/packages/sandstorm-ui-powerbox/powerbox-client.js
@@ -1,3 +1,21 @@
+// Sandstorm - Personal Cloud Sandbox
+// Copyright (c) 2016 Sandstorm Development Group, Inc. and contributors
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const PowerboxOptions = new Mongo.Collection("powerboxOptions");
+
 SandstormPowerboxRequest = class SandstormPowerboxRequest {
   constructor(db, requestInfo) {
     check(requestInfo, Match.ObjectIncluding({
@@ -5,7 +23,7 @@ SandstormPowerboxRequest = class SandstormPowerboxRequest {
       origin: Match.Any,
       rpcId: Match.Any,
       // For reasons I don't understand, Match.Optional does not work here.
-      query: Match.OneOf(undefined, [Object]),
+      query: Match.OneOf(undefined, [String]),
       saveLabel: Match.Optional(Match.Any),
       sessionId: String,
       grainId: String,
@@ -21,6 +39,14 @@ SandstormPowerboxRequest = class SandstormPowerboxRequest {
     // State to track UI status.
     this._filter = new ReactiveVar("");
     this._selectedProvider = new ReactiveVar(undefined);
+
+    this._requestId = Random.id();
+  }
+
+  subscribe(tmpl) {
+    if (this._requestInfo.query) {
+      tmpl.subscribe("powerboxOptions", this._requestId, this._requestInfo.query);
+    }
   }
 
   finalize() {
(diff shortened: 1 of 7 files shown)


shell/packages/sandstorm-ui-powerbox/powerbox-server.js
```diff
@@ -1,16 +1,53 @@
...
-    if (!db.isAdmin(this.userId)) {
-      throw new Meteor.Error(403, "User must be an admin to powerbox offer frontendrefs");
+
+    let descriptor;
+    if (frontendRefVariety.ipNetwork) {
+      if (!db.isAdmin(this.userId)) {
+        throw new Meteor.Error(403, "User must be an admin to powerbox offer IpNetwork");
+      }
+
+      descriptor = "EAZQAQEAABEBF1EEAQH_QCAqemtXgqkAAAA";
```
zarvox: I would be happier if these were constructed programmatically to clarify intent and to make it easier for the next person who needs to produce such a descriptor to figure out how it should look.  Since this is server code, you have access to all the schemas and type IDs, which would make this clearer.

shell/packages/sandstorm-ui-powerbox/powerbox-client.js
```diff
@@ -504,11 +318,7 @@ Template.powerboxRequest.helpers({
 
   requestedInterfaceIsImplementedByFrontendRef: function () {
     const ref = Template.instance().data.get();
-    return (
-      ref.requestedInterfaceMatchesTag({ id: "12214421258504904768" }) || // IpNetwork
-      ref.requestedInterfaceMatchesTag({ id: "16369547182874744570" }) || // IpInterface
-      ref.requestedInterfaceMatchesTag({ id: "15831515641881813735" }) // UiView
-    );
+    return true;
```
zarvox: This seems inaccurate.  If it's no longer needed, consider removing it?

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
```diff
@@ -31,8 +68,240 @@ Meteor.methods({
...
+    } else {
+      return false;
+    }
+  }
+}
+
+const specialCaseTypes = {
+  // This object maps tag IDs to functions which return lists of matches for that tag.
+
+  15831515641881813735(db, userId, value) {  // UiView
```
zarvox: I'd be marginally happier if these referred to the type ids programmatically, but I see how that would be more awkward here.

shell/packages/sandstorm-ui-powerbox/powerbox-client.js
```diff
@@ -193,133 +133,103 @@ SandstormPowerboxRequest = class SandstormPowerboxRequest {
...
+      cardData.lastUsed = ownerData.lastUsed;
+      cardData.apiTokenId = apiToken._id;
+
+      cardData.callback = () => () => {
+        Meteor.call("getViewInfoForApiToken", grainCard.tokenId, (err, result) => {
+          if (err) {
+            console.log(err);
+            this._error.set(err.toString());
+          } else {
+            this.selectGrain(grainCard, result || {});
```
zarvox: I thought `cachedViewInfo` for a running grain would never be falsy; is that mistaken?

shell/packages/sandstorm-ui-powerbox/powerbox-client.js
```diff
@@ -151,12 +92,7 @@ SandstormPowerboxRequest = class SandstormPowerboxRequest {
             this._requestInfo.source.postMessage({
               rpcId: this._requestInfo.rpcId,
               token: apiToken,
-              descriptor: {
-                tags: [
-                  { id: "15831515641881813735" }, // UiView
-                ],
-                quality: "acceptable",
-              },
+              descriptor: "EAZQAQEAABEBF1EEAQH_5-Jn6pjXtNsAAAA",
```
zarvox: Please leave a comment about what this blob represents, to make it clearer to the next person to read this code.  Otherwise, they have to pull out a urlsafe base64 decoder, pycapnp or node-capnp, and the Sandstorm schemas to understand/verify the intent of this line.

zarvox:
Garply, retest this please

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
```diff
@@ -31,8 +68,240 @@ Meteor.methods({
...
+  // Performs a powerbox query, returning options to present to the user in the powerbox.
+  // `descriptorList` is an array of `PowerboxDescriptor`s, each individually serialized in
+  // packed format and base64-encoded. The publish populates a pseudo-collection called
+  // `powerboxOptions`. Each item has the following fields:
+  //
+  //   _id: Object with two fields: `requestId` (as passed when subscribing) and `optionId` (a
+  //       unique identifier for the option).
+  //   matchQuality: "preferred" or "acceptable" ("unacceptable" options aren't returned).
+  //   frontendRef: If present, selecting this option means creating a simple frontendRef. The
+  //       field value should be passed back to the method `newFrontendRef` verbatim. The fromat
```
zarvox: s/fromat/format/

shell/packages/sandstorm-ui-powerbox/powerbox-client.js
```diff
@@ -193,133 +133,103 @@ SandstormPowerboxRequest = class SandstormPowerboxRequest {
...
-            };
-          },
-        });
+  filteredCardData() {
+    const cards = PowerboxOptions.find({ requestId: this._requestId }).map(cardData => {
+      // Use ID as title if we don't find anything better.
+      cardData.title = cardData._id;
+
+      if (cardData.grainId) {
+        this.extendWithGrainInfo(cardData);
```
zarvox: I find this logic a little hard to follow, in that it doesn't tell me anything about the universe of fields I can expect to find in the eventual return value.  I can tell from the comments in the serverside publish for `PowerboxOptions` what values come from there, but it's not clear at a glance what `extendWithGrainInfo` will add, remove, or change, or under what conditions I can expect those values to be present or absent.

The previous code was a bit more explicit here; the final object structure was all in one place.  Documenting which fields are guaranteed to be present in the value returned by `filteredCardData()` would be a welcome addition.

Maybe these cardData instances are large enough that they deserve a class extraction, which would give some clearer guarantees about what data this type contains?

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
```diff
@@ -31,8 +68,240 @@ Meteor.methods({
...
+
+      return { descriptor, matches };
+    });
+
+    // TODO(someday): The implementation of matchQuality here is not quite right. In theory, we're
+    //   supposed to compare descriptors to determine which ones are more specific than which
+    //   others, and prefer the most-specific match. For now, though, I've implemented a heuristic:
+    //   consider each descriptor in order. If it is "unacceptable", have it cancel out any
+    //   matches seen previously. Otherwise, take the maximum match quality. This will usually
+    //   produce the same results as long as "unacceptable" descrtiptors are placed last in the
```
zarvox: s/descrtiptors/descriptors/

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
kentonv: Whoa. You commenting on this line made me realize that `jscs --fix` actually injected a bug here by removing the quotes around these numbers! Ugh.

zarvox: Oh wow, because if that value is treated as a number, we lose precision because we only have 53-bit numerics.  Amazing.

That might be worth filing a bug against JSCS.

shell/packages/sandstorm-ui-powerbox/powerbox-client.js
kentonv: Not totally sure. I think it's usually true, but let's say you create a grain that fails before returning ViewInfo. Presumably you'll be left with a grain table entry with no cachedViewInfo. Better to be safe here.

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
kentonv: https://github.com/jscs-dev/node-jscs/issues/2183

I've disabled this check for now.

shell/packages/sandstorm-ui-powerbox/powerbox-client.js
kentonv: Oops, done. (I had modified this function quickly while debugging then forgot about it...)

shell/packages/sandstorm-ui-powerbox/powerbox-client.js
kentonv: Added comments. Refactoring into a class might be a good idea but not going to do it right now.

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
kentonv: Done.

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
```diff
@@ -39,13 +46,13 @@ Meteor.methods({
         throw new Meteor.Error(403, "User must be an admin to powerbox offer IpNetwork");
       }
 
-      descriptor = "EAZQAQEAABEBF1EEAQH_QCAqemtXgqkAAAA";
+      descriptor = encodePowerboxDescriptor({tags: [{id: "12214421258504904768"}]});
```
zarvox: Bonus points (optional):

```js
const Ip = Capnp.importSystem("sandstorm/ip.capnp");
```

...

```js

descriptor = encodePowerboxDescriptor({
  tags: [
    { id: Ip.IpNetwork.typeId },
  ],
});
```

and likewise for IpInterface.

zarvox:
Did you test the IpNetwork/IpInterface/UiView interface-specifying request flows?  The change in API presumably breaks my test app, so I should probably update it accordingly.

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
```diff
@@ -1,16 +1,60 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const Capnp = Npm.require("capnp");
+const Grain = Capnp.importSystem("sandstorm/grain.capnp");
+
+function encodePowerboxDescriptor(desc) {
+  return Capnp.serializePacked(Grain.PowerboxDescriptor, desc)
+              .toString("base64")
+              .replace("+", "-")
```
zarvox: This first arg needs to be regex to replace all the instances rather than just the first.  Ditto the following line.

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
```diff
@@ -31,8 +75,240 @@ Meteor.methods({
...
+  const results = {};
+  const db = this.connection.sandstormDb;
+
+  if (descriptorList.length > 0) {
+    const descriptorMatches = descriptorList.map(packedDescriptor => {
+      // Decode the descriptor.
+      // TODO(now): Also single-segment? Canonical?
+
+      // Use URL-safe base64.
+      unUrlsafe = packedDescriptor.replace("-", "+").replace("_", "/");
```
zarvox: Same story as the encoding side: the replacements need to be regex to replace all instances of "-" rather than just the first.

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
kentonv: I honestly forgot we had .typeId constants. I've gone and replaced all the hard-coded IDs with these.

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
kentonv: Fixed.

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
kentonv: Turns out Node documentedly accepts either character set when decoding, so I removed the replacements altogether.

kentonv:
> Did you test the IpNetwork/IpInterface/UiView interface-specifying request flows? The change in API presumably breaks my test app, so I should probably update it accordingly.

Yes, I tested them. Your test app is not part of the test suite, I take it?

zarvox:
> Yes, I tested them. Your test app is not part of the test suite, I take it?

Great!  I was just checking, since you didn't mention it in the PR overview.

The powerbox request/offer bits of my test app are not yet a part of the test suite.  (The app is used for a number of other tests like backup/restore and other postMessage-y things.)

zarvox:
This all looks good to me.

PR #1681, opened 2016-03-21
Add docs on new LDAP fields

paulproteus:
Thanks to @jparyani for writing the code, and @TC01 for letting us know this was required.

### DIFF
--- a/docs/administering/for-work.md
+++ b/docs/administering/for-work.md
@@ -27,6 +27,26 @@ To enable either of these login providers, take the following steps:
 
 You should now see a checkbox that allows you to enable LDAP login.
 
+Implementation notes for LDAP that may apply to your site:
+
+- A typical LDAP configuration (e.g. Active Directory) will use an **LDAP Base Search Dn** of
+  `ou=People` followed by the LDAP version of your domain name. For `example.com`, this would be
+  `ou=People,dc=example,dc=com`.
+
+- When logging in with LDAP, you might run hit this error: `Exception while invoking method 'login'
+  SizeLimitExceededError`. In our testing, this seems to occur with LDAP sites where multiple LDAP
+  objects match the username. In this case, you probably need to add a custom **LDAP Search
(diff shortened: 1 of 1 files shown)


paulproteus:
Self-merging, as this is 100% docs.

ocdtrekkie:
I feel like many Windows admins might still be confused here. In five years of working with AD, I think I've actually stared at the actual LDAP string like... once. And it was when I was troubleshooting broken replication between domain controllers across geographic sites. It's not something many will see commonly.

Is it at all possible to replicate the workflow of entertain the domain (sometimes it's not even example.com, usually you can enter EXAMPLE), and have it find the domain, and then prompt you for authorized credentials to connect to it?

If that's not possible, it'd be ideal to document a way that a Windows admin could see their LDAP string from their domain controller.

jparyani:
@ocdtrekkie: We do want to add an upfront option that is "this is an AD server" and auto-configure based on that. I didn't get a chance to test an AD install and get this working, but it's definitely something we're interested in doing. It'll probably happen soon after SAML support is done.

ocdtrekkie:
Awesome, @jparyani

zarvox:
If I remember correctly, AD lets you specify the bind user as something like `\\EXAMPLE\Directory Admin` at the LDAP protocol level and will map that to the appropriate CN/DN/DC values, so this may "just work".

It's not clear to me if looking up EXAMPLE => example.com is something we can expect to Just Work on a Linux machine in a Windows world, or if there's additional legwork needed, but it almost certainly won't work if a user is just spinning up a random cloud instance which isn't served by the AD-controlled DHCP/DNS server.

But yeah, I see a place for a "setting up LDAP" guide.  I am semi-inspired by https://support.aerofs.com/hc/en-us/articles/204861930-How-Do-I-Configure-ActiveDirectory-And-LDAP-With-My-AeroFS-Appliance- .

PR #1685, opened 2016-03-22
Fix lots of bugs in SturdyRef handling.

kentonv:
Nothing exploitable (at least with today's non-powerbox-using apps).

- SandstormPermissions.createNewApiToken()'s check()s lazily used Match.ObjectIncluding() to match LocalizedString, possibly allowing garbage through (ignored garbage, but this is being stored).
- SandstormPermissions.createNewApiToken() did not support all ApiTokenOwner types, so could't safely be used to implement save().
- SandstormCoreImpl.restore() added a membrane reqirement that the token was valid, only to have save() (makeChildTokenInternal()) attempt to filter it out of the requirements later. It seems more natural to say that this requirement is implicit, so doesn't need to be in the list.
- SandstormCoreImpl.restore() was not enforcing that the token was owned by the calling grain.
- SandstormCoreImpl.makeChildToken() was passing the grainId into makeChildTokenInternal() where it was being incorrectly stored on the token record, possibly changing the meaning of the token or at least creating confusion. ApiToken.grainId is supposed to identify the grain that hosts the capability, not the grain that created the child token.
- SandstormCoreImpl.makeChildToken() was not correctly handling UiView capabilities.
- The various save() implementations had varying, inconsistent implementations. Some of them would set the parent token to the root token rather than the child that the capability had been restored from, or not setting a parent at all. Many of them were not correctly saving the requirements list that had been passed to restore(). I've now unified all the save()s into a single implementation.
- restoreInternal() was only using the `requirements` passed to it in the case of an app ref (objectId), and even had a comment saying this, but this is not correct -- these requirements apply to all object types.
- When an offer template token was first used, `expiresIfUnused` was being set `null` instead of being removed (inefficient).

@jparyani I may have forgotten why we did things a certain way. Let me know if you remember something I don't.

@zarvox Where can I find your picker powerbox test app? (Ideally already compiled)

### DIFF
--- a/shell/server/core.js
+++ b/shell/server/core.js
@@ -371,6 +371,9 @@ restoreInternal = (originalToken, ownerPattern, requirements, tokenId) => {
 
   // The capability we restore must implement SystemPersistent, and we already know what the
   // implementation will look like. So, construct it here.
+  //
+  // TODO(cleanup): It would probably be a lot cleaner to have a common base class that these
+  //   inherit, and to pass down some common parameters to the constructor.
   const persistentMethods = {
     save(params) {
       return inMeteor(() => {
--- a/shell/server/core.js
+++ b/shell/server/core.js
@@ -32,13 +32,17 @@ class SandstormCoreImpl {
     const _this = this;
     return inMeteor(() => {
       const hashedSturdyRef = hashSturdyRef(sturdyRef);
-      const token = ApiTokens.findOne(hashedSturdyRef);
-      const requirements = [
-        {
-          tokenValid: hashedSturdyRef,
-        },
-      ];
+      const token = ApiTokens.findOne({
+        _id: hashedSturdyRef,
+        "owner.grain.grainId": this.grainId,
+      });
 
+      if (!token) {
+        throw new Error("no such token");
+      }
+
+      // Honor `requiredPermissions`.
+      const requirements = [];
       if (requiredPermissions && token.owner.grain.introducerIdentity) {
         requirements.push({
           permissionsHeld: {
@@ -49,9 +53,9 @@ class SandstormCoreImpl {
         });
       }
 
-      return restoreInternal(hashedSturdyRef,
+      return restoreInternal(sturdyRef,
                              { grain: Match.ObjectIncluding({ grainId: _this.grainId }) },
-                             requirements, sturdyRef);
+                             requirements, hashedSturdyRef);
     });
   }
 
@@ -85,7 +89,9 @@ class SandstormCoreImpl {
   makeChildToken(parent, owner, requirements) {
     const _this = this;
     return inMeteor(() => {
-      return makeChildTokenInternal(hashSturdyRef(parent), owner, requirements, _this.grainId);
+      return {
+        token: new Buffer(makeChildTokenInternal(parent, owner, requirements)),
+      };
     });
   }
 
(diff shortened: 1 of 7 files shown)


shell/server/core.js
```diff
@@ -142,59 +155,11 @@ class NotificationHandle {
...
-          this._token.grainId,
-          this._token.petname,
-          this._token.roleAssignment,
-          owner);
-
-      return { sturdyRef: new Buffer(ret.token) };
-    });
-    return res;
+  constructor(persistentMethods) {
+    _.extend(this, persistentMethods);
```
zarvox: I do not like this approach of dynamically adding functions to `this`, as a general scheme.  It makes it much more difficult to navigate around and understand the code using types, and makes it harder to tell what `save()` will actually do, since now you have to find where this instance was created and look at the arguments to the constructor.

I haven't finished reading the rest of this PR yet, but is this a necessary change to support some important behavior, or can we instead pass in and save the appropriate data needed to implement `save()` to the constructor, as the previous implementation did?

zarvox:
The source for my test app lives at https://github.com/zarvox/sandstorm-test-python .  I've updated it to follow the new scheme in #1675.

A prebuilt SPK of the latest sandstorm-test-python master which works with current Sandstorm master ( 432be37489b210b19d9677f29befcd5c910a0f76 as of this writing ) can be found at https://alpha-hlngxit86q1mrs2iplnx.sandstorm.io/test-6.spk

zarvox:
Maybe I misunderstood the docs at https://github.com/sandstorm-io/sandstorm/blob/master/shell/packages/sandstorm-db/db.js#L371..L377 but I thought that the permissions computation requires `grainId` to be set on the token, but you seem to indicate that that field is only intended to identify the grain hosting the capability.  @dwrensha can you clarify how you intended for the pseudo-capnp to be serialized into the ApiToken db structure?

shell/server/core.js
kentonv: The problem is that the implementation of `save()` is likely to change in the future, and also we're likely to want to implement `addRequirements()` as well.

Another strategy, I suppose, would be to have all these classes inherit a common base class and then pass around an encapsulated version of the state it needs.

kentonv:
> I thought that the permissions computation requires grainId to be set on the token, but you seem to indicate that that field is only intended to identify the grain hosting the capability

Yes. The grainId identifies the grain to which the capability points, *not* the grain which requested creation of the token. (Before this change, makeChildToken() implemented the latter logic.)

kentonv:
Garply, retest this please.

shell/server/core.js
```diff
@@ -32,13 +32,17 @@ class SandstormCoreImpl {
     const _this = this;
     return inMeteor(() => {
       const hashedSturdyRef = hashSturdyRef(sturdyRef);
-      const token = ApiTokens.findOne(hashedSturdyRef);
-      const requirements = [
-        {
-          tokenValid: hashedSturdyRef,
```
jparyani: Are we sure we want to drop this requirement? I think it may be reasonable to do so, but I'm worried about accountability (ie. giving users the ability to revoke a capability that's been granted, even if it's been re-saved by another owner).

jparyani:
LGTM other than my one comment about changing a requirement on restore.

I do agree with Drew that passing around save methods seems a bit dirtier than doing it with inheritance and passing state instead, but it's not a showstopper.

shell/server/core.js
kentonv: I'm not dropping the requirement. It is more natural to treat this requirement as implied rather than explicitly list it. The requirements in this list will be explicitly checked during restoreInternal(), but checking the parent token there would be redundant since restoreInternal() is already going to follow parent tokens in order to figure out what it's restoring, and will already fail if any of them are no longer valid. I've changed comments in supervisor.capnp to indicate that the parent token requirement is implied and won't be listed explicitly. Since the parent token is passed into Supervisor.restore() as a separate argument, it doesn't need to be listed again under `requirements`.

kentonv:
I agree that the current "mixin" strategy for persistence is ugly and we should probably instead have a common base class that implements the methods (requiring some common constructor parameters to be passed down).

For now I've left a TODO(cleanup) to this effect.

PR #1686, opened 2016-03-22
Link to newer LDAP Search Filter RFC.

dwrensha:
According to [the introduction of RFC 4515](https://tools.ietf.org/search/rfc4515#section-1):

> This document replaces RFC 2254.


jparyani:
Technically the library we're using, ldapjs, says it supports RFC 2254 and doesn't mention 4515 (see. http://ldapjs.org/filters.html).

kentonv:
So it sounds like we shouldn't change this? @dwrensha what is your opinion?

dwrensha:
Matching the library's documentation sounds like a prudent plan.

PR #1687, opened 2016-03-22
Mention https port in reverse proxy docs

paulproteus:
These docs changes adjust a few things:

- Mention how `HTTPS_PORT` relates to other things in the reverse proxy docs.

- Adjust `wildcard.md` to tell people how to set up wildcard DNS, rather than just that we're so cool to use it.

- Adjust reverse proxy docs to clarify that HTTPS behind a reverse proxy is optional.


paulproteus:

PR #1689, opened 2016-03-22
Fix pluralization error

paulproteus:
Thanks @zarvox 

While at it, clarify that DNS != certificates.


paulproteus:

Issue #1690, opened 2016-03-22
Share-by-identity sends emails that result in Sandstorm seeming to hang before ever loading the grain


paulproteus:
Steps to reproduce:

- Alice shares a grain (by identity) with Bob

- Bob visits their email

- Bob clicks the link

Expected behavior:

- Bob sees the grain

Actual behavior:

- Bob sees the Sandstorm UI but the grain never seems to start loading, let alone finish loading

Current diagnosis:

- Data availability race w/ Iron Router??

Note: We've seen this happen at least once. I don't know yet if it's reliable.

paulproteus:
Last I heard @zarvox was working on this, so assigning to him for now. I hope that's OK!

paulproteus:
FWIW, one thing that occurs to me is that this is a showstopper-level bug that, when I worked at Eventbrite, would merit a code push the same day. This is not to say we will/won't be able to do that at Sandstorm, but maybe it suggests that we should have a label and/or a process that describes these kinds of urgent & important bugs.

kentonv:
Release pushed. Oasis hotfixed.

PR #1692, opened 2016-03-23
Fix bug in direct share links. Add test.

dwrensha:
Fixes https://github.com/sandstorm-io/sandstorm/issues/1690.

### DIFF
--- a/shell/client/grain-client.js
+++ b/shell/client/grain-client.js
@@ -1490,6 +1490,9 @@ Router.map(function () {
           return this.next();
         } else if (tokenInfo.invalidToken) {
           this.state.set("invalidToken", true);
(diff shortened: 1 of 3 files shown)


dwrensha:
The problem: the GrainView gets created and then, before it can be pushed onto the global grains list, gets destroyed in `_redirectFromShareLink()`. Only then does the GrainView get pushed onto the list, but by then it's dead, so its presence there causes trouble when the `/grain/` route tries to open the grain.

The solution is to detect earlier that we need to redirect, avoiding the need to even construct a GrainView before the redirect.

dwrensha:
Garply, retest this please.

kentonv:
Garply, retest this please.

Issue #1694, opened 2016-03-23
Some apps fail to run


jvasile:
I installed Sandstorm today and all is well for many apps.  Others just fail to work right out the gate.  Telescope and Groove Basin both load an initial screen and then keep refreshing or reloading forever without end.  When I hit the debug log, I see the same line over and over: { type: 'sandstorm', userId: 'randomstring' }

Maybe this is an implementation bug or maybe it's a config problem on my end, but it's definitely a UX fail.  I was left without much ability to hunt down the error.

dwrensha:
Hmm. This sounds like what happens when websockets aren't working for some reason.

paulproteus:
It could be that now is a good time to figure out how to integrate "maaan
your websockets are broken" into a message kind of like the admin warning
that complains about wildcard DNS. I could possibly be the person to write
that.

jvasile:
It might be useful to test web sockets upon initial install, the same way wildcard sub domains get tested.

I'm not sure where websockets enter the picture, but here's another data point in case it helps: I'm running Sandstorm on a box that sits behind a reverse proxy located on a different box.  The box running Sandstorm can make outgoing connections to the Internet but is firewalled against incoming connections except those coming from the reverse proxy.  The broken web sockets issue might be related to this.

paulproteus:
Hi @jvasile ,

I suppose there are at least two things to find out here.

First, how can we solve your immediate problem?

Second, how can we help people like you solve this problem without having to get in touch with us?

I'll work on the first right now, and turn the second into a separate GitHub issue.

paulproteus:
If you're using a reverse proxy, which you say you are, is it nginx? If so, note that you'll need a stanza like:

```


# This magic stanza is ESSENTIAL to avoid breaking WebSockets.
#
# Specifically, for WebSocket forwarding, we want to forward the `Connection` header.
# This "map" declaration helps with that.
map $http_upgrade $connection_upgrade {
  default upgrade;
  ''      close;
}
```

in your nginx settings. See also: https://github.com/sandstorm-io/sandstorm/blob/master/docs/administering/sample-config/nginx-example.conf & https://docs.sandstorm.io/en/latest/administering/reverse-proxy/

If your reverse proxy is Apache2, then I am honestly not yet sure how to enable WebSockets while reverse proxying. Let me know which you're using, and we can figure things out.

paulproteus:
For nginx, you'll also need something like:

```
  location / {
[...]

PR #1695, opened 2016-03-23
Document Apache reverse proxy configuration

paulproteus:
Refs #1694


paulproteus:

PR #1696, opened 2016-03-24
In grain-invite emails, fill in the "from" field with a more appropriate value.

dwrensha:
This uses `SandstormDb.getReturnAddress()` for the envelope sender and something like "Alice \<alice@example.com\>" for the inner "from" field. It also fills the subject line with "grain title - Invitation to collaborate".

I tested this out with some display names containing non-ASCII characters and it seems to work; at least, when gmail received the messaged the headers were properly encoded as per [RFC 2047](https://tools.ietf.org/html/rfc2047).

One snag is that SimpleSmtp (or something that it depends on) does not like to see certain characters such as "<" in the display name, and encoding or escaping them does not seem to help. (Whereas according to my reading of [RFC 2822](https://tools.ietf.org/html/rfc2822), "<" ought to be allowed.) So I drop such characters.

### DIFF
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -546,6 +546,11 @@ limitations under the License.
     </div>
     {{#if completionState.clear}}
     <div class="button-container" role="presentation">
+      {{#with invitationExplanation}}
+        {{!-- Display the entire message in the hover text, in case "overflow: ellipsis" causes
+              part of the message to be hidden in the main view. --}}
+        <p class="invitation-explanation" title="{{.}}">{{.}}</p>
+      {{/with}}
       <button>Send</button>
     </div>
     {{/if}}
--- a/shell/server/grain-server.js
+++ b/shell/server/grain-server.js
@@ -392,8 +392,8 @@ Meteor.methods({
 
       const accountId = this.userId;
       const outerResult = { successes: [], failures: [] };
-      const envelopeFrom = globalDb.getReturnAddress();
-      const fromEmail = globalDb.getPrimaryEmailWithDisplayName(accountId, identityId);
+      const fromEmail = globalDb.getReturnAddressWithDisplayName(identityId);
+      const replyTo = globalDb.getPrimaryEmail(accountId, identityId);
       contacts.forEach(function (contact) {
         if (contact.isDefault && contact.profile.service === "email") {
           const emailAddress = contact.profile.intrinsicName;
@@ -411,7 +411,7 @@ Meteor.methods({
             SandstormEmail.send({
               to: emailAddress,
               from: fromEmail,
-              envelopeFrom: envelopeFrom,
+              replyTo: replyTo,
               subject: title + " - Invitation to collaborate",
               text: message.text + "\n\nFollow this link to open the shared grain:\n\n" + url +
                 "\n\nNote: If you forward this email to other people, they will be able to " +
@@ -452,7 +452,7 @@ Meteor.methods({
               SandstormEmail.send({
                 to: email.email,
                 from: fromEmail,
-                envelopeFrom: envelopeFrom,
+                replyTo: replyTo,
                 subject: title + " - Invitation to collaborate",
                 text: message.text + "\n\nFollow this link to open the shared grain:\n\n" + url +
                   "\n\nNote: You will need to log in with your " + loginNote +
@@ -505,8 +505,8 @@ Meteor.methods({
       const identity = globalDb.getIdentity(identityId);
       globalDb.addContact(grainOwner._id, identityId);
 
-      const envelopeFrom = globalDb.getReturnAddress();
-      const fromEmail = globalDb.getPrimaryEmailWithDisplayName(Meteor.userId(), identityId);
+      const fromEmail = globalDb.getReturnAddressWithDisplayName(identityId);
+      const replyTo = globalDb.getPrimaryEmail(Meteor.userId(), identityId);
 
       // TODO(soon): In the HTML version, we should display an identity card.
       let identityNote = "";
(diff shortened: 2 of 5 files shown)


zarvox:
As a privacy note, we should probably indicate which email address we're going to use as the From: address to the user, since we're now revealing one of their email addresses to the person they share with in a way we didn't before.

dwrensha:
There could be some text next to the send button, like this maybe:

invitefrom

cc @neynah

kentonv:
I think the note should probably be gray rather than red, unless @neynah disagrees. Similar to the note shown under the "open" button in the email body.

kentonv:
(I agree that we should add the note. Otherwise the change looks good.)

neynah:
@kentonv Yeah I think it should be grey(just because red is associated with errors & warnings) and left-aligned with the input box and icons.

What happens if the text is really long? If the email is long I think the full email should go on a second line instead of breaking it with a hyphen.

dwrensha:
I've implemented the note about revealing your email. I made the text black and small and italic, to match the explanatory note in the "Get shareable link" tab. It looks like this:

invite-from-actual

When the email is too long, the text gets cut off with `text-overflow: ellipsis`. The full text is in the `title` attribute, so it can be read on mouse-over.

I also updated the Send button and the Create button to use `%button-primary`.

dwrensha:
Garply, retest this please.

paulproteus:
A bunch of us did some analysis here: "Outbound email policy" <https://oasis.sandstorm.io/shared/jOtMuOe4V-lNj0FBYJby4mxnsxOaND-iaooPZVBlo4d>.

Implications for #1696 and #1719

- It's not generally safe (in the case of DMARC rules enforcing DKIM on From headers) to use the user's From address, and additionally, it's going to get eaten by GMail SMTP and rejected by Zoho.
- However, using the user's *name* in the From: is generally safe.
- Using the user's email address as reply-to is generally safe.
- In the mockups, show this message: "Invitation will be from david@sandstorm.io" rather than "Invitation will be sent from david@sandstorm.io"
    - Rationale: We're not literally using it as the "From:" header.
    - I got this string from an IRL discussion with @neynah and @zarvox. I don't think either care a great deal about changing the string, but it's nice if we can. The important thing we want to convey is that 
- We should continue using the server's email address as the envelope-from. However, due to DMARC+DKIM it's IMHO not super safe to use it as "From". Once we can affirmatively check if a domain's "From" header is using DMARC to enforce a DKIM policy that would make it OK for us to use that doman's "From" header, then I would prefer we switch to the "From" header, but I personally would like to see these land.
- This proposal is a blend of #1696 and #1719. I'm hopeful @kentonv will +1 this note.

kentonv:
OK, sounds like we should switch to Reply-To.

Are there other e-mails we currently send "From" the user? Request access, maybe? I guess those should be Reply-To instead.

dwrensha:
I've implemented the proposal outlined by @paulproteus.

Looks like Garply was a bit overeager to approve this branch. The js-style check ran, but not the integration tests yet, as far as I can tell.

jparyani:
Garply, retest this please

shell/client/shell.html
```diff
@@ -546,6 +546,11 @@ <h4 class="who-has-access">Who has access</h4>
     </div>
     {{#if completionState.clear}}
     <div class="button-container" role="presentation">
+      {{#with invitationExplanation}}
```
zarvox: This is a newer Spacebars feature, so we're far from reliably using it in the rest of the codebase, but a clearer way to write this would use the second syntax at https://github.com/meteor/meteor/blob/devel/packages/spacebars/README.md#with where you bind a name to that value, rather than hoisting the whole thing into your current data context.

```
{{#with explanation=invitationExplanation}}
<p class="invitation-explanation" title="{{explanation}}">{{explanation}}</p>
{{/with}}
```

shell/server/grain-server.js
```diff
@@ -557,8 +551,8 @@ Meteor.methods({
 
       SandstormEmail.send({
         to: emailAddress,
-        envelopeFrom: envelopeFrom,
```
zarvox: Might be nice to be retain the explicitness about what `envelopeFrom` we intend to use here, as the next pass of changes(picking a different From: address) would cause that to break if we didn't add the envelopeFrom back in.

Ditto the other `SandstormEmail.send` callsites.

shell/client/shell.html
zarvox: Argh, Meteor moved the docs.  Now it's just http://docs.meteor.com/#/full/pkg_spacebars and I can't link to the particular fragment I wanted to.

The Meteor guide recommends using `{{#let}}` instead of `{{#with}}`, so let's go with that: http://guide.meteor.com/blaze.html#builtin-block-helpers

kentonv:
Garply, retest this please

kentonv:
@zarvox's comments look like style tweaks. @dwrensha if you happen to get to them before I decide to do a release, great, otherwise I'll merge this tomorrow and we can tweak it later.

shell/client/shell.html
dwrensha: @zarvox: Both of your suggestions act subtly differently from the code that I've written. Note that `invitationExplanation` can be null (e.g. for dev users). In that case, `{{#with invitationExplanation}}<p>...</p>{{/with}}` does not render the `<p>` element at all --- it's as if the `{{#with}}` is wrapped in an `{{#if}}`. In contrast, with the variable binding that you suggest, the `<p>` is always rendered.

Note that we do use this pattern [elsewhere](https://github.com/sandstorm-io/sandstorm/blob/8a001ccf19cf59d9b484fd7f989d8e5e7e48be48/shell/packages/sandstorm-accounts-ui/account-settings.html#L191-L200).

kentonv:
Merging now for release.

Issue #708, opened 2015-08-22
libreboard not working


berndblume:
I just installed sandstorm on Ubuntu 15.04. Everything works like a charm, draw.io, etherpad, etc. all work fine.
I also installed libreboard. After opening a "New LibreBoard", I see the libreboard screen, and in the title bar it says "Untitled LibreBoard", and then the MB keep counting for minutes. The page is reloaded every second or so, and the MBs go up and up. In the demo, it does work - it stops at 1.6 MB or so. But on my own server: counting and counting.
You can try yourself at http://sandstorm.hau.to

kentonv:
That's strange! Anything interesting in the debug log (click the monitor icon in the top bar while the app is open)?

berndblume:
Hi Kanton,
Here's a screen shot of the log.
Thanks, Bernd

[image1.PNG]

-----
Sent from an iPad

On Aug 23, 2015, at 12:04 AM, Kenton Varda <notifications@github.com<mailto:notifications@github.com>> wrote:


That's strange! Anything interesting in the debug log (click the monitor icon in the top bar while the app is open)?

-

paulproteus:
Hi @berndblume ,

Any chance you could send me an invite to your server, via https://storm.debian.net/admin/invites ? I'm asheesh@sandstorm.io.

paulproteus:
Thanks! I received your invite and can test now!

paulproteus:
OK, I think I know what's going on. When I visit your server, and I try to visit a LibreBoard instance, I see this message in my browser's Javascript console:

> WebSocket is closed before the connection is established.

Also, the LibreBoard reloads itself in an infinite loop.

The infinite loop causes the LibreBoard to keep taking up slightly more disk space as it reloads.

@berndblume , my guess is that you need to configure your frontend web server (nginx? apache?) to proxy web sockets. https://github.com/sandstorm-io/sandstorm/blob/master/nginx-example.conf shows one way to do that for nginx; if you're not using nginx, I can perhaps help you by providing some sample config you can use.

https://github.com/sandstorm-io/sandstorm/issues/316 discusses this issue in more detail.

berndblume:
Hi Asheesh,

that might be it, indeed! :-)

I am using nginx as a reverse proxy, but I was unaware I have to take special action in order to have nginx proxy web sockets.

I will check the links you provided and will let you know if I was able to fix it or not.

Thanks, Bernd


[...]

Issue #504, opened 2015-06-15
Document how to set BASE_URL for apache2 with https


arthurlutz:
Hi,

I'm trying to setup sandstorm in a docker behind an apache web server. But I'm having trouble and cannot find any documentation about this. 

What should we put in the BASE_URL and WILDCARD_HOST ? In the initial version there is port information, can this be removed when using https ? 

I'm using : 
```
-BASE_URL=http://local.sandstorm.io:6080\n\
-WILDCARD_HOST=*.local.sandstorm.io:6080\n\
+BASE_URL=https://sandstorm.mydomain.im\n\
+WILDCARD_HOST=*.sandstorm.mydomain.im\n\
```
And I end up getting : 

```
Error looking up DNS TXT records for host '127.0.0.1': queryTxt ENOTFOUND

This Sandstorm server's main interface is at: https://sandstorm.mydomain.im 
```
Is this because the apache proxy makes sandstorm see the incoming connexion as 127.0.0.1 ?

arthurlutz:
Explaining a problem is a great way to think about it. I solved it by adding a ```ProxyPreserveHost On``` to the apache configuration!

Now I'm a bit stuck with the wildcard dns and StartSSL certificate. Any tips ? I'm probably going to try out sandcats.io as a temporary solution.

arthurlutz:
OK, using sandstorm over http for the moment to bypass the wildcard certificate problem (could find a way to add ssl exceptions on frames for each grain and that would get me going with https and a certificate exception for each grain).

Now am bumping into the websocket problem. Am stuck with apache for a number of reasons, is there a way to proxy websocket like you suggest with nginx ? I've activated mod_proxy_wstunnel https://httpd.apache.org/docs/2.4/mod/mod_proxy_wstunnel.html but I guess I'm missing some configuration for it to work (for example libreboard fails to load on ws error).

paulproteus:
Hi Arthur! Good to know you're experimenting with these things.

You could try setting the ProxyPass line not to http://something but more
like

ProxyPass ws://something

See e.g.
https://stackoverflow.com/questions/27526281/websockets-and-apache-proxy-how-to-configure-mod-proxy-wstunnel

I haven't tested this, so I hope you'll let me know if it works!

paulproteus:
Related:

http://blog.cafarelli.fr/2013/04/backporting-apache-support-for-websockets-reverse-proxy-aka-getting-gateone-to-work-behind-apache/
​
[...]

PR #1697, opened 2016-03-24
Provide a warning to admins when WebSockets are broken

paulproteus:
Refs #1694 #708 #504

### DIFF
--- a/shell/client/admin-client.js
+++ b/shell/client/admin-client.js
@@ -142,10 +142,10 @@ Template.admin.helpers({
     // so long as we have verified that the attributes we expect are in place.
     if (Meteor &&
         Meteor.connection &&
-	Meteor.connection._stream &&
-	Meteor.connection._stream.socket &&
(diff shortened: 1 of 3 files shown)


paulproteus:
Here's what it looks like when you get both warnings:

screenshot from 2016-03-23 19 04 30

If you want to imagine what it's like to get just this one warning, pretend it's the only one there!

paulproteus:
FWIW, the purpose of this pull request is to be a quick hack that people hopefully see. I would perhaps be happier with a fix that makes a real "Notification", but I think this is 100% worth merging as-is for now. cc: @jvasile

paulproteus:
@kentonv ready for your review/merge. Not necessarily urgent, but it's always easier to address feedback sooner rather than later.

kentonv:
Please let your editor know not to insert tabs. :)

I fixed it for this PR, so merging now.

PR #1701, opened 2016-03-24
eke out some more coverage in the incognito test

dwrensha:
Adds coverage for the functionality added in https://github.com/sandstorm-io/sandstorm/pull/1484.


dwrensha:
I incorporated this into #1702.

PR #1705, opened 2016-03-25
Fix typo in troubleshooting guide

azirbel:
Fixes a typo.

Before

> If possible, configure the app to use a base URL of '', literally the empty string. Then it will send HTTP redirects without specifying a base URL. If that isn't possible, Sandstorm apps should look at the
> 
> Sandstorm apps should look at the Host: header for a base URL if they need one.

After

> If possible, configure the app to use a base URL of '', literally the empty string. Then it will send HTTP redirects without specifying a base URL. If that isn't possible, Sandstorm apps should look at the `Host:` header for a base URL.

Also made the word-wrapping in this file consistent; let me know if you'd rather I not touch this, and I'll put it back.

### DIFF
--- a/docs/developing/troubleshooting.md
+++ b/docs/developing/troubleshooting.md
@@ -1,8 +1,13 @@
 This page lists common problems and work-arounds.
 
-This page needs expansion! If you encounter a new problem and solve it, feel free to add it here. If you encounter a problem you don't know how to solve, try asking on Stack Overflow, and be sure to add the "sandstorm.io" tag to your question. Or, talk to us on IRC (#sandstorm on freenode) or [sandstorm-dev](https://groups.google.com/group/sandstorm-dev).
+This page needs expansion! If you encounter a new problem and solve it, feel
+free to add it here. If you encounter a problem you don't know how to solve,
+try asking on Stack Overflow, and be sure to add the "sandstorm.io" tag to your
+question. Or, talk to us on IRC (#sandstorm on freenode) or
+[sandstorm-dev](https://groups.google.com/group/sandstorm-dev).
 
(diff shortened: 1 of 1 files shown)


garply:

docs/developing/troubleshooting.md
```diff
@@ -35,16 +40,20 @@ very likely you are seeing this error.
 If possible, configure the app to use a base URL of `''`, literally
 the empty string. Then it will send HTTP redirects without
 specifying a base URL. If that isn't possible, Sandstorm apps should
-look at the
+look at the `Host:` header for a base URL.
```
paulproteus: Funny thing is, they should actually use `X-Sandstorm-Base-Url` instead, since `Host` doesn't specify a URL scheme (http vs. https). I guess. I need to verify that it really is `X-Sandstorm-Base-Url` specifically, though. I'll merge this as-is, but consider testing/changing that too if you want.

azirbel: Haha, ok! I'll keep it in mind, though I don't think I can verify it myself yet. I have another couple levels of understanding to go through first.

Don't feel any pressure to merge this if it would be better to close it and make a more correct PR. That's fine too.

paulproteus:
Thanks @azirbel ! Great seeing you tonight.

docs/developing/troubleshooting.md
paulproteus: https://docs.sandstorm.io/en/latest/developing/troubleshooting/ is now up
to date. Thanks again!

kentonv:
FWIW the combination of `Host` and `X-Forwarded-Proto` form a more-standard alternative to `X-Sandstorm-Base-Url` and are perfectly reasonable for apps to rely on.

PR #1711, opened 2016-03-26
Added icon preview graphic to docs

### LINKED ISSUE #1564: docs: icon size info could be clearer, in app market publishing guide
@phildini had a problem where had received only one icon from @neynah and wasn't sure about the following things:

- Relative to what directory does the `embed` occur? (Answer: `.sandstorm/`)

- Will it follow symlinks? (Answer: Yes, I guess?)

- Which icons truly _need_ to be filled out? Does he need to specify image sizes for an SVG? (Answers: Just the first one. No.)

Also @neynah I think we have four image formats because we have different notions of how they should be styled, like different border-radius or similar. If that's true, we should write it down somewhere! Or maybe we already do write it down somewhere and I should link to it from the app market publishing docs?

neynah:
This partially fixes #1564. @paulproteus is going to re-host the image.

image

### DIFF
--- a/docs/developing/publishing-apps.md
+++ b/docs/developing/publishing-apps.md
@@ -131,6 +131,8 @@ Your app's manifest, or package definition file, (`sandstorm-pkgdef.capnp`) cont
 
 You can embed both SVGs or PNGs, and Sandstorm will use the best version provided for the use in question. Using PNGs requires a slightly different structure, which you can find an example of [here](https://github.com/dwrensha/sharelatex/blob/sandstorm-app/sandstorm-pkgdef.capnp#L26).
 
+<img src="https://inozs8hbcttx5fdveukc.oasis.sandstorm.io/docs/docs-package-icons.png"></img>
+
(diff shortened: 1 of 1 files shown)


docs/developing/publishing-apps.md
```diff
@@ -131,6 +131,8 @@ Your app's manifest, or package definition file, (`sandstorm-pkgdef.capnp`) cont
 
 You can embed both SVGs or PNGs, and Sandstorm will use the best version provided for the use in question. Using PNGs requires a slightly different structure, which you can find an example of [here](https://github.com/dwrensha/sharelatex/blob/sandstorm-app/sandstorm-pkgdef.capnp#L26).
 
+<img src="https://inozs8hbcttx5fdveukc.oasis.sandstorm.io/docs/docs-package-icons.png"></img>
```
paulproteus: If possible, I have some suggestions here:

- Use one image tag per image, so that it breaks apart (which IMHO it should) on narrow screens

- Use Markdown image syntax, like in docs/administering/faq.md

paulproteus:
One upside of Markdown image syntax is that it gives you the chance to add ALT text for accessibility.

@neynah is this something you're interested in doing? If not, I can merge this as-is instead; let me know.

PR #1652, opened 2016-03-15
Somewhat clarify the prod-vs-dev choice in the install script.

dwrensha:
This step always makes me pause, because I'm like "I want a full development server so that I can hack on Sandstorm itself -- which option does that fall under?"

PR #1712, opened 2016-03-26
Clarify that dev install can be used for Sandstorm dev

paulproteus:
Hopefully this addresses the concern that resulted in #1652

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -635,8 +635,8 @@ choose_install_mode() {
   if [ -z "${CHOSEN_INSTALL_MODE:-}" ]; then
     echo "Sandstorm makes it easy to run web apps on your own server. You can have:"
     echo ""
-    echo "1. A full server with automatic setup (press enter to accept this default)"
-    echo "2. A development server, for writing apps."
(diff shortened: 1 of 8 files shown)


paulproteus:
I'll probably have to update installer-tests; will do that in a commit on top. Curious for @dwrensha 's +1 -1 on this text.

dwrensha:
+1

I like "typical install" better than "full install".

It is a bit awkward to use the word "development" twice in option 2, but I can't think of any improvements that also keep the "localhost-based" bit.

kentonv:
Is the word "live" a possible alternative to "production"? "A live server"...

Just an idea, but I leave it to @paulproteus to decide.

paulproteus:
FWIW, I'm -0.5 on "live". I don't know that it adds any information beyond
"typical".

I'll land this tomorrow, after double-checking with installer-tests.

paulproteus:
Self-merging because 100% install.sh and installer-tests.

The installer-tests pass on my laptop.

PR #1714, opened 2016-03-27
Fix bug that prevented ldapNameField from getting set.

dwrensha:


dwrensha:
For reference, the "ldapSearchUsername" value gets used [here](https://github.com/sandstorm-io/sandstorm/blob/205d931b444f72e438a5e34b8f0362d9601591cd/shell/client/admin-client.js#L261).

PR #1715, opened 2016-03-27
Implement email verification via Powerbox

kentonv:


shell/packages/sandstorm-db/db.js
```diff
@@ -415,9 +423,17 @@ ApiTokens = new Mongo.Collection("apiTokens");
...
-//        ipInterface :Bool;
+//       notificationHandle :Text;
+//       ipNetwork :Bool;
+//       ipInterface :Bool;
+//       emailVerifier :group {
+//         services :List(String);
+//       }
+//       verifiedEmail :group {
+//         verifierId :Text;
+//         tabId :Text;
```
zarvox: If I'm reading this correctly, this means that if your app requests a `VerifiedEmail`, and then the user terminates that session and comes back later, they're going to have a different `tabId`, so the VerifiedEmail won't verify if passed to `EmailVerifier` any more.

This basically makes `save()` and `restore()` on `VerifiedEmail` useless for apps.  Is the idea that the app would request `VerifiedEmailSendPort` once, verify the email immediately before saving it durably, and then only ever call `EmailSendPort`'s methods on the capability you received thereafter?

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
```diff
@@ -184,7 +235,65 @@ specialCaseTypes[Ip.IpInterface.typeId] = function (db, userId, value) {
...
+      services = {};
+      verifierInfo.services.forEach(service => services[service] = true);
+    }
+  }
+
+  const user = Meteor.users.findOne(userId);
+  const emails = {};  // map address -> true, for uniquification
+  Meteor.users.find({_id: {$in: SandstormDb.getUserIdentityIds(user)}}).forEach(identity => {
+    if (!services || services[identity.profile.service]) {
+      SandstormDb.getVerifiedEmails(identity).forEach(email => emails[email.email] = true);
```
zarvox: Consider wrapping the `email[email.email] = true` statement in a block, to make it clearer you don't actually care about the return value, but rather a side effect.

Consider writing the whole thing in pure-functional style with `_.chain()`:

```javascript
return _.chain(Meteor.user.find({_id: {$in: SandstormDb.getUserIdentityIds(user)}}).fetch())
  .filter(identity => !services || services[identity.profile.service])
  .map(identity => SandstormDb.getVerifiedEmails(identity).map(email -> email.email))
  .flatten()
  .uniq()
  .value();
```

shell/packages/sandstorm-db/db.js
```diff
@@ -368,6 +368,14 @@ ApiTokens = new Mongo.Collection("apiTokens");
 //                  database when that happens. This field contains the boolean true to signify that
 //                  it has been set.
 //       ipInterface: Ditto IpNetwork, except it's an IpInterface.
+//       emailVerifier: An EmailVerifier capability that is implemented by the frontend. The
+//                      value is an object containing the field `services`, which itself is a
+//                      list of names of identity providers that are trusted to verify addresses.
+//                      If `services` is omitted or falsy, all configured identity providers are
```
zarvox: I might rename `services` to `restrictToServices` or `filterServices` to make it clearer in the consuming code that this is an optional filter, not an exhaustive list.  Not a blocker, though.

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
```diff
@@ -41,28 +44,61 @@ Meteor.methods({
...
 
       descriptor = encodePowerboxDescriptor({ tags: [{ id: Ip.IpNetwork.typeId }] });
+      requirements.push({ userIsAdmin: Meteor.userId() });
     } else if (frontendRefVariety.ipInterface) {
       if (!db.isAdmin(this.userId)) {
         throw new Meteor.Error(403, "User must be an admin to powerbox offer IpInterface");
       }
 
       descriptor = encodePowerboxDescriptor({ tags: [{ id: Ip.IpInterface.typeId }] });
+      requirements.push({ userIsAdmin: Meteor.userId() });
```
zarvox: As a matter of practicality, I'm not sure you want this requirement, given the intended use case of IpNetwork and IpInterface for drivers:

1. Most of the time, admins won't have their admin privileges revoked.
2. We have UI in the admin panel for admins to review and revoke all IpNetwork and IpInterface sturydyrefs.
3. If an admin leaves the company or changes roles and no longer needs admin privileges, this requirement necessitates urgently reissuing the capability to any drivers they may have granted the privilege to.
4. I assume that most users will not actually feel a need to reissue the capabilities that that admin initially set up.
5. We currently lack a system for methodically reissuing such capabilities, and building an automated one will require explicit support from apps/drivers.

I suspect that in practice, having IpNetwork/IpInterface auto-revoke when the admin becomes a non-admin is likely to be **much** more surprising and damaging to real deployments than letting them persist.

The way I'd been thinking about things, IpNetwork and IpInterface are owned by the deployment, not the admin that granted them, and admins have an interface by which they may ask the deployment to create/modify/delete them.  Maybe that's not quite the pure ocap model, but it is how I'd been modeling the problem space, especially given that admins have/need the power to review/revoke each other's IpNetwork/IpInterface grants.

shell/packages/sandstorm-ui-powerbox/powerbox-client.js
```diff
@@ -161,6 +161,28 @@ SandstormPowerboxRequest = class SandstormPowerboxRequest {
           cardData.title = "Admin: grant all outgoing network access";
         } else if (cardData.frontendRef.ipInterface) {
           cardData.title = "Admin: grant all incoming network access";
+        } else if (cardData.frontendRef.emailVerifier) {
+          const services = cardData.frontendRef.emailVerifier.services;
+          if (services) {
+            const name = services[0];
```
zarvox: Weird that this is a list, but we only ever look at the first element of said list.  Does this make more sense as an optional String instead?

zarvox:
Solid-looking implementation.  I am excited to see this happening!

* One workflow question about email verification which can probably be answered with "Yeah, that's the intent, deal with it" which might benefit from some inline comments about intended workflow for whoever writes developer docs on this later.
* One policy question about IpNetwork/IpInterface being tied to whether the issuer retains admin privileges or not
* A couple minor style/naming/clarity comments
* Fix the JSCS complaints :)

shell/packages/sandstorm-db/db.js
kentonv: If the app wants to persist a `VerifiedEmail`, it would need to store the `tabId` that the user had at the time it was requested. Remember that the point here is to ensure that the verification occurred in the `tabId` that the app thought it did. That said, I don't think apps would commonly store a `VerifiedEmail` long-term; once the verification handshake has completed, they may as well store the address itself.

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
kentonv: This requirement is not new. If it's wrong, we should remove it in a separate change. (But I'd argue, as usual, that what we really want is some sort of UI where when an admin is revoked, the revoker can audit what connections they made and choose whether or not to revoke them.)

shell/packages/sandstorm-ui-powerbox/powerbox-client.js
kentonv: It's somewhat of a temporary UI restriction that we don't currently support selecting multiple services. E.g. I might want to trust Google and email but not Github, or something. Perhaps in the future this can be accomplished through whatever multi-select UI we come up with for the powerbox, or perhaps what I really should have done here is made the first selection be "verify based on login provider" which would then have a custom-UI second step where the user chooses providers. I decided it wasn't worth doing now but figured there was no harm in making the underlying representation support it, so that we can easily twiddle the UI later.

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
kentonv: Agreed with the braces.

`_.chain()` is neat but I'm sad that Underscore's implementation of uniq() is quadratic-time. (They can't assume that the elements are strings, therefore they can't assume that they can be hash keys, therefore they just give up.) It also bothers me that Underscore in general adds all kinds of dynamic type checking tests and branches... :/

shell/packages/sandstorm-db/db.js
kentonv: I agree that naming would be slightly better but the difference isn't enough that I want to do the search-and-replace. :) (Wish Javascript had refactoring IDEs...)

kentonv:
Thanks for the quick review on a weekend. I'm going to merge this in the hopes that if it goes out tonight, everything will be in place for the feature key vendor app to launch whenever it's done (it's close!).

PR #1719, opened 2016-03-28
In grain-invite emails, fill in the "from" header with the server's configured Return Address.

### LINKED ISSUE #1717: Use Return-address in SMTP configuration in From-Header for sharing
Hi and thanks for the great work.

I have Sandstorm using my own mail server on another machine. Sandstorm correctly authorizes via SMTP and sending mails in general works. However, Sandstorm uses a sandcats domain and tries to send invitations from no-reply@...sandcats.io - my mail-server rejects these message, because it is not responsible for this domain. (Even if I configured it to allow sending mails from there, other servers would probably mark mails from this address send by my server as spam.)

I'd expect sandstorm to use the Return-address specified in the SMTP configuration.

Is there another way to specify the address used by sandstorm?

dwrensha:
An alternative to #1696. Fixes #1717. See [this sandstorm-dev thread](https://groups.google.com/forum/#!topic/sandstorm-dev/qSDba86W7xI) for another user who ran into similar issues.

This sends all "request access" and "grain invitation" emails from the server's configured value for "Return Address". Our original plan was to use the server's Return Address for the bounce address and then to use the sending user's actual verified email address for the inner "from" header. There are two problems with that plan:

  1. Some people want to use existing SMTP setups that refuse to send things with arbitrary values in the "from" header.

  2. Writing arbitrary values in the "from" header will probably make the emails likely to end up in a spam folder. My understanding is that the reason that Google Docs can get away with sending invites that come from the inviter's gmail address (rather than `invitations@googledocs.com` or something) is that Google controls both pieces, and therefore can do whatever DKIM signing is necessary. Once Sandstorm has a more complete email setup, we should be capable of something similar, but until then I don't see how the recipients of these emails are supposed to verify that we're not just spoofing the "from" header. I mean, we basically would be spoofing it, albeit with informal permission from the inviter.

It possible that I'm missing something important in my analysis. Maybe someone with a better understanding of DKIM and such can explain what I'm getting wrong?

### DIFF
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -1331,6 +1331,17 @@ _.extend(SandstormDb.prototype, {
     const setting = Settings.findOne({ _id: "organizationLdap" });
     return setting ? setting.value : false;
   },
+
+  getReturnAddressWithDisplayName: function (identityId) {
+    check(identityId, String);
+    const identity = this.getIdentity(identityId);
+
+    // First remove any instances of characters that cause trouble for SimpleSmtp. Ideally,
+    // we could escape such characters with a backslash, but that does not seem to help here.
+    const sanitized = identity.profile.name.replace(/"|<|>|\\|\r/g, "");
+    const displayName = sanitized + " (via " + this.getServerTitle() + ")";
+    return "\"" + displayName + "\" <" + this.getReturnAddress() + ">";
+  },
 });
 
 SandstormDb.escapeMongoKey = (key) => {
--- a/shell/server/grain-server.js
+++ b/shell/server/grain-server.js
@@ -388,9 +388,8 @@ Meteor.methods({
       }
 
       const accountId = this.userId;
-      const identity = globalDb.getIdentity(identityId);
-      const sharerDisplayName = identity.profile.name;
       const outerResult = { successes: [], failures: [] };
+      const fromEmail = globalDb.getReturnAddressWithDisplayName(identityId);
       contacts.forEach(function (contact) {
         if (contact.isDefault && contact.profile.service === "email") {
           const emailAddress = contact.profile.intrinsicName;
@@ -407,8 +406,8 @@ Meteor.methods({
           try {
             SandstormEmail.send({
               to: emailAddress,
-              from: "Sandstorm server <no-reply@" + HOSTNAME + ">",
-              subject: sharerDisplayName + " has invited you to join a grain: " + title,
+              from: fromEmail,
+              subject: title + " - Invitation to collaborate",
               text: message.text + "\n\nFollow this link to open the shared grain:\n\n" + url +
                 "\n\nNote: If you forward this email to other people, they will be able to " +
                 "access the share as well. To prevent this, remove the link before forwarding.",
@@ -447,8 +446,8 @@ Meteor.methods({
                   " to access this grain.";
               SandstormEmail.send({
                 to: email.email,
-                from: "Sandstorm server <no-reply@" + HOSTNAME + ">",
-                subject: sharerDisplayName + " has invited you to join a grain: " + title,
+                from: fromEmail,
+                subject: title + " - Invitation to collaborate",
                 text: message.text + "\n\nFollow this link to open the shared grain:\n\n" + url +
                   "\n\nNote: You will need to log in with your " + loginNote +
                   " to access this grain.",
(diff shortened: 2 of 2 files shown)


hannesrauhe:
Your assumption is right. Usually there are three things other mail servers check when receiving e-mail: 

1. The public key provided by the DNS for the domain matches the mail-signature.
2. The sendings servers IP matches the one provided by the domain's SPF-record
3. There reverse DNS entry of the sending IP points to the domain.

So 1. and 2. can probably be set correctly for sandcats.io subdomains when creating the certificate.
3 can usually only be set by the ISP that hosts the machine sandstorm is running on. Unfortunately, this is also the most important setting, some mail servers just reject mail if the reverse DNS entry is missing.
2 is also important, usually mail ends up in spam folders if the SPF record is not set or wrong. DKIM adds to the spam score, but is not that important. (based on my personal experiments, maybe others do see different priorities.)

dwrensha:
@hannesrauhe: in your setup, do you need both the bounce address and the inner `From` field to be set to the configured Return Address?

To be more concrete: suppose a Sandstorm server is at `https://foo.sandcats.io`, its configured Return Address is `sandstorm-mailer@example.com`, and Alice, whose confirmed email address is `alice@alice.biz`, is sharing a grain with Bob.

In today's implementation, Alice's invitation email will be sent from `no-reply@foo.sandcats.io`. That is, both the `From` field and the bounce address will be set to `no-reply@foo.sandcats.io`. I think we all agree that the bounce address ought to be `sandstorm-mailer@example.com` instead, which I believe would allow @hannesrauhe's points (2) and (3) to be taken care of.

But what about the inner `From` field? It would maybe be nice to set it to `alice@alice.biz`, as I do in #1696, but then it seems like I have no way to take care of @hannesrauhe's point (1). Therefore, the current pull request proposes using `sandstorm-mailer@example.com` for the inner `From` field.

kentonv:
This is not correct AFAIK. Setting the envelope sender correctly should allow you to set any "from" address you want, and if the receiver doesn't trust you, it will display the message as being from e.g. "Bob via sandstorm.io".

If this weren't true, all mailing lists would not work. Mailing lists by and large leave the "From" header unchanged but set the envelope sender to the mailing list.

DKIM, etc. should be triggered off the envelope sender, not the From header.

Note that services like Sendgrid (or Mailgun, or Mandrill) automatically replace the envelope sender with their own address, but do not change the "From". Sendgrid deals with making sure you pass all the usual spam checks. The only problem for sandcats is that we currently don't provide a way for people to add the necessary DNS entries for Sendgrid to do its job, so we ought to fix that.

hannesrauhe:
I'd be happy to try it out with my setup in the next couple of days, up until now I haven't played around with envelope sender, especially not from different domains, but @kentonv's statement makes sense.

hannesrauhe:
I've made some experiments and can confirm what @kentonv said. The "From:" header can contain any address as long as the envelop sender (`MAIL FROM:`) is the one given in the SMTP configuration.
I tried this with my own mail server as well as with some commercial SMTP servers.

Interestingly Thunderbird on the receiving side only shows the "From": header without giving any hint about the "real" sender which can be found in the first line of the source:

`Return-Path: <sandstorm-mailer@example.com>`

GMail shows some information as described [here](https://support.google.com/mail/answer/1311182?hl=en).

dwrensha:
Thanks for investigating, @hannesrauhe. After a closer look at [RFC 6376](https://tools.ietf.org/html/rfc6376) I now favor #1696.

hannesrauhe:
Agreed!

paulproteus:
A bunch of us did some analysis here: "Outbound email policy" <https://oasis.sandstorm.io/shared/jOtMuOe4V-lNj0FBYJby4mxnsxOaND-iaooPZVBlo4d>.

Implications for #1696 and #1719

- It's not generally safe (in the case of DMARC rules enforcing DKIM on From headers) to use the user's From address, and additionally, it's going to get eaten by GMail SMTP and rejected by Zoho.
- However, using the user's *name* in the From: is generally safe.
- Using the user's email address as reply-to is generally safe.
- In the mockups, show this message: "Invitation will be from david@sandstorm.io" rather than "Invitation will be sent from david@sandstorm.io"
    - Rationale: We're not literally using it as the "From:" header.
    - I got this string from an IRL discussion with @neynah and @zarvox. I don't think either care a great deal about changing the string, but it's nice if we can. The important thing we want to convey is that 
- We should continue using the server's email address as the envelope-from. However, due to DMARC+DKIM it's IMHO not super safe to use it as "From". Once we can affirmatively check if a domain's "From" header is using DMARC to enforce a DKIM policy that would make it OK for us to use that doman's "From" header, then I would prefer we switch to the "From" header, but I personally would like to see these land.
- This proposal is a blend of #1696 and #1719. I'm hopeful @kentonv will +1 this note.

dwrensha:
Closing in favor of #1696.

PR #1721, opened 2016-03-29
docs: Stop referring to Mailpile

paulproteus:
Thanks for noticing, @garrison !


paulproteus:

This may take until tomorrow to show up live on docs.sandstorm.io.

PR #1722, opened 2016-03-29
Fix 1711

paulproteus:
This merges the commit by @neynah and also adds a few other docs fixes on top, as well as copying her image into my Davros.


paulproteus:

PR #1726, opened 2016-03-29
Fix mobile popup positioning

### LINKED ISSUE #1725: sharing popup is misaligned in mobile view
Half of the popup is offscreen, so it is basically unusable. It looks like this:

offscreen

It looks like this problem was introduced in https://github.com/sandstorm-io/sandstorm/pull/1637.

zarvox:
We rely on `position: static` on mobile to ignore the `left: $41px` or `right:
0px` inline styles which we compute to place the popup near its controlling
element for desktop.

Since we benefit from using `position: relative` for the contents of the frame,
split this into two nested divs:

`div.frame-container` is responsible for the placement and size of the popup as
well as border and drop-shadow.

`div.frame` is responsible for including all of the interesting content of that
popup, which can be `position: relative` to allow for absolutely-positioned
items within the frame.

Fixes #1725.

### DIFF
--- a/shell/client/styles/_account.scss
+++ b/shell/client/styles/_account.scss
@@ -1,4 +1,4 @@
-body>.popup.login>.frame {
+body>.popup.login>.frame-container>.frame {
   width: 200px;
   border: none;
   min-width: 200px;
@@ -15,7 +15,7 @@ body>.popup.login>.frame {
   }
 }
 
-body>.popup.account>.frame {
+body>.popup.account>.frame-container>.frame {
   width: 250px;
   border: none;
 
@@ -1093,7 +1093,7 @@ body>.popup,
   }
 }
 
-body>.popup.billing-prompt>.frame {
+body>.popup.billing-prompt>.frame-container>.frame {
   overflow-y: auto;
   -webkit-overflow-scrolling: touch;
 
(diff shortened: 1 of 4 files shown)


dwrensha:
This appears to fix the mobile view, but makes the popups very wide in the desktop view:
wide

zarvox:
Yeah, this makes it so the `<input type=text>` and the `<select>` don't have to awkwardly stack if they don't have to, and makes more space available to the who-has-access table.

We can add the `max-width: 50%` back in, or just for that field, but I did make that change intentionally.  There's not a whole lot of benefit to us showing the user a popup which blocks the primary content and all interaction with the grain frame, but then constraining that popup's size to be some arbitrary fraction of the screen size.  In my mind, it's better to let the popup take as much space as it can reasonably make use of.  In particular, powerbox interactions with nested grains will likely want to take up as much of the screen as can be claimed without making it impossible to dismiss the powerbox.

Now, if the criticism is "well, this form doesn't get better with more space", then I can probably be swayed to add a cap back for this form.  What would you say the appropriate max-width for this particular popup is?

dwrensha:
Hm. The change caught me by surprise, but I suppose I don't have any strong arguments against giving the popups more space.

It is a bit weird that the sharing popup can change size when I switch between its two tabs.

Also, it looks like some awkward stacking can still occur:
stacking

kentonv:
I agree that it's weird for the sharing dialog's width to be variable depending on its content, and for the different tabs to have different widths.

I also think that very wide textareas are not great for usability.

Can we go back to having a well-defined width of some sort? It doesn't necessarily need to be 50%.

PR #1727, opened 2016-03-30
Restore previous Share Access popup width

zarvox:
In response to comments on #1726.


paulproteus:
Garply, retest this please

paulproteus:
The build machine was running low on disk, but I believe it should be OK now.

PR #1728, opened 2016-03-30
Fix installer tests

paulproteus:


paulproteus:

PR #1734, opened 2016-03-31
Fix up more popup styling

zarvox:
This also fixes a bug in the powerbox popup where at small screen heights
content would overflow the .frame-container by using `max-height: inherit`.

This also fixes an issue where I wrote `flex-begin` rather than `flex-start` in
four places.  It would be nice if we had a CSS/SCSS linter.

### DIFF
--- a/shell/client/styles/_account.scss
+++ b/shell/client/styles/_account.scss
@@ -1,33 +1,39 @@
-body>.popup.login>.frame-container>.frame {
-  width: 200px;
-  border: none;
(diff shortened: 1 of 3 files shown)


paulproteus:
+1 to CSS/SCSS linter.

zarvox:
I'm going to wager the build failure is not related to these style changes.

zarvox:
Garply, retest this please

jparyani:
Garply, retest this please

PR #1735, opened 2016-03-31
Consolidate organization membership settings

zarvox:
Reduces client-server/DB roundtrips when checking org membership and makes for more straightforward settings code.  The setup wizard uses this merged settings layout.

@jparyani I'd appreciate a sanity check on this.

### DIFF
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -1011,18 +1011,19 @@ _.extend(SandstormDb.prototype, {
     }
 
     const orgMembership = this.getOrganizationMembership();
-    const googleEnabled = orgMembership.google.enabled;
-    const googleDomain = orgMembership.google.domain;
-    const emailEnabled = orgMembership.emailToken.enabled;
-    const emailDomain = orgMembership.emailToken.domain;
-
+    const googleEnabled = orgMembership && orgMembership.google && orgMembership.google.enabled;
+    const googleDomain = orgMembership && orgMembership.google && orgMembership.google.domain;
+    const emailEnabled = orgMembership && orgMembership.emailToken && orgMembership.emailToken.enabled;
+    const emailDomain = orgMembership && orgMembership.emailToken && orgMembership.emailToken.domain;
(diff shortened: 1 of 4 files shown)


jparyani:
Garply, retest this please

jparyani:
LGTM assuming Garply passes it

shell/packages/sandstorm-db/db.js
```diff
@@ -1010,20 +1010,22 @@ _.extend(SandstormDb.prototype, {
       return false;
     }
 
-    const googleDomain = this.getOrganizationGoogle();
-    const emailDomain = this.getOrganizationEmail();
+    const orgMembership = this.getOrganizationMembership();
+    const googleEnabled = orgMembership.google.enabled;
```
kentonv: It looks like `getOrganizationMembership()` could return `undefined` in which case this will throw.

shell/packages/sandstorm-db/migrations.js
```diff
@@ -496,6 +496,36 @@ function smtpPortShouldBeNumber() {
   }
 }
 
+function consolidateOrgSettings() {
+  const orgGoogleDomain = Settings.findOne({ _id: "organizationGoogle" });
+  const orgEmailDomain = Settings.findOne({ _id: "organizationEmail" });
+  const orgLdap = Settings.findOne({ _id: "organizationLdap" });
+  const orgSaml = Settings.findOne({ _id: "organizationSaml" });
```
kentonv: If all the above returned `undefined` then should we skip the rest of the function?

shell/packages/sandstorm-db/db.js
zarvox: Strictly speaking, the migration guarantees that `orgMembership` will exist and be in a well-defined structure.  But if you prefer to skip the migration if none of the org settings are set yet as your comment suggests below, then I suppose I'd need to guard here.

Another option would be to have `SandstormDb.prototype.getOrganizationMembership` guarantee that it returns a populated structure with the usual default values.

zarvox:
Rebased.

zarvox:
There's a bit of a backlog on Jenkins though.

PR #1739, opened 2016-04-01
Login improvements

kentonv:
See individual commits for details.

### DIFF
--- a/shell/packages/sandstorm-db/profile.js
+++ b/shell/packages/sandstorm-db/profile.js
@@ -211,11 +211,11 @@ SandstormDb.fillInProfileDefaults = function (user) {
   } else if (profile.service === "ldap") {
     const setting = Settings.findOne({ _id: "ldapNameField" });
     const key = setting ? setting.value : "";
-    profile.name = profile.name || user.services.ldap.rawAttrs[key] || "Name Unknown";
     profile.handle = profile.handle || user.services.ldap.username;
+    profile.name = profile.name || user.services.ldap.rawAttrs[key] || profile.handle;
   } else if (profile.service === "saml") {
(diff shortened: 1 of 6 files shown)


shell/packages/sandstorm-db/profile.js
```diff
@@ -211,10 +212,10 @@ SandstormDb.fillInProfileDefaults = function (user) {
     const setting = Settings.findOne({ _id: "ldapNameField" });
     const key = setting ? setting.value : "";
     profile.name = profile.name || user.services.ldap.rawAttrs[key] || "Name Unknown";
-    profile.handle = profile.handle || filterHandle(profile.name);
+    profile.handle = profile.handle || user.services.ldap.username;
   } else if (profile.service === "saml") {
     profile.name = profile.name || user.services.saml.displayName || "Name Unknown";
-    profile.handle = profile.handle || filterHandle(profile.name);
+    profile.handle = profile.handle || user.services.saml.email.split("@")[0];
```
jparyani: This should use the `emailToHandle` function that's already defined elsewhere in this file.

shell/packages/sandstorm-db/profile.js
```diff
@@ -211,10 +212,10 @@ SandstormDb.fillInProfileDefaults = function (user) {
     const setting = Settings.findOne({ _id: "ldapNameField" });
     const key = setting ? setting.value : "";
     profile.name = profile.name || user.services.ldap.rawAttrs[key] || "Name Unknown";
```
jparyani: Perhaps instead of "Name Unknown", we should use `username` here.

jparyani:
LGTM other than a couple of minor comments.

PR #1740, opened 2016-04-01
Update Asheesh release-certificate for April 2016

paulproteus:


paulproteus:
Self-merging because I own my certificate, I suppose.

PR #1742, opened 2016-04-01
Add ldap email field

jparyani:
"mail" is by convention usually where this field resides (as part of the
inetOrgPerson objectClass). It could theoretically be in a different
field though.


shell/client/shell.html
```diff
@@ -1072,7 +1072,9 @@ <h4 class="share-with-others">Share with others</h4>
         <label>LDAP Search Bind Password: (optional, needed if specifying bind Dn above): <input type="password" name="ldapSearchBindPassword" value="{{ldapSearchBindPassword}}" ></label>
         <label>LDAP Search Filter (optional, see <a href="http://www.ietf.org/rfc/rfc2254.txt">RFC2254</a>): <input id="ldapFilter" type="text" name="ldapFilter" value="{{ldapFilter}}" placeholder="eg. (&(email=*@bar.com)(!(ou=students))"></label>
         {{/if}}
-        <label>LDAP Name Field: <input id="ldapNameField" type="text" name="ldapNameField" value="{{ldapNameField}}"></label></p>
+        <label>LDAP Name Field: <input id="ldapNameField" type="text" name="ldapNameField" value="{{ldapNameField}}"></label>
```
kentonv: While we're at it, can we call this "LDAP Given Name Field" to be a little clearer?

jparyani:
Done

PR #1747, opened 2016-04-01
Make install succeed when /usr/local/bin is unwritable

### LINKED ISSUE #1660: Installer: Handle non-writable /usr/local
Trying to come up with the ultimate, hands-off, don't-have-to-touch-it Sandstorm installation for organizations with no or little tech competency. I think CoreOS and its automatic security updates might make a great choice since both the OS and Sandstorm could upgrade themselves without interaction.

Only, unfortunately /usr/local/bin isn't writable under CoreOS--or, at least, the installer fails trying to write the /usr/local/bin symlink. Would it be possible for the installer to either a) skip creating the /usr/local/bin symlink if writing fails, or b) put it in /opt/sandstorm/bin? I don't know if the systemd unit file accesses the binary directly, but I'm pretty sure /etc/systemd/system is writable so that much should work, and the unit could be updated to always access it directly rather than via the symlink.

Thanks.

paulproteus:
Close #1660

Note that this prints an ugly warning when this occurs. If we want smarter handling, someone can file a new bug. I think this addresses @ndarilek's need that `install.sh` should succeed on CoreOS.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1360,9 +1360,24 @@ install_sandstorm_symlinks() {
     return
   fi
 
+  local FAILED_TO_WRITE_SYMLINK="no"
+
   # Install tools.
-  ln -sfT $PWD/sandstorm /usr/local/bin/sandstorm
-  ln -sfT $PWD/sandstorm /usr/local/bin/spk
+  ln -sfT $PWD/sandstorm /usr/local/bin/sandstorm || FAILED_TO_WRITE_SYMLINK=yes
+  ln -sfT $PWD/sandstorm /usr/local/bin/spk || FAILED_TO_WRITE_SYMLINK=yes
+
+  # If /usr/local/bin is not actually writeable, even though we are root, then bail on this for now.
+  # That can happen on e.g. CoreOS; see https://github.com/sandstorm-io/sandstorm/issues/1660
+  # the bash "-w" does not detect read-only mounts, so we use a behavior check above.
+  if [ "${FAILED_TO_WRITE_SYMLINK}" = "yes" ] ; then
+    echo ""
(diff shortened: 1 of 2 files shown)


paulproteus:
Self-merging as this is is 100% install.sh & installer-tests/.

install.sh
```diff
@@ -1360,9 +1360,24 @@ install_sandstorm_symlinks() {
...
+  ln -sfT $PWD/sandstorm /usr/local/bin/spk || FAILED_TO_WRITE_SYMLINK=yes
+
+  # If /usr/local/bin is not actually writeable, even though we are root, then bail on this for now.
+  # That can happen on e.g. CoreOS; see https://github.com/sandstorm-io/sandstorm/issues/1660
+  # the bash "-w" does not detect read-only mounts, so we use a behavior check above.
+  if [ "${FAILED_TO_WRITE_SYMLINK}" = "yes" ] ; then
+    echo ""
+    echo "*** WARNING: /usr/local/bin was not writeable. To run sandstorm or spk manually, use:"
+    echo " - $PWD/sandstorm"
+    echo " - $PWD/spk"
```
kentonv: This is incorrect: There is no `/opt/sandstorm/spk`. Instead, users would need to type `/opt/sandstorm/sandstorm spk`.

paulproteus: Right, thanks. Although FWIW, there is a

/opt/sandstorm/latest/bin/spk


Maybe I'll suggest that instead.

kentonv: > Maybe I'll suggest that instead.

I'd actually prefer you not as I think that symlink was intended to be a legacy compatibility thing.

paulproteus: OK, great! (-:​

PR #1748, opened 2016-04-02
SAML tweaks

kentonv:
See individual commits.


kentonv:
These are minor so I'm self-merging but please look over them when you have a moment.

PR #1750, opened 2016-04-02
install.sh: Fix sandstorm path; add hash -r

paulproteus:
This is a follow-up pull request to #1747


paulproteus:
Self-merging as this is 100% install.sh & installer-tests.

PR #1752, opened 2016-04-03
Remove assertion that Sandstorm shows up on $PATH

paulproteus:
This will hopefully make installer-tests be green.


paulproteus:

PR #1753, opened 2016-04-03
Remove ip_tables red herring warning

paulproteus:
Note that this does not remove the underlying code. I can do that if we
really want, but for what it is worth, the warning causes much more heartache
than the unused code.


paulproteus:
@kentonv requesting your review/merge.

PR #1757, opened 2016-04-04
Rearrange some symlinks and add version.h to get libsodium to build.

dwrensha:
This makes the directory struct more closely match the source libsodium repo, so that [these relative includes](https://github.com/jedisct1/libsodium/commit/c3b68c12d60b92d7344f75936c00c928db815f5c) have a chance of working.

The new symlink to the `sodium/` directory requires that we have provide a `version.h`, which is usually generated at configure-time from `version.h.in`.

### DIFF
--- a/src/sodium/crypto_core_hchacha20.h
+++ b/src/sodium/crypto_core_hchacha20.h
@@ -1,0 +1,1 @@
+../../deps/libsodium/src/libsodium/include/sodium/crypto_core_hchacha20.h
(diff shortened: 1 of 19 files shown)


dwrensha:
Another option for dealing with version.h might be to make it a symlink to https://github.com/jedisct1/libsodium/blob/master/builds/msvc/version.h

kentonv:
Superseded by #1758

PR #1761, opened 2016-04-05
Setup wizard

zarvox:
To explain some of the unusual code and layout: this changeset started out as an attempt to build the new admin UI, and switched to being the setup wizard partway through.  A significant amount of code for the former was reusable for the latter; rather than try to separate out the two changes, I include some unrelated unlinked routes (`/admin-new/*`) and supporting code (`shell/client/admin/*`).  There are some things that will need to be reworked depending on context (setup wizard vs admin settings).

We're going to need the new admin panel code and layouts in the not-so-distant future anyway, and I didn't really want to deal with figuring out exactly what depends on what, pulling the pieces apart, only to merge them back together later.

Since there are a bunch of UI states that we didn't fully consider in the design phase, I had to wing some of the copy, which I am happy to replace with better text.

I need sleep now, but I'll post screenshots in the morning.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1526,7 +1526,7 @@ print_success() {
     echo "Visit this link to configure it:"
   fi
   echo ""
-  echo "  ${BASE_URL:-(unknown; bad config)}/admin/settings/$ADMIN_TOKEN"
+  echo "  ${BASE_URL:-(unknown; bad config)}/setup/token/$ADMIN_TOKEN"
   if [ "${SANDCATS_HTTPS_SUCCESSFUL}" = "yes" ] ; then
     echo ""
     echo "(If your browser shows you an OCSP error, wait 10 minutes for it to auto-resolve"
--- a/shell/client/admin-client.js
+++ b/shell/client/admin-client.js
@@ -27,7 +27,8 @@ const resetResult = function (state) {
   state.set("powerboxOfferUrl", null);
 };
 
-const getToken = function () {
+// Export for use in other admin/ routes
+getToken = function () {
   const state = Iron.controller().state;
   const token = state.get("token");
   if (!token) {
@@ -125,7 +126,7 @@ Template.admin.helpers({
   },
 
   featureKeyActive: function () {
-    return Router.current().route.getName() == "adminFeatureKey";
+    return Router.current().route.getName() == "adminFeatureKeyPage";
   },
 
   wildcardHostSeemsBroken: function () {
@@ -987,7 +988,8 @@ Template.featureKeyUploadForm.events({
       if (err) {
         instance.error.set(err.message);
       } else {
-        instance.data.successCb && instance.data.successCb();
+        instance.error.set(undefined);
+        instance.data && instance.data.successCb && instance.data.successCb();
       }
     });
   },
@@ -1005,7 +1007,7 @@ Template.featureKeyUploadForm.events({
           if (err) {
             instance.error.set(err.message);
           } else {
-            instance.data.successCb && instance.data.successCb();
+            instance.data && instance.data.successCb && instance.data.successCb();
           }
         });
       });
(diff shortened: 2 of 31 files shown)


kentonv:
I'm testing it!

1. In login providers, enabled Google and SAML.
2. In Organization, checked only SAML.
3. Tried to click "save and continue".

Result:

    Exception while invoking method 'saveOrganizationSettings' Error: Match error: Missing key 'enabled' in field membership.emailToken
    at check (packages/check/match.js:33:1)
    at [object Object].Meteor.methods.saveOrganizationSettings (server/admin-server.js:192:5)
    at maybeAuditArgumentChecks (livedata_server.js:1698:12)
    at livedata_server.js:708:19
    at livedata_server.js:706:40
    at livedata_server.js:704:46
    at tryCallTwo (/programs/server/npm/promise/node_modules/meteor-promise/node_modules/promise/lib/core.js:45:5)
    at doResolve (/programs/server/npm/promise/node_modules/meteor-promise/node_modules/promise/lib/core.js:171:13)

kentonv:
Sending test email can take a while during which there's no visual feedback that the click was registered.

kentonv:
When the first-login profile editor shows up, it has a "cancel" button, which logs the user out, but at this point they've already consumed the admin token, so they land at "setup token required".

I think we should probably hide the "cancel" button here.

kentonv:
At the end of the admin flow we present "add users" as an option, leading to the "send invites" route. We may want to hide this in cases where organizational user management has been enabled, since sending invites isn't necessary and will actually allow users to create account under non-organization identities, which is confusing.

kentonv:
The installer needs to be updated to print the new URL at the end.

kentonv:
On the "Organization" step, the domain boxes should be automatically filled in with the domain of the contact email on the feature key.

kentonv:
Also on the subject of the first-login profile editor, nothing requires me to click the "continue" button to submit the profile before I click the "finish" button to continue to the next step. If I click "finish" first then I am re-presented with the profile editor later, and none of the edits I had made are saved.

I think we should remove both the "continue" and "back" buttons from the editor and have the "finish" button submit the editor instead.

zarvox:
> On the "Organization" step, the domain boxes should be automatically filled in with the domain of the contact email on the feature key.

This is possibly semantically confusing, since we allow jumping between steps - the UI would make it *look like* the domain value is configured, but in practice the value would be unset until the user saves the setting.  This would be surprising behavior.

There's some options here:

* fill the field, and disallow jumping between setup steps
* leave the field and allow jumping between setup steps and let displayed settings not match actual effective settings, with the understanding that this may be surprising
* fill in the field, but also determine whether you're allowed to jump between steps by tracking if any admins have explicitly proceeded past that step of the wizard, by storing some durable state in the DB

I find myself favoring the first option.  Do you have a preference?

kentonv:
@zarvox If the check box isn't checked then the domain has no effect anyway, right? So does it really make much difference what's displayed there initially?

zarvox:
@kentonv Good point, in this case it doesn't matter.

zarvox:
@neynah points out that after initial setup, the "log in to become an admin" button doesn't do anything, but it should probably show the login UI.

neynah:
I spotted a strange error on "Create an admin account" but now I can't reproduce it since I'm already logged in but this is what I did:
- I enabled "Email" identity provider.
- then skipped the e-mail setup
- then tried to log in with some random e-mail and I got an error in red text (Invalid something (I don't remember the specific string))
- then I went back and configured github
- when I arrived on "create admin account" the 2nd time, it still showed the "Invalid" error but I was still able to log in successfully with github

zarvox:
Task list:

- [X] fix saveOrganizationSettings
- [X] infer org domain from feature key
- [X] update url printed by installer
- [ ] Make intro page let you log in if server is already set up and you are not logged in
- [ ] Incorporate assorted @paulproteus copy changes
- [ ] Make confirm profile buttons go away/reuse wizard button positions.
- [ ] Hide "Add Users" on the setupWizardSuccess page if any org management option is enabled.

zarvox:
All right, that's fixes added for everything discussed.

zarvox:
https://github.com/zarvox/sandstorm/commit/6123ae3cd6d92e705cac84ec6c1eedf2be3fd50f (not included in this branch) catches up with the backend/UI changes in #1765.

kentonv:
I'm assigning Jason to look over the code in detail over the next few days, but I'm going to merge now.

PR #1767, opened 2016-04-05
Update permissions computation to account for permissionsHeld membrane requirements.

dwrensha:
This is a continuation of https://github.com/sandstorm-io/sandstorm/pull/1620, which accidentally got closed.

I've implemented a selective forward-chaining algorithm that avoids the performance issues of the previous implementation.

### DIFF
--- a/shell/packages/sandstorm-permissions/permissions.js
+++ b/shell/packages/sandstorm-permissions/permissions.js
@@ -20,7 +20,15 @@ const Url = Npm.require("url");
 SandstormPermissions = {};
 
 class PermissionSet {
-  // A wrapper around an array of booleans.
+  // A wrapper around an array of booleans representing a set of permissions like "read" or
+  // "write". This might represent the permissions held by some user on some grain, or it might
+  // represent the permissions that one user has chosen to share with another user.
+  //
+  // In our model, permissions are independent. You can have "read" without "write" and you can
+  // have "write" without "read". Many apps don't actually allow arbitrary permutations, and
+  // instead define "roles" like "editor" or "viewer", where "editor" implies both read and write
+  // permission while "viewer" implies only read. Roles, however, are just aliases for sets of
+  // permissions; all actual computation is done on permissions.
 
   constructor(array) {
     if (!array) {
@@ -33,9 +41,21 @@ class PermissionSet {
   }
 
   static fromRoleAssignment(roleAssignment, viewInfo) {
+    // Create a PermissionSet based on a ViewSharingLink.RoleAssignment and a UiView.ViewInfo (as
+    // defined in grain.capnp). ViewInfo defines a mapping from roles to permission sets for a
+    // particular grain type. RoleAssignment represents the permissions passed from one user to
+    // another -- usually it specifies a single role, but sometimes also specifies permissions to
+    // add or remove as well.
+    //
+    // A falsy value for `roleAssignment` is considered equivalent to a "none" value, which means
+    // that no role was explicitly chosen, so the default role should be assigned.
+
     let result = new PermissionSet([]);
 
     if (!roleAssignment || "none" in roleAssignment) {
+      // No role explicitly chosen, e.g. because the app did not define any roles at the time
+      // the sharing took place. Assign the default role, if there is one.
+
       if (viewInfo.roles) {
         for (let ii = 0; ii < viewInfo.roles.length; ++ii) {
           const roleDef = viewInfo.roles[ii];
@@ -46,6 +66,8 @@ class PermissionSet {
         }
       }
     } else if ("allAccess" in roleAssignment) {
+      // All permissions are shared, even if there is no explicitly-defined role for this.
+
       let length = 0;
       if (viewInfo.permissions) {
         length = viewInfo.permissions.length;
@@ -58,6 +80,8 @@ class PermissionSet {
 
       result = new PermissionSet(array);
     } else if ("roleId" in roleAssignment && viewInfo.roles && viewInfo.roles.length > 0) {
+      // A specific role was chosen.
+
       const roleDef = viewInfo.roles[roleAssignment.roleId];
       if (roleDef) {
         result = new PermissionSet(roleDef.permissions);
@@ -65,6 +89,7 @@ class PermissionSet {
     }
 
     if (roleAssignment) {
+      // Add or remove specific permissions. This is uncommon.
       result.add(new PermissionSet(roleAssignment.addPermissionSet));
       result.remove(new PermissionSet(roleAssignment.removePermissionSet));
     }
@@ -133,6 +158,23 @@ class PermissionSet {
 
 class Clause {
   // A conjunction of permissions for identities on grains.
+  //
+  // This typcially represents a list of `MembraneRequirement`s, as defined in `supervisor.capnp`.
+  // These represent conditions under which some connection formed between grains remains valid.
+  // When a capability travels from grain to grain, it passes across these connections -- if any
+  // of the connections becomes invalid (is revoked), then the capability must be revoked as well.
+  // The word "membrane" comes from the concept of revokable membranes; the capability is passing
+  // across such membranes as it travels.
+  //
+  // For example, a clause might represent the statement "Alice has read access to Foo, Bob has
+  // write access to Foo, and Bob has read access to Bar". Specifically, this example situation
+  // would come about if:
+  // - Bob used his read access to Bar to extract a capability from it.
+  // - Bob embedded that capability into Foo, using his write access.
+  // - Alice extracted the capability from Foo, using her read access.
+  // If any of these permissions are revoked, then the capability needs to be revoked as well.
+  //
+  // TODO(cleanup): Consider renaming to RequirementSet for readability.
 
   constructor() {
     this.identityPermissions = {};
@@ -144,8 +186,11 @@ class Clause {
     // be used as a key for such a set.
     //
     // TODO(perf): Investigate other approaches. We do not need this hash to be cryptographically
-    //   secure; we just need it to avoid accidental collisions. Maybe we should instead represent
+    //   opaque, but we do need it to avoid collisions. Maybe we should instead represent
     //   sets of clauses as ordered maps, or maybe as proper hash maps with bucketing?
+    //   Note that a representation that allows for efficient matching by subset or superset would
+    //   be useful for some computations marked with TODO(perf) later in this file, but I'm not
+    //   sure what such a representation would look like.
     const hasher = Crypto.createHash("sha256");
     Object.keys(this.identityPermissions).sort().forEach((grainId) => {
       hasher.update(grainId + "{");
@@ -178,7 +223,8 @@ class Clause {
   }
 
   addMembraneRequirements(membraneRequirements) {
-    // Updates this clause to include the permissions required by `membraneRequirements`.
+    // Updates this clause to include the permissions required by `membraneRequirements`, which
+    // is a decoded Cap'n Proto List(MembraneRequirement).
 
     if (!membraneRequirements) return;
     membraneRequirements.forEach((requirement) => {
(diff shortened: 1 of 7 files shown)


dwrensha:
Garply, retest this please.

dwrensha:
Garply, retest this please.

jparyani:
Garply, retest this please

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -291,100 +252,173 @@ class RequirementSet {
...
-    for (let grainId in other.identityPermissions) {
-      for (let identityId in other.identityPermissions[grainId]) {
-        let otherPermissions = other.identityPermissions[grainId][identityId];
-        this._ensureEntryExists(grainId, identityId);
-        this.identityPermissions[grainId][identityId].add(otherPermissions);
+class Variable {
+  // Our permissions computation can be framed as a propositional HORNSAT problem; this `Variable`
+  // class represents a variable in that sense. There is a variable for every (grain ID, vertex ID,
+  // permission ID) triple. In any given computation, we only explicitly construct those
+  // variables that know might actually be relevant to the result.
```
kentonv: that know -> that we know

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -291,100 +252,173 @@ class RequirementSet {
...
+        this.appDefined.push({ doesNotCarry: null });
       }
     }
   }
-}
 
-class ConditionalPermissionSet {
-  // A PermissionSet that is contingent upon some requirements.
+  requirementsAreMet() {
+    return this.numUnmetRequirements <= 0;
```
kentonv: Presumably a negative value here suggests a horrible bug where some requirements may be being ignored? Maybe `decrementRequirements()` should throw if the count goes negative?

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -291,100 +252,173 @@ class RequirementSet {
       delete this.identityPermissions[grainId];
     }
 
-    return { vertex: { grain: { _id: grainId, identityId: identityId } },
-             permissions: permissions, };
+    return { grainId: grainId, identityId: identityId, permissionSet: permissions, };
+  }
+}
+
+const permissionIdPattern = Match.OneOf({ canAccess: null }, { appDefined: Match.Integer });
```
kentonv: This feels bulky to me. Possible alternative: Match.OneOf(Number, "canAccess"), i.e. the value is either a number or the special string value "canAccess". You can then represent a permission set as an object whose keys are numbers or "canAccess". This actually makes more sense than an array in that a permission array is really an infinite array with an all-false tail, so now we have an infinite object where unset fields are considered false...

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -291,100 +252,173 @@ class RequirementSet {
...
-    // membrane requirements M. The result does not carry any information about the vertices
-    // A and B, even though the input `token` may contain information about them, for example
-    // in the `parentToken` field.
-
-    const permissions = PermissionSet.fromRoleAssignment(token.roleAssignment, viewInfo);
-    const result = new ConditionalPermissionSet(permissions, new RequirementSet());
-    result.requirements.addMembraneRequirements(token.requirements);
-    result.tokensUsed[token._id] = true;
-    return result;
+  directIsMet(permissionId) {
```
kentonv: These methods could use doc comments.

shell/packages/sandstorm-permissions/permissions.js
dwrensha: I've switched to a plain object. Now a permission ID is a stringified integer >= -1, where "-1" means "can access".  This does simplify a few things!

dwrensha:
Garply, retest this please.

dwrensha:
I've addressed the comments that have come in so far.

I don't know what the deal is with the integration tests... Garply is acting strangely.

jparyani:
Garply, retest this please.

jparyani:
That was my bad. I was resizing the disk and had to reboot the machine.

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -256,18 +256,17 @@ class RequirementSet {
   }
 }
 
-const permissionIdPattern = Match.OneOf({ canAccess: null }, { appDefined: Match.Integer });
-// As we compute a flow of permissions, we need to be able to explicitly refer to not only
-// the permissions enumerated in a grain's `ViewInfo`, but also the usually-implicit "can
-// access the grain at all" permission. A permission ID, defined here, allows us to refer
-// to either of these types of permissions.
+// A permission ID is a stringified integer greater than or equal to -1. The permission "-1"
```
kentonv: Why stringified?

Why "-1" rather than "canAccess"?

dwrensha: Stringified because as far as I understand, keys in plain javascript objects are always strings. I considered using an `Array` rather than a plain object here, but that seemed weird.

-1 because that makes it somewhat easier to iterate through the permissions and to find the maximum key.

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -315,65 +314,60 @@ class Variable {
...
 
-    // The following fields have a "permission status" value, one of:
-    // {unmet: null}, {met: null}, or { doesNotCarry: null }.
-    this.canAccess = { unmet: null };
-    this.appDefined = [];
+    this.receivedPermissions = { "-1": false };
+    // A map from permission ID to boolean. If a permission ID is present in this map,
+    // then this token's roleAssignment includes that permission. If the corresponding
+    // value is `true`, then we've proven that this token receives this permission and
+    // therefore propagates it.
```
dwrensha: "propagates it if the requirements are met", I should probably say.

shell/packages/sandstorm-permissions/permissions.js
kentonv: > keys in plain javascript objects are always strings

Sort of but not exactly. An array in Javascript is actually an object:

    > var x = ["foo", "bar"]
    > typeof x
    'object'
    > Object.keys(x)
    [ '0', '1' ]
    > x.baz = "qux"
    > x
    [ 'foo', 'bar', baz: 'qux' ]
    > Object.keys(x)
    [ '0', '1', 'baz' ]

Similarly, objects can have numeric keys:

    > ({0: "foo", 1: "bar"})[1]
    'bar'

Theoretically a non-string key is converted to a string for all purposes, but that's not actually what happens under the hood. v8 treats numeric keys (including string keys that happen to be canonical representations of integers) differently from strings and optimizes for them. Internally, the difference between an object and an array is actually pretty fuzzy.

But in any case, it is perfectly fine to use numbers as keys, or use an array with extra non-numeric fields.

> -1 because that makes it somewhat easier to iterate through the permissions and to find the maximum key.

But it's significantly less self-documenting, and if we end up with other pseudo-permissions in the future it gets uglier...

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -399,136 +433,447 @@ class Context {
...
+          const reqPermissions = requirement.permissionsHeld.permissions || [];
+
+          forEachPermission(reqPermissions, (permissionId) => {
+            const variable = this.getVariable(reqGrainId, reqVertexId, permissionId);
+            if (!variable.value) {
+              numUnmetRequirements += 1;
+              variable.requirementsTailList.push(tokenId);
+            }
+          });
+        }
```
kentonv: Hmm, what about other types of requirements? `tokenValid` and `userIsAdmin`?

dwrensha: I'm actually not sure that I understand the meaning of `tokenValid` requirements. Are they satisfied if the token merely exists and hasn't been revoked? Or (assuming the token is a UiView token) do they also require receives some set of permissions, similar to the `permissionsHeld` requirement? In the latter case, we would need to add a field, I think.

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -399,136 +433,447 @@ class Context {
...
+    if (needToExploreUnmet) {
+      this.unmetRequirements.addRequirements(token.requirements);
+    }
+
+    this.activeTokens[tokenId] = activeToken;
+
+    return true;
+  }
+
+  _ensureVariableExists(grainId, vertexId, permissionId) {
```
kentonv: I think you can collapse _ensureVariableExists() into getVariable() at this point and save some redundant operations. Callers can call getVariable() and ignore the result if they just wanted to make sure the variable exists...

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -399,136 +433,447 @@ class Context {
     }
   }
 
-  addGrain(db, grainId) {
-    check(db, SandstormDb);
-    check(grainId, String);
-    const grain = db.getGrain(grainId);
-    this.grains[grainId] = grain;
+  addGrains(db, grainIds) {
```
kentonv: It looks like various callers of this method could conceivably pass in an empty grainIds, in which case the database query is a waste.

shell/packages/sandstorm-permissions/permissions.js
kentonv: A `tokenValid` requirement comes about when a capability was transmitted across a non-UiView capability. That is, the `tokenValid` requirement should always point to a non-UiView token. The requirement passes if restoring that token would currently succeed. This requires checking that the token and its ancestors (parentToken chain) all exist and are not revoked, and that all of their respective requirements are satisfied.

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -399,136 +433,447 @@ class Context {
...
+    check(vertexId, String);
+    check(permissionSet, PermissionSet);
+    check(db, Match.OneOf(undefined, SandstormDb));
+    // If `db` is not provided, then this function will make no database queries.
+
+    if (db) {
+      this.addGrains(db, [grainId]);
+    }
+
+    this.activateRelevantTokens(grainId, vertexId, permissionSet);
```
kentonv: This method appears to take only two arguments, but you're passing a third here.

kentonv: Incidentally it seems useful for it to take permissionSet as an argument, too, especially when checking requirements. E.g. if we have a requirement that Bob has write access then we might waste time following tokens that only provide read access. But it also seems like this would be more complicated to implement since now you need partially-activated tokens, or something.

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -399,136 +433,447 @@ class Context {
...
+    });
+
+    while (stack.length > 0) {
+      const current = stack.pop();
+      const variable = this.getVariable(current.grainId, current.vertexId, current.permissionId);
+      const tokenId = variable.responsibleTokenId;
+      if (tokenId) {
+        const token = this.tokensById[tokenId];
+
+        let sharerId = (token.parentToken && "t:" + token.parentToken) || ("i:" + token.identityId);
```
kentonv: `(a && b) || c` => `a ? b : c`?

kentonv:
OK I did the commentary thing again. See/merge: https://github.com/sandstorm-io/sandstorm/commit/4dddb34eedbdd8cca0ed2b11620fcf136615c3a5

kentonv:
Note: Github apparently buggily claims that comments were on an "outdated diff" when really they were on a diff that was too large for Github to display by default.

dwrensha:
Updated to handle `tokenValid` and `userIsAdmin` requirements. I also updated `checkRequirements()` in core.js, which was not validating parent tokens.

dwrensha:
rebasing to resolve conflicts...

PR #1768, opened 2016-04-05
Remove dockerfiles

jparyani:
They have been unsupported for a while. We've seen many users become confused/frustrated when they don't work, and their initial purpose as a testing environment is no longer relevant as we just do a local install now.


paulproteus:
+1 to removing things that don't work. Maybe one day we can have something
that does work, but best to stop confusing people for now, in my opinion.

PR #1779, opened 2016-04-06
Allow muting the app dev message with a particular key in localStorage

zarvox:
I'm tired of seeing this every time the page refreshes and I care about what's
being logged but instead see mostly this giant note with yellow background and
larger type than console.log() and links that are large click targets that make
it harder to focus the JS console input.

Disable the note from the JS console:

    localStorage.setItem("muteDevNote", true)


paulproteus:
+1, thanks for looking into this.

jparyani:
Garply, retest this please

PR #1781, opened 2016-04-06
For performance, use native C++ objects in the `spk dev` FUSE layer.

dwrensha:
With these changes, I have observed dev-mode GitLab to launch six times faster than it does with the currently-released `spk` tool.

The improvement comes from:

   1. avoiding the use of exceptions in the `ENOENT` case of `lookup()`,
   2. abandoning the Cap'n Proto FUSE interfaces in favor of native C++ classes.


src/sandstorm/fuse.c++
```diff
@@ -511,49 +492,51 @@ private:
...
+            return kj::mv(reply);
+          }
+        });
         break;
       }
 
-      case FUSE_GETATTR:
-        addReplyTask(header.unique, EIO, node.getAttributesRequest(capnp::MessageSize {4, 0}).send()
-            .then([this](auto&& response) -> kj::Own<ResponseBase> {
+      case FUSE_GETATTR: {
```
paulproteus: Not a showstopper by any means, but just because I'm curious and want to learn more C++: What's the meaning of this `{` here? Does it add something to this `case`?

paulproteus: Oh, I guess it's to avoid `response` and `reply` from being in-scope elsewhere. http://stackoverflow.com/questions/4241545/c-switch-case-curly-braces-after-every-case

I need a rubber-duck emoji.

dwrensha: In this case the brackets are not actually needed. `response` and `reply` are inside the scope of a lambda, so they would not cause problems. I think these brackets got added because at one point I did in fact need the scope to define some variables, but that need went away with further edits. In any case, the brackets don't hurt.

src/sandstorm/fuse.c++
```diff
@@ -408,10 +398,9 @@ private:
   }
 
   bool dispatch(struct fuse_in_header& header, kj::ArrayPtr<const kj::byte> body) {
-    auto iter = nodeMap.find(header.nodeid);
-    KJ_REQUIRE(header.nodeid == 0 || iter != nodeMap.end(),
+    auto node_iter = nodeMap.find(header.nodeid);
```
kentonv: camelCase?

src/sandstorm/fuse.c++
kentonv: The worst part about switch statements and their crazy fall-though semantics is that sometimes it's useful.

src/sandstorm/fuse.c++
```diff
@@ -866,26 +853,25 @@ inline int64_t toNanos(const struct timespec& ts) {
...
     KJ_SYSCALL(ifd = open(path.cStr(), O_RDONLY), path);
     fd = kj::AutoCloseFd(ifd);
   }
 
-protected:
-  kj::Promise<void> read(ReadContext context) override {
-    auto params = context.getParams();
-    auto size = params.getSize();
-    auto offset = params.getOffset();
+  kj::Own<fuse::File> addRef() override {
```
kentonv: Nitpick: If providing an addRef() method then I'd inherit `kj::Refcounted` privately. If inheriting `kj::Refcounted` publicly then callers might as well call `kj::addRef()` directly.

kentonv: Actually, no, forget I said that, that's wrong... `addRef()` exists for the sake of people using the abstract interface. Inheriting Refcounted privately wouldn't work because `kj::refcounted<T>()` wouldn't be able to access the private parent class.

src/sandstorm/fuse.h
```diff
@@ -17,13 +17,89 @@
...
 
-#include <sandstorm/fuse.capnp.h>
 #include <kj/time.h>
 #include <kj/io.h>
 #include <kj/function.h>
+#include <kj/refcount.h>
 
 namespace kj { class UnixEventPort; }
 
+namespace fuse {
```
kentonv: `fuse` should be nested in `sandstorm`.

src/sandstorm/fuse.h
```diff
@@ -17,13 +17,89 @@
...
+    BLOCK_DEVICE = 1,
+    CHARACTER_DEVICE = 2,
+    DIRECTORY = 3,
+    FIFO = 4,
+    SYMLINK = 5,
+    REGULAR = 6,
+    SOCKET = 7,
+  };
+
+  struct Attributes {
```
kentonv: Would it make any sense to use `struct fuse_attr` directly here?

I suppose it's probably not worth the refactoring effort, but curious what you think.

kentonv:
All my comments are worthless so I'll just go ahead and merge this.

src/sandstorm/fuse.h
dwrensha: Maybe! One potential downside is that we would then need to have `#include <linux/fuse.h>` in this header.

PR #1782, opened 2016-04-06
docs: Indicate that Sandstorm for Work can be turned on/off

paulproteus:


paulproteus:

PR #1783, opened 2016-04-07
Meteor 1.3

zarvox:
Not too much that had to change, surprisingly.  Things I did:

* Ran `meteor update` and committed the output
* Fixed up an incorrect path in a `require()` in `meteor-bundle-main.js` (Meteor 1.3 changed how their modules are laid out to properly support ES2015 modules)
* Regenerated all of our `npm-shrinkwrap.json` files.  Meteor generates these when a Meteor package specifies a dependency from npm.  Meteor 1.3 uses a newer version of npm (2.14.22 vs 1.4.28) which 
* Added `shell/package.json` (and `shell/npm-shrinkwrap.json`) so we can add direct dependencies on npm packages.
* Switched our intro.js dependency (which seemingly broke due to stricter module export rules that Meteor 1.3 enforces) to one pulled in via npm.  It actually points at our fork of intro.js, which has a couple CSS fixes.  This was a nice exercise, and demonstrates 1) the ease with which we can pull in npm deps, and 2) the ease with which we can maintain forks of things when we need to.

I'll probably switch some of our other just-an-npm-wrapper packages to the new installation model as time goes on.  For reference, adding a dependency is simply:

```bash
cd shell
meteor npm install --save intro.js
meteor npm shrinkwrap
```

to install the dependency from npm, and then to use something from the package:

```js
import { introJs } from "intro.js";
```

in the file that uses `introJs`.

One last happy note: live reload seems **significantly faster**.  I'm getting <1s rebuild times, and I'm seeing the page refresh in the browser in ~3s.  The isobuild perf team did some good work this release! :)


zarvox:
The stylechecks failed because the trick I tried with the dangling-symlink-until-npm-install-runs seems to trip up `jscs`, even though it's not linking to a `.js` file.  This looks like it can be fixed by excluding the path in `.jscsrc`.

The build failed because we're out of disk space again.

zarvox:
Remaining js-style complaint is fixed by https://github.com/sandstorm-io/sandstorm/pull/1785.

jparyani:
Garply, retest this please

kentonv:
I'm going to hold this one until Monday so that we have a week to notice problems before it lands in a release.

paulproteus:
I promised to test this for Drew, so I'll run it on rose.sandcats.io for a
little while. Expect a report in a day or so.

paulproteus:
I'm running this on rose.sandcats.io now. I can use my server with HTTPS and open grains. If I run into any issues, I'll let you know.

zarvox:
Garply, retest this please

dwrensha:
Garply, retest this please.

paulproteus:
Sweet!​

PR #1784, opened 2016-04-07
Feature key form: don't allow submitting an empty feature key text

### LINKED ISSUE #1766: Feature key disappears after clicking "Upload new feature key"
I clicked "upload new feature key"
image
I couldn't cancel what I was doing so I clicked "verify"
image
Then my feature key was removed essentially. It was unexpected that if I clicked "upload new feature key" I'd "lose" my existing feature key without being able to confirm or cancel.
image

Since I clicked "upload new feature key" and the input box appeared below. I didn't think to click "back" in the browser. Because I didn't feel like I navigated forward anywhere.

@zarvox says we should disable the "verify" button until valid key has been entered & we may also want to make this a dismissable modal.

This probably won't get fixed until the new admin UI gets shipped but I thought I'd make a note of it.

zarvox:
Due to a hilarious dual-purposing, the default empty string value is
considered by the backend to mean "please delete the current feature key".

We have a button for that, with an appropriate warning.  We should not let you
accidentally remove your feature key in this manner.

Fixes #1766.

### DIFF
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -1477,10 +1477,10 @@ limitations under the License.
       {{/if}}
     {{/with}}
     <label>Feature key:
-    <textarea class="feature-key" placeholder="Paste feature key here..."></textarea>
+      <textarea class="feature-key" placeholder="Paste feature key here..." value="{{text}}"></textarea>
     </label>
     <div class="feature-key-button-row">
-      <button class="check-key">Verify</button>
+      <button class="check-key" {{htmlDisabled}}>Verify</button>
     </div>
     <label>Or select a file to upload: <input name="feature-key" type="file" /></label>
   </form>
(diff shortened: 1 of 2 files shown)


shell/client/shell.html
```diff
@@ -1477,10 +1477,10 @@ <h3 id="app-stats">Apps</h3>
       {{/if}}
     {{/with}}
     <label>Feature key:
-    <textarea class="feature-key" placeholder="Paste feature key here..."></textarea>
+      <textarea class="feature-key" placeholder="Paste feature key here..." value="{{text}}"></textarea>
     </label>
     <div class="feature-key-button-row">
-      <button class="check-key">Verify</button>
+      <button class="check-key" {{htmlDisabled}}>Verify</button>
```
kentonv: I believe blaze supports `disabled={{value}}` where `value` is a boolean.

kentonv:
Garply, retest this please.

kentonv:
I'll merge this and make the change.

kentonv:
d33a912

I notice the setup wizard code has a bunch of cases of this. Perhaps they should be cleaned up too.

PR #1788, opened 2016-04-07
Send a new SandstormCore capability on Supervisor.keepAlive().

kentonv:
This is important if the front-end has independently shut down and started up again; any existing capability will be disconnected.

As part of this, I had to refactor the CapabilityRedirector stuff a bit. Previously, were actually creating two levels of redirectors, one of which was never being redirected.

While I was there, I also made CapabilityRedirector directly detect disconnects in case the caller can't (as is the case after keepAlive()).

(This problem has been causing the feature key vendor to fail a lot.)


jparyani:
Garply, retest this please

jparyani:
LGTM

PR #1789, opened 2016-04-08
install.sh: fix issue 1777

paulproteus:
While we're at it:

- Stop sending people to the admin token flow in the case of dev accounts being on. (EDIT: Somehow my change to do this caused syntax errors in bash on jessie, but not on unstable? Skipping for now.) (RE-EDIT: I forgot a `fi` for bash to close the `if`. Fixed.)

- Stop asking `-u` users if they want option (1) or (2) when only (2) will work. (Send them straight into option 2.)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -628,12 +628,21 @@ detect_init_system() {
 }
 
 choose_install_mode() {
+  echo -n 'Sandstorm makes it easy to run web apps on your own server. '
+
   if [ "yes" = "$USE_DEFAULTS" ] ; then
     CHOSEN_INSTALL_MODE="${CHOSEN_INSTALL_MODE:-2}"  # dev server mode by default
   fi
 
+  if [ "no" = "${PREFER_ROOT:-}" ] ; then
+    echo ""
+    echo "NOTE: Showing you all options, including development options, but omitting "
+    echo "      init script automation, because you chose to install without using root."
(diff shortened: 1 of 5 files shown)


paulproteus:

install.sh
```diff
@@ -628,12 +628,21 @@ detect_init_system() {
...
+
   if [ "yes" = "$USE_DEFAULTS" ] ; then
     CHOSEN_INSTALL_MODE="${CHOSEN_INSTALL_MODE:-2}"  # dev server mode by default
   fi
 
+  if [ "no" = "${PREFER_ROOT:-}" ] ; then
+    echo ""
+    echo "NOTE: Showing you all options, including development options, but omitting "
+    echo "      init script automation, because you chose to install without using root."
+    CHOSEN_INSTALL_MODE="${CHOSEN_INSTALL_MODE:-2}"  # dev server mode by default
```
kentonv: Please don't default to a dev server when not root!

It's totally legitimate for people to want to install a live server with sandcats and everything as non-root.

kentonv: I consider this release-blocking.

kentonv: https://github.com/sandstorm-io/sandstorm/pull/1800

kentonv: Uggghh apparently I'm an idiot -- see #1800.

PR #1796, opened 2016-04-08
Offer templates should avoid extra newlines on the clipboard in Firefox

### LINKED ISSUE #1795: Offer templates copy with leading and trailing spaces on Android/Firefox
So, I was setting up ownCloud/Davros, and I kept getting wrong password errors, and was very confused. This was fast enough the template was not expiring. I figured out there were a few spaces before and after everything I copied from offer templates via the clipboard button. Leading spaces accidentally in the URL field weren't a problem, but leading spaces I failed to remove at the beginning of the password were.

zarvox:
There appears to be a Selection edge case in Firefox that inserts spurious
newlines into the value copied onto the clipboard if you attempt to select
everything contained by a `<pre>` within a `<div>` if the `<div>` contains newlines,
but the `<pre>` does not.

Removing the contained newlines makes Firefox do the right thing.

Fixes #1795.

### DIFF
--- a/shell/public/offer-template.html
+++ b/shell/public/offer-template.html
@@ -64,9 +64,10 @@ body {
     </style>
   </head>
   <body>
-    <div id="container">
-      <pre id="text"></pre>
-    </div>
+    <!-- Since <pre> doesn't bound selection, this needs to all be on the same line,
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
You're awesome! Thanks!

PR #1797, opened 2016-04-08
Update Selenium to version 2.53.0.

dwrensha:
Firefox 45.0.1 causes errors [like this](https://bsr44jl5i2yrk3lgqhmd.oasis.sandstorm.io/debug.txt) when I run `make test`. Updating Selenium makes the problem go away.


dwrensha:
It looks like Selenium's behavior may have changed in a few places and now some of the tests are failing. I'm looking into it.

jparyani:
Ok, I'll hold off on updating the build server to the new selenium version (it runs selenium manually in the background).

dwrensha:
I've been unable to reproduce the failures. They seem to have been a one-time thing when I updated. Now I'm only seeing the occasional email EADDRINUSE error, which has been a known problem.

jparyani:
K I'll go ahead and update the build server.

PR #1799, opened 2016-04-09
install.sh: if adminToken does not exist, do not try to chown it

paulproteus:
Note that without this fix, new dev installs are broken. Thanks to @Daemoen
for showing up on IRC and sharing the problem.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1356,10 +1356,17 @@ set_permissions() {
   fi
 
   # Set ownership of files.  We want the dirs to be root:sandstorm but the contents to be
-  # sandstorm:sandstorm.
-  chown -R $SERVER_USER:$GROUP var/{log,pid,mongo} var/sandstorm/{apps,grains,downloads,adminToken}
-  chown root:$GROUP var/{log,pid,mongo,sandstorm} var/sandstorm/{apps,grains,downloads,adminToken}
-  chmod -R g=rwX,o= var/{log,pid,mongo,sandstorm} var/sandstorm/{apps,grains,downloads,adminToken}
+  # sandstorm:sandstorm. The adminToken file might not exist, so tread carefully there.
+  local VAR_SANDSTORM_CONTENTS=var/sandstorm/{apps,grains,downloads}
+  local ADMIN_TOKEN_PATH="var/sandstorm/adminToken"
+  if [ -e "${ADMIN_TOKEN_PATH}" ] ; then
+    VAR_SANDSTORM_CONTENTS="${VAR_SANDSTORM_CONTENTS} ${ADMIN_TOKEN_PATH}"
+  fi
+
+  # We use $VAR_SANDSTORM_CONTENTS without quotes around it so that each entry gets passed to chown.
+  chown -R $SERVER_USER:$GROUP var/{log,pid,mongo} $VAR_SANDSTORM_CONTENTS
+  chown root:$GROUP var/{log,pid,mongo,sandstorm} $VAR_SANDSTORM_CONTENTS
+  chmod -R g=rwX,o= var/{log,pid,mongo,sandstorm} $VAR_SANDSTORM_CONTENTS
 }
 
 install_sandstorm_symlinks() {


dwrensha:
The problem was introduced in https://github.com/sandstorm-io/sandstorm/commit/563c67ec85b19a334c9c26a7594445a9ed2632aa, presumably?

kentonv:
Hmm, why were people using an unreleased install.sh? I mean it's good they were because it caught this bug, but seems odd.

kentonv:
This change does not work. {} expansion does not occur in variable assignments. This change broke the installer for everyone.

paulproteus:
A few thoughts.

**How was Daemoen using an unreleased install.sh?** I think this is due to doing a `git clone` and running `vagrant up`. This uses VirtualBox file sharing to map the latest `install.sh` into the Linux virtual machine, 

**On the topic of {} expansion not occurring during variable assignments.** Let me test that now.

```
$ cat /tmp/ok.sh
#!/bin/bash
THING="ok"
THING2="Curly braces are ${THING}"
echo "$THING2"
$ bash /tmp/ok.sh
+ THING=ok
+ THING2='Curly braces are ok'
+ echo 'Curly braces are ok'
Curly braces are ok
```

So I'm not sure which line you are referring to when you say `{}` expansion doesn't work in variable assignments. Can you point me at a line number in `install.sh` that has the problem you're referring to, so I can understand better?

More generally: I am 100% OK with digging into these `install.sh` after waking up tomorrow, so that it can be in a good state for the release. I'm rather sorry about the code pushes of ambiguous quality. I need to fix `installer-tests`, which I can also spend time doing tomorrow. I'll make sure to state clearer rationales and how I've tested things, as part of that.

@dwrensha: Yes, this does seem to be me to be introduced in https://github.com/sandstorm-io/sandstorm/commit/563c67ec85b19a334c9c26a7594445a9ed2632aa . I think I need the conditional logic in this PR because on one hand, I don't want to store an admin token if we're not going to print it (and not going to prompt people to use it), and yet not having an admin token means the file list passed to chown does have to change. I read the JS code within Sandstorm itself, and I _believe_ that creating an empty-string admin token file would possibly allow an empty string to be used as an admin token, which would be bad news, so I shied away from creating an empty-string admin token.

dwrensha:
@paulproteus: the problem is the `local` definition of `VAR_SANDSTORM_CONTENTS`. I find bash's behavior here to be rather surprising!

```
#!/bin/bash                                                                                          

set_permissions() {
    local VAR_SANDSTORM_CONTENTS=var/sandstorm/{apps,grains,downloads}
    echo $VAR_SANDSTORM_CONTENTS # var/sandstorm/downloads                                           

    VAR_SANDSTORM_CONTENTS=var/sandstorm/{apps,grains,downloads}
    echo $VAR_SANDSTORM_CONTENTS # var/sandstorm/{apps,grains,downloads}                             

    local VAR_SANDSTORM_CONTENTS=var/sandstorm/\{apps,grains,downloads\}
    echo $VAR_SANDSTORM_CONTENTS # var/sandstorm/{apps,grains,downloads}                             
}

set_permissions
```

kentonv:
local VAR_SANDSTORM_CONTENTS=var/sandstorm/{apps,grains,downloads}

expands to:

    local VAR_SANDSTORM_CONTENTS=var/sandstorm/apps \
          VAR_SANDSTORM_CONTENTS=var/sandstorm/grains \
          VAR_SANDSTORM_CONTENTS=var/sandstorm/downloads

The assignments are evaluated in order so the last one wins.

I actually pushed a release last night since today is a LAN party. Since there had been installer changes, I decided to test setting up Sandstorm on a fresh Digital Ocean instance. I actually impressed myself with how easy it was -- except that when I tried to install an app, it failed with some permissions error!

Turns out this was the culprit.

Perhaps we need a test that actually goes through the whole server install process and then tries to install and run an app. Or we need to do that manually whenever we change the script.

PR #1800, opened 2016-04-09
Don't default to dev server when using install.sh -u.

kentonv:
There is no reason to expect that non-root installs want a dev server. In fact, dev servers mostly don't work with non-root installs (although I'd prefer not to default to live either because I often use this flow for testing).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -638,7 +638,6 @@ choose_install_mode() {
     echo ""
     echo "NOTE: Showing you all options, including development options, but omitting "
     echo "      init script automation, because you chose to install without using root."
-    CHOSEN_INSTALL_MODE="${CHOSEN_INSTALL_MODE:-2}"  # dev server mode by default
   fi
 
   if [ -z "${CHOSEN_INSTALL_MODE:-}" ]; then


kentonv:
Self-merging so that I can do a release!

kentonv:
@paulproteus Apologies, I realize now that path (1) never actually worked with -u. Which I think is a bug, but it was silly for me to revert your change given that (1) currently doesn't work anyway. I am not sure why I managed not to actually try taking path 1 before submitting this pull request -- I only checked that the prompt came back.

kentonv:
(Feeling pretty silly right now!)

kentonv:
This technically means that install.sh -u gives you the option between:

1. A path that immediately errors out.
2. A path that installs a server that won't do anything useful, because dev mode requires root.

Which makes me wonder why we are supporting this mode...

zarvox:
FWIW, I recently used -u to install a non-root, dev mode server which I've been using for shell development.  Since I'm not using `spk dev`, this is precisely the right combination for my purposes.  So I found path 2 useful, despite the can't-use-fuse-for-app-dev issue.

PR #1804, opened 2016-04-09
Fix installer

kentonv:


kentonv:
Self-merging because already deployed.

PR #1807, opened 2016-04-10
Fix bug where hidden paths actually do show up in the results of Directory.read()

dwrensha:
Currently, if my `sourceMap` has some hidden paths and I do `ls -l .` inside a `spk dev` grain, I see results like:

```
ls: cannot access 'hidden': No such file or directory
ls: cannot access 'home': No such file or directory
ls: cannot access 'sys': No such file or directory
total 3500
lrwxrwxrwx  1 65534 65534       7 Dec 14 12:48 bin -> usr/bin
drwxr-xr-x  2 65534 65534    4096 May 28  2014 boot
drwxr-xr-x  2  1000 david     120 Apr 10 09:37 dev
drwxr-xr-x 72 65534 65534    4096 Apr  8 10:31 etc
d?????????  ? ?     ?           ?            ? hidden
d?????????  ? ?     ?           ?            ? home
...
```

The hidden directories correctly return `ENOENT` when they are `stat()`ed, but they still show up in the results of `Directory.read()`.

This patch fixes the problem, which as far as I can tell is as old as `spk dev` itself. Although `HidingDirectory` has always existed, apparently it's never been used!

This patch also fixes a typo in the name of the class `SimpleDirectory`.

### DIFF
--- a/src/sandstorm/union-fs.c++
+++ b/src/sandstorm/union-fs.c++
@@ -77,7 +77,7 @@ protected:
   kj::Own<fuse::Node> delegate;
 };
 
-class SimpleDirecotry: public fuse::Directory, public kj::Refcounted {
+class SimpleDirectory: public fuse::Directory, public kj::Refcounted {
   // Implementation of fuse::Directory that is easier to implement because it just calls a
   // method that returns the whole content as an array.
 
@@ -158,7 +158,7 @@ private:
   }
 };
 
-class UnionDirectory final: public SimpleDirecotry {
+class UnionDirectory final: public SimpleDirectory {
   // Directory that merges the contents of several directories.
 
 public:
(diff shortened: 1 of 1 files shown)


dwrensha:
I have experimentally confirmed that this bug was present on [v0.149](https://github.com/sandstorm-io/sandstorm/releases/tag/v0.149), so presumably it is not a regression.

paulproteus:
As a side note: if there's an spk dev feature that's never been used, so
we'll never notice if it breaks again, is it better to outright remove it?

dwrensha:
The feature in question (the ability to hide paths) most certainly does get used, and the important part of its functionality does work. Part of its functionality, however, is broken in a confusing way: `readdir()` and `stat()` give inconsistent results on an immutable filesystem.

PR #1809, opened 2016-04-11
docs: Clarify that Sandstorm respects navigation within a grain

paulproteus:


paulproteus:

PR #1814, opened 2016-04-12
administering.md: Add new install.sh docs to human-curated ToC

paulproteus:


paulproteus:

PR #1818, opened 2016-04-12
Make sure that the fibers package targets the correct v8 version.

dwrensha:
If the system's `node --version` is not 0.10, then the built Sandstorm bundle will crash with errors like 

```
Error: `/programs/server/node_modules/fibers/bin/linux-x64-v8-3.14/fibers.node` is missing. Try reinstalling `node-fibers`
```

This fixes the problem. See also https://github.com/sandstorm-io/meteor-spk/pull/16.

### DIFF
--- a/make-bundle.sh
+++ b/make-bundle.sh
@@ -87,8 +87,12 @@ cp -r shell-build/bundle bundle
 rm -f bundle/README
 cp meteor-bundle-main.js bundle/sandstorm-main.js
 
-# Meteor wants us to do "npm install" in the bundle to prepare it.
-(cd bundle/programs/server && "$METEOR_DEV_BUNDLE/bin/npm" install)
+# Meteor wants us to do `npm install` in the bundle to prepare it.
+# The fibers package builds native extensions, choosing the target v8 version based on
+# the version of `/usr/bin/env node`. We need to make it does not pick up the wrong binary,
+# so we prepend `METEOR_DEV_BUNDLE/bin` to `PATH`.
+(cd bundle/programs/server && \
+ PATH=$METEOR_DEV_BUNDLE/bin:$PATH "$METEOR_DEV_BUNDLE/bin/npm" install)
 
 # Copy over key binaries.
 mkdir -p bundle/bin


kentonv:
The Meteor tool now has a `meteor npm` command. Maybe we should be using that?

kentonv:
Indeed, I ran into the same problem in meteor-spk, and switching to `meteor npm` solved the problem without the need for PATH hacks.

zarvox:
Possibly, though that possibly risks that in the future (say, Meteor 1.4) we'd wind up using whatever the latest `meteor npm` installed on your system uses, rather than the one specified by the version used in `.meteor/release`, which might not be what you intend.  I'm not sure what Meteor's compatibility guarantees are.

kentonv:
So, the `meteor` tool decides which version of itself to run by climbing the directory tree looking for a `.meteor/release`. I guess this doesn't quite work for us since our `.meteor` directory is under `shell`, but the bundle is not. However, Meteor allows you to override the release with `--release`. E.g.:

    $ meteor --version
    Meteor 1.3.1
    $ meteor --release 1.2 --version
    Meteor 1.2

So perhaps we should pass `--release`?

kentonv:
This does the right thing, FWIW:

    meteor --release $(<shell/.meteor/release) --version

dwrensha:
Yeah, `meteor npm` looks like a better idea. Unfortunately, `meteor --version 1.3.1 npm` does not work.

I've updated this pull request to instead call `meteor npm install` from the `./shell` directory.

kentonv:
`meteor --version 1.3.1 npm` is not the right syntax. You want `meteor --release 1.3.1 npm`. Can you try again?

kentonv:
(To me clear, the `--version` flag in my examples was there to instruct Meteor to print out its own version, to verify that `--release` was taking effect.)

dwrensha:
Oops, typo. This is what I mean to say:

```
$ meteor --release 1.3.1 npm
'npm' is not a Meteor command. See 'meteor --help'.
```

kentonv:
Huh. Indeed, it seems like the `npm` command is implemented differently from everything else. Annoying.

I'm not sure your current approach is correct. You're using `--prefix` to tell it where to put the installed modules, but won't it end up installing dependencies of `shell/package.json` instead of `bundle/programs/server/package.json`? Are these equivalent?

dwrensha:
Gwah. Good point. I've switched this back to my original proposal, which seems like the simplest solution here.

Note that basically the only thing that `meteor npm` does is to prepend to the path in the same way: https://github.com/meteor/meteor/blob/release/METEOR%401.3.1/tools/cli/dev-bundle-bin-commands.js#L28

PR #1833, opened 2016-04-13
Admin panel app sources page

zarvox:


shell/client/styles/_partials.scss
```diff
@@ -51,6 +51,7 @@
   font-size: 13px;
   text-align: center;
   text-decoration: none;
+  font-family: 'Source Sans Pro', sans-serif;
```
kentonv: How about `font-family: inherit`?

kentonv:
Otherwise looks good.

zarvox:
Fixed.

Issue #1188, opened 2015-11-21
Configure advice for Google auth slightly wrong


pdehaye:
One of the URLs suggested when setting up Google login is slightly wrong. 

For step 5:

Set Authorized Javascript Origins to: https://my-awesome-subdomain.sandcats.io/

Google complains about this, it should not include the trailing "/"

frigginglorious:
Thought it would be an easy tackle, but I was grepping around in files for that dialogue and couldn't find anything. I asked in IRC and this is what I got:

> kentonv: the instructions you're trying to update are in Meteor
> kentonv: the Meteor `google` or `accounts-google` package.

dwrensha:
Fixed by #1835.

Issue #953, opened 2015-10-01
Provide our own OAuth setup instructions


kentonv:
Meteor's instructions, at least for Google, are constantly out-of-date as Google is constantly changing their interface. We could respond faster to Google's interface changes if we controlled the instruction content directly.

paulproteus:
+1

Not to mention OW the Meteor ones are ugly.

dwrensha:
Fixed by #1835.

PR #1835, opened 2016-04-13
IdP settings: vendor in and improve Google/GitHub OAuth instructions

### LINKED ISSUE #1756: Google OAuth setup instructions need update
The instructions for Google OAuth setup do not match the Google menus and screens, as Google recently changed their interface.

zarvox:
We need to be able to keep this up to date in our releases,
since Google changes their workflow more frequently than
Meteor manages to update their setup instructions.

This change includes the copy updates from @ocdtrekkie aka Jacob Weisz from
https://github.com/meteor/meteor/pull/6757

This change strips the trailing slash from the Javascript origin in the Google
instructions, since Google's form complains if you include a path in that
scheme/host/port.

Fixes #1756, #1188, #953.

### DIFF
--- a/shell/client/admin/identity-providers.html
+++ b/shell/client/admin/identity-providers.html
@@ -54,6 +54,21 @@
 {{/modalDialogWithBackdrop}}
 </template>
 
+<template name="googleLoginSetupInstructions">
+  <p>
+    First, you'll need to get a Google Client ID. Follow these steps:
+  </p>
+  <ol>
+    <li>Visit <a href="https://console.developers.google.com/" target="blank">https://console.developers.google.com/</a></li>
+    <li>"Create Project", if needed. Wait for Google to finish provisioning.</li>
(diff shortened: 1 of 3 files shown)


zarvox:
Hmmm, strictly speaking, this only fixes the instructions in the new admin UI and setup wizard, so maybe this shouldn't close all those issues yet.  They should get cleaned up eventually, though, and our current process allows for closing other tickets before the fix is shipped in a release, so I don't feel *too* bad.

Also, big thanks @ocdtrekkie for coming up with up-to-date Google OAuth setup instructions!

paulproteus:
I agree that it's livable that the fix is, for now, only available in "new
admin". Having said that, I *believe* this also makes the new text part of
the admin initial onboarding flow? If so, I think it's also fair to call
this commit a fix to the issue, since that's where most people will see it.

zarvox:
Affirmative, this change does fix the text in the initial admin onboarding flow.

PR #1837, opened 2016-04-14
Clickable identity cards in share-by-wrong-identity.

### LINKED ISSUE #1624: In share-by-identity, it's possible to "lose" the grain someone shared with you
Steps to reproduce, according to @mrdomino:

- Create a grain

- Use share-by-identity to share with someone already in your contacts, call them Bob

- Bob clicks a link from an email, and opens the share link as one account and then sign in with another i guess not-yet-linked account

- Bob can wind up on the grains page without access to the grain that Bob thought was shared with you.

Clarifying question from me to @mrdomino :

- Can you explain step 3 further? Particularly, what does it mean to "open the share link as one account" - does that mean they started out _not_ logged in, then viewed the share link, then logged out, then logged in with a another not-yet-linked account?

dwrensha:
Alice shares to Bob by identity. Bob clicks the link in the email, but is either not signed in or is signed in to a different account. In this situation, we currently display an identity card for the shared-to identity and we indicate that Bob needs to sign is as that identity in order to access the grain. With the changes in this pull request, that identity card becomes a button. If Bob clicks on it, he initiates sign-in as that identity.

For example, for a github identity, the identity card button looks like this:

github

and one click is sufficient to sign in. For an email identity, the sign-in sequence looks like:

email
email1
email2

Fixes #1624.

This also fixes a few other things that I noticed were broken, such as linking of LDAP or SAML identities, and email notifications for share-by-identity to LDAP or SAML identities.

### DIFF
--- a/shell/server/grain-server.js
+++ b/shell/server/grain-server.js
@@ -121,10 +121,34 @@ Meteor.publish("tokenInfo", function (token) {
         let identity = globalDb.getIdentity(apiToken.owner.user.identityId);
         let metadata = apiToken.owner.user.denormalizedGrainMetadata;
         if (identity && metadata) {
+          const login = {};
+          // TODO(cleanup): This switching logic does not belong here. Perhaps it should go
+          //   in sandstorm-db/profile.js or be added as a method on Accounts.identityServices[].
+          if (identity.profile.service === "google") {
+            login.oneClick = { method: "loginWithGoogle",
+                               args: [{ loginHint: identity.services.google.email }], };
+          } else if (identity.profile.service === "github") {
+            login.oneClick = { method: "loginWithGithub", args: [] };
+          } else if (identity.profile.service === "email") {
+            login.form = { initializer: identity.services.email.email };
+          } else if (identity.profile.service === "ldap") {
+            login.form = { initializer: identity.services.ldap.username };
+          } else if (identity.profile.service === "saml") {
+            login.oneClick = { method: "loginWithSaml", args: [{ provider: "default" }] };
+          } else if (identity.profile.service === "dev") {
+            // TODO(cleanup): The dev login method is unusual in that it lives on the global
+            //   `window` object rather than the global `Meteor` object. Perhaps we should
+            //   move it to `Meteor`.
+            login.oneClick = { method: "loginDevAccount", args: [identity.services.dev.name],
+                               windowMethod: true, };
+          }
+
           this.added("tokenInfo", token,
                      { identityOwner: _.pick(identity, "_id", "profile"),
                        grainId: grainId,
-                       grainMetadata: metadata, });
+                       grainMetadata: metadata,
+                       login: login,
+                     });
         } else {
           this.added("tokenInfo", token, { invalidToken: true });
         }
@@ -440,6 +464,10 @@ Meteor.methods({
                 loginNote = "Github account with username " + intrinsicName;
               } else if (contact.profile.service === "email") {
                 loginNote = "email address " + intrinsicName;
+              } else if (contact.profile.service === "ldap") {
+                loginNote = "LDAP username " + intrinsicName;
+              } else if (contact.profile.service === "saml") {
+                loginNote = "SAML ID " + intrinsicName;
               } else {
                 throw new Meteor.Error(500, "Unknown service to email share link.");
               }
--- a/shell/server/grain-server.js
+++ b/shell/server/grain-server.js
@@ -121,33 +121,11 @@ Meteor.publish("tokenInfo", function (token) {
         let identity = globalDb.getIdentity(apiToken.owner.user.identityId);
         let metadata = apiToken.owner.user.denormalizedGrainMetadata;
         if (identity && metadata) {
-          const login = {};
-          // TODO(cleanup): This switching logic does not belong here. Perhaps it should go
-          //   in sandstorm-db/profile.js or be added as a method on Accounts.identityServices[].
-          if (identity.profile.service === "google") {
-            login.oneClick = { method: "loginWithGoogle",
-                               args: [{ loginHint: identity.services.google.email }], };
-          } else if (identity.profile.service === "github") {
-            login.oneClick = { method: "loginWithGithub", args: [] };
-          } else if (identity.profile.service === "email") {
-            login.form = { initializer: identity.services.email.email };
-          } else if (identity.profile.service === "ldap") {
-            login.form = { initializer: identity.services.ldap.username };
-          } else if (identity.profile.service === "saml") {
-            login.oneClick = { method: "loginWithSaml", args: [{ provider: "default" }] };
-          } else if (identity.profile.service === "dev") {
-            // TODO(cleanup): The dev login method is unusual in that it lives on the global
-            //   `window` object rather than the global `Meteor` object. Perhaps we should
-            //   move it to `Meteor`.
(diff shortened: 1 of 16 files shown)


shell/server/grain-server.js
```diff
@@ -121,10 +121,34 @@ Meteor.publish("tokenInfo", function (token) {
         let identity = globalDb.getIdentity(apiToken.owner.user.identityId);
         let metadata = apiToken.owner.user.denormalizedGrainMetadata;
         if (identity && metadata) {
+          const login = {};
+          // TODO(cleanup): This switching logic does not belong here. Perhaps it should go
+          //   in sandstorm-db/profile.js or be added as a method on Accounts.identityServices[].
+          if (identity.profile.service === "google") {
+            login.oneClick = { method: "loginWithGoogle",
```
kentonv: Hmm, specifying the exact client-side method to call in JSON from the server seems at best ugly, at worst scary.

As just one problem: Although we currently update the client and server in lockstep, eventually we may want to move towards making sure a new server is backwards-compatible with old clients, in order to avoid disrupting the user as much as we do today.

I think you should be returning a structure here that identifies the service to use by its regular service name ("google", "github", etc.) and has fields specifying any specific parameters needed to direct login, and then have a separate cascading if/else on the client side (or better yet, add a new method to the objects in `Accounts.identityServices`).

kentonv:
I think the UI needs a bit more explanatory text as to what is going on. Maybe @neynah and @zarvox could comment.

One thing that should be included: If you are currently logged in, the text should make clear that you'll be logged out to log in as the new account.

kentonv:
Hmm, if the identity is a non-login identity, then when I attempt to log in, I am immediately presented with a "that is not a login identity for any account" error. Three things about this:

1. This error dialog itself shows identity cards to suggest logging in. It should use your new code to make those cards clickable!
2. If I was already logged in before the interaction, maybe I should be given the option to link the identity to my account.
3. The wrong-identity error should probably be styled similarly to this non-login-identity error.

Note: I was originally going to suggest that the wrong-identity UI should immediately show me all the login identities I could use rather than directing me to log in as the non-login identity which we already know won't work. However, on further thought I think this could be too much of an information leak. E.g. I could share a grain to a pseudonym identity with leading text that causes them to reply to the invite, then I could click on the button and potentially see their real identity presented to me. So, I think it's necessary that the person prove that they control the target identity first before they can find out what login identities are associated with it.

dwrensha:
Updated. Now when you are logged in with the wrong identity, you see: 
logged-in

and when you are not logged in you see:
not-logged-in

The identity cards on the identity-login interstitial are now clickable. Looks like:
not-login

and if I click the email identity it looks like: 
not-login-expanded

shell/server/grain-server.js
dwrensha: I've addressed this by adding the methods `getLoginId()` and `initiateLogin()` to the objects in `Accounts.identityServices`.

neynah:
@dwrensha I'm working on some styling right now.

Could you provide a little more context on why users should also be able to unlink identities on this UI rather than just under "Account Settings"?

jparyani:
Garply, retest this please

neynah:
Is the "You have authenticated as this identity" identity clickable or is it just displayed for information purposes?

image

dwrensha:
> Could you provide a little more context on why users should also be able to unlink identities on this UI rather than just under "Account Settings"?

One scenario: someone other than you got hold of your credentials and linked the identity to their account. You've reset your credentials to lock them out, but you still need a way to unlink the identity from their account. This page is where you would do that.

dwrensha:
> Is the "You have authenticated as this identity" identity clickable or is it just displayed for information purposes?

Just for information purposes.

neynah:
> One scenario: someone other than you got hold of your credentials and linked the identity to their account. You've reset your credentials to lock them out, but you still need a way to unlink the identity from their account. This page is where you would do that.

If you unlink identities one by one on this page, you won't be able to unlink the remaining identity right? If so, we can show the same text "You are not allowed to unlink your only logged in identity".

I guess it's odd to me that even though you haven't successfully logged in with one of your login identities, you can still unlink any of the login identities. To me it seems like an action you should only be able to take if you are logged in.

If we are allowing unlinking of logged in identities from this page then I think we should make it clear that unlinking these identities will prevent you from logging in with these identities. I'll think of a way to display that.

kentonv:
@dwrensha It looks like there may be some more work to do styling this. Can you break out the LDAP/SAML fixes so we can commit those right away?

dwrensha:
> Can you break out the LDAP/SAML fixes so we can commit those right away?

Eh, I *could* do that. But I don't the they styling here as-is is a regression from what we currently have.

kentonv:
Yeah I guess you're right.

kentonv:
I'll merge this now, but let's keep thinking about how to make the UX clearer.

dwrensha:
@neynah: Maybe the confusing part here is that the "unlink" button has a different meaning from what it does on the account settings page. Here, "unlink" means "unlink my currently-authenticated identity from the account where this other identity can log in" --- it's quite a mouthful! The button has hover-text "Unlink your identity from this account", which I hope clarifies a bit.

Note that clicking "unlink" can in fact cause multiple of the cards in the list to go away, if they belong to the same account. It might be good to have some visual indication of this.

Once you've clicked all of the "unlink" buttons on this page, you've unlinked your identity from all accounts that have it linked. Then the page automatically creates a new account for the identity and logs you in to it. Maybe that's confusing too!

dwrensha:
@kenton: thanks!

neynah:
> @neynah: Maybe the confusing part here is that the "unlink" button has a different meaning from what it does on the account settings page. Here, "unlink" means "unlink my currently-authenticated identity from the account where this other identity can log in" --- it's quite a mouthful! The button has hover-text "Unlink your identity from this account", which I hope clarifies a bit.

That does, thanks. I told @kentonv I was also confused because all the David's below seemed legitimate! :) What do you think about "Unlink and revoke log in"?

> Note that clicking "unlink" can in fact cause multiple of the cards in the list to go away, if they belong to the same account. It might be good to have some visual indication of this.

If identity A, B, and C belong to the same account, unlinking any one of them will cause unlinking all of them? If so, maybe we should visually group them and show 1 "unlink" button for that group.

> Once you've clicked all of the "unlink" buttons on this page, you've unlinked your identity from all accounts that have it linked. Then the page automatically creates a new account for the identity and logs you in to it. Maybe that's confusing too!

Oh! Yeah that seems surprising. One option could be to add a string roughly saying: "Unlinking all these identities will require you to log in to a new account."

shell/shared/admin.js
```diff
@@ -80,8 +114,33 @@ Accounts.identityServices.saml = {
     return serviceEnabled("saml") && globalDb.isFeatureKeyValid();
   },
 
+  getLoginId: function (identity) {
+    return user.services.saml.id;
+  },
+
+  initiateLogin: function (loginId) {
+    Meteor.loginWithSaml();
```
kentonv: Turns out `loginWithSaml()` currently requires the parameter `{provider:"default"}`, so this doesn't work. Fixing...

neynah:
@dwrensha Not that this is urgent, but I figured I'd post it here anyway if we ever get around to any restyling.

- all cards share the same width & height except the buttony ones below have rounded edges
- also moved the "unlink" button to be below each card
- faded the border on the top identity card to deemphasize it (not necessary)
- added padding below the "Go Back" button

pull-1837

PR #1839, opened 2016-04-14
Autocomplete contacts inside an org

jparyani:
This is enabled by default, but can be disabled by an admin if needed.


shell/client/shell.html
```diff
@@ -1102,6 +1102,8 @@ <h4 class="share-with-others">Share with others</h4>
       <p>
         <label><input type="checkbox" name="disallowGuests" value="value" checked={{disallowGuests}}> Disallow collaboration with users outside the organization.</label><br>
         <span class="details">Users will be required to log in as a member of your organization to view sharing links. Guest access will be disabled.</span>
+        <label><input type="checkbox" name="autoShareContacts" value="value" checked={{autoShareContacts}}> Automatically share contacts between the organization.</label><br>
```
zarvox: I think that `value="value"` is spurious here (and in the checkbox above)

shell/server/admin-server.js
```diff
@@ -209,6 +209,7 @@ Meteor.methods({
       },
       settings: {
         disallowGuests: Boolean,
+        shareContacts: Boolean,
```
zarvox: Please update the setup wizard, lest this break initial setup.

PR #1856, opened 2016-04-15
Fix saml login

jparyani:
@dwrensha can you sign off on this? Also, let me know if you need help setting up SAML for local testing.


dwrensha:
Thanks. Looks good to me.

PR #1868, opened 2016-04-18
Implement reordering of grains in navbar by dragging

andresilva:
This PR adds reordering of grains in the navbar by drag-and-drop. It looks like this:

dragging-grains

The "shadow" element that's created is added to the body, so the inherited style from `navitem-grain` doesn't work, and I had to replicate the existing styles for the "shadow" element. This is obviously not a good solution and should be fixed (I don't know sass).

This is a feature that wasn't requested so I understand if you're not interested in merging this.

### DIFF
--- a/shell/client/styles/_topbar.scss
+++ b/shell/client/styles/_topbar.scss
@@ -62,56 +62,6 @@ body>.topbar-update {
   }
 }
 
-body>li.navitem-grain {
-  height: 32px;
-  line-height: 32px;
-  list-style-type: none;
-
-  a {
-    color: inherit;
-    display: block;
-    overflow-x: hidden;
-    overflow-y: hidden;
-    padding-left: 48px;
-    padding-right: 32px;
-    position: relative;
-    text-decoration: none;
-    text-overflow: ellipsis;
-    white-space: nowrap;
-  }
-
-  background-color: $topbar-background-color-hover;
-  color: $topbar-foreground-color-hover;
-
-  &.current {
-    background-color: $topbar-background-color-active;
-    color: $topbar-foreground-color-active;
-  }
-
-  .app-icon {
-    @extend %pseudo-img-tag;
-    position: absolute;
-    top: 4px;
-    left: (($navbar-width-desktop-shrunk - 24px) / 2);
-    width: 24px;
-    height: 24px;
-  }
-
-  .close-button {
-    @extend %unstyled-button;
-    position: absolute;
-    top: 1px;
-    right: 0px;
-    width: 24px;
-    height: 24px;
-    margin: 2px;
-    display: block;
-    background-image: url("/close.svg");
-    border: 1px transparent;
-    border-radius: 4px;
-  }
-}
-
 body>.topbar {
   position: fixed;
   top: 0;
--- a/shell/client/styles/_topbar.scss
+++ b/shell/client/styles/_topbar.scss
@@ -62,6 +62,56 @@ body>.topbar-update {
   }
 }
 
+body>li.navitem-grain {
+  height: 32px;
+  line-height: 32px;
+  list-style-type: none;
+
+  a {
+    color: inherit;
+    display: block;
+    overflow-x: hidden;
+    overflow-y: hidden;
+    padding-left: 48px;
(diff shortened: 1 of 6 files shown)


garply:

andrebeat:
Also the "shadow" element isn't properly styled when the navbar is shrinked (it would involve copying more css).

shell/packages/sandstorm-ui-topbar/topbar.js
```diff
@@ -44,6 +46,10 @@ Template.sandstormTopbarBlockReload.onDestroyed(function () {
   }
 });
 
+Template.sandstormTopbar.onRendered(function () {
+  this.drag = dragula([document.getElementById("navbar-grains")]);
```
zarvox: `onRendered()` gets called potentially many times for a single Template instance. Given that there's a cleanup hook in `onDestroyed`, does it make sense to check if `this.drag` is `undefined`, and if not, `destroy()` that instance before assigning the new one here?

kentonv:
Whoa, awesome! I've been wanting this.

Since this is a non-trivial contribution, can I get you to sign our CLA? Note that it is the same CLA used by the Apache Foundation so hopefully not objectionable. https://sandstorm.io/cla/

shell/client/styles/_topbar.scss
```diff
@@ -62,6 +62,56 @@ body>.topbar-update {
   }
 }
 
+body>li.navitem-grain {
```
zarvox: I'm curious why this block is needed.  Does `dragula` just move the original grabbed DOM node to the root of the document while it's in motion?

zarvox: Ahhh, missed the note below your screenshot in the PR cover message.  Never mind.

shell/packages/sandstorm-ui-topbar/topbar.js
andrebeat: I didn't know that. Fixed in e88dd35.

andrebeat:
I've changed the shadow element to be added to `navbar-grains` instead of `document.body`. This has the advantage of not requiring any CSS changes to work properly.

andrebeat:
@kentonv Done.

zarvox:
As I read more about how dragula works, I think we might need to add a hook to tell `GrainViewList` to reorder the `GrainView`s it contains, rather than just letting dragula rearrange the DOM on its own.  I think as-is this may do strange things on subsequent renders, which might happen when you open or close a grain, or otherwise trigger a navbar rerender.

I think we'd need something like: 
```javascript
// ...
this.drag = dragula( /* args as written in PR... */);
dragula.on("drop", function (el, target, source, sibling) {
  /* do something to reorder our JS representation of the tab order */
}));
```

andrebeat:
Yes, I agree it could lead to strange side-effects. I'll look into it.

andrebeat:
@zarvox I've updated the PR so that the JS representation is kept in sync with the DOM representation. It seems to be working since the order is now kept if you refresh the page.

kentonv:
Excited about this -- but currently it doesn't work correctly for me in Chrome 50 / Linux. When I start dragging something, I see the dragged copy appear lower down the page rather than under the cursor. (It moves as I move the cursor, it's just offset downwards for some reason.) Aside from rendering in the wrong place, the drag/drop action does appear to work.

screenshot from 2016-04-22 13-42-03

Also: If I begin dragging and then move the mouse cursor over the app's iframe and drop there, the release-button event seems to be lost, and the drag action keeps going. When I then return the mouse cursor to the sidebar, it acts like I still have the button held down. (To be fair, this kind of bug is incredibly common in apps that implement drag & drop in my experience.)

andrebeat:
I only tested on OSX on both Firefox and Chrome. I'll try to replicate on a Linux VM. Regarding the dragging over the iframe I think that's the downside of restricting the shadow element to `navbar-grains` instead of `document.body` as it was originally.

kentonv:
Hmm, it's extremely rare for Chrome to have a Linux-specific rendering bug. Am I really the only one seeing this in Chrome?

PR #1875, opened 2016-04-19
Fix powerbox offer of UiView

zarvox:
The addition of `lastUsed` to `user` in `ApiTokenOwner` caused the field to propagate to `createNewChildToken()` where the presence of its default-value caused a shape check to fail.

`lastUsed` is useful in contexts other than `ApiTokenOwner`, is not actually needed as a part of any of the protocols that use `ApiTokenOwner`, and is better off removed from the `.capnp` and added as a top-level field on `ApiTokens`.

Usually, removing a field from a `.capnp` would be **strictly forbidden**, but the commit message of dfc260cea1c7a24cbefb95aa0b87c8d7bf6f9347 explains why this is actually okay in this particular case.

### DIFF
--- a/shell/server/stats-server.js
+++ b/shell/server/stats-server.js
@@ -93,6 +93,7 @@ computeStats = function (since) {
 
     const counts = ApiTokens.aggregate([
       { $match: {
+          "owner.user": { $exists: true },
           lastUsed: timeConstraint,
           grainId: { $in: grainIds },
         },
--- a/shell/server/stats-server.js
+++ b/shell/server/stats-server.js
@@ -93,7 +93,7 @@ computeStats = function (since) {
 
     const counts = ApiTokens.aggregate([
       { $match: {
-          "owner.user.lastUsed": timeConstraint,
+          lastUsed: timeConstraint,
           grainId: { $in: grainIds },
         },
       },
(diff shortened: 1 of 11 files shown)


jparyani:
Garply, retest this please.

shell/server/stats-server.js
```diff
@@ -93,7 +93,7 @@ computeStats = function (since) {
 
     const counts = ApiTokens.aggregate([
       { $match: {
-          "owner.user.lastUsed": timeConstraint,
+          lastUsed: timeConstraint,
```
kentonv: Seems like we should now add `"owner.user": {$exists: true}` to this query, since it's intended to only count user-owned tokens?

zarvox: Good catch; fixed in afc9e166cd02898b3ce6ae901a859dc15893c71b

dwrensha:
> The addition of lastUsed to user in ApiTokenOwner caused the field to propagate to createNewChildToken() where the presence of its default-value caused a shape check to fail.

@zarvox: I'm having trouble following this. Can you link to where the exception was being thrown?

Do we have any tests for this stuff yet? Last I remember @zarvox had an app that could be used to manually test, but it wasn't hooked up to the integration tests. Is that still the situation?

zarvox:
https://github.com/sandstorm-io/sandstorm/blob/master/shell/packages/sandstorm-permissions/permissions.js#L455..L456 would fail because the user object contained `lastUsed: "0"` in addition to `identityId` and `title`, because the default value for the UInt64 `lastUsed` field is 0, but since it's a 64-bit int, it must be converted to a base-10 string to avoid loss of precision.

We do not have test automation for this at present.  I am given to understand that @kentonv wants to change how we do testing to reference fewer external artifacts and have a test app which is maintained more directly with the Sandstorm codeline; possibly within the Sandstorm repo itself.

kentonv:
@zarvox My hopes that we clean up our test code shouldn't stop us from writing additional tests in the meantime. :)

PR #1876, opened 2016-04-19
Use a longer WILDCARD_HOST self-test delay

paulproteus:
@zarvox and @dwrensha and @paulproteus have all experienced false-positives
on this self-test. Four seconds might be long enough.


paulproteus:
@kentonv , this is ready for your review/merge.

paulproteus:
Yay, thanks! Sorry to @zarvox that he noticed this a while ago and I only
now fixed it.

Issue #1065, opened 2015-10-21
more helpful [400] error text


jadeqwang:
Let's say I'm trying to share access to something that I don't have access to (total edge case, I know, but there might be other ways to trigger it). I get this error:

![image](https://cloud.githubusercontent.com/assets/1584904/10652136/d982f8fc-7808-11e5-97a5-e1b5e8c4c38e.png)

A more helpful error might be: "Error: [400] You do not have permission to share this grain."
That is, unless there are other cases where this error message comes up. @dwrensha ?

zarvox:
I think the clearer thing to do would be to disable the "Share access" forms entirely when you're looking at a view for a grain that you don't actually have access to.

jadeqwang:
:+1:

neynah:
Tested this today:
1. Attempted to access grain while logged out
2. Clicked "Share access"
3. I knew it wouldn't go anywhere useful but I copied the link since it told me to
4. Arrived at "Invalid token: null"

At the very least I think we shouldn't provide a link in this case.

![1](https://cloud.githubusercontent.com/assets/855341/14054959/00c54b88-f29b-11e5-9225-c377db0a5aef.jpg)
![2](https://cloud.githubusercontent.com/assets/855341/14054961/0487d16e-f29b-11e5-98d4-9df200cb48e8.jpg)
![3](https://cloud.githubusercontent.com/assets/855341/14054966/088d2b38-f29b-11e5-83ba-6e9ff96731f3.jpg)

Issue #1869, opened 2016-04-18
"undefined" in share link for anonymous users


mitar:
See `undefined` in the share link:

![sandstorm_build_wiki_ _alpha sandstorm io_-_2016-04-18_16 24 17](https://cloud.githubusercontent.com/assets/585279/14622888/28bf5cce-0582-11e6-9381-041cdee7d247.png)

dwrensha:
Good catch. This specifically happens for grains that still use the old sharing model.

zarvox:
I guess the right thing to do for old-sharing-model grains is to give the current route instead?  (Or migrate all known old-sharing-model grains to the new model.)

dwrensha:
If you're logged in, you see "This grain uses the old sharing model. You can share it by copy/pasting the URL from the location bar", which is reasonable.

ocdtrekkie:
Do we have a ballpark on what's actually left of the old sharing model grains? They predate Oasis, so AFAIK, it'd be entirely on Alpha and people's own self-hosted servers, right?

neynah:
I also found "null" in the share link (while logged out): https://github.com/sandstorm-io/sandstorm/issues/1065#issuecomment-201542062

paulproteus:
![screenshot from 2016-04-27 18 20 27](https://cloud.githubusercontent.com/assets/25457/14872830/e0f69dce-0ca4-11e6-9ff9-c04e29c25949.png)


There is no longer a "Share" button for anonymous users in this case, so closing the issue! Thanks @dwrensha for fixing it, presumably in #1879

mitar:
There is no title either anymore? ;-)

paulproteus:
In the case of viewing a grain to which you do not have access, while not logged in, yes, there is no title. But I think that is OK! :)

mitar:
My example was from: https://alpha.sandstorm.io/grain/kB7bofATTL2jGuKSFwmqdA/GithubPRCommands

Where you do have access.

paulproteus:
@mitar d'oh. I meant to leave the image as a comment on this bug: https://github.com/sandstorm-io/sandstorm/issues/1065#issuecomment-201542062

PR #1879, opened 2016-04-19
More sensible text in share-access popup for various corner cases.

dwrensha:
Fixes https://github.com/sandstorm-io/sandstorm/issues/1065 and https://github.com/sandstorm-io/sandstorm/issues/1869.

### DIFF
--- a/shell/client/grain-client.js
+++ b/shell/client/grain-client.js
@@ -463,8 +463,15 @@ Template.grainSharePopup.helpers({
     return !Accounts.getCurrentIdentityId();
   },
 
+  hasAccess: function () {
+    return !!Sessions.findOne(globalGrains.getActive().sessionId());
+  },
+
   currentTokenUrl: function () {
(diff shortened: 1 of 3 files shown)


dwrensha:
If you do not have permission to access the grain, the popup looks like this:
loggedout

If you are not logged in and the grain is using the old sharing model, the popup looks like this:
public

zarvox:
Thanks for looking into this. :)

Does it perhaps make more sense to hide the Share access button (and the webkey button?) entirely in these cases (or at least in the unauthorized-to-access-grain case)?  We already remove the rest of the grain-specific options in the toolbar that you can't make any use of if you're not the grain owner.

(Happy with the outcome either way.)

dwrensha:
I do think it makes more sense to hide the buttons. I've added a commit which does that.

paulproteus:
I would like to remark that this patchset is making my day. Yippee!

PR #1893, opened 2016-04-22
Improve grain renaming behavior. Fixes #1087

### LINKED ISSUE #1087: Grain titles being petnames confuses people
Currently, Sandstorm grain titles follow a petname model: when you receive a document shared from another user, initially the title is a copy of the title as seen by the sender, but subsequently you can change the title, and the changes are only visible to you (and people you later share with).

This is totally different from how most cloud services work, and it confuses people.

Petnames have some nice properties, though, and I'd be sad to simply throw them out. In particular:
* If someone sends me an overly-generically-titled grain (e.g. "My todo list", or "Untitled Etherpad"), I'd like to rename it for the purposes of my grain list.
* If I'm looking for a document that I've opened before, I'll probably be looking for the title I remember it having. If the owner can change that, I might be confused.
* There may be a phishing threat if a malicious user can rename a grain in my grain list to spoof some other grain that they know I have.

Note that petnames are analogous to what happens when someone e-mails you a file: the filename is initialized based on the sender's filename, but once on your system, the sender can't change it, only you can. Another analogy is to variable names in a programming language: when a value is passed from one function to another, the name often changes, and this is considered a good thing, since the name is meant to identify the value *relative to* the context.

But that's all theory. Back here in reality, it seems that users do expect that the title (as shown in the top bar, at least) is part of the grain's _content_ and therefore is up to the grain owner to choose and change at will. We've seen plenty of examples of people being confused by the current behavior -- clearly it needs to change somehow. The question is, can we come up with a compromise that keeps the nice properties above but also makes sense to people?

Some questions to think about:
* Do people have different expectations for the topbar title vs. the grain list title? My issues mostly revolve around how the grain is shown in the grain list, but I think most of the complains focus on what appears in the topbar when the grain is open. Could these be disassociated somehow?
* Could we solve my problems by showing more information in the grain list and top bar that provide sufficient context to avoid confusion? For example:
    * What if we always show the author's name? "From Bob: My todo list" largely solves the problem of bad/ambiguous titles. It solves the phishing problem *if* we separately prevent people from spoofing names, which may require a petname system for identities, but that's something we probably want to do anyway. There is a possible problem, though, in that it's not clear that we are always allowed to reveal owners' identities to sharing recipients.
    * Could we solve the searchability problem by matching search queries against all of the document's previous titles? Could we show "Was: [old title]" in the UI somewhere?
* Should recipients still be given the option to specify their own title if they actively choose to do so?

kentonv:
Historically, when a grain is shared, the title would be copied and become the receiver's "petname" for the grain, displayed whenever the receiver opens the grain. If the owner renamed the grain post-sharing, the receiver would never see the rename, only their petname. If the receiver renamed the grain, it would change their petname. This was generally confusing to people.

However, the alternative of "always show the owner's title" is also problematic since:
- If the title changes in the recipient's grain list, they may be confused as to how this new grain that they don't remember got here.
- If the owner's title is poorly-chosen or similar to other grains in the receiver's grain list, the receiver may want to retitle the grain for themselves (hence the petname behavior).

This change implements a compromise by displaying both the petname and the owner's title:
- If the petname and original title are the same, we just display it.
- If the sharing recipient has explicitly modified their petname, we show their petname as the title, with subtext of "renamed from: [upstream title]". This allows the user to assign their own title if they need to disambiguate the grain, though most of the time they probably won't.
- If the sharing recipient has not explicitly modified their petname (i.e. it is a clone of the title from the time of sharing), but the upstream title has changed, we show the upstream title, with subtext of "was: [petname]". This allows the user to understand how this grain whose title they don't recognize arrived in their grain list.

### DIFF
--- a/shell/packages/sandstorm-db/migrations.js
+++ b/shell/packages/sandstorm-db/migrations.js
@@ -556,6 +556,45 @@ function extractLastUsedFromApiTokenOwner() {
   });
 }
 
+function setUpstreamTitles() {
+  // Initializes the `upstreamTitle` and `renamed` fields of `ApiToken.owner.user`.
+
+  const apiTokensRaw = ApiTokens.rawCollection();
+  const aggregateApiTokens = Meteor.wrapAsync(apiTokensRaw.aggregate, apiTokensRaw);
+
+  // First, construct a list of all *shared* grains. We will need to do a separate update()
+  // for each of these, so we'd like to skip those which have nothing to update.
+  const grainIds = aggregateApiTokens([
+    { $match: { "owner.user": { $exists: true }, grainId: { $exists: true } } },
+    { $group: { _id: "$grainId" } },
+  ]).map(grain => grain._id);
+
+  let count = 0;
+  Grains.find({ _id: { $in: grainIds } }, { fields: { title: 1 } }).forEach((grain) => {
+    if (count % 100 == 0) {
+      console.log(count + " / " + grainIds.length);
+    }
+
+    ++count;
+
+    // For ApiTokens whose petname titles do not match the upstream title, we need to set
+    // `upstreamTitle` and `renamed`. We have no way of knowing which of the two names was the
+    // original name, so we don't know whether it was the owner or the receiver who renamed their
+    // copy post-sharaing. We assume it was the owner, because:
+    // 1. It's probably unusual for people to try to rename a grain they don't own.
+    // 2. This results in UI that is reasonably non-confusing even if we guessed wrong. Namely,
+    //    the user sees "Owner's title (was: User's title)", which is reasonably OK even if it
+    //    was the user who had renamed their copy. They can rename it again if they like. On the
+    //    other hand, if we guessed wrongly in the other direction, the user would see
+    //    "User's title (renamed from: Owners title)", which would be wrong if it was in fact the
+    //    owner who renamed post-sharing.
+    ApiTokens.update({
+      grainId: grain._id,
+      "owner.user.title": { $exists: true, $ne: grain.title },
+    }, { $set: { "owner.user.upstreamTitle": grain.title } }, { multi: true });
+  });
+}
+
 // This must come after all the functions named within are defined.
 // Only append to this list!  Do not modify or remove list entries;
 // doing so is likely change the meaning and semantics of user databases.
@@ -585,6 +624,7 @@ const MIGRATIONS = [
   consolidateOrgSettings,
   unsetSmtpDefaultHostnameIfNoUsersExist,
   extractLastUsedFromApiTokenOwner,
+  setUpstreamTitles,
 ];
 
 function migrateToLatest() {
--- a/shell/server/grain-server.js
+++ b/shell/server/grain-server.js
@@ -335,6 +335,9 @@ Meteor.methods({
       if (grain) {
         if (grain.userId === this.userId) {
           Grains.update({ _id: grainId, userId: this.userId }, { $set: { title: newTitle } });
+
+          // Denormalize new title out to all sharing tokens.
+          ApiTokens.update({ grainId: grainId }, { $set: { "owner.user.upstreamTitle": newTitle } });
         } else {
           if (!globalDb.userHasIdentity(this.userId, identityId)) {
             throw new Meteor.Error(403, "Current user does not have identity " + identityId);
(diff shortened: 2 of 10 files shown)


shell/server/grain-server.js
```diff
@@ -335,6 +335,9 @@ Meteor.methods({
       if (grain) {
         if (grain.userId === this.userId) {
           Grains.update({ _id: grainId, userId: this.userId }, { $set: { title: newTitle } });
+
+          // Denormalize new title out to all sharing tokens.
+          ApiTokens.update({ grainId: grainId }, { $set: { "owner.user.upstreamTitle": newTitle } });
```
zarvox: Probably you only want to do this for grains where `"owner.user": { $exists: true }`

kentonv:
screenshot from 2016-04-22 11-06-27
screenshot from 2016-04-22 11-07-06
screenshot from 2016-04-22 11-07-02

shell/server/grain-server.js
kentonv: Good catch, thanks.

paulproteus:
One item discussed in https://github.com/sandstorm-io/sandstorm/issues/1087 is a Google Calendar-style pop-up when a user is about to rename something in a way that prevents them from getting updates from the item owner. I don't see any code in this PR to address that; is that intended to indicate that we don't want to have that pop-up/notification/warning?

If the pop-up/notification/warning is desired, but out-of-scope for now for expediency, then I can file an issue to track that change whenever we get to it.

Generally, I'm excited about this change!

kentonv:
@paulproteus It turns out that the popup for grain renames already says: "Set a new personal title (does not change the owner's title for this grain)" I didn't feel that it was necessary to change it, but if you'd like to change it then feel free to create a new PR for that. :)

The popup is a `window.prompt()` currently which is pretty sad but designing a better popup is I think definitely out of scope for this PR.

paulproteus:
+1 I think that is fine!​ If I can come up with a better string I can
submit a PR. I guess I forgot what it said, which shows how often I rename
grains other people share with me.

neynah:
I would suggest adding a little extra padding between the "was" grain name and the bottom of the black bar since its close proximity to the bottom of the bar makes it harder to read. I'd also slightly deemphasizing the previous grain title further by changing the font colour to something like #9b9b9b so that it's not overwhelming but still readable.

By adding the extra string of text, the top bar does feel a bit inconsistent since the grain name is no longer aligned with "Sandstorm", "Share access" and "Kenton Varda" but if we make those padding tweaks it should help a bit. The "Share Access" & purple notification icons also seem too large and inconsistent with the other icon sizes but that's a separate issue. :)

improve-grain-renaming-behaviour

ocdtrekkie:
Stupid question: Is it obvious when the upstream renames a grain that anyone downstream from them will see it? (Is this even important?) Like if I had shared a grain with my boss, and then later renamed it "Retard Idea" because I thought the idea was bad.

I would not do this, but I can imagine there are people who will.

Does anyone care? Does my question matter?

dwrensha:
Can a user ever own a grain-hosted non-main-uiview capability? If so, what do the new fields in `ApiTokenOwner.user` mean in that case?

kentonv:
@ocdtrekkie Interesting thought, but my guess is that most users expect that titles are synchronized, and those that don't expect it would at least anticipate it as a possibility. I'm going to say this is a non-issue for now, but we can revisit if it is observed in the wild.

kentonv:
@dwrensha We don't support non-main UiViews yet. I'm actually totally unsure how titles will work on such UiViews. Would they be based on the grain's main title, or would they be independent? If independent, who is the "owner" who can set the title? Is it the owner of the main grain, or is the non-main UiView "owned" by the person who created it (who might not be the main grain owner)? I think, though, that once we answer those questions, the new fields in `ApiTokenOwner.user` would have a straightforward meaning relative to the sub-view's "title", whatever that may be.

dwrensha:
@kentonv I suppose the same answer applies to capabilities that aren't UiViews at all?

kentonv:
@dwrensha Capabilities that aren't UiViews at all cannot have `user` as the owner, at least in our current plan. Having `user` as the owner implies appearing in the grain list. Non-UiView capabilities are passed grain-to-grain via the Powerbox, but never go directly into the grain list.

shell/packages/sandstorm-db/migrations.js
```diff
@@ -556,6 +556,45 @@ function extractLastUsedFromApiTokenOwner() {
...
 
+function setUpstreamTitles() {
+  // Initializes the `upstreamTitle` and `renamed` fields of `ApiToken.owner.user`.
+
+  const apiTokensRaw = ApiTokens.rawCollection();
+  const aggregateApiTokens = Meteor.wrapAsync(apiTokensRaw.aggregate, apiTokensRaw);
+
+  // First, construct a list of all *shared* grains. We will need to do a separate update()
+  // for each of these, so we'd like to skip those which have nothing to update.
+  const grainIds = aggregateApiTokens([
```
dwrensha: Would this be simpler with [`distinct()`](https://mongodb.github.io/node-mongodb-native/api-generated/collection.html#distinct)?

dwrensha: (maybe not worth changing if you've already tested this as-is)

dwrensha:
This looks good to me.

shell/packages/sandstorm-db/migrations.js
kentonv: TIL. But yeah, would be annoying to test again and it's a one-off anyhow.

elimisteve:
> ...my guess is that most users expect that titles are synchronized

Not that anyone asked, but to me it seems strange that they _wouldn't_ be synchronized.  If I share a Google Doc with someone, the idea that we're sharing one and the same document, but somehow there are 2 different titles, seems bizarre.  On Sandstorm this seems equally bizarre, though I can see how coming back to the grain list and not recognizing a title (because it was changed by someone who changed it) is a motivating example.

kentonv:
@elimisteve See extended discussion on #1087 for background here.

PR #1896, opened 2016-04-23
Fix various corner cases with grain renaming.

dwrensha:
We should actually write `upstreamTitle` when initially creating tokens.

A user can have multiple incoming tokens for a single grain. The `lastUsed` update in sandstorm-backend.js needs `{ multi: true }`, or else it just picks a single token nondeterministically.

The token used in the grainlist should be the oldest, not the most recently used. It would be weird if merely receiving a new share caused the title to change.

When the user updates the title on an incoming token, all incoming tokens for that grain should get renamed, for the sake of robustness.


kentonv:
Nice catches.

PR #1905, opened 2016-04-25
Docs talk about the wiki less

paulproteus:
Thanks @neynah for pointing out that this is a smart change to make.


paulproteus:
Self-merged because 100% docs

PR #1907, opened 2016-04-25
Prevent demo users from sharing by email. Increment dailySentMailCount when sharing by email.

### LINKED ISSUE #1906: Disable e-mail sharing for demo users and rate-limit for everyone else
It looks like someone has discovered that they can send spam messages through the Sandstorm sharing dialog. :(

We should:
* Disallow demo users from sending sharing e-mail. They'll need to share by link only.
* Rate-limit sharing emails sent by real users, probably using the same mechanism as we use to rate-limit email from apps.

This is somewhat of an emergency as the longer this spammer continues their activity the more reputation we lose.

dwrensha:
Closes #1906.

demo user

### DIFF
--- a/shell/client/grain-client.js
+++ b/shell/client/grain-client.js
@@ -1019,6 +1019,10 @@ Template.shareableLinkTab.helpers({
   },
 });
 Template.emailInviteTab.helpers({
+  isDemoUser: function () {
+    return globalDb.isDemoUser();
+  },
+
   completionState: function () {
     const instance = Template.instance();
     return instance.completionState.get();
--- a/shell/client/shell.html
+++ b/shell/client/shell.html
@@ -548,8 +548,12 @@ limitations under the License.
     {{> selectRole viewInfo=viewInfo}}
     </p>
     <div>
+      {{#if isDemoUser}}
+      <textarea class="personal-message" disabled="true">Demo users cannot send sharing e-mails for spam reasons. If you create an account, you'll be able to use this UI. Otherwise, try creating a link.</textarea>
+      {{else}}
       <textarea class="personal-message"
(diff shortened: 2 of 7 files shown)


neynah:
I would probably display it as help/detail text rather than having it in a box and change it to something simpler like:

"Sharing emails cannot be sent while in demo mode. Create an account to send sharing emails, or get a shareable link instead."

I don't think mentioning spam enhances the demo-users experience.

dwrensha:
@neynah more like this?

simpler

neynah:
Can you match the bottom padding of that text with the top padding? Otherwise, +1! :)

kentonv:
Hmm, but I like having the UI shown (but disabled) so that people can see what they would be getting with a full account.

How about:
* Place the text in a red box (styled like the error messages in the admin UI).
* Show the regular sharing UI -- but with all inputs disabled -- under the error message.

neynah:
Just talked to Kenton in person & I agree that his suggestion is an improvement. :+1:

dwrensha:
I pushed a commit that looks like:
red box

kentonv:
Hmm, almost there, but should the input boxes be gray or something so that they look more "disabled"?

dwrensha:
pushed: grayed

kentonv:
Great!

ocdtrekkie:
Should the "get shareable link" tab be the default one when you expand this tab, if sharing by email is disabled? So the currently available sharing feature is most visible to the user when they're testing out Sandstorm?

kentonv:
@ocdtrekkie has a good point.

I'm planning to push a hotfix later tonight. If a change to the default tab (for demo users) arrives before that I'll include it, otherwise it'll be in the next regular weekly release.

PR #1914, opened 2016-04-27
Whitelist Mercurial headers in web sessions

### LINKED ISSUE #1897: Feature Request: Whitelist Mercurial HTTP headers
I would like to package Mercurial as a Sandstorm app.  The Mercurial wire protocol sends most of its data using HTTP headers.  They follow the pattern `X-HgArg-N`, where `N` is an integer greater than zero.  Each header payload will be ~1KB by default, so you can pick an upper bound on how high you want N to be.

(I tried hacking around this by disabling this mode of operation, but it blows out the URL length quickly, and it doesn't appear to have a retry mechanism.)

zombiezen:
Fixes #1897

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -1526,7 +1526,18 @@ class Proxy {
 
     context.additionalHeaders = [];
     WebSession.Context.headerWhitelist.forEach((headerName) => {
-      if (request.headers[headerName]) {
+      if (headerName.endsWith('*')) {
+        const prefix = headerName.substr(0, headerName.length-1);
+        for (const h in request.headers) {
+          if (!h.startsWith(prefix)) {
+            continue;
+          }
+          context.additionalHeaders.push({
+            name: h,
+            value: request.headers[h],
+          });
+        }
+      } else if (request.headers[headerName]) {
         context.additionalHeaders.push({
           name: headerName,
           value: request.headers[headerName],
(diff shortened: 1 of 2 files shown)


garply:

paulproteus:
Garply, ok to test

paulproteus:
Ross, do you happen to know if 16 is the max number of headers that the
Mercurial client sends? If that's not a maximum, then we'll probably
confuse it by accepting only the first 16 and dropping the rest.

Thanks for writing this!

zombiezen:
AFAICT, it's unbounded ([source](https://selenic.com/hg/file/87d4a6c5567e/mercurial/httppeer.py#l128)).  However, it does seem like [newer clients](https://www.mercurial-scm.org/pipermail/mercurial-devel/2016-March/080830.html) may start supporting using POST body as arguments, but this will take a while to get to all downstream package repos etc.

I picked 16 as 16KB seems to be a common header size limit (at least for [IIS](https://www.mercurial-scm.org/wiki/HgWebInIisOnWindows)).  The size of the header payload appears to be proportional to the number of "heads" -- leaf nodes in the history DAG.  For most repositories, this will be only a few.

kentonv:
Maybe we should extend the header whitelist to support a trailing wildcard? Wouldn't be too hard...

paulproteus:
Kenton, I think your idea is a fine one. Ross, are you possibly interested
in looking at the C++ changes needed to implement it? If not, then we
should figure out the best way to support hg within grains. Let me know.

zombiezen:
I'm not entirely sure how to compile/test this myself, but this should be the general patch.

kentonv:
Thanks!

Issue #1916, opened 2016-04-28
error when running `run-dev.sh`: Assertion `loop->watchers[w->fd] == w' failed.


dwrensha:
```
node: ../deps/uv/src/unix/core.c:701: uv__io_stop: Assertion `loop->watchers[w->fd] == w' failed.
```

Looks like this is somehow related to https://github.com/sandstorm-io/sandstorm/pull/1877.

paulproteus:
Thanks for catching that. I'll fix that shortly.​

PR #1917, opened 2016-04-28
Fix SMTP listening in run-dev mode.

dwrensha:
Fixes https://github.com/sandstorm-io/sandstorm/issues/1916.

### DIFF
--- a/meteor-bundle-main.js
+++ b/meteor-bundle-main.js
@@ -18,6 +18,7 @@ var url = require('url');
 var forge = require('./programs/server/npm/node_modules/meteor/meteor-node-forge/node_modules/node-forge');
 
 function sandstormMain() {
+  global.SANDSTORM_SMTP_LISTEN_HANDLE = { fd: 3 };
   monkeypatchHttpAndHttps();
 
   // Delegate to Meteor.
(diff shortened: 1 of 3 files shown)


dwrensha:
In run-dev mode, [meteor-bundle-main.js](https://github.com/sandstorm-io/sandstorm/blob/80e91c59f8d7b62e8700a4bcef9306a31026427c/meteor-bundle-main.js) does not get run at all.

paulproteus:
One problem is that this will crash if SMTP_LISTEN_PORT=25, right? But maybe that is just life.

paulproteus:
I'll manually test my guess now, one sec.

paulproteus:
In manual testing with this patch:

- With `SMTP_LISTEN_PORT=25` in sandstorm.conf: `run-dev.sh` now starts up OK, but doesn't listen on port 25. I can live with that.

- With `SMTP_LISTEN_PORT=30025` in sandstorm.conf: `run-dev.sh` now starts up OK and does in fact listen on port 30025.

So I'm +1 on this PR. Thanks @dwrensha for writing it.

Non-showstopper feedback:

- It would be nice run-dev.sh were to open the port and pass it in, as an fd, so that the code wouldn't have to be aware of `run-dev.sh`. But I don't think that's crucial.

kentonv:
Turns out this was also necessary for blackrock, btw!

PR #1923, opened 2016-04-29
Small e-mail fixes

mitar:
Just some things I noticed while reading through the code.

Not tested.

### DIFF
--- a/shell/server/drivers/mail.js
+++ b/shell/server/drivers/mail.js
@@ -216,13 +216,15 @@ hackSendEmail = (session, email) => {
           "Please feel free to contact us if this is a problem for you.");
     }
 
-    // Overwrite the 'from' address with the grain's address.
-    if (!email.from) {
-      email.from = {};
-    }
-
     const grainAddress = session._getAddress();
     const userAddress = session._getUserAddress();
+
+    // Overwrite the 'from' address with the grain's address.
+    if (!email.from) {
+      email.from = {
+        address: grainAddress
+      };
+    }
 
     // First check if we're changing the from address, and if so, move it to reply-to
     if (email.from.address !== grainAddress && email.from.address !== userAddress.address) {
@@ -260,7 +262,7 @@ hackSendEmail = (session, email) => {
       mc.addHeader("references", email.references);
     }
 
-    if (email.messageId) {
+    if (email.inReplyTo) {
       mc.addHeader("in-reply-to", email.inReplyTo);
     }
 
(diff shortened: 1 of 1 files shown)


garply:

kentonv:
Hmm, it's clear the code does not match the comments, but I wonder if it is the code or the comments that is wrong. Maybe we actually intended to require that the grain specify an address matching one of the two, without any default? @jparyani do you remember?

(The unrelated `inReplyTo` change is clearly a bug, of course.)

mitar:
In documentation it says:

> When you send e-mail from a Sandstorm app, we allow you to set the "From" header either to your app's random address or to your verified Sandstorm login address. 

It  does not "feel" like you have to set it, but you are allowed to set it?

kentonv:
Yeah that line is really ambiguous, but the documentation was meant to be descriptive, not prescriptive, so that doesn't help us figure out what was the original intended behavior here...

mitar:
How does app figure out what is its valid outgoing address? For example, how can a Meteor app know that? Is this easy? Then we can require it to set the right from address.

kentonv:
Apps normally call `HackSessionContext#getPublicId()` and then form the address as publicId@hostname.

https://github.com/sandstorm-io/sandstorm/blob/master/src/sandstorm/hack-session.capnp#L35

kentonv:
Or to get the user's address, `HackSessionContext#getUserAddress()`:

https://github.com/sandstorm-io/sandstorm/blob/master/src/sandstorm/hack-session.capnp#L63

Keep in mind `HackSessionContext` is a born-deprecated API.

mitar:
Hm, and that is not available for Meteor apps, no? Is it even possible to send e-mails from Meteor apps?

kentonv:
You can use node-capnp to make Cap'n Proto requests from Meteor apps.

The documentation describes how to get at HackSessionContext: https://docs.sandstorm.io/en/latest/developing/email-from-apps/

shell/server/drivers/mail.js
```diff
@@ -216,15 +216,17 @@ hackSendEmail = (session, email) => {
...
-      email.from = {};
+      email.from = {
+        address: grainAddress
+      };
     }
 
-    const grainAddress = session._getAddress();
-    const userAddress = session._getUserAddress();
-
-    // First check if we're changing the from address, and if so, move it to reply-to
```
mitar: I also removed this comment. It is not true at all.

mitar:
There could be a Meteor package for that. ;-)

kentonv:
Thanks!

Issue #1918, opened 2016-04-29
Support for <grainid>+<custom>@alpha.sandstorm.io e-mail addresses


mitar:
It would be great if all inbound and outbound e-mails could also use `+<custom>` suffixes to the address. Then app/user could use this to make even finer control to which outside service it gives which e-mail address. For example, imagine an mailing list reader, or reusable e-mail addresses app, it could create custom e-mail addresses every time and then know where the e-mail is coming from.

paulproteus:
@kentonv would you be willing to provide a +1/-1 for if you're OK with `publicId+suffix` email addresses, as well as `publicId` email addresses?

As a clarification for @mitar it's not {{grainId}} before the @ sign currently, but instead {{publicId}}.

To some extent this is documented at https://docs.sandstorm.io/en/latest/developing/email-from-apps/ , but that page mentions `publicId` 0 times, so I should probably improve it.

mitar:
Ah, yes, this is also OK with me. :-)

kentonv:
Sounds fine to me, although these ephemeral emails are a hack that should go away entirely at some point...

paulproteus:
Great!

Also: what's wrong with these ephemeral email addresses? I keep wanting to
make a mailinator replacement based on them.​ (Although arguably the
Roundcube app already is that, but anyway.)

mitar:
Cool. What is the planned replacement for them?

kentonv:
The replacement is some UI that lets you choose pretty addresses, perhaps at a separate domain which is also somehow managed through Sandstorm. :)

mitar:
But ephemeral addresses which can be used without user interaction are useful as well. But yes, for meaningful addresses users should be consulted.

paulproteus:
@zarvox remarked to me in person that the app offering an `EmailSendPort` somehow would be more idiomatic, but I am OK with the pull request getting merged.

mitar:
> that the app offering an EmailSendPort somehow would be more idiomatic

I completely do not understand what you are saying here. Care to rephrase it?

PR #1925, opened 2016-04-29
Email suffix

mitar:
Addressing #1918.

It assumes #1923 will be merged so it contains those commits as well (it was not really possible to separate changes).

(Not yet tested, I have to figure out how to send and receive test e-mails from Sandstorm.)


garply:

shell/server/drivers/mail.js
```diff
@@ -116,7 +119,13 @@ Meteor.startup(function () {
           // there will be an nginx frontend verifying hostnames anyway. Grain public IDs are
           // globally unique anyway, so an e-mail meant for another server presumably won't match
           // any ID at this one anyway.
-          return deliverTo.slice(0, deliverTo.indexOf("@"));
+          const match = PUBLIC_ID_REGEX.exec(deliverTo);
+          if (match) {
+            return match[1];
+          }
+          else {
+            throw new Error("Invalid recipient e-mail address: " + deliverTo);
```
mitar: We could also just return `null` which would then more or less ignore this recipient and log an error.

kentonv:
Hey @mitar, can you sign our CLA?

https://sandstorm.io/cla/

(It's the same CLA used by the Apache Foundation, FWIW.)

mitar:
Done.

shell/server/drivers/mail.js
```diff
@@ -116,7 +119,13 @@ Meteor.startup(function () {
           // there will be an nginx frontend verifying hostnames anyway. Grain public IDs are
           // globally unique anyway, so an e-mail meant for another server presumably won't match
           // any ID at this one anyway.
-          return deliverTo.slice(0, deliverTo.indexOf("@"));
+          const match = PUBLIC_ID_REGEX.exec(deliverTo);
+          if (match) {
+            return match[1];
+          }
+          else {
```
kentonv: nitpick: Please have `} else {` all on one line.

mitar: Does your linting find these?

garply:

shell/server/drivers/mail.js
kentonv: `make stylecheck` runs JSCS for linting, but I don't know exactly what it catches.

mitar:
Rebased on master.

shell/server/hack-session.js
```diff
@@ -33,6 +33,8 @@ const Grain = Capnp.importSystem("sandstorm/grain.capnp");
 
 const Url = Npm.require("url");
 
+const escapeRegExp = require('lodash.escaperegexp');
```
mitar: Not completely sure if this is how one should do it in new Meteor. But I will test it once I get to run it. :-)

mitar: (Or do you have some other way to get NPM packages on the server side. In my Meteor apps I use only packages, never top-level `client`, `server` directories, then it is easy to depend on NPM packages from Meteor packages.)

kentonv:
I feel like the regexes are overkill here and are complicated enough to be difficult to read and reason about. Plus bringing in a regex escaping library as a dependency feels heavyweight and risky (if it has a bug, does it become a security problem?).

Can we instead write a function like:

    function removePlusSuffix(addr) {
      const parts = addr.split('@');
      parts[0] = parts[0].split('+')[0];
      return parts.join("@");
    }

?

mitar:
Regex is also used to check if e-mail address is valid or not, and if envelope should be replaced or not.

If you worry about escaping, then I would suggest I just escape string by replacing all characters one by one with their unicode escape sequence (`\uhhhh`). This will make regex uglier, but it is completely fail safe.

kentonv:
I would be much more comfortable using a regex specifically to check whether the suffix is valid. The rest of the email has to be an exact match anyway.

    function removeValidPlusSuffix(addr) {
      const parts = addr.split('@');
      const plusParts = parts[0].split('+');
      if (plusParts.length > 2 ||
          (plusParts.length == 2 && !plusParts[1].match(/^[a-zA-Z0-9_.-]+$/))) {
        throw new Error("Email +suffix can only contain letters, numbers, ., _, and -.");
      }
      parts[0] = plusParts[0];
      return parts.join("@");
    }

mitar:
Hm, I am not checking only if suffix is valid. I will annotate inline.

shell/server/drivers/mail.js
```diff
@@ -226,9 +236,9 @@ hackSendEmail = (session, email) => {
       };
     }
 
-    if (email.from.address !== grainAddress && email.from.address !== userAddress.address) {
+    if (!allowedOutgoingAddressesRegex.test(email.from.address) && email.from.address !== userAddress.address) {
```
mitar: So primary use of regex is here.

shell/server/drivers/mail.js
```diff
@@ -246,7 +256,12 @@ hackSendEmail = (session, email) => {
     });
 
     const envelope = mc.getEnvelope();
-    envelope.from = grainAddress;
+    // We allow envelope address to have a suffix as well. In this way apps can use unique envelope
+    // addresses when sending out e-mails which allow better handling of e-mail bounces - app can
+    // match bounce with the sent e-mail.
+    if (!allowedOutgoingAddressesRegex.test(envelope.from)) {
```
mitar: And here. This is why regex is the easiest to use.

mitar:
I removed the dependency and make regex escaping bullet-proof and future-proof. :-)

paulproteus:
@kentonv I believe this deserves a fresh look.

paulproteus:
Hi @mitar ,

Thanks for submitting this. I chatted with @kentonv about it, and here's my understanding of what's needed.

- Avoid `(?:` in regular expressions. Rationale: It's an unusual escape sequence, which makes it harder for other developers less familiar with regular expressions to understand the meaning of this code. In this case, since `(?:` means a group that isn't captured, it would presumably be straightforward to make it a capturing group, and ignore it upon processing the results of the regular expression.
- Avoid creating regular expressions at runtime, even if we are 100% sure that they are the right regular expression. Rationale: The escaping logic is tricky, and so for aesthetic reasons, we'd prefer not to have that in the Sandstorm codebase.

This suggests the following change:

- In `shell/server/drivers/mail.js`, avoid the non-capturing group. Also for readability, consider using `[+]` instead of `\\+` since the former is easier to reason about in my personal opinion. Alternatively, write a function that calls `indexOf("+")` as well as `indexOf("@")`, and returns whichever is lower.

- In `shell/server/hack-session.js`, rename `_getAllowedOutgoingAddressesRegex()` to `_getAllowedOutgoingAddressesChecker()` and return a function rather than a regular expression.

- Implement the function without using dynamically-created regular expressions, e.g. string splitting or looking at string indices and/or doing equality checks.

- Remove `escapeRegExp()` from `shell/server/hack-session.js`.

I recognize that regular expression may be better for performance, but we have these aesthestic preferences, at least for now.

**Unrelated docs request:** Since I'm the owner of the docs, I also have a request in the documentation part of this patch. The main change you made was to add this sentence:

> Moreover, your app will receive any e-mail to an address with an arbitrary suffix, like "JBuaKxjkwiJq7oksS+foobar@alpha.sandstorm.io".

I would propose that you:

- Change the text to mention "sub-addressing" or "plus-addressing". This allows someone skimming the document to have a term to Google for.

- Move that to its own paragraph, immediately after the paragraph it's in now. Making the discussion of sub-addressing later will, I believe, help people skim the documentation more effectively. In the docs as they exist in your patch, it's easy to think that sub-addressing is essential for setting up inbound email in my opinion.

Proposed text:

> These Sandstorm email addresses support sub-addressing using the `+` character. For example, if an app can receive email at JBuaKxjkwiJq7oksS@alpha.sandstorm.io, then it can also receive email at JBuaKxjkwiJq7oksS+foobar@alpha.sandstorm.io. This allows app authors to generate email addresses that are specific to a particular purpose.

Thanks for your patience with us on this long review!

mitar:
> I recognize that regular expression may be better for performance, but we have these aesthestic preferences, at least for now.

Sorry, but for me this is a no go. It is not about performance. It is about security. I am a computer security person. And I do not believe in re-implementing from scratch popular algorithms. Manual parsing of strings is exactly what makes C problematic. And regular expressions make it much cleaner and predictable. You have a semantic description of behavior and you leave to the engine to execute it.

Also, reading manual parsing code vs. regular expression is doubtful which one is easier to read for a beginner programmer.

paulproteus:
Hi @mitar. I nearly agree with you. The main difficulty with this argument is that the regex construction code is dynamic and hard to evaluate/audit. Maybe there's a way to use a regex without the dynamic construction code, I wonder. I'll reread and see what I can think of.

paulproteus:
In the case of `PUBLIC_ID_REGEX`, the regex is reasonably clear to me except for the non-capturing group. I believe @kentonv would agree. I hope you will agree that a preference against non-capturing groups, due to the concern that other contributors might not know what `(?:` does, is purely aesthetic.

In the case of the regex in `_getAllowedOutgoingAddressesRegex()`, it seems to me that the algorithm is reasonably easy to describe, and I think much easier to read as a function that splits a string based on `@`, followed by a split based on `+`, followed by equality checks. I can't think of a way to do that using one regular expression, so long as `this._getPublicId()` can use non-regex-safe characters, and so long as `ROOT_URL` can use non-regex-safe characters. But perhaps one way around this is to avoid the dynamic regex creation, and instead match all characters in that position, then assert that they equal what one would expect. For example:

```js
> /^([^+]*)([+]*[^+]*)@(.*)$/.exec('examplePublicId+exampleSubAddress@example.com')
[ 'examplePublicId+exampleSubAddress@example.com',
  'examplePublicId',
  '+exampleSubAddress',
  'example.com',
  index: 0,
  input: 'examplePublicId+exampleSubAddress@example.com' ]
```

Then it is possible to check that the value that comes out of `exec()` has the right data in [1] and [2] and [3].

I wonder if that would work for you. Let me know.

mitar:
Just a short comment, I am a bit busy at the moment with other things. I hope to get to this back at some point.

mitar:
@kentonv, @paulproteus, I have addressed PR feedback from both of you. Please review the pull requests again. Sorry that it took so long.

shell/server/drivers/mail.js
```diff
@@ -223,15 +240,23 @@ hackSendEmail = (session, email) => {
...
         userAddress.address + ". Yours was: " + email.from.address);
     }
 
     // Unpack fields
     const { from, to, cc, bcc, replyTo, subject, text, html } = email;
 
+    // We allow envelope address to have a suffix as well. In this way apps can use unique envelope
+    // addresses when sending out e-mails which allow better handling of e-mail bounces - app can
+    // match bounce with the sent e-mail.
+    let envelopeFrom = email.envelopeFrom || from.address;
```
kentonv: Currently, `envelopeFrom` will never be present since it is not defined in `email.capnp`. There's no way for an app to fill this in.

We could add it, but I'm not sure if this is the right interface. If the app can't actually set an arbitrary `envelopeFrom` then it is confusing to let them specify one and then fail when they specify the wrong one. Maybe instead we should have a field `bounceId` which is strictly the ID to place after the `+`? Another idea would be to extend the email interfaces such that Sandstorm can capture bounces and make some RPC callback to the app, which might be a nicer interface for apps, but requires a lot more work.

Are you actually anticipating the need to handle bounces in an app? If not then maybe we can punt this for now.

mitar: > Currently, envelopeFrom will never be present since it is not defined in email.capnp. There's no way for an app to fill this in.

Oh. I am not yet familiar enough with Sandstorm. But yes, then we have to extend the API a bit.

> if the app can't actually set an arbitrary envelopeFrom then it is confusing to let them specify one and then fail when they specify the wrong one.

I think this is pretty reasonable restriction, especially because we are restricting `From` as well. In the same sense it will fail if they set `From` to an arbitrary value.

> Maybe instead we should have a field bounceId which is strictly the ID to place after the +?

I think this is too specific. We can go with that, but then the same holds for `From`?

I think the current approach is the right balance between prescribing, restricting and allowing. Currently developer can choose:
* `From` as login e-mail address
* `From` as sandbox random e-mail address
* `From` as sandbox random e-mail address with custom suffix, if they want to expose the suffix to users (imagine mailing list service, where people should reply to the mailing list)
* any of the above for `From`, but customizing envelope from, to handle automatic replies in general

I think there is some advantage of envelope having same restrictions to  sandbox random e-mail address with optional custom suffix.

Anyway, it is your decision. I think we should try to overthink things here.

>  Another idea would be to extend the email interfaces such that Sandstorm can capture bounces and make some RPC callback to the app, which might be a nicer interface for apps, but requires a lot more work.

Yes, that is also an option.

> Are you actually anticipating the need to handle bounces in an app? If not then maybe we can punt this for now.

I wanted to use this primarily for sending notifications for posts to a forum out. And allowing recipients to reply to the e-mail to make a post. Similar to how GitHub allows replying to notifications about issues. Detecting bounces is useful to know issues with e-mails used in the forum.

But there are many open questions here if this is really the best approach. I have not yet checked new Sandstorm notifications API, so maybe that could be good enough. And there is also a limit on number of e-mails allowed to be send through Sandstorm. And also there is an issue of me needing to send e-mails in background, as a digest, when maybe I do not have user context/capability anymore. Also, e-mails used in the app should be all verified by Sandstorm already so bounces should be low.

So maybe at the end it will not work out, but this was the first step which I was thinking of improving in this aspect. I see it just as allowing more with a bit of code changes.

kentonv: The restriction on `From` is different since we actually allow the app to specify the user's e-mail address (rather than the grain's) as `From`. But it fundamentally does not make sense to allow overriding the envelope sender address to anything other than the grain's address, because the whole purpose of the envelope sender address is to point back at the sending entity.

Incidentally, depending on the e-mail provider in use, this may be moot. Sendgrid, for example, does not send bounces back to the sender; it adds the recipient address to the bounce list and silently drops all future messages to that recipient. So, on Oasis, your app will never receive bounces.

Generally I would like to shield apps from having to understand the complexity of e-mail processing. As you point out, Sandstorm already verifies addresses, so bounces will be uncommon. But if they do happen, that's a problem for Sandstorm, independent of the app. That is to say, regardless of what any app does, Sandstorm is going to need some sort of mechanism of its own for dealing with broken e-mail addresses. I'd prefer if apps can rely on Sandstorm taking care of these things and not try to do it themselves. (However, Sandstorm doesn't take care of these things today, so... this is admittedly vaporware.)

On another note, the notifications API still needs a lot of work, but our intent for that API is that it should allow "github-like" email notifications without the app having to understand any specifics of e-mail. See: https://github.com/sandstorm-io/sandstorm/tree/master/roadmap/platform/activity

mitar: OK, I agree about long term goals here. But my question what can I do in short term, until all that is in place? It seems for now the best way to have github-like notification is to implement them myself (I will have them anyway for running app outside Sandstorm) and get small changes in Sandstorm in place to allow this. But then it seems it will not work anyway on Oasis because of Sendgrid and general sending limits and lack of background e-mail sending context or in general background tasks. So maybe I should just give up with this approach and wait for Sandstorm to have notifications API and until then the app will not have this feature when running inside Sandstorm.

kentonv: I'd integrate with the Sandstorm activity API now, which will produce bell notifications, and then add e-mail notification reply processing once it lands in Sandstorm.

That said, I think the only thing that doesn't work is bounce processing. So another option would be to implement everything except that.

mitar: OK, so then these changes to e-mail sending are not necessary and I can use notifications API.

PR #1928, opened 2016-04-29
Avoid using system node when installing webfont packages.

### LINKED ISSUE #1922: Cannot make sandstorm on clean VM
```
Finding meteor-tool installation (can take a few seconds)... meteor-tool@1.3.1
cd icons && /home/user/.meteor/packages/meteor-tool/.1.3.1.1ult9zk++os.linux.x86_64+web.browser+web.cordova/mt-os.linux.x86_64/dev_bundle/bin/npm install
/usr/bin/env: node: No such file or directory
make: *** [icons/node_modules] Error 127
```

So Meteor is installed, it finds `npm`, but not `node`. I suspect `node` is not added to `PATH`?

dwrensha:
Fixes #1922.

Meteor's bundled npm script now does

```
#!/usr/bin/env node
```

whereas it previously did

```
`dirname "$0"`/node "$0" "$@"
```

therefore we should set the PATH so that the correct node binary is found.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -253,7 +253,7 @@ tmp/.shell-env: tmp/.ekam-run $(IMAGES) shell/client/changelog.html shell/client
 	@cd shell/ && meteor npm install
 
 icons/node_modules: icons/package.json
-	cd icons && $(METEOR_DEV_BUNDLE)/bin/npm install
+	cd icons && PATH=$(METEOR_DEV_BUNDLE)/bin:$$PATH $(METEOR_DEV_BUNDLE)/bin/npm install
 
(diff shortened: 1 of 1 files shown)


zarvox:
If we called `meteor npm` (rather than explicitly executing the one we look up in the dev bundle) would the correct `node` wind up on `PATH`?

dwrensha:
That would use *a* Meteor node, but I think possibly not the one corresponding to the Meteor version in shell/.meteor/release.

mitar:
I think the PATH approach is the safest.

PR #1929, opened 2016-04-29
New admin user list and user details pages

### LINKED ISSUE #1822: Admin Panel Redesign: User Accounts
Required to ship User Accounts:
- [x] change "Guest" -> "Visitor"
- [x] "created" and "last used" dates should align vertically center
- [x] look into user accounts: DOM representation
- [x] add label "Show" above/beside role filter checkboxes
- [x] stack & left-align role filter checkboxes
- [x] add number count to role filter labels
- [x] figure out a way to communicate what "A" "U" "V" icons are
- [x] uncheck "Visitors" by default

Out of scope:
- [ ] pagination so that the table doesn't extend forever

Current mock-up:
user-accounts4

Required to ship Manage User:
- [ ] change radio buttons to buttons that say "Upgrade to Admin", "Demote to User", etc.
- [ ] display more labels about status of each identity: "Can log in", "Org Member", "Demo"
- [ ] add save button at the bottom (disabled until changes are detected)
- [ ] change "Guest" -> "Visitor"
- [ ] any user can be upgraded to Admin
- [ ] cannot demote Admin to User or Visitor, cannot demote User to Visitor

Out of scope:
- [ ] deactivate identity
- [ ] assume role
- [ ] display identity IDs / other server-relevant details
- [ ] show warning for demoting roles
- [ ] enable some sort of re-parenting of grains for demoted Admins

Current mock-up:
manage-user4
[Previous mock-up](https://cloud.githubusercontent.com/assets/855341/14655098/a4cfffea-0635-11e6-9467-d0e0f6302272.png)
Note: status labels will get pushed to the next line only when there isn't enough horizontal space available

zarvox:
User list:

user-list

Click on any particular user (or the "Manage" link for keyboard users), see details about them:

user-details-self

You cannot modify your own account, as shown above, but you can modify other users:

user-details-avatar-multi-emails

You can't demote org members to Visitor:

cannot-demote-org-member

Closes #1822.

### DIFF
--- a/shell/server/admin-server.js
+++ b/shell/server/admin-server.js
@@ -476,7 +476,8 @@ Meteor.publish("allUsers", function (token) {
 Meteor.publish("adminUserDetails", function (userId) {
   if (!authorizedAsAdmin(undefined, this.userId)) return [];
 
-  // Reactive publish of any identities owned by
+  // Reactive publish of any identities owned by the account with id userId,
+  // as well as that user object itself.
   const identitySubs = {};
   const accountId = userId;
 
@@ -554,6 +555,8 @@ Meteor.publish("adminUserDetails", function (userId) {
     });
   });
 
+  // Meteor's cursor.observe() will synchronously call all of the added() callbacks from the initial
+  // query, so by the time we get here we can report readiness.
   this.ready();
 });
 
--- a/shell/server/admin-server.js
+++ b/shell/server/admin-server.js
@@ -473,6 +473,90 @@ Meteor.publish("allUsers", function (token) {
   return Meteor.users.find();
 });
 
+Meteor.publish("adminUserDetails", function (userId) {
+  if (!authorizedAsAdmin(undefined, this.userId)) return [];
+
+  // Reactive publish of any identities owned by
+  const identitySubs = {};
+  const accountId = userId;
+
+  const unrefIdentity = (identityId) => {
+    const observeHandle = identitySubs[identityId];
+    identitySubs[identityId] = undefined;
+    observeHandle.stop();
+    this.removed("users", identityId);
+  };
+
+  const refIdentity = (identityId) => {
+    const cursor = Meteor.users.find({ _id: identityId });
+    const observeHandle = cursor.observe({
+      added: (doc) => {
+        this.added("users", doc._id, doc);
+      },
+
+      changed: (doc) => {
+        this.changed("users", newDoc._id, newDoc);
+      },
+
+      removed: (doc) => {
+        this.removed("users", oldDoc._id);
+      },
+    });
+
+    identitySubs[identityId] = observeHandle;
+  };
+
+  const accountCursor = Meteor.users.find({ _id: accountId });
+  const accountSubHandle = accountCursor.observe({
+    added: (newDoc) => {
+      const newIdentities = SandstormDb.getUserIdentityIds(newDoc);
+      newIdentities.forEach((identityId) => {
+        refIdentity(identityId);
+      });
+
+      this.added("users", newDoc._id, newDoc);
+    },
+
+    changed: (newDoc, oldDoc) => {
+      const newIdentities = SandstormDb.getUserIdentityIds(newDoc);
+      const oldIdentities = SandstormDb.getUserIdentityIds(oldDoc);
+
+      // Those in newDoc - oldDoc, ref.
+      const identitiesAdded = _.difference(newIdentities, oldIdentities);
+      identitiesAdded.forEach((identityId) => {
+        refIdentity(identityId);
+      });
+
+      // Those in oldDoc - newDoc, unref.
+      const identitiesRemoved = _.difference(oldIdentities, newIdentities);
+      identitiesRemoved.forEach((identityId) => {
+        unrefIdentity(identityId);
+      });
+
+      this.changed("users", newDoc._id, newDoc);
+    },
+
+    removed: (oldDoc) => {
+      this.removed("users", oldDoc._id);
+      const oldIdentities = SandstormDb.getUserIdentityIds(oldDoc);
+      oldIdentities.forEach((identityId) => {
+        unrefIdentity(identityId);
+      });
+    },
+  });
+
+  this.onStop(() => {
+    accountSubHandle.stop();
+    // Also stop all the identity subscriptions.
+    const subs = _.values(identitySubs);
+    subs.forEach((sub) => {
+      sub.stop();
+    });
+  });
+
+  this.ready();
+});
+
 Meteor.publish("activityStats", function (token) {
   if (!authorizedAsAdmin(token, this.userId)) return [];
   return ActivityStats.find();
--- a/shell/server/admin-server.js
+++ b/shell/server/admin-server.js
@@ -482,6 +482,13 @@ Meteor.publish("adminUserDetails", function (userId) {
   const accountId = userId;
 
   const unrefIdentity = (identityId) => {
+    if (!identitySubs[identityId]) {
+      // should never happen, but if somehow you attempt to unref an identity that we don't have a
+      // subscription to, then don't crash
+      console.error("attempted to unref untracked identity id:", identityId);
+      return;
+    }
+
     const observeHandle = identitySubs[identityId];
     identitySubs[identityId] = undefined;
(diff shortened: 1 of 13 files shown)


neynah:
Looks great @zarvox. Can we left-align the column title "Properties"?

paulproteus:
Presumably we should rename 'guest' to 'visitor' in other places too before releasing this, to avoid confusion, including:

- Any error messages that the app provides

- In the docs

I'm OK with that being a separate pull request, but thought I should mention it for completeness. Since this is the code change that renames the concept from "Guest" to "Visitor", I'd prefer if @zarvox were the one to do those other fixes, but I'm also OK with him delegating that to a willing party. (I can be that willing party.)

paulproteus:
Oh, then again, this isn't live until we switch from "oldadmin" to "newadmin", right? So it's OK to delay that until then, I figure.

neynah:
Icons on the Users list should be top-aligned and the "Manage" link should also be styled as a button. Otherwise, looks great!

neynah:
On Users / Test it seems like only the verified e-mail is linked—is this intentional?

zarvox:
> Looks great @zarvox. Can we left-align the column title "Properties"?

Discussed in person, left-aligned both the column header and the contents for readability.

> (some things about guest vs visitor)

Yeah, let's do that when we do the switchover.

> Icons on the Users list should be top-aligned

Discussed in person, implemented.  Kind of a pain to get the paddings to line up nicely.

> and the "Manage" link should also be styled as a button. Otherwise, looks great!

Done.

I believe that addresses all the review comments given.

shell/server/admin-server.js
```diff
@@ -473,6 +473,90 @@ Meteor.publish("allUsers", function (token) {
   return Meteor.users.find();
 });
 
+Meteor.publish("adminUserDetails", function (userId) {
+  if (!authorizedAsAdmin(undefined, this.userId)) return [];
+
+  // Reactive publish of any identities owned by
```
kentonv: Finish sentence?

shell/server/admin-server.js
```diff
@@ -473,6 +473,90 @@ Meteor.publish("allUsers", function (token) {
...
+      changed: (doc) => {
+        this.changed("users", newDoc._id, newDoc);
+      },
+
+      removed: (doc) => {
+        this.removed("users", oldDoc._id);
+      },
+    });
+
+    identitySubs[identityId] = observeHandle;
```
kentonv: If two accounts share an identity then I think this will leak the old handle.

kentonv: I think you'll need some refcounting...

shell/server/admin-server.js
```diff
@@ -473,6 +473,90 @@ Meteor.publish("allUsers", function (token) {
...
+
+  const unrefIdentity = (identityId) => {
+    const observeHandle = identitySubs[identityId];
+    identitySubs[identityId] = undefined;
+    observeHandle.stop();
+    this.removed("users", identityId);
+  };
+
+  const refIdentity = (identityId) => {
+    const cursor = Meteor.users.find({ _id: identityId });
```
kentonv: So we're subscribing to each identity separately? Will be interesting to see what happens on Oasis... I guess the only solution is pagination, though.

shell/server/admin-server.js
```diff
@@ -473,6 +473,90 @@ Meteor.publish("allUsers", function (token) {
...
+  this.onStop(() => {
+    accountSubHandle.stop();
+    // Also stop all the identity subscriptions.
+    const subs = _.values(identitySubs);
+    subs.forEach((sub) => {
+      sub.stop();
+    });
+  });
+
+  this.ready();
```
kentonv: I forget, does observe() call callbacks for the initial set of results synchronously? If so comment here as a reminder. If not then it's too early to call ready()...

kentonv:
"Make Admin" / "Make User" / "Make Visitor" seems like confusing wording to me... like I'm making a new one, or something. Maybe "Promote to Admin", "Promote to User", "Demote to Visitor"?

shell/server/admin-server.js
zarvox: Note that this publish is only used on the user details page, so you're only calling this `subscribe()` on a single account ID.

shell/server/admin-server.js
zarvox: I don't think so - this publish only operates on a single account at a time.  So two accounts get two separate subscribe() calls, and two separate `identitySubs` instances, and two separate `observeHandle`s.

You might argue that I should be sharing the subscriptions between the publishes, and you'd probably be right.  I can rework this code to do so, but given that you're fairly unlikely to be looking at too many of these things at once in the first place, and it's only available to admins, I didn't think it was worth the complexity in the first pass.

shell/server/admin-server.js
zarvox: Yes.  http://docs.meteor.com/#/full/observe says

> Before observe returns, added (or addedAt) will be called zero or more times to deliver the initial results of the query.

I'll add the comment.

shell/server/admin-server.js
zarvox: D'oh.

shell/server/admin-server.js
zarvox: For reference, the user-list page subscribes to the "allUsers" publish five lines above the start of this function.

kentonv: Ohh, yeah, I should have read more carefully.

shell/server/admin-server.js
kentonv: OK, I think you should add some defensive code to detect a double-subscribe (and double-unsubscribe). If somehow someone ends up with the same identity listed twice on their account, it shouldn't cause the server to leak mongo subscriptions. Plus I could easily imagine this code changing later on to be used in a different way, and the name `refIdentity` sort of implies refcounting, so I'd hate for someone to assume that's what it does and have a silent leak...

kentonv: To be clear, the code could just throw an exception... though then we have to ask whether the context is exception-safe. Where is RAII when you need it?

zarvox:
Stylecheck failure was introduced by #1923 which Garply never ran tests on, since Mitar isn't on the whitelist and no one flagged the patch as okay to test.  Added a fix to the end of this patch series.

shell/server/admin-server.js
mitar: > it shouldn't cause the server to leak mongo subscriptions

Duplicate subscriptions are de-duplicated by Meteor. I would not worry about such type of optimizations. It seems premature optimizations to me.

mitar:
Whitelist me. ;-)

shell/server/admin-server.js
kentonv: @mitar Avoiding memory leaks is not "optimization". I think you may misunderstand the problem here.

mitar: Oh, oh. You are right.

You are manually doing related publish here? This is so error prone. Why not using some packages which are trying really hard to do this properly? Why reinventing the wheel? (Of course I have my [own](https://github.com/peerlibrary/meteor-reactive-publish), but not everyone will like my approach. I hear [this one](https://github.com/englue/meteor-publish-composite) is pretty popular as well.

zarvox:
Added guards.

neynah:
@zarvox Not urgent but I forgot to mention "Show" should be inline with the search box. Also, some padding(10-15pxish) below the "Visitor" checkbox would reduce clutter.

Also, the search box takes up more space than it does in our current UI(although I don't believe this is currently standardized). Maybe we should chat at some point about how much horizontal space a search bar should take. I'm thinking 50-70% depending on the size of the screen.

zarvox:
I can be talked down from having the search bar take up the full available width, but in the absence of more precise specifications in the mockups, I will generally attempt to approximate a grid and assign any remaining space to the interactive elements or important presentational content, since larger click targets and readable text are generally preferable.

That said, some things benefit from having their width capped - readability is improved around 65 characters per line, and so forth.

If you think this sort of issue is likely to come up in the future, let's come up with a way to communicate this sort of intent more clearly in the mockups.  From my side, it's hard to infer from a static picture with no gridlines, size comments, or sub-boxes what is intended to happen at varying widths!

PR #1931, opened 2016-04-29
Fix websocket connections over API endpoints

jparyani:


paulproteus:
@ndarilek , I believe we were discussing the question of if websockets over API endpoints would work.

I thought they did already, but I was wrong. In @jparyani 's testing, he found they don't.

So - now they do! Once this is out in a live build, and servers have auto-updated (probably by Monday May 2), you'll now be able to use WebSockets over a grain's API hostname, and then layer whatever protocol you want on top of that.

PR #1946, opened 2016-05-05
WIP notification / activity log design

kentonv:
This is just an interface.

This has changed significantly from the design doc:
* The interface is now centered around an activity feed, logging events occurring on a grain. This is a common feature seen across many apps. Having a unified activity feed provided by Sandstorm seems valuable. Also, "activity feed" and "audit log" are basically the same thing.
* Events can generate notifications under various circumstances. It seems like in all the use cases we've been talking about, notifications are tied to an event, so this seemed more natural than defining separate activity() and notify() methods that need to be called.
* This also makes it easier to see how a user might tweak their notification settings -- e.g. amplifying or muting certain kinds of events. One can even imagine being able to define rules like: "Notify me via bell menu when I'm assigned to a card in Wekan, send me an e-mail if it was Jade that assigned me, and mute all notifications from this one grain that I don't care about." Obviously, we shouldn't rush into providing that much complexity, but the point is that it is _possible_ if the demand arises.

TODO: I have defined `ActivityTypeDef`, which is similar to `PermissionDef` or `RoleDef`: the application would provide a list of activity types upfront, then tag each event with the appropriate type. However, I have not figured out where the app should actually report this list. `ViewInfo` (next to permission and role definitions) seems like the obvious place, but has the problem that it makes it hard for Sandstorm to know if all grains from the app are expected to have the same set of activity types, which in turn makes it hard for Sandstorm to present the list of activity types to a user for them to choose notification preferences. Putting the definitions in the app manifest would fix that, but a non-main UiView might want to offer different activity types. OTOH, maybe we never really need to construct such a list, but instead give the user an interface like "mute messages like this" which they use reactively? Hmm.

### DIFF
--- a/src/sandstorm/activity.capnp
+++ b/src/sandstorm/activity.capnp
@@ -1,0 +1,180 @@
+# Sandstorm - Personal Cloud Sandbox
+# Copyright (c) 2016 Sandstorm Development Group, Inc. and contributors
+# All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+@0xa4e001d4cbcf33fa;
+
+$import "/capnp/c++.capnp".namespace("sandstorm");
+
+using Util = import "util.capnp";
+
+struct ActivityEvent {
+  # Describes an event in a grain's activity feed / log.
+  #
+  # Call SessionContext.activity() to post new activity. The activity is attributed to the user
+  # whose SessionContext was used.
+
+  path @0 :Text;
+  # Path (as in, URL) within the grain where the user can see this activity. Users inspecting the
+  # event will be linked to this location. Visiting this path implicitly marks the event as "read".
+  #
+  # This should NOT include a leading '/'. (So, an empty string -- the default -- goes to the grain
+  # root).
+  #
+  # Users are able to subscribe to events at specific paths to get notified on activity.
+  # Subscribing to the root (empty path) implies subscribing to all events, and subscribing to a
+  # specific path also implies subscribing to any path that has the subscription path as a prefix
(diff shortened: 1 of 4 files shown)


ocdtrekkie:
"maybe we never really need to construct such a list, but instead give the user an interface like "mute messages like this" which they use reactively?" sounds sensible. It's hard to know the frequency or importance of a type of notification you've never seen.

zarvox:
I found this whole discussion about dealing with too much email from Phabricator interesting and relevant, particularly the validation of the "suicide hotline" approach to helping users avoid receiving unwanted notifications of particular classes: https://secure.phabricator.com/T9161

kentonv:
Perhaps the underlying mechanism for notification settings is filters, like email filters. Then the UX is:

* Every notification has a "filter notifications like these" or maybe just "unsubscribe" button / link which launches into a wizard-ish UI for creating a filter to mute that notification.
* Apps place "subscribe" or "watch" buttons in-app which invoke a postMessage API which opens a dialog where the user is again presented with the filter wizard, this time initialized to match the thing the user is looking at and to amplify notifications.

Note that I don't think we should attempt to build this UI for v1 of notifications. For now we'll trust apps to choose good defaults. I also figure we'll initially implement bell notifications only to sidestep the spam question.

src/sandstorm/activity.capnp
```diff
@@ -0,0 +1,180 @@
...
+  #   in the bell menu after the user clicks on the notification (Google+ style).
+  # TODO(someday): Support notifications that can receive text replies. The replies are delivered
+  #   to a provided capability. When notifications are delivered via email, Sandstorm can
+  #   automatically support email replies. When notifications are delivered via the bell menu,
+  #   Sandstorm can render an inline reply textarea (like Google+ notifications).
+  # TODO(someday): Support rich interactive notifications, e.g. the ability to play/pause music
+  #   via buttons.
+}
+
+interface NotificationTarget {
```
kentonv: Moved type needs explicit type ID.

src/sandstorm/activity.capnp
```diff
@@ -0,0 +1,180 @@
...
+  # TODO(someday): Expand on this and move it into `grain.capnp` when notifications are
+  #   fully-implemented.
+
+  addOngoing @0 (displayInfo :NotificationDisplayInfo, notification :OngoingNotification)
+             -> (handle :Util.Handle);
+  # Sends an ongoing notification to the notification target. `notification` must be persistent.
+  # The notification is removed when the returned `handle` is dropped. The handle is persistent.
+}
+
+interface OngoingNotification {
```
kentonv: Moved type needs explicit type ID.

PR #1948, opened 2016-05-06
Explain that mkdir is required

### LINKED ISSUE #1934: No such file or directory; name = /var/mail/tmp/_Qa8lI_Valn8JIBtmb2kcd1
Hm, it seems that receiving e-mails do not work if app does not create before necessary spool directory?

There is no `/var/mail/tmp` nor `/var/mail/new` and it seems http-bridge code currently assumes there is.

paulproteus:
Close #1934


paulproteus:

PR #1949, opened 2016-05-06
Add an authentication method for API websockets

jparyani:
It turns out that sending an Authorization header is techincally not
allowed for cross origin websockets. To work around this, we will allow
websockets to provide the token as part of the path.

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -975,7 +975,7 @@ const apiTokenForRequest = (req, hostId, isWebsocket) => {
   } else if (auth && auth.slice(0, 6).toLowerCase() === "basic " &&
              apiUseBasicAuth(req, hostId)) {
     token = (new Buffer(auth.slice(6).trim(), "base64")).toString().split(":")[1];
-  } else if (isWebsocket && req.url.startsWith("/.sandstorm-api-token/")) {
+  } else if (isWebsocket && req.url.startsWith("/.sandstorm-token/")) {
     let parts = req.url.slice(1).split("/"); // remove leading / and split
     if (parts.length < 2) {
       token = undefined;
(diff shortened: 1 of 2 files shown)


shell/server/proxy.js
```diff
@@ -975,6 +975,15 @@ const apiTokenForRequest = (req, hostId) => {
   } else if (auth && auth.slice(0, 6).toLowerCase() === "basic " &&
              apiUseBasicAuth(req, hostId)) {
     token = (new Buffer(auth.slice(6).trim(), "base64")).toString().split(":")[1];
+  } else if (isWebsocket && req.url.startsWith("/.sandstorm-well-known/token/")) {
```
paulproteus: Request:

Use `/.well-known/sandstorm/token` not `/.sandstorm-well-known`

Rationale: `/.well-known/` is standardized so apps should know strange things might happen there.

paulproteus:
I'm excited about this. I guess we should document it somewhere, presumably on the "exporting a HTTP API" docs page. Jason are you OK being the one to add it to that doc? If not, I can.

paulproteus:
(Also per IRL conversation, no need to use the literal `.well-known/` even though I :heart:  it)

jparyani:
Docs added. Feel free to tweak them @paulproteus

kentonv:
`.sandstorm-api-token/` feels verbose. Part of me wants to remove it entirely -- we could assume that if no authorization header is present, then the first path component is the token. However, I suppose that might not be great in the future if we want opening a webkey directly to display an API explorer of some sort.

How about shortening it to `.sandstorm-token/`?

On another note, don't tokens sometimes contain `/` characters? Do we need to instruct people to URL-escape those? Or should we perhaps switch to a URL-safe encoding (for newly-created tokens)?

I don't think we need this to be websocket-specific -- we might as well make it work for regular requests as well.

jparyani:
I just checked, and tokens today will never contain a '/'. We create token with `Random.secret` ( see https://github.com/sandstorm-io/sandstorm/blob/ecee5769cc8958d38474a3fb3b4f266ab27ee43c/shell/packages/sandstorm-permissions/permissions.js#L1468), and this is defined in Meteor as using its definition of base64 characters (see https://github.com/meteor/meteor/blob/2c5adbe98ae62a059476db6322c7b573ac117b4d/packages/random/random.js#L89).

Other comments have been addressed.

Issue #1915, opened 2016-04-27
Grain Management UI


neynah:
**Required to ship:**
Grain list:
- [ ] Grains archived will have an archive icon, grains trashed will have a trash icon. These will be displayed somewhere along each grain name.
- [ ] "Delete permanently" button should be disabled unless the grain is trashed. Only trashed grains can be permanently deleted.
- [ ] Disabled "Delete permanently" button will display a tooltip saying something like "Only trashed grains can be permanently deleted. Please trash the selected grains."
- [ ] Any grain can be archived or trashed; if at least one of the grains selected is already currently in the same archived or trashed state, we still allow the user to click the "Archived" or "Trash" button.
- [ ] Ability to select all and deselect all
- [ ] Search searches all lists: main list, archived, trashed.
- [ ] Trashed items have a 30-day expiry? We should also display a message above the grain list saying trashed grains that they will expire after 30-days. Perhaps a good time would be immediately after a user moves a grain to trash, or when they filter & view only trashed grains.

Grain view(opening a grain):
- [ ] If the grain is trashed by the owner/upstream, the owner sees an error: "File is in the trash, to access this grain restore it to the main list." Followed by primary button "Restore to main list" and secondary button "Go back to grain list"
- [ ] If the grain is trashed by the owner/upstream, the sharee sees an error: "You can no longer access this grain as it is in someone's trash." Followed by primary button "Go back to grain list".
- [ ] If the grain is deleted by the owner/upstream, the owner and sharee see an error: "No grain found with ID...". Followed by primary button "Go back to grain list".
- [ ] If grain is trashed/deleted grain preview will not display, top bar actions will not display.

**Out of scope:**
Grain list:
- keyboard shortcuts

Grain view:
- enabling sharees to request the owner to restore a grain
- advanced search operators
- labels/tags

Here's a preliminary mock-up so far:
![grain-mgmt3](https://cloud.githubusercontent.com/assets/855341/14897235/06e828ae-0d36-11e6-901b-2e8767c2f60e.png)
![grain-mgmt5](https://cloud.githubusercontent.com/assets/855341/15001527/14991ede-114d-11e6-872c-db705c2cf7ba.png)

Previous mock-ups:
[v1](https://cloud.githubusercontent.com/assets/855341/14859715/4df1bad0-0c5a-11e6-93ff-c7bbdb30c1d5.png)
[...]

PR #1950, opened 2016-05-06
Implement trash bin and bulk grain action buttons.

dwrensha:
This implements most of #1915.

The grain list looks like: 
grains

When you click "View Trash", you get:
trash

The "Delete permanently" button updates its text depending on what's selected. If only "shared with me grains" are selected, it says "Forget permanently". If both kinds of grains are selected, it says "Delete/forget permanently".
forget

The app details page gets a similar update:
app-details

If you search for something in the main list and matches only exist in the trash list, you get a pointer like this: 
no-match

If you try to open a grain that is in your trash, you get an error page like this:
in-trash-interstitial

If you try to open a grain that is not owned by you and is in its owner's trash, you get an error page like this: 
owner-trash

This also updates the permissions computation so that when a grain gets trashed or deleted, any sessions that depended on that grain are immediately closed.

### DIFF
--- a/shell/packages/sandstorm-permissions/permissions.js
+++ b/shell/packages/sandstorm-permissions/permissions.js
@@ -519,7 +519,7 @@ class Context {
     check(grainIds, [String]);
     if (grainIds.length == 0) return; // Nothing to do.
 
-    db.collections.grains.find({ _id: { $in: grainIds } }).forEach((grain) => {
+    db.collections.grains.find({ _id: { $in: grainIds }, trashed: { $exists: false } }).forEach((grain) => {
       this.grains[grain._id] = grain;
       if (!this.userIdentityIds[grain.userId]) {
         this.userIdentityIds[grain.userId] = SandstormDb.getUserIdentityIds(
@@ -528,7 +528,9 @@ class Context {
     });
 
     const query = { grainId: { $in: grainIds },
-                    revoked: { $ne: true }, objectId: { $exists: false }, };
+                    revoked: { $ne: true },
+                    trashed: { $exists: false },
+                    objectId: { $exists: false }, };
     db.collections.apiTokens.find(query).forEach((token) => this.addToken(token));
   }
 
@@ -921,7 +923,9 @@ class Context {
     // This function works by walking backwards in the sharing graph, following this trail of
     // "responsible tokens".
     //
-    // Returns the result as a list of token IDs.
+    // Returns an object with two fields:
+    //    tokenIds: list of IDs of the responsible tokens.
+    //    grainIds: list of IDs of all relevant grains.
 
     check(grainId, String);
     check(vertexId, String);
@@ -948,7 +952,7 @@ class Context {
     const neededTokens = {}; // TokenId -> bool
 
     forEachPermission(this.getPermissions(grainId, vertexId).array, (permissionId) => {
-      stack.push({ grainId: grainId, vertexId: vertexId, permissionId: permissionId });
+      pushVertex(grainId, vertexId, permissionId);
     });
 
     while (stack.length > 0) {
@@ -983,7 +987,7 @@ class Context {
       }
     }
 
-    return Object.keys(neededTokens);
+    return { tokenIds: Object.keys(neededTokens), grainIds: Object.keys(visited) };
   }
 }
 
@@ -1012,6 +1016,7 @@ function computeRelevantTokens(context, grainId, vertexId) {
   check(vertexId, String);
 
   const grain = context.grains[grainId];
+  if (!grain) return { tokenIds: [], ownerEdges: [] };
   const viewInfo = grain.cachedViewInfo || {};
   const ownerIdentityIds = context.userIdentityIds[grain.userId];
 
@@ -1157,6 +1162,20 @@ SandstormPermissions.mayOpenGrain = function (db, vertex) {
   return !!context.tryToProve(grainId, vertexId, emptyPermissions, db);
 };
 
+class CompoundObserveHandle {
+  constructor() {
+    this._handles = [];
+  }
+
+  push(handle) {
+    this._handles.push(handle);
+  }
+
+  stop() {
+    this._handles.forEach((handle) => handle.stop());
+  }
+}
+
 SandstormPermissions.grainPermissions = function (db, vertex, viewInfo, onInvalidated) {
   // Computes the set of permissions received by `vertex`. Returns an object with a
   // `permissions` field containing the computed permissions. If the field is null then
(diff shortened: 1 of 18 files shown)


ocdtrekkie:
I am so excited about this I can't keep it to myself, and have to tell you. :D

neynah:
I'm guessing Select all(list) as well as secondary buttons to "Return to Main list" (in the last 2 screenshots) are still being worked on?

Otherwise, looks great @dwrensha!

shell/server/grain-server.js
```diff
@@ -273,12 +273,75 @@ Meteor.publish("grainSize", function (grainId) {
   });
 });
 
+const GRAIN_DELETION_MS = 1000 * 60 * 60 * 24 * 30; // thirty days
+SandstormDb.periodicCleanup(86400000, () => {
+  const trashExpiration = new Date(Date.now() - GRAIN_DELETION_MS);
+  Grains.find({ trashed: { $lt: trashExpiration } }).forEach((grain) => {
+    Grains.remove({ _id: grain._id });
```
kentonv: Maybe I'm missing something but it looks like this only removes the grain metadata from mongo, not the actual grain data from the back-end?

shell/server/grain-server.js
```diff
@@ -273,12 +273,75 @@ Meteor.publish("grainSize", function (grainId) {
...
+      grainId: grain._id,
+      $or: [
+        { owner: { $exists: false } },
+        { owner: { webkey: null } },
+      ],
+    });
+
+    ApiTokens.remove({ trashed: { $lt: trashExpiration } });
+
+    // Don't need to update demo stats: demo grains never get here!
```
kentonv: But DeleteStats isn't just for demo users...

kentonv:
Hmm, it occurs to me that trash could be confusing for people hitting quota limits. Maybe we need an "empty trash" button added to the "out of quota" prompt? Or maybe we should automatically remove items from the trash if the user is out of quota?

kentonv:
Hmm, I see that there's no "select / unselect all" checkbox at the top of the check box column. I feel like this is desirable, especially for "unselect all" -- If I select a bunch of things then change my mind, I have to painstakingly click each one to unselect it, or refresh the whole page.

On a similar note, there probably needs to be a one-click "empty trash" button in the trash. Currently if I want to clear everything (e.g. to free up quota) I have to click to select each item...

kentonv:
Can we make the check-box click targets bigger? Like, additionally listen for clicks on the whole table cell. When going down the list selecting things I find myself clicking in the space between two checkboxes a lot.

kentonv:
Bug: When confirming installation of a new app, the "view trash" button is visible but doesn't do anything.

neynah:
@paulproteus mentioned something about displaying "Grains: 4 (plus 2 in trash)".

I think we should have the number alongside "View trash". So "View trash (2)" for example.

shell/server/grain-server.js
dwrensha: Good catch! Also, the clearing of `ApiTokens` should be outside of this iteration. Fixed.

shell/server/grain-server.js
dwrensha: Fixed.

dwrensha:
> Hmm, it occurs to me that trash could be confusing for people hitting quota limits. Maybe we need an "empty trash" button added to the "out of quota" prompt? Or maybe we should automatically remove items from the trash if the user is out of quota?

Automatically deleting things sounds scary to me. Putting a button on the prompt sounds fine. That will need to go in a separate pull request agains the blackrock repo.

dwrensha:
> Hmm, I see that there's no "select / unselect all" checkbox at the top of the check box column. I feel like this is desirable, especially for "unselect all" -- If I select a bunch of things then change my mind, I have to painstakingly click each one to unselect it, or refresh the whole page.

I've implemented "select all" and "unselect all". If anything is selected, the `select-all` checkbox is checked, and clicking on it unselects everything.

> On a similar note, there probably needs to be a one-click "empty trash" button in the trash. Currently if I want to clear everything (e.g. to free up quota) I have to click to select each item...

My feeling is that "select all" is sufficient to address this concern.

If we decide otherwise, where should the "empty trash" button go?

dwrensha:
I've addressed all comments. With the new additions, the UI looks like:

select-all

ocdtrekkie:
Select All + Delete Permanently = Empty Trash seems reasonable. I don't like "delete everything" to ever be a one click operation.

zarvox:
> My feeling is that "select all" is sufficient to address this concern.
> 
> If we decide otherwise, where should the "empty trash" button go?

FWIW, I favor having the "empty the trash" button, since I think it's a reasonable thing to have, and shortens the "I don't always delete things, but when I do, I want them gone" workflow.  I expect that many users will only visit the trash list to reduce their quota impact, and the best thing for them is to just finish deleting the things they already expressed intent to delete, rather than making them figure out multiselect, especially if they haven't bothered learning how to use that feature yet.  I believe that there's probably a decently-sized fraction of Gmail users who have never used the checkboxes/multiselect feature and work entirely from clicking threads and the buttons in the button row; this affordance would serve that class of users well.  (These users also benefit from retaining the trash button in the viewing-a-grain topbar view - they can navigate to the item, and then take the action they expect to be able to take on it.)

I'd support a separate "empty the trash" small, link-styled button in the "Grains in the trash will be permanently removed after 30 days." paragraph, like how Gmail adds the option when you have items in the trash.  Perhaps:

```html
<p>
  <a onclick="emptyTrash();">Empty trash now</a> (Grains that have been in the trash for more than 30 days will be removed permanently.  To keep a grain, restore it to the grain list.)
</p>
```

> I don't like "delete everything" to ever be a one click operation.

To be fair, with this set of changes, It's not -- you have to send everything to the trash first, and then empty the trash.  The data there is already going to expire if you leave it alone; this is an affordance to allow the user to converge on that state sooner because the

neynah:
I'm +1 on the "empty all trash" link as long as there's a prompt to confirm that you really do want to delete all the (#) grains. This behaviour is very common & expected. Also, people misclick things all the time. :)

dwrensha:
I've implemented zarvox's suggestion:

emptytrash

When you click, you get a popup confirmation:

emptytrash-popup

zarvox:
:100: 

Someday (not today) I hope to replace all of our uses of `window.confirm()` with app-modal dialogs like those used on the (new) identity provider settings page.

dwrensha:
One more update: now the trash bin has the route "/grain#trash". This makes it possible to link directly to it, which should make it easy to implement @neynah's suggestion in https://github.com/sandstorm-io/sandstorm/issues/1915#issuecomment-217980290.

shell/server/grain-server.js
```diff
@@ -273,12 +273,83 @@ Meteor.publish("grainSize", function (grainId) {
...
+
+    if (grain.lastUsed) {
+      DeleteStats.insert({
+        type: "grain",  // Demo grains can never never get here!
+        lastActive: grain.lastUsed,
+        appId: grain.appId,
+      });
+    }
+
+    waitPromise(globalBackend.deleteGrain(grain._id, grain.userId));
```
kentonv: For robustness, this call should probably happen before removing the table entry. Otherwise, a badly-timed crash could lead to the grain storage still existing with no way to clean it up.

Probably we should make the same change for the explicit delete flow, too...

dwrensha: Done. I've also verified that if the grain is gone on the filesystem, the rest of this method still executes as intended. I'm not sure about Blackrock, though.

kentonv: Yes, the method was intentionally designed to be idempotent for this reason.

shell/packages/sandstorm-ui-grainlist/grainlist-client.js
```diff
@@ -357,6 +357,14 @@ Template.sandstormGrainTable.events({
     }
   },
 
+  "click td.select-grain": function (event, instance) {
```
kentonv: This doesn't seem to have worked AFAICT: clicking just outside the checkbox does not check it for me.

kentonv: Wait, never mind. I was running a different version of the code than I thought I was.

shell/packages/sandstorm-permissions/permissions.js
```diff
@@ -1203,54 +1222,71 @@ SandstormPermissions.grainPermissions = function (db, vertex, viewInfo, onInvali
...
+      objectId: { $exists: false },
+    });
 
-          removed(oldApiToken) {
+    if (onInvalidated) {
+      observeHandle = new CompoundObserveHandle();
+      observeHandle.push(tokenCursor.observe({
+        changed(newApiToken, oldApiToken) {
+          if (newApiToken.trashed ||
+              !_.isEqual(newApiToken.roleAssignment, oldApiToken.roaleAssignment) ||
```
kentonv: We really need to switch to a type-safe language...

PR #1952, opened 2016-05-07
New Admin: System Status page

### LINKED ISSUE #1825: Admin Panel Redesign: System Stats
Required to ship: 
- [ ] white-space: pre-wrap
- [ ] link Sandstorm version number to "About Sandstorm"
- [ ] for "Current usage" show demo accounts separately
- [ ] restyle "Download full log" button to primary

Out of scope:
- [ ] change "users" -> "user accounts" or "people active"

Current mock-ups:
systemstatus2

zarvox:
This marks the start of me attempting to use ES6 modules to deduplicate code,
as described in http://guide.meteor.com/structure.html#example-app-structure
Expect future work to attempt to avoid duplicating functions in various modules,
and to see existing mostly-UI packages to move out of Meteor packages and
into ES6 modules.

system-status-page

One nice affordance I managed to get to work is that the log starts out scrolled to the **end** of the file, and will continue to follow the end as logs stream in if scrolled to the bottom, but will preserve the scroll position if you scroll up (so new input doesn't prevent you from seeing older input if you want).  I intend to replicate this behavior for the "Show Debug Log" pane some time, as the most recent logs are almost always the most interesting to see, and by default you won't see them until you switch focus and scroll because they are added below the fold.

Closes #1825.

### DIFF
--- a/shell/server/admin/system-status-server.js
+++ b/shell/server/admin/system-status-server.js
@@ -11,6 +11,7 @@ const hashSessionId = function (sessionId) {
 // An in-memory map of random id -> timestamp, used to authorize requests to download the system
 // log.
 const SYSTEM_LOG_DOWNLOAD_TOKENS = {};
+const SYSTEM_LOG_DOWNLOAD_TOKEN_VALIDITY_DURATION = 60000; // 60 seconds
 
 Meteor.methods({
   adminGetServerLogDownloadToken() {
@@ -21,6 +22,11 @@ Meteor.methods({
 
     const token = Random.id();
     SYSTEM_LOG_DOWNLOAD_TOKENS[token] = Date.now();
+    setTimeout(() => {
+      if (SYSTEM_LOG_DOWNLOAD_TOKENS[token]) {
+        delete SYSTEM_LOG_DOWNLOAD_TOKENS[token];
+      }
+    }, SYSTEM_LOG_DOWNLOAD_TOKEN_VALIDITY_DURATION * 2);
     return token;
   },
 });
@@ -34,7 +40,8 @@ Router.map(function () {
       const response = this.response;
       const issueDate = SYSTEM_LOG_DOWNLOAD_TOKENS[token];
       delete SYSTEM_LOG_DOWNLOAD_TOKENS[token]; // Clear the token from the in-memory map.
-      if (issueDate === undefined || issueDate + 60000 < Date.now()) {
+      if (issueDate === undefined ||
+          issueDate + SYSTEM_LOG_DOWNLOAD_TOKEN_VALIDITY_DURATION < Date.now()) {
         // Require download to start within 60 seconds of the adminGetServerLogDownloadToken
         // method being called.
         response.writeHead(404, { "Content-Type": "text/plain" });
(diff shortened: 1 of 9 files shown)


zarvox:
Fixed JSCS issues.

shell/server/admin/system-status-server.js
```diff
@@ -1,9 +1,61 @@
...
+
+Router.map(function () {
+  this.route("adminDownloadServerLog", {
+    where: "server",
+    path: "/admin/status/server-log/:tokenId",
+    action() {
+      const token = this.params.tokenId;
+      const response = this.response;
+      const issueDate = SYSTEM_LOG_DOWNLOAD_TOKENS[token];
+      SYSTEM_LOG_DOWNLOAD_TOKENS[token] = undefined; // Clear the token from the in-memory map.
```
kentonv: I think you need to use `delete` to actually remove the map entry.

shell/server/admin/system-status-server.js
```diff
@@ -1,9 +1,61 @@
...
+      if (issueDate === undefined || issueDate + 60000 < Date.now()) {
+        // Require download to start within 60 seconds of the adminGetServerLogDownloadToken
+        // method being called.
+        response.writeHead(404, { "Content-Type": "text/plain" });
+        return response.end("Invalid server log download token.");
+      }
+
+      const logFilePath = SANDSTORM_LOGDIR + "/sandstorm.log";
+      // Lazily assume that this logfile won't be "too large" and we can just load it into memory
```
kentonv: I bet some people have some very very long log files, since we don't (yet) do any rotation of these.

kentonv: Streaming should be pretty easy FWIW -- I think it's just a single function call on an input stream to tell it to pump to an output stream...

shell/server/admin/system-status-server.js
kentonv: Also I guess these never get cleaned up if the adminDownloadServerLog() call doesn't occur. I guess this is such a small memory leak we figure it doesn't matter?

zarvox: Yeah, it's tiny, requires an admin to take an action to trigger, goes away on frontend restart, and so on.

I guess I could `setTimeout()` a function to remove the newly-added entry from the map after, say, ~2x the valid duration, though, which wouldn't be much of a hassle.

shell/server/admin/system-status-server.js
zarvox: Yeah, there's a couple catches like "what if the file is being extended as we stream it" but that should be easy enough.

I'd missed the `fd` option to [fs.createReadStream](https://nodejs.org/docs/latest-v0.10.x/api/fs.html#fs_fs_createreadstream_path_options) the first time I read through the docs, which made me fearful of TOCTOU mismatch.

I'm implement streaming now.

zarvox:
Addressed all of the issues discussed.

shell/server/admin/system-status-server.js
```diff
@@ -1,9 +1,73 @@
...
+const SYSTEM_LOG_DOWNLOAD_TOKEN_VALIDITY_DURATION = 60000; // 60 seconds
+
+Meteor.methods({
+  adminGetServerLogDownloadToken() {
+    const db = this.connection.sandstormDb;
+    if (!db.isAdminById(this.userId)) {
+      throw new Meteor.Error(403, "User must be admin to download system log.");
+    }
+
+    const token = Random.id();
```
kentonv: Hmm, technically we ought to be using Random.secret() for these or passing a parameter to Random.id() to make it generate a longer ID.

Issue #1291, opened 2015-12-15
Multiple identities: Does it use the last one I used with a grain?


ocdtrekkie:
I did link a second identity to my account on Alpha, and set it as my login identity, but I found it really odd when I was suddenly no longer able to vote on the Sandstorm App Committee page, until I realized it was using my Github identity instead of my Google one, and there was a dropdown to change it on the top right.

I'm confused why it used my Github identity, since I only had interacted with that grain via my Google identity prior.

kentonv:
Hmm, indeed, I observed this to: I visited sandstorm.io/vote and found that it defaulted to a different identity than I have ever used with that grain before.

@dwrensha it's supposed to prefer an identity I've used in the past, right?

dwrensha:
I suspect the weirdness comes from this being a "public" grain on the old sharing model. There are likely some corner cases that can arise there that I haven't considered yet.

kentonv:
Oh right... I forgot about the old sharing model.

dwrensha:
The answer to "Does it use the last one I used with a grain?" is "yes", and that's working as intended for non-legacy grains.

For legacy grains, after #1965 we will use the current identity for opening, which I think is the best we can do.

PR #1965, opened 2016-05-11
Eliminate some out-of-scope usages of globalGrains.

dwrensha:
After this change, `Accounts.getCurrentIdentityId()` and `Accounts.setCurrentIdentityId()` do not deal with the grainview list at all. If you care about the active identity on a grainview, you need to call `globalGrains.getActive().identityId()`.

Closes https://github.com/sandstorm-io/sandstorm/issues/1291.


dwrensha:
(This change is motivated by the need to call `Accounts.getCurrentIdentityId()` from inside `GrainView.revealIdentity()`.)

PR #1968, opened 2016-05-11
install.sh: Only warn about 15 min token if there is a token (and other small string changes)

paulproteus:


paulproteus:

PR #1970, opened 2016-05-12
install.sh: Explain why we failed to re-exec ourselves

### LINKED ISSUE #631: "Oops, I couldn't figure out how to switch to root." with non-`install.sh` name.
This is really minor but anyway:
With both full server and dev paths:
```
To set up Sandstorm, we will need to use sudo.
Sandstorm's database and web interface won't run as root.
OK to continue? [yes] yes
Oops, I couldn't figure out how to switch to root. Please re-run the installer 
as root.
```
Strange, I'm on Ubuntu 15.04, `sudo` works perfectly.
Aha:
```
  elif [ "$(basename $SCRIPT_NAME)" == install.sh ] && [ -e "$0" ]; then
```
I wanted to peek at the script before I run it (was TLDR so runned anyway ;-)
but I was lazy and saved as `s.sh` not `install.sh`...

paulproteus:
Close #631 

I chose to make this message both technically clear ("Could not detect the script's filename") because when someone triggers it, it might cue them to realize that the reason they triggered it is that they chose a strange filename.

I chose to provide a specific command to run at the end so that people who are seeing it, but don't know why, can get past it too.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -369,7 +369,11 @@ rerun_script_as_root() {
   fi
 
   # Don't know how to run the script. Let the user figure it out.
-  REPORT=no fail "E_CANT_SWITCH_TO_ROOT" "Oops, I couldn't figure out how to switch to root. Please re-run the installer as root."
+  REPORT=no fail "E_CANT_SWITCH_TO_ROOT" "ERROR: This script could not detect its own filename, so could not switch to root. \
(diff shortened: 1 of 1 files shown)


paulproteus:
I'll self-merge as soon as installer-tests pass for me locally.

Thanks to @cben for filing this a while ago.

PR #1971, opened 2016-05-12
install.sh: Pass --user-group to useradd

### LINKED ISSUE #1306: Install script fails to create group sandstorm for some configurations
On a fresh install of openSUSE Tumbleweed (rolling release) the install script fails with the message:
`...chown: invalid group: ‘sandstorm:sandstorm’`

After the failure, the user sandstorm exists but the group sandstorm does not.  The failure to create the group is linked the a setting in /etc/login.defs containing the variable USERGROUPS_ENAB which has a default setting of "no" in openSUSE.  Changing the value to "yes" allows the install script to succeed.

The install script should verify the group exists prior to creating the user and assign the user to that group when executing the useradd command.

Props to dwrensha on the IRC for identifying the source of the problem.

paulproteus:
Close #1306

Context: This change, with test included, results in install.sh working properly on OpenSuSE where it
not before.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1107,8 +1107,15 @@ create_server_user_if_needed() {
     return
   fi
 
-  # OK!
-  useradd --system "$SERVER_USER"
+  # OK! Let's proceed.
+  #
+  # useradd comes from the passwd package, whereas adduser is a Debian-specific utility
+  # <https://packages.debian.org/adduser>, so we prefer useradd here. Per the man page for useradd,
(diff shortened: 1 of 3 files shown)


paulproteus:
Thanks to the following people who helped me understand what was going on here:

- @dwrensha for being helpful on IRC
- @medic15 for initially filing the bug and explaining `USERGROUPs_ENAB`
- @jbigte for leaving a fresh note on the issue to say it was affecting them.

Since this test passes locally for me, and I ran at least one of the other installer-tests in the hopes of making sure it didn't break those, I will merge it now.

PR #1978, opened 2016-05-12
install.sh: Use sysctl.d by default

paulproteus:
Thanks @zarvox for writing this patch and @jparyani for remarking on a clean solution.


paulproteus:
Self-merging because 100% docs & installer-tests.

paulproteus:
Oh, and installer.

PR #1979, opened 2016-05-13
Add documentation explaining how to get a shell within a grain

paulproteus:
While we're at it:

- Add mention to docs/README.md on the need for graphviz.


paulproteus:
I'm going to *NOT* self-merge this, at least until I get a +1 from @zarvox.

Note to @zarvox to help with review:

- This pull request depends on https://github.com/sandstorm-io/vagrant-spk/pull/165 and me doing a release. I can take responsibility for doing the merges/releases/docs deploy/etc in the right order.

docs/vagrant-spk/debugging.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+**Option 3. Enable filesystem tracing for this app.** In `.sandstorm/sandstorm-pkgdef.capnp`, look for this
+line, which disables tracing.
+
+```bash
+alwaysInclude = [ "." ] ,
+```
+
+Replace it with this line, which enables tracing.
```
zarvox: This isn't quite right - `alwaysInclude` and tracing being enabled are orthogonal.  See [sandstorm-test-python](https://github.com/zarvox/sandstorm-test-python/blob/master/.sandstorm/sandstorm-pkgdef.capnp) for an example which uses both tracing and `alwaysInclude`.

It's `fileList` that needs to be provided to enable tracing.  Combined with the `sourcePath = "/"` below, that allows mapping in any file on the system.

Suggest reworking this section.

paulproteus: +1 I didn't know this, will fix!

docs/vagrant-spk/debugging.md
```diff
@@ -0,0 +1,245 @@
...
+line is present. If not, you should add it.
+
+```bash
+      ( sourcePath = "/" ),
+```
+
+Finally, look for the terminal window where you are running `vagrant-spk dev`. Use Ctrl-C to terminate
+it, then run it again.
+
+This will dramatically bloat your sandstorm-files.list, so I recommend you **undo this change before
```
zarvox: Do we use "I" or "we" in the docs?  I think I've seen both; we should probably attempt to be consistent.  (I prefer "we" in documentation maintained by multiple individuals, unless there's also clearly attributed authorship for each article/unit, so we know who is "I".)

zarvox:
I would appreciate a couple of things, then you should feel free to merge:

* Add a note that `enter-grain`'s interface may change over time and is not considered frozen/stable.  Such changes may be necessary to achieve future behavior that is plumbed through Sandstorm properly, or to make the tool easier to integrate into automation (right now it always asks you which grain you want to enter, which is unpredictable/harder to tool around).
* Correct the section explaining the `sandstorm-pkgdef.capnp` fields, as noted.
* Create some future task, with no particular due date, to make a decision on I vs. we, and audit the docs for consistency of usage.

paulproteus:
> Add a note that enter-grain's interface may change over time and is not considered frozen/stable. Such changes may be necessary to achieve future behavior that is plumbed through Sandstorm properly, or to make the tool easier to integrate into automation (right now it always asks you which grain you want to enter, which is unpredictable/harder to tool around).

+0

I'd prefer to freeze enter-grain as-is, and then add `internal-enter-grain` which is used for automation, but let's just do it the way you suggest, which is also OK with me.

> Correct the section explaining the sandstorm-pkgdef.capnp fields, as noted.

+1

> Create some future task, with no particular due date, to make a decision on I vs. we, and audit the docs for consistency of usage.

+1

paulproteus:
@zarvox I'm going to take the liberty of doing

```
$ git commit -m 'Rename this doc for easier-to-remember URL'
[add-docs-on-vagrant-spk-shell 5b8586a] Rename this doc for easier-to-remember URL
 1 file changed, 0 insertions(+), 0 deletions(-)
 rename docs/{vagrant-spk => developing}/debugging.md (100%)
```

If you object, let me know. I  presume you don't care much about this choice of URL.

zarvox:
+1.

Arguably, the packaging tutorials also belong under `/developing` rather than `/vagrant-spk` as well?

paulproteus:
The packaging tutorials are in `vagrant-spk` for URL permanance for now, though I could leave a redirect stub behind, I suppose. A year-ish ago, I was wondering if there'd be a `vagrant-spk` competitor by now, though there doesn't seem to be.

w/r/t redirect stubs: I've done it before in the case of moving the security practices docs: See https://github.com/sandstorm-io/sandstorm/blob/master/docs/generate.sh#L113 . Can't say I _love_ it, but that's how it works anyway.

paulproteus:
BTW @zarvox I managed to stop saying "I" :)

paulproteus:
I also adjusted the "Option 3" to be "make the whole filesystem be available" rather than "enable tracing." The one new `sourcePath` is enough to make things work, and I think it's sensible to keep the changes minimal so that people can easily undo them.

paulproteus:
I'll merge this once I make the vagrant-spk changes not be experimental. (Therefore there's still 1-24 hours in which to provide further feedback if desired, pre-push.) Thanks @zarvox for your review.

paulproteus:
Self-merging b/c 100% docs

PR #1981, opened 2016-05-14
New Admin: Network Capabilities

### LINKED ISSUE #1823: Admin Panel Redesign: Network Capabilities
Changes required to ship:
- [ ] link grain title to grain
- [ ] add checkbox filters to show "active" and "revoked", uncheck "revoked" but default
- [ ] add confirmation for revoking, something like: "Are you sure, this will break this app..."
- [ ] for the user, display "Temporarily unavailable, try again later" if capability has been revoked"

Out of scope:
- [ ] where in cap'n p does "grain title" get saved
- [ ] Admin may have access to any key
- [ ] pagination so the tables don't extend forever
- [ ] Users will need a place to revoke capabilities

Current mock-up:
networkcapabilities3
([Previous mock-up can be viewed here.](https://cloud.githubusercontent.com/assets/855341/14691046/fe3a7d02-0703-11e6-89a6-cebb305ea0cb.png))

zarvox:
I'm not super thrilled with the fact that:

1. There's no way to sanely assign width to the grain title vs. grain owner boxes, short of just making both fixed width.  (Note that the `.identity-card` is already a fixed width box.)
2. We only show grain owner if it's different from who granted the capability.  The two people I've shown this design to were both confused by this behavior.

But I think this is still better than what existed before (provides more context about the capability, offers search/filtering), and we're likely to make more changes in the future that will obsolete this design, so I think we can roll with this.

admin-capabilities

Closes #1823.

### DIFF
--- a/shell/client/admin/admin.html
+++ b/shell/client/admin/admin.html
@@ -50,7 +50,3 @@
     {{/if}}
   </div>
 </template>
-
-<template name="newAdminNetworkCapabilities">
-  <h1>Network Capabilities</h1>
-</template>
--- a/shell/client/admin/network-capabilities-client.js
+++ b/shell/client/admin/network-capabilities-client.js
@@ -15,13 +15,13 @@ const tokenIntroducer = function (token) {
   return identityId && lookupIdentityById(identityId);
 };
 
-const capDetails = function(cap) {
+const capDetails = function (cap) {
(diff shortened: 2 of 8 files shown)


kentonv:
It looks like there's a new scss file missing from the commit (hence the build failure).

zarvox:
D'oh.  Fixed.

PR #1988, opened 2016-05-16
Fix a few bugs pertaining to grainviews and login/logout.

### LINKED ISSUE #1984: Signing in from a share link to an account that already has access to the grain leads to error
Steps to reproduce:
1. Create grain as Alice.
2. Create share link.
3. Open share link in a browser that is not logged in.
4. Once opened, click to sign in as Bob via Google or Github.
5. See the grain open correctly.
6. Sign out.
7. Browse to the share link again (while signed out).
8. Sign in as Bob again.

Expected result: Grain opens again as in step 5.
Actual result: Blank-ish screen:

screenshot from 2016-05-14 17-29-00

(Also, after several minutes, the grain spontaneously loads, without being listed on the sidebar!)

I suspect this may be a race condition similar to that fixed in #1980, but it is not fixed by #1980. Assigning @dwrensha since he fixed that issue and may have context. I think this is high-priority to fix in the next week.

dwrensha:
Fixes #1984. The problem was that, during login, we would attempt to open a session during the half-logged in state. This would result in two separate redirects from `/shared/:tokenId` to `/grain/:grainId`, the second of which would remove the grainview added by the first, due to [this logic](https://github.com/sandstorm-io/sandstorm/blob/d54130627e0c18fb33604dffbe65b5618d22ed31/shell/packages/sandstorm-ui-grainview/grainview.js#L514).

This also fixes a problem where we were sometimes attaching multiple Blaze views for the same grain at the same time. (You could see this behavior by doing `Meteor.logout()` while visiting a grain.) The fix here is somewhat crude: in `GrainView.reset()`, before attaching a new Blaze view, we verify that the current `GrainView` is in fact in grainview list. This cover all of the problem cases I've seen, where race conditions could lead to a GrainView being `reset()` after it has been removed from the list.

### DIFF
--- a/shell/client/grain-client.js
+++ b/shell/client/grain-client.js
@@ -1469,6 +1469,9 @@ Router.map(function () {
     },
 
     onBeforeAction: function () {
+      // Don't do anything for non-account users.
+      if (Meteor.userId() && !Meteor.user().loginIdentities) return;
+
       // Only run the hook once.
       if (this.state.get("beforeActionHookRan")) return this.next();
 
@@ -1542,6 +1545,9 @@ Router.map(function () {
     },
 
     onBeforeAction: function () {
+      // Don't do anything for non-account users.
+      if (Meteor.userId() && !Meteor.user().loginIdentities) return;
+
       // Only run the hook once.
       if (this.state.get("beforeActionHookRan")) return this.next();
       this.state.set("beforeActionHookRan", true);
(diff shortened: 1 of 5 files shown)


shell/client/grain-client.js
```diff
@@ -1570,6 +1576,8 @@ Router.map(function () {
           Router.go("/grain/" + tokenInfo.grainId + path, {}, { replaceState: true });
         } else if (tokenInfo.grainId) {
           const grainId = tokenInfo.grainId;
+          const identityChosenByLogin = this.state.get("identity-chosen-by-login");
+          this.state.set("identity.chosen-by-login", undefined);
```
kentonv: `identity.chosen` -> `identity-chosen`?

dwrensha: Fixed.

PR #1990, opened 2016-05-16
Clarify what is being uploaded by the Upload... button on the apps page

### LINKED ISSUE #1985: App Upload confused with "Upload File"
People using Sandstorm for the first time initially thought that the "Upload" button on the Apps page would let them upload a local file.

Although when you're familiar with Sandstorm, the fact that it's on the Apps page makes it clear that it is for uploading a new application, it makes sense to me that this is confusing especially when coming to Sandstorm as an alternative to Google Docs.

As upload app is a fairly advanced feature, maybe it can even be hidden or de-emphasized.

zarvox:
Closes #1985.


paulproteus:
+1 from me, :balloon: to @jacksingleton for filing this & to @zarvox for fixing

PR #1996, opened 2016-05-17
Improve error logging in WILDCARD_HOST self-test

### LINKED ISSUE #1987: wildcard host self-test should report the actual error if there is one
Currently we throw away the error: https://github.com/sandstorm-io/sandstorm/blob/v0.163/shell/client/admin-client.js#L1179-L1195

paulproteus:
Close #1987

Note that since the self-test is performed client-side, it is difficult to get
detailed information about XHR failures. Therefore, this commit adds documentation
with information on how decode errors that I have seen with Chrome and Firefox.

### DIFF
--- a/shell/client/admin-client.js
+++ b/shell/client/admin-client.js
@@ -1185,8 +1185,8 @@ const adminRoute = RouteController.extend({
                     console.log("Sandstorm WILDCARD_HOST self-test failed. Details:");
                     console.log(error);
                     console.log(
-                      "Further details may be available in the JS console provided by your web " +
-                        "browser. Look for errors involving a domain starting with selftest-*.");
+                      "Error messages above or below this one may relate to this error; look for " +
+                        "messages involving a domain starting with selftest-*.");
                     console.log(
                       "See also docs: https://docs.sandstorm.io/en/latest/administering/faq/#why-do-i-see-an-error-when-i-try-to-launch-an-app-even-when-the-sandstorm-interface-works-fine");
                     console.log(
--- a/shell/client/admin-client.js
+++ b/shell/client/admin-client.js
@@ -1198,7 +1198,7 @@ const adminRoute = RouteController.extend({
                       looksGood = true;
                     } else {
                       console.log(
-                        "Sandstorm WILDCARD_HOST self-test failed. Received status code:")
+                        "Sandstorm WILDCARD_HOST self-test failed. Received status code:");
                       console.log(response.statusCode);
                       looksGood = false;
                     }
--- a/shell/client/admin-client.js
+++ b/shell/client/admin-client.js
@@ -1182,11 +1182,24 @@ const adminRoute = RouteController.extend({
                   let looksGood;
                   if (error) {
                     looksGood = false;
+                    console.log("Sandstorm WILDCARD_HOST self-test failed. Details:");
+                    console.log(error);
+                    console.log(
+                      "Further details may be available in the JS console provided by your web " +
+                        "browser. Look for errors involving a domain starting with selftest-*.");
+                    console.log(
+                      "See also docs: https://docs.sandstorm.io/en/latest/administering/faq/#why-do-i-see-an-error-when-i-try-to-launch-an-app-even-when-the-sandstorm-interface-works-fine");
+                    console.log(
+                      "Slow DNS or intermittent Internet connectivity can cause this message " +
+                        "to appear unnecessarily; in that case, reloading the page should make " +
+                        "it go away.");
                   } else {
                     if (response.statusCode === 200) {
                       looksGood = true;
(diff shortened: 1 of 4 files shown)


paulproteus:
@Tuxick I have done some research, and I think this is the best I can do. I'm curious if this would have helped you. I would like to log more details, but the info is obscured from Javascript: http://stackoverflow.com/questions/31058764/determine-if-ajax-call-failed-due-to-insecure-response-or-connection-refused

paulproteus:
Here's a screenshot of the JS console and the admin page, after this change.

screenshot from 2016-05-16 21 56 06

paulproteus:
@kentonv this mostly-string-fix pull request is ready for your review/merge.

@zarvox in case you are curious, this commit also adds a note to the `console.log()` that false-positives might occur.

@dwrensha curious for your +1/-1 on this.

shell/client/admin-client.js
```diff
@@ -1182,11 +1182,24 @@ const adminRoute = RouteController.extend({
                   let looksGood;
                   if (error) {
                     looksGood = false;
+                    console.log("Sandstorm WILDCARD_HOST self-test failed. Details:");
+                    console.log(error);
+                    console.log(
+                      "Further details may be available in the JS console provided by your web " +
```
dwrensha: This message is somewhat confusing to me. The only place this message ever shows up is in the JS console, right? So why is it telling me to check the JS console? I can imagine someone reading this and thinking: "Is there another super secret JS console that I don't know about that has all the answers?!"

paulproteus: FWIW what I'm trying to say is:

- The JS console will have text that comes from Sandstorm.

- The JS console MIGHT have other, colorized text that comes from your web
browser.

- Sandstorm can't see that error, but you, a human operator of the web
browser, can indeed see that error!

- So you should look at it and see what it says.

See also the screenshot I provided, where Chromium provides more details
about the error than Sandstorm can perceive.

I'm 100% open to better ways to phrase this; hopefully the general intent
is now clear.

dwrensha:
I see that the error shows up as `network(...)`. Is "network" actually the only sensible string that we can get out of these errors? Do you see anything if you click on the `...`?

shell/client/admin-client.js
kentonv: I agree the message is confusing, since it seems to be referring to the JS console as something far away.

Perhaps say: "Error messages above this one may relate to this error."

dwrensha:
So, the two common ways this can fail are "connection refused" and "insecure response". Even if it's not possible to distinguish those in the javascript, I think it's still valuable to print the actual error in the red banner. There might be other possible errors, and they might have meaningful text associated with them.

paulproteus:
`Error: network` is the full error string reported to JS in the "Connection refused' case, the `...` refers to the fact that a stacktrace is available. That's a pretty depressing state of affairs, but per the experimentation in http://stackoverflow.com/questions/31058764/determine-if-ajax-call-failed-due-to-insecure-response-or-connection-refused I believe it's all we can get from the client side.

We could move this test to the server side, which has its own problems, and given the choice, I prefer it to be a client-side test.

We could use both the server and the client... which I think is not better than just using the client.

So that's why I need the user to look at their browser. David, when you say, "Even if it's not possible to distinguish those in the javascript, I think it's still valuable to print the actual error in the red banner", how would I print the actual error in the red banner, since my client-side JS code can't perceive any difference between the two errors?

@kentonv I think your string is good; I'll adjust to be more like that. I might say "Above or below" since I don't know for sure where the browser's message will show up.

paulproteus:
I went with:

>                      "Error messages above or below this one may relate to this error; look for " +
>                        "messages involving a domain starting with selftest-*.");

dwrensha:
> how would I print the actual error in the red banner, since my client-side JS code can't perceive any difference between the two errors?

In the case where `HTTP.call` returns an error, you could report `error.toString()`, and in the case where the status code isn't 200, you could report the returned status code.

But maybe that would only serve to clutter the red banner in the vast majority of cases?

tuxick:
On 05/17/2016 07:45 PM, Asheesh Laroia wrote:
> I went with:
>
>     |"Error messages above or below this one may relate to this error;
>     look for " + "messages involving a domain starting with selftest-*."); |
>

uh look where?


Tony

kentonv:
Should any of those `console.log`s actually be `console.error`? (Stylistic choice and I don't have a strong opinion, just wondering if you considered it.)

Should we mention to the user that they can open the JS console using F12 (I think that works on all common browsers, whereas other hotkeys don't)?

paulproteus:
I'm going to make some further string fixes. Temporarily closing the PR to avoid people accidentally merging it as-is.

PR #1953, opened 2016-05-07
more context links to tutorials

techtonik:

PR #1997, opened 2016-05-17
Further clarity improvements in tutorial

paulproteus:
- When linking to the sample app, indicate that the link is to an online view,
  rather than the files on your own computer.

- When talking about platform stacks, consistently call them platform stacks.

Refs #1953


paulproteus:

PR #2000, opened 2016-05-17
Add reference docs on config file

### LINKED ISSUE #1395: Add reference documentation for config file
I had an ancient Sandstorm installation which I wanted to bring up to current standards. I couldn't find any docs that would help me update my config file (to include such things as HTTPS and Sandcats configuration), so I ended up rerunning the installer.

While a specific tutorial on “getting from version X to version Y” would be obsolete quickly, what would have helped that I didn't find is _reference_ documentation on the contents of the Sandstorm config file. From that I could have drawn my own conclusions about what needs revising.

There should be documentation which describes each element that can be in the config file.

paulproteus:
Close #1395

### DIFF
--- a/docs/administering/config-file.md
+++ b/docs/administering/config-file.md
@@ -1,0 +1,178 @@
+This document provides **detailed technical documentation**, also known as reference documentation,
+on the available options within Sandstorm's configuration file, `sandstorm.conf`.
+
+## Overview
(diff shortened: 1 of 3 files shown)


paulproteus:
Merging & pushing now, since this is 100% docs.

PR #2005, opened 2016-05-19
New admin reorganization

zarvox:
* Adds a root admin panel page
* Divides admin panel pages into Configuration (server settings) and Management (day-to-day operational task support).
* Moves to breadcrumb-based hierarchical navigation for individual admin pages (which works very nicely for e.g. `Admin / Users / Invite`)
* Corrects a good deal of inconsistent capitalization in the product.  The standard we (attempt to) follow is to capitalize the first letter of a heading, label, or button, but no other words, unless they are proper nouns.  For example, "Sandstorm for Work" and "Sandstorm" are always capitalized as written here.
  * I don't think "Trash" and "Main list" qualify as proper nouns.
  * App market is not a proper noun when we write it on our blog, so I assumed it needn't be here either.
  * I didn't bother fixing capitalization in the old admin panel since I intend to deprecate/remove it Soon™.

admin-panel-root

admin-users-invite-breadcrumbs

### DIFF
--- a/shell/client/admin-client.js
+++ b/shell/client/admin-client.js
@@ -437,10 +437,6 @@ Template.adminSettings.helpers({
     return globalDb.getSamlPublicCert();
   },
 
-  smtpUrl: function () {
-    return Iron.controller().state.get("smtpUrl");
-  },
-
   smtpConfig() {
     const config = Settings.findOne({ _id: "smtpConfig" });
     return config && config.value;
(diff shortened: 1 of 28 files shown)


paulproteus:
I a​dded some of this info here:
https://github.com/sandstorm-io/sandstorm/wiki/English-language-style-guide

Feel free to expand/correct etc., but I figure we should make sure this
info lives somewhere so that when we are wondering what the standard is, we
have a place to look.

paulproteus:
Also wow, this is great, looking at the screenshots.

PR #2007, opened 2016-05-20
New admin panel

zarvox:
This change replaces the old admin UI with the new admin UI (which previously lived at `/admin-new`), and places the old admin UI at `/admin-old`.  We intend to remove the latter after some settling period to make sure users aren't missing essential functionality that was available in the old UI.

This also replaces the old powerbox IpNetwork and IpInterface tests with new ones that use the appropriate powerbox descriptors.

Merging this change marks the completion of the admin panel project.  I'm pretty excited about how much more well-organized, consistent, and altogether *usable* things feel.  And it's quite pretty!  A huge thanks to @neynah, @kentonv, and @paulproteus for voluminous discussion and feedback, both on the admin panel and the setup wizard.

I want to tag @ndarilek here - this is a significant change to basically all the admin functionality.  I've done my best to make sure that every interactive element is reachable for keyboard-only users, that success and failure messages grab focus when they appear, and that markup is reasonably structured.  If there's something you can't find, or something which we've made unpleasantly hard for screenreader users, or something we could do better, please let me know so I can promptly fix it before we remove `/admin-old`. :)

Screenshot below demonstrates the IpNetwork test app fetching any file you name over (unencrypted) HTTP.

ipnetwork-fetch-anything-over-http

### DIFF
--- a/shell/client/admin-client.js
+++ b/shell/client/admin-client.js
@@ -1213,39 +1213,39 @@ const adminRoute = RouteController.extend({
 
 Router.map(function () {
   this.route("adminSettings", {
-    path: "/admin/settings/:_token?",
+    path: "/admin-old/settings/:_token?",
     controller: adminRoute,
   });
   this.route("adminUsers", {
-    path: "/admin/users/:_token?",
+    path: "/admin-old/users/:_token?",
     controller: adminRoute,
   });
   this.route("adminStats", {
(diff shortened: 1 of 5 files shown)


ndarilek:
Just wanted to say that, while I haven't done an exhaustive check, I finally popped into the new admin panel. From an accessibility perspective, it looks great. Thanks for getting that right! :)

PR #2010, opened 2016-05-20
Grainlog smarter autoscroll

zarvox:
The previous grainlog behavior attempted to tail the log, but failed to do so
reliably for various reasons:

* not dealing with fractional pixels or "almost at the bottom" scroll states
* recomputing whether the scroll position was at the bottom *after* the additional
  content had been added (causing the window to never scroll to the bottom on
  content change)
* didn't pay attention to resize events, which can knock the view off the "bottom"
  of the log

Reusing the design from the new admin system log page, we fix all of these issues
and make the log window actually reliably tail the log when scrolled to the bottom,
while leaving the scroll position untouched if you scroll elsewhere.


kentonv:
Awesome! Can you jscs --fix?

zarvox:
semicolon added

I'm looking into the recent spat of test failures on the build server, but the failing ipnetwork/ipinterface tests seem to work locally for me.

PR #2012, opened 2016-05-21
Fix IpNetwork/IpInterface tests

zarvox:
v0.0.8 of the test app had a bug where I used `elem.innerText` (nonstandard)
when I should have used `elem.textContent` (standard).

This broke in CI, which runs iceweasel 38, which did not have support for `elem.innerText`.  (Support was added somewhere between firefox 38 and firefox 46, which is the version available in Fedora, which worked fine.)


zarvox:
Self-merging because this is test-only code, and this unbreaks the build.

PR #2014, opened 2016-05-23
Add missing period in maintenance message description

zarvox:


zarvox:
Self-merging trivial string correction with Kenton's verbal approval.

PR #2016, opened 2016-05-24
Add UiView.PowerboxTag struct and fill it in for powerbox request results.

dwrensha:
This also updates the `newApiToken()` Meteor method to no longer allow creation of grain-owned tokens, mainly to prevent the caller form specifying arbitrary values in `owner.user.introducerIdentity`.

### DIFF
--- a/shell/packages/sandstorm-ui-powerbox/powerbox-server.js
+++ b/shell/packages/sandstorm-ui-powerbox/powerbox-server.js
@@ -302,8 +302,8 @@ Meteor.publish("powerboxOptions", function (requestId, descriptorList) {
   // packed format and base64-encoded. The publish populates a pseudo-collection called
   // `powerboxOptions`. Each item has the following fields:
   //
-  //   _id: Object with two fields: `requestId` (as passed when subscribing) and `optionId` (a
-  //       unique identifier for the option).
+  //   _id: Unique identifier string.
+  //   requestId: The value of `requestId` that was passed in when subscribing.
   //   matchQuality: "preferred" or "acceptable" ("unacceptable" options aren't returned).
   //   frontendRef: If present, selecting this option means creating a simple frontendRef. The
   //       field value should be passed back to the method `newFrontendRef` verbatim. The format
--- a/shell/packages/sandstorm-ui-powerbox/powerbox-server.js
+++ b/shell/packages/sandstorm-ui-powerbox/powerbox-server.js
@@ -112,6 +112,69 @@ Meteor.methods({
     const sturdyRef = waitPromise(saveFrontendRef(frontendRefVariety, apiTokenOwner, requirements)).sturdyRef.toString();
     return { sturdyRef, descriptor };
   },
+
+  fulfillUiViewRequest(identityId, grainId, petname, roleAssignment, saveLabel, ownerGrainId) {
+    check(identityId, String);
+    check(ownerGrainId, String);
+
+    const db = this.connection.sandstormDb;
+    if (!this.userId || !db.userHasIdentity(this.userId, identityId)) {
+      throw new Meteor.Error(403, "Not an identity of the current user: " + identityId);
+    }
+
+    const provider = {
+      identityId,
+      accountId: this.userId,
+    };
+
+    const grain = db.getGrain(grainId);
+    if (!grain) {
+      throw new Meteor.Error(404, "Grain not found: ", grainId);
+    }
+
+    let title = grain.title;
+    if (grain.userId !== this.userId) {
+      const sharerToken = db.collections.apiTokens.findOne({
+        grainId: grainId,
+        "owner.user.identityId": identityId,
+      }, {
+        sort: {
+          lastUsed: -1,
+        },
+      });
+      if (sharerToken) {
+        title = sharerToken.owner.user.title;
+      } else {
+        title = "shared grain";
+      }
+    }
+
+    const metadata = db.getDenormalizedGrainInfo(grainId);
+
+    const descriptor = encodePowerboxDescriptor({
+      tags: [
+        { id: "15831515641881813735",
+          value: Capnp.serialize(Grain.UiView.PowerboxTag, { metadata, title }),
+        },
+      ],
+    });
+
+    const owner = {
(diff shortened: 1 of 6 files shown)


shell/packages/sandstorm-ui-powerbox/powerbox-server.js
```diff
@@ -112,6 +112,69 @@ Meteor.methods({
...
+      } else {
+        title = "shared grain";
+      }
+    }
+
+    const metadata = db.getDenormalizedGrainInfo(grainId);
+
+    const descriptor = encodePowerboxDescriptor({
+      tags: [
+        { id: "15831515641881813735",
```
zarvox: I'd prefer if this used `Grain.UiView.typeId` instead of the string literal.

dwrensha: Done.

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
```diff
@@ -112,6 +112,69 @@ Meteor.methods({
     const sturdyRef = waitPromise(saveFrontendRef(frontendRefVariety, apiTokenOwner, requirements)).sturdyRef.toString();
     return { sturdyRef, descriptor };
   },
+
+  fulfillUiViewRequest(identityId, grainId, petname, roleAssignment, saveLabel, ownerGrainId) {
+    check(identityId, String);
```
kentonv: Either comment which inputs are type-checked elsewhere or check them here too to be safe.

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
```diff
@@ -200,8 +263,12 @@ specialCaseTypes[Grain.UiView.typeId] = function (db, userId, value) {
 
   // TODO(someday): Allow `value` to specify app IDs to filter for.
 
-  const sharedGrainIds = db.userApiTokens(userId).map(token => token.grainId);
-  const ownedGrainIds = Grains.find({ userId: userId }, { fields: { _id: 1 } }).map(grain => grain._id);
+  const sharedGrainIds = db.userApiTokens(userId).fetch()
```
kentonv: Should userApiTokens() and userGrains() maybe be updated to ignore trash by default, or given a parameter to control the query?

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
```diff
@@ -302,8 +369,8 @@ Meteor.publish("powerboxOptions", function (requestId, descriptorList) {
   // packed format and base64-encoded. The publish populates a pseudo-collection called
   // `powerboxOptions`. Each item has the following fields:
   //
-  //   _id: Object with two fields: `requestId` (as passed when subscribing) and `optionId` (a
-  //       unique identifier for the option).
+  //   _id: Unique identifier string.
```
kentonv: The old `optionId` was only unique within a particular query, not across queries. In fact, optionIds for singleton options are often human-readable names. Hence the need for a compound _id. Do you really need to change this?

kentonv: Oh, I see, you are changing the comment to match the implementation. But in fact the implementation is wrong and the comment is right... >_>

src/sandstorm/grain.capnp
```diff
@@ -336,6 +336,29 @@ interface SandstormApi(AppObjectId) {
...
+
+  appTitle @0 :Util.LocalizedText;
+  # A copy of the app name for the corresponding UIView for presentation in the grain list.
+
+  union {
+    icon :group {
+      format @1 :Text;
+      # Icon asset format, if present.  One of "png" or "svg"
+
+      assetId @2 :Text;
```
kentonv: Hmm, in order to use `assetId` the app would need to infer the server's static asset host, which seems problematic and possibly even ambiguous depending on the WILDCARD_HOST pattern in use.

I'm not sure what the best alternative is. We don't really have a namespace for assets that apps can manipulate, currently. Complicating matters, we need something that can go both ways (requests and offers).

Arguably the correct representation is actually a capability. Maybe:

    interface StaticAsset {
      getUrl @0 () -> (url :Text);
      getContent @1 () -> (mimeType :Text, data :Data);
    }

(Note: Having `getUrl()` return a string is XSS-prone; we should actually use a representation that can only express protocol schemes that have sane same-origin policy properties -- probably only `http` or `https`.)

And then `SandstormApi` has a new method:

    createStaticAsset (mimeType :Text, content :Data) -> (asset :StaticAsset);

This would allow an app to mint a new asset if needed.

Hopefully this API would turn out useful for other purposes later on.

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
dwrensha: When I try to implement the ID as an object containing `optionId` and `requestId`, I get the error:

```
Error: Meteor does not currently support objects other than ObjectID as ids

```

src/sandstorm/grain.capnp
dwrensha: Maybe it should be 

```capnp
interface StaticAsset {
  getId @0 () -> (assetId :Text);
  getContent @1 () -> (mimeType :Text, data :Data);
}
```

where the `assetId` is the path component of an asset URL. Then these asset ID's would be what we put in `ViewInfo`, and we could add a `staticAssetHost` field to `WebSession.Params`. When an app wants to display an asset, it combines the `staticAssetHost` from the current session with the asset ID.

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
dwrensha: I've added a TODO about this in #2027.

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
dwrensha: Done in #2027.

dwrensha:
Closing because I've been continuing work over on #2027.

PR #2019, opened 2016-05-26
Authorization code

dwrensha:
This implements the first step of the changes proposed in #1892.

The second step will be to update a few places in powerbox-server.js to use the new `clientPowerboxRequest` owner (this will need to build on #2016's new `fulfillUiViewRequest` method), and to update the impls of `restore()` and `claimRequest()` in core.js to enforce the restrictions on the new owner variant.

There are at least two things that need to happen before this second step can land:
  1. We need to update the feature key vendor app to use the new API.
  2. We need to update the sandstorm-test-python app to use the new API

We had talked about maybe putting in both steps at once. I would be open to that option, but would need to coordinate with @zarvox to get the python app updated. The main advantage to this approach is that it would open the possibility of changing the token parameter of `claimRequest()` to `Text`, which does seem desirable, but I'm not sure if it's worth the coordination that it will require.

### DIFF
--- a/shell/lib/db-deprecated.js
+++ b/shell/lib/db-deprecated.js
@@ -64,6 +64,8 @@ if (Meteor.isServer) {
     connection.sandstormDb = globalDb;
   });
   SandstormDb.periodicCleanup(5 * 60 * 1000, SandstormPermissions.cleanupSelfDestructing(globalDb));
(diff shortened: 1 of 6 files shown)


kentonv:
Garply, retest this please

kentonv:
Superseded by #2027

Issue #1986, opened 2016-05-15
"What is a grain?"


jacksingleton:
During a recent user feedback session a very common question that came up was "What is a grain?"

This group of people were using Sandstorm for the first time, after being invited to the server via email.

ocdtrekkie:
I know for a while, they were replacing the 'grain' term with 'file' in the UI, but that was probably more confusing, because you could put files in a grain, and some grains aren't really used like 'files' in a traditional sense. Your inbox or website could be a grain. 

I think people will need to learn what a grain is. I think it's probably 'a unit container of information' at some level.

jacksingleton:
Yeah I don't have a perfect answer :)

As an admin part of me wants to be able to rename "Grain" myself. Although I'm not sure what to call it instead -- Document is the best I can think of for now. I think the most intuitive name will vary from group to group, but can also see the benefit of keeping it Grain across all Sandstorm instances.

We might also be able to improve the first use experience to better explain what a grain is, although I think there are limits to how much people will read.

ocdtrekkie:
My notion is that Sandstorm will eventually (inevitably?) become popular enough that grain is simply a term people will learn when learning basic computing. 

But then, I'm occasionally exceesingly optimistic about things I like. :)

ocdtrekkie:
And yeah, I think naming it differently server to server would lead to difficulties in troubleshooting, documentation in apps, etc.

paulproteus:
BTW, random idea:

When we send an email invite, we can look up the nounPhrase for this grain
(e.g. "Untitled Etherpad document" => "Subject: Jack Singleton has shared a
document with you")

ocdtrekkie:
That seems sensible, Asheesh.

techtonik:
If `document` is data, then `grain` is data+logic, right?

dwrensha:
>  When we send an email invite, we can look up the nounPhrase for this grain (e.g. "Untitled Etherpad document" => "Subject: Jack Singleton has shared a document with you")

Currently the subject line would be "Untitled Etherpad document - invitation to collaborate".

I think the confusion is coming from the link in the email body, which currently reads "Open Shared Grain". We could change that text to "Open Shared [nounPhrase]".

neynah:
[...]

PR #2021, opened 2016-05-26
Say "Open" not "Open Shared Grain" in emails

paulproteus:
See also: #1986

Curious for thoughts from @neynah @jacksingleton @zarvox and anyone else. Seems like an easy and satisfying change to me.

### DIFF
--- a/shell/server/grain-server.js
+++ b/shell/server/grain-server.js
@@ -524,7 +524,7 @@ Meteor.methods({
             roleAssignment, { webkey: { forSharing: true } });
           const url = origin + "/shared/" + result.token;
           const html = message.html + "<br><br>" +
-              emailLinkWithInlineStyle(url, "Open Shared Grain") +
+              emailLinkWithInlineStyle(url, "Open") +
               "<div style='font-size:8pt;font-style:italic;color:gray'>" +
               "Note: If you forward this email to other people, they will be able to access " +
               "the share as well. To prevent this, remove the button before forwarding.</div>";
@@ -571,7 +571,7 @@ Meteor.methods({
               }
 
               const html = message.html + "<br><br>" +
-                  emailLinkWithInlineStyle(url, "Open Shared Grain") +
+                  emailLinkWithInlineStyle(url, "Open") +
                   "<div style='font-size:8pt;font-style:italic;color:gray'>" +
                   "Note: You will need to log in with your " + loginNote +
                   " to access this grain.";


zarvox:
Actually, I wanted the button to say "Open &lt;grainTitle&gt;" but GitHub swallowed the tag in the initial writing, and you only read the email, so you didn't see the edit that corrected it.

zarvox:
Other options include "Open on Sandstorm" or "Open on ${serverTitle}".

paulproteus:
> Open `<grainTitle>`

hah, oops!

I do honestly think that "Open" is better than "Open on {{serverTitle}}", and "Open {{grainTitle}}" has the downside that my typical grain titles of 15 words would make that button sort of ridiculous. I'm fine tabling this pull request until tomorrow, though, during blur day.

zarvox:
If your typical grain title is 15 words, that's going to overflow the subject line (and all other pieces of UI) no matter where you put it.  I'm not sure that should be considered Sandstorm's problem; in fact, I think we should have affordances that encourage people to choose short, meaningful titles for grains, which are helpful in various contexts (referring to a grain title out loud, visibility in the grain list, etc.).  I am curious if there are things Sandstorm could do that would help make shorter titles more viable for your workflows, or why you choose such long titles today!

I particularly like adding the context about what is being opened or where this button will take you because I believe it's particularly valuable to add context for users who are first discovering Sandstorm through this invitation email.  For comparison, Google Docs' invitation emails read "Open in Docs".

I do feel pretty strongly that adding some context, be it grain-specific or service-specific, is better than just the bare string "Open", especially keeping in mind that we learned "people can't reliably tell open-the-noun from open-the-verb" from previous usertests back when the "Apps" and "Grains" navbar items were labeled "New" and "Open".

paulproteus:
@zarvox all very sensible concerns. I'll think about it more and close this to avoid confusion, but happy to see more discussion here or on #1986

neynah:
The subject line currently contains the grain title. If we want to provide additional context for people opening this grain, we can display the grain title again in the e-mail itself and have a button that just says "Open" or "Open in Sandstorm". The problem with "Open in Sandstorm" is that we provide no information as to what Sandstorm is.

E.g. Nena Nguyen is inviting you to collaborate on {{grainTitle}}. "Open" button.

FWIW, button strings are often relatively short in length whether they provided additional context or not. I think this is good practice. If we did display the grain title in the button itself we should decide on a max-width and cut off the text. Typically, long strings or text that will vary significantly in length should be displayed in full outside of the button. Looking through Google Drive's buttons, I haven't seen one that displays the document titles within the buttons themselves. Also, wetransfer's emails display the sender, the sender's message, the file name as well as a button that says "Download". The button string doesn't need to be long as the body of the email should provide more context about what is being "Opened" or "Downloaded".

(This issue closed halfway through me writing this but I'm deciding to post it anyway.)

PR #2022, opened 2016-05-26
docs: Explain why we don't need user-agent whitelist anymore

### LINKED ISSUE #2013: docs: HTTP APIs docs say that basic auth requires a whitelist, but this is now false
See this quote in the current docs:

> The Sandstorm code maintains a whitelist of User-Agent strings that are allowed to use Basic auth

https://docs.sandstorm.io/en/latest/developing/http-apis/#bearer-tokens-vs-basic-auth

Context:

- frew on IRC said: "I see that there's the API option, but will that work with, for example, the TTRSS mobile app which doesn't know about the extra Authorization header?". I believe the docs failed them, so I can fix that doc.

paulproteus:
Close #2013

Also rework the rest of the http-apis.md document to be clearer, hopefully.


paulproteus:

PR #2024, opened 2016-05-31
Update Asheesh release-certificate for June 2016

paulproteus:


paulproteus:
Self-merging because I'm under the impression that I own my release-certificate file.

PR #2025, opened 2016-06-01
docs: When installing VirtualBox on Windows, stop implying people should fiddle with %PATH%

paulproteus:
Thanks @ocdtrekkie for noticing this.


paulproteus:
And thanks @ocdtrekkie for reporting it!

paulproteus:

Issue #1866, opened 2016-04-18
Default avatar is not passed to the app


mitar:
<img width="957" alt="screen shot 2016-04-18 at 12 56 39 pm" src="https://cloud.githubusercontent.com/assets/585279/14617951/287ebaba-0565-11e6-8a68-e89a67317534.png">

My test user does not have an avatar set. Sandstorm display the default Avatar and says this will be the picture to identify the user, but that avatar URL is not send to the app to be really used.

mitar:
And I know that it is proposed that if picture is `null`, app should be using those same avatars, but I would claim that if Sandstorm shows some picture, that picture should be propagated to the app.

dwrensha:
From https://docs.sandstorm.io/en/latest/developing/auth/ :

>X-Sandstorm-User-Picture: The URL of the user's profile picture. [...] If this header is missing, the user has no profile picture. In this case, it is recommended that apps use identicon.js, with the user's ID (from X-Sandstorm-User-Id) as the input, to produce identicons consistent with those that Sandstorm's own UI would produce. Note that you should NOT hash the ID; just pass the hex ID directly to the Identicon constructor as the hash argument.

mitar:
Yes, but this requires change to the app, just to be consistent with Sandstorm? Wouldn't it be better if Sandstorm would just provide that default avatar and this is it?

kentonv:
The identicon is generated client-side using identicon.js. There is nothing for the server to pass.

mitar:
I know, but just to make it easier for apps, it could generate it on the server and passed it.

I made necessary changes and now package can run on the server as well: https://github.com/stewartlord/identicon.js

I am generating it myself now in the app, but it would be really better if it would be done by the Sandstorm: https://github.com/mitar/council-app/blob/master/packages/core/documents/user.coffee#L37

kentonv:
Unfortunately identicon.js produces PNGs that seem to be essentially uncompressed, probably because it doesn't use a real PNG library but rather a hack meant for use in client-side data URLs / blobs. If you can fix that too, then maybe we're getting somewhere.

mitar:
Done: https://github.com/mitar/council-app/commit/c76748268af4b20e63262c4cd42c125d3b25ec78

paulproteus:
Hi @mitar,

I'm not sure I understand what you mean when you say "done" here, and I'm interested in seeing this ticket progress forward. Particularly:

[...]

PR #2027, opened 2016-06-03
claimRequest() and other powerbox-related updates

dwrensha:
This pulls together the work from #2016 and #2019.

The ip-networking integration tests are working after applying https://github.com/zarvox/sandstorm-test-python/pull/4.

I have not yet updated the other powerbox integration tests, which are based on https://github.com/jparyani/sandstorm-test-app. So this will not pass Jenkins.

I've split out a new file powerbox.capnp. Doing so isn't strictly necessary for this change, but if we want to do that eventually, now seems like an opportune time.

I've not yet addressed all of the comments in #2019, but I wanted to get this pull request up for discussion.

### DIFF
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -1215,7 +1215,7 @@ _.extend(SandstormDb.prototype, {
     user = user || Meteor.user();
     if (user && (Meteor.isServer || user.pseudoUsage)) {
       if (Meteor.isClient) {
-        // Filled by pseudo-subscription to "getMyUsage". WARNING: The subscription is currenly
+        // Filled by pseudo-subscription to "getMyUsage". WARNING: The subscription is currently
         // not reactive.
         return user.pseudoUsage;
       } else {
@@ -1603,6 +1603,7 @@ if (Meteor.isServer) {
   const Crypto = Npm.require("crypto");
   const ContentType = Npm.require("content-type");
   const Zlib = Npm.require("zlib");
+  const Url = Npm.require("url");
 
   const replicaNumber = Meteor.settings.replicaNumber || 0;
 
@@ -1802,6 +1803,55 @@ if (Meteor.isServer) {
 
   // Cleanup tokens every hour.
   SandstormDb.periodicCleanup(3600000, cleanupExpiredAssetUploads);
+
+  SandstormDb.prototype.getIconUrls = function (icon) {
+    check(icon, Match.OneOf(undefined, null,
+                            { format: String, assetId: String,
+                              assetId2xDpi: Match.Optional(String), }));
+    const root =
+        Url.parse(process.env.ROOT_URL).protocol + "//" + this.makeWildcardHost("static") + "/";
+
+    result = {};
+    if (icon.assetId) {
+      result.url =  root + icon.assetId;
+    }
+
+    if (icon.assetId2xDpi) {
+      result.assetId2xDpi = root + icon.assetId2xDpi;
+    }
+
+    return result;
+  };
+
+  SandstormDb.prototype.userGrainTitle = function (grainId, accountId, identityId) {
+    check(grainId, String);
+    check(accountId, Match.OneOf(String, undefined, null));
+    check(identityId, String);
+
+    const grain = this.getGrain(grainId);
+    if (!grain) {
+      throw new Error("called userGrainTitle() for a grain that doesn't exist");
+    }
+
+    let title = grain.title;
+    if (grain.userId !== accountId) {
+      const sharerToken = this.collections.apiTokens.findOne({
+        grainId: grainId,
+        "owner.user.identityId": identityId,
+      }, {
+        sort: {
+          lastUsed: -1,
+        },
+      });
+      if (sharerToken) {
+        title = sharerToken.owner.user.title;
+      } else {
+        title = "shared grain";
+      }
+    }
+
+    return title;
+  };
 
   const packageCache = {};
   // Package info is immutable. Let's cache to save on mongo queries.
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -1721,7 +1721,8 @@ if (Meteor.isServer) {
   SandstormDb.prototype.addStaticAsset = addStaticAsset;
 
   SandstormDb.prototype.refStaticAsset = function (id) {
-    // Increment the refcount on an existing static asset.
+    // Increment the refcount on an existing static asset. Returns the asset on success.
+    // If the asset does not exist, returns a falsey value.
     //
     // You must call this BEFORE adding the new reference to the DB, in case of failure between
     // the two calls. (This way, the failure case is a storage leak, which is probably not a big
@@ -1732,11 +1733,10 @@ if (Meteor.isServer) {
     const existing = StaticAssets.findAndModify({
       query: { hash: hash },
       update: { $inc: { refcount: 1 } },
-      fields: { _id: 1, refcount: 1 },
+      fields: { _id: 1, content: 1, mimeType: 1 },
     });
-    if (!existing) {
-      throw new Error("refStaticAsset() called on asset that doesn't exist");
-    }
+
+    return existing;
   };
 
   SandstormDb.prototype.unrefStaticAsset = function (id) {
@@ -1810,23 +1810,81 @@ if (Meteor.isServer) {
   // Cleanup tokens every hour.
   SandstormDb.periodicCleanup(3600000, cleanupExpiredAssetUploads);
 
-  SandstormDb.prototype.getIconUrls = function (icon) {
-    check(icon, Match.OneOf(undefined, null,
-                            { format: String, assetId: String,
-                              assetId2xDpi: Match.Optional(String), }));
-    const root =
-        Url.parse(process.env.ROOT_URL).protocol + "//" + this.makeWildcardHost("static") + "/";
+  SandstormDb.prototype.updateCachedViewInfo = function (grainId, viewInfo) {
+    check(grainId, String);
+    // TODO(soon): Typecheck viewInfo. We should be able to factor out some of the checking code
+    //   that's currently in sandstorm-permissions/permissions-test.js.
 
-    result = {};
-    if (icon.assetId) {
-      result.url =  root + icon.assetId;
+    function validateAsset(icon, field) {
+      if (icon[field]) {
+        const asset = this.refStaticAsset(icon[field]);
+        if (asset) {
+          if (asset.content.length() > 64 * 1024 || !asset.mimeType.startsWith("image/")) {
+            this.unrefStaticAsset(icon[field]);
+            delete icon[field];
+          }
+        } else {
+          delete icon[field];
+        }
+      }
     }
 
-    if (icon.assetId2xDpi) {
-      result.assetId2xDpi = root + icon.assetId2xDpi;
+    const icon = (viewInfo.metadata || {}).icon || {};
+    validateAsset(icon, "assetId");
+    validateAsset(icon, "assetId2xDpi");
+
+    const oldGrain = Grains.findAndModify({
+      query: { _id: grainId },
+      update: { $set: { cachedViewInfo: viewInfo } },
+    });
+
+    const oldIcon = ((oldGrain.cachedViewInfo || {}).metadata || {}).icon || {};
+    if (oldIcon.assetId) {
+      this.unrefStaticAsset(oldIcon.assetId);
     }
 
-    return result;
+    if (oldIcon.assetId2xDpi) {
+      this.unrefStaticAsset(oldIcon.assetId2xDpi);
+    }
+  };
+
+  SandstormDb.prototype.deleteGrains = function (query, backend, type) {
+    check(type, Match.OneOf("grain", "demoGrain"));
+
+    Grains.find(query).forEach((grain) => {
+      waitPromise(backend.deleteGrain(grain._id, grain.userId));
+      Grains.remove({ _id: grain._id });
+      this.removeApiTokens({
+        grainId: grain._id,
+        $or: [
+          { owner: { $exists: false } },
+          { owner: { webkey: null } },
+        ],
+      });
+
+      this.removeApiTokens({ "owner.grain.grainId": grain._id });
+
+      if (grain.lastUsed) {
+        DeleteStats.insert({
+          type: "grain",  // Demo grains can never get here!
+          lastActive: grain.lastUsed,
+          appId: grain.appId,
+        });
+      }
+
+      if (grain.cachedViewInfo) {
+        const icon = (grain.cachedViewInfo.metadata || {}).icon || {};
+        if (icon.assetId) {
+          this.unrefStaticAsset(icon.assetId);
+        }
+
+        if (icon.assetId2xDpi) {
+          this.unrefStaticAsset(icon.assetId2xDpi);
+        }
+      }
+
+      Meteor.call("deleteUnusedPackages", grain.appId);
+    });
   };
 
   SandstormDb.prototype.userGrainTitle = function (grainId, accountId, identityId) {
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -1810,44 +1810,6 @@ if (Meteor.isServer) {
   // Cleanup tokens every hour.
   SandstormDb.periodicCleanup(3600000, cleanupExpiredAssetUploads);
 
-  SandstormDb.prototype.updateCachedViewInfo = function (grainId, viewInfo) {
-    check(grainId, String);
-    // TODO(soon): Typecheck viewInfo. We should be able to factor out some of the checking code
-    //   that's currently in sandstorm-permissions/permissions-test.js.
-
-    function validateAsset(icon, field) {
-      if (icon[field]) {
-        const asset = this.refStaticAsset(icon[field]);
-        if (asset) {
-          if (asset.content.length() > 64 * 1024 || !asset.mimeType.startsWith("image/")) {
-            this.unrefStaticAsset(icon[field]);
-            delete icon[field];
-          }
-        } else {
-          delete icon[field];
-        }
-      }
-    }
-
-    const icon = (viewInfo.metadata || {}).icon || {};
-    validateAsset(icon, "assetId");
-    validateAsset(icon, "assetId2xDpi");
-
-    const oldGrain = Grains.findAndModify({
-      query: { _id: grainId },
-      update: { $set: { cachedViewInfo: viewInfo } },
-    });
-
-    const oldIcon = ((oldGrain.cachedViewInfo || {}).metadata || {}).icon || {};
-    if (oldIcon.assetId) {
-      this.unrefStaticAsset(oldIcon.assetId);
-    }
-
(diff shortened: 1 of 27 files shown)


dwrensha:
@kentonv [this](https://github.com/sandstorm-io/sandstorm/blob/b7bad661698b561439491622df37c9e1d6a18386/shell/server/core.js#L498-L507) is the code that I was talking about earlier today. It sets the `owner.user.title` field based on the `owner.grain.saveLabel` field, which feels wrong to me. In my understanding, `saveLabel` is used by the app to explain to the user why it saved the capability. I can imagine it being weird if such an explanation ended up in a user's title for a grain.

kentonv:
@dwrensha I agree, that does look wrong. It seems like the code meant to use `tokenInfo.owner.user.title`.

dwrensha:
Hm... it also occurs to me that `ApiTokenOwner.grain.introducerIdentity` is obsolete after this change. It's role gets fulfilled instead by `ApiTokenOwner.clientPowerboxRequest.introducerIdentity`.

zarvox:
> Hm... it also occurs to me that `ApiTokenOwner.grain.introducerIdentity` is obsolete after this change. It's role gets fulfilled instead by `ApiTokenOwner.clientPowerboxRequest.introducerIdentity`.

I'm not so sure.

Would making that change require that other users who want to review/audit the capabilities owned by a grain also be able to see `ApiTokens` with owner `clientPowerboxRequest`?  I'm thinking about what information you'd need to build a compelling "audit connections" UI.

Does this still make sense if the powerbox request is initiated server-side, or do we intend to drop that flow entirely?  Seems we'd need a place for the introducer in that workflow too, but you wouldn't necessarily want to chain to a `clientPowerboxRequest` there.

dwrensha:
> Would making that change require that other users who want to review/audit the capabilities owned by a grain also be able to see ApiTokens with owner clientPowerboxRequest? I'm thinking about what information you'd need to build a compelling "audit connections" UI.

No, `clientPowerboxRequest` tokens are short-lived and would not need to show up in an audit UI.

The `introducerIdentity` field is only needed to determine which user to apply `requiredPermissions` to.

The clientside flow goes like this:
  1. Grain makes `postMessage` powerbox request on Alice's session, gets back a `requestToken` that has `introducerIdentity` set to Alice's identity ID.
  2. Grain calls `SandstormApi.claimRequest(requestToken, requiredPermissions)` on the server side. The returned capability has a `permissionsHeld` requirement so that it will stop working when Alice no longer holds `requiredPermissions`.
  3. The live capability gets saved. Note that here `grain`, `webkey` and `user` are all possibilities for the owner.

In a hypothetical server-side-only flow, the first two steps are merged into a single `SessionContext.request(query, requiredPermissions)` call.

It's the sturdyref created in step 3 that I think you're talking about showing up in an audit UI. The `requirements` field will hold the needed information. It's true that a capability can accumulate a large number of membrane requirements if it is passed through many powerbox requests. However, the way we represent that is through a chain of parent tokens; each link along the way will contain only the requirements picked up during the powerbox interaction which minted that token.

zarvox:
It is not obvious to me if or why I should assume that some (the last?) item in the `requirements` array is, in fact, the introducer's identity, should I wish to show that to the user.

If we do make that guarantee, we should probably document it and assert about it in the code.   I have always assumed `requiredPermissions`/`requirements` are an implementation detail of the permissions computation algorithm, rather than something with a guaranteed layout suitable for presentation to a user.  Happy to have that clarified, though!

This suggests that we will need some *pretty complicated* join logic to pull profiles for identities referenced by any number of ApiTokens, but perhaps that was already true.

zarvox:
One issue that I'm not sure this addresses is "how do I associate an identity (not just an account) with an `IpNetwork` I've granted to a grain, so I can show a sane thing in the UI?"  There, the `Supervisor.MembraneRequirement` is just `userIsAdmin: accountId`.

Maybe userIsAdmin needs to be a struct of (accountId, identityId), rather than just a Text?

dwrensha:
> There, the Supervisor.MembraneRequirement is just userIsAdmin: accountId.

To pass the `IpNetwork` to a grain, you need to go through a powerbox flow, which must be associated to a `SessionContext`, which will have a particular identity associated with it, and should attach a `permissionsHeld` requirement.

zarvox:
Ahhh, so for a hypothetical `IpNetwork` in the grain-specific audit interface, I'll look only for tokens with `owner.grain` with the matching grainId, and can always infer the introducer identity from the `requirements`.

Okay, so it sounds like the guarantee is:

* An ApiToken with owner `grain` is guaranteed to have as the last entry in its `requirements` a `permissionsHeld` struct which represents the account/identity responsible for creating the connection.

Does that sound right?

(Also, thanks for helping me understand all the intent here!)

dwrensha:
> An ApiToken with owner grain is guaranteed to have as the last entry in its requirements a permissionsHeld struct which represents the account/identity responsible for creating the connection.

Note quite. I would say that the entire `requirements` list is relevant, not just the final element in the list.

Consider the following interaction.
  1. Admin Alice grants an `IpNetwork` to grain G, which saves the cap as a sturdyref T with a `requirements` list including "Alice's account has admin privileges" and "Alice can access G".
  2. Alice shares G to Bob.
  3. Bob opens grain H, which makes a powerbox request for an `IpNetwork`. Bob chooses to allow G to fulfill the request, and to do that G uses the sturdyref T from step 1.
  4. H saves its new `IpNetwork` the resulting sturdyref has a `requirements` list consisting only of the `permissionsHeld` requirement "Bob can access H". The new sturdyref U points to T as its parent token.

The `IpNetwork` cap has passed through several powerbox interactions. At each step along the chain the local `requirements` list consists only of those requirements relevant to the most recent powerbox interaction. But to restore the cap, all of the requirements along the chain of parent tokens must be fulfilled.

The "requirements relevant to the most recent powerbox interaction" seems to me to be exactly the information needed for a an audit UI. I suppose it is plausible that in future revisions of the powerbox this may no longer be the case (e.g. if we support capabilities that don't autorevoke when the sharer loses access to the intermediary grain). Then we will need to add some additional way to capture the data needed in the audit UI. But I say let's not worry about that until it becomes a problem.

dwrensha:
After some updates to the copy/paste powerbox flow and [these changes](https://github.com/jparyani/sandstorm-test-app/pull/2) to the powerbox test app, the integration tests are now passing.

dwrensha:
Note that I'm not quite happy with `ViewInfo.getIconUrls()`. As discussed in https://github.com/sandstorm-io/sandstorm/pull/2016#discussion_r66898963, maybe this should be `ViewInfo.getIconAssetIds()` and we should add a `staticHost` field to `WebSession.Params`.

dwrensha:
In the latest commit, `ViewInfo` gets a `metadata: DenormalizedGrainMetadata` field and `WebSession.Params` gets a `staticAssetPath: Text` field. This allows a collections app to read icon asset IDs and then turn the into URLs suitable for including in an `` tag. Each appearance of an asset ID in a `Grains.cachedViewInfo` increments that asset's reference count.

This approach still does not seem quite satisfactory, because a grain has no way either to prevent an asset from being deleted or even to get notified when an asset is deleted, short of repeatedly calling `getViewInfo()`. This suggests to me that we should try something like the following:

```capnp
interface StaticAsset {
    getUrl @0 () -> (url :Text);
}

 SandstormApi {
    // ....
    createStaticAsset @8 (mimeType :Text, encoding :Text, content :Data)
        -> (asset :StaticAsset);
}

struct GrainMetadata {
    appTitle @0 :LocalizedText;
    union {
         icon :group {
             format @1 :Text;
             assetId @2 :StaticAsset;
             assetId2xDpi @3 :StaticAsset;
             # If present, the asset ID for the equivalent asset as assetId at twice-resolution
         }
         appId @4 :Text;
    }
}

struct ViewInfo {
    /// ...
    metadata @5 :GrainMetadata;
} 

```

That is, we would put the actual `StaticAsset` capabilities in the `ViewInfo`. Then a grain would be able to prevent an asset from being deleted, simply by holding on to a reference to it; when a `StaticAsset` is restored, it increments the refcount in the database, and when it is dropped, it decrements the refcount.

I think we would also want to prevent grains from providing their own implementations of `StaticAsset`, to avoid possible cross-site scripting attacks. I think such prevention should be possible via a `capnp::MembranePolicy`.

@kentonv: I'm interested in your thoughts on all of this.

dwrensha:
Another idea for keeping assets valid: we could add a method like

```capnp
interface SandstormApi {
    // ....
    holdStaticAsset @9 (assetId :Text) -> (handle :Util.Handle);
}
```

where the asset's refcount is incremented for as long as `handle` is held. This would allow us to pass around asset IDs like in my proposed `ViewInfo.metadata`. A grain that wants to use an asset calls `holdStaticAsset()` to check that it's still valid and to ensure that it stays valid.

shell/server/core.js
```diff
@@ -176,17 +176,40 @@ class NotificationHandle {
...
+    const viewInfo = this._grain.cachedViewInfo;
+    return inMeteor(() => {
+      const pkg = Packages.findOne({ _id: this._grain.packageId });
+      const manifest = pkg.manifest || {};
+      if (!viewInfo.appTitle) {
+        viewInfo.appTitle = manifest.appTitle || {};
+      }
+
+      if (!viewInfo.grainIconUrl) {
+        // TODO(security, now): Do we need to do some kind of filtering here to prevent
```
kentonv: This TODO wouldn't be caught by our release script, FWIW. You need to do `TODO(now) TODO(security)`, or just say that `TODO(now)` takes precedence.

shell/server/core.js
```diff
@@ -158,17 +187,46 @@ class NotificationHandle {
...
   // All other UiView methods are currently unimplemented, which, while not strictly correct,
   // results in the same overall behavior, since users can't call restore() on a PersistentUiView,
   // and grains can't call methods on UiViews because they lack the "is human" pseudopermission.
 }
 
-const makePersistentUiView = function (persistentMethods) {
-  return new Capnp.Capability(new PersistentUiViewImpl(persistentMethods), PersistentUiView);
+const makePersistentUiView = function (persistentMethods, grainId) {
+  check(grainId, String);
+  const grain = Grains.findOne({ _id: grainId });
```
kentonv: Why not do this query on-demand when `getViewInfo()` is invoked? Seems like the view info can become outdated if someone restores the capability and keeps it around for a while.

src/sandstorm/supervisor.capnp
```diff
@@ -274,15 +256,28 @@ struct ApiTokenOwner {
 
       introducerIdentity @9 :Text;
       # The identity ID through which a user's powerbox action caused the grain to receive this
-      # token. This is the identity against which the `requiredPermissions` parameter
-      # to `restore()` will be checked. This field is only intended to be filled in by the
-      # front-end during a powerbox request; a regular `save()` call produces a capability that
-      # has no "introducer".
+      # token. The token will become invalid if this identity loses access to the grain. This field
```
kentonv: Is there a good reason not to declare `grain.introducerIdentity` entirely obsolete and ignore it?

shell/packages/sandstorm-db/db.js
```diff
@@ -1810,23 +1810,81 @@ if (Meteor.isServer) {
...
 
-  SandstormDb.prototype.getIconUrls = function (icon) {
-    check(icon, Match.OneOf(undefined, null,
-                            { format: String, assetId: String,
-                              assetId2xDpi: Match.Optional(String), }));
-    const root =
-        Url.parse(process.env.ROOT_URL).protocol + "//" + this.makeWildcardHost("static") + "/";
+  SandstormDb.prototype.updateCachedViewInfo = function (grainId, viewInfo) {
+    check(grainId, String);
+    // TODO(soon): Typecheck viewInfo. We should be able to factor out some of the checking code
```
kentonv: I guess type checking viewInfo isn't critical because we normally receive it via Cap'n Proto, which already enforces the type, right? Mention that here.

shell/packages/sandstorm-db/db.js
```diff
@@ -1810,23 +1810,81 @@ if (Meteor.isServer) {
   // Cleanup tokens every hour.
   SandstormDb.periodicCleanup(3600000, cleanupExpiredAssetUploads);
 
-  SandstormDb.prototype.getIconUrls = function (icon) {
-    check(icon, Match.OneOf(undefined, null,
-                            { format: String, assetId: String,
-                              assetId2xDpi: Match.Optional(String), }));
-    const root =
-        Url.parse(process.env.ROOT_URL).protocol + "//" + this.makeWildcardHost("static") + "/";
+  SandstormDb.prototype.updateCachedViewInfo = function (grainId, viewInfo) {
```
kentonv: Probably worth adding some doc comment explaining what this method is doing.

I guess all the refcounting here only happens if the app defines grain-specific overrides for the icons, which it usually doesn't, right?

src/sandstorm/supervisor.capnp
zarvox: I want an introducer identity to visualize which admin granted a grain an admin capability -- right now you can't get that from just a `userIsAdmin` membrane requirement.

I'd be okay with `userIsAdmin` turning from a `Text` into a struct or group with an `accountId` and an `identityId` and saying the requirements are the canonical location for this information.  But the issuing identity needs to be stored somewhere.

kentonv:
OK, reviewed everything.

I'm worried that referencing assets by string ID rather than by capability is going to create obstacles somewhere down the road, particularly when we start having federation between Sandstorm servers where the same asset probably has different IDs on different servers.

I think we will want to be aware when asset references move between servers, so that data can automatically be synced in the background, etc. We obviously only have that ability with capabilities.

Let me think about this more tomorrow.

src/sandstorm/supervisor.capnp
kentonv: I believe that an admin capability will, in the new model, always have two `MembraneRequirements` -- a `userIsAdmin` and a `permissionsHeld`. The latter tells you the identity under which the user was acting when they injected the capability into the grain.

zarvox: Okay, that works for me!

dwrensha: > Is there a good reason not to declare `grain.introducerIdentity` entirely obsolete and ignore it?

It would be nice if we could continue to support existing grain-owned tokens. Most of these will have an introducer identity and therefore should be auto-revoked if that identity loses access to the grain. To accomplish this, we could either migrate these tokens so that the data in `introducerIdentity` field moves into the `requirements` field, or we could follow my strategy here, which is to continue honoring the `grain.introducerIdentity` field if it is present.

shell/packages/sandstorm-db/db.js
dwrensha: The refcounting happens when the grain specifically fills in `ViewInfo.metadata`. I wouldn't want to rule out the possibility that this could become the usual case in the future, e.g. if it becomes common for apps to provide multiple types of actions and to present a different grain icon for each action.

src/sandstorm/supervisor.capnp
zarvox: FWIW, I personally tend to favor migrations where possible over needing to remember to support both structures forever.  It turns an ongoing maintenance cost/conceptual debt (how should the code behave when it encounter tokens that have this shape + educating/reminding devs that once upon a time we didn't enforce the invariants that new tokens uphold) into one that can be paid down all-at-once.

I think it's higher cost to teach future devs (including my future self, who will likely have forgotten) about various inconsistencies than it is to remove them from existence.

That said, I'm also wary of asking other people to write possibly tricky code, and it seems the other implementation already lives, so I'm reasonably happy with either of David's suggestions, and if I want to put my money where my mouth is, I could write a migration for this at ~any time in the future.

kentonv: @dwrensha The problem is existing tokens were not only expecting `introducerIdentity` to have basic access but were also expecting to check it against some required permissions specified at restore() time. The new interface doesn't give any way to specify said required permissions. So, you are loosening the requirements here either way. Why is it OK to loosen them somewhat but not completely?

I think the only self-consistent options here are:
1. Support exactly what we had before, including keeping a way to specify requiredPermissions on restore.
2. Declare that because no one was actually using these yet (except feature-key-vendor), we don't care. In this case probably we should refuse to restore any token that has `introducerIdentity` at all, on the basis that we can no longer correctly enforce whatever it was trying to express. Or write a migration that deletes these tokens.

I vote for 2.

kentonv:
As discussed in our meeting, let's take the simple approach for now of wrapping the asset IDs in capabilities that *aren't* persistable. The only thing you can do with these capabilities is call a further method to obtain the asset ID. The purpose here is to give us the opportunity to extend the behavior later without breaking backwards compatibility.

For now, we probably should not allow an app to specify its own assets, since I'm not sure currently how that would work. We can add that ability later when it is needed.

src/sandstorm/supervisor.capnp
dwrensha: > Why is it OK to loosen them somewhat but not completely?

I was somehow under the impression that any apps that do use the powerbox today would not yet need full support for `grain.requiredPermissions`. But apparently that's not the case even for the feature key vendor app.

I agree that we should refuse to restore such tokens.

shell/server/core.js
dwrensha: Good point. I think makes sense to also do the query here so we can throw an error early if the grain has been deleted.

dwrensha:
Ready for re-review.

In this version, the `StaticAsset` capability can represent either some data in the `StaticAssets` collection or an identicon that will be served from an "identicon" host, which acts similarly to the "static" host, but does not require any backing data. This idea came up in a separate discussion: https://github.com/sandstorm-io/sandstorm/issues/1866#issuecomment-228872122.

src/sandstorm/grain.capnp
```diff
@@ -160,27 +160,11 @@ interface SandstormApi(AppObjectId) {
...
-      assetId @2 :Text;
-      # The asset ID associated with the grain-size icon for this token
-
-      assetId2xDpi @3 :Text;
-      # If present, the asset ID for the equivalent asset as assetId at twice-resolution
-    }
-    appId @4 :Text;
-    # App ID, needed to generate a favicon if no icon is provided.
-  }
+  getUrl @0 () -> (url :Text);
```
kentonv: This is XSS-prone: If the StaticAsset capability is implemented by another app or other malicious source, it can return a `javascript:` URL or similar, which is likely to be evaluated in the context of the calling app. You can say it is the caller's responsibility to filter, but most won't.

We can prevent this problem by changing the return to something like:

    (https: Bool, hostPath :Text)

The URL is then formed as:

    (https ? "https://" : "http://") + hostPath

shell/server/pre-meteor.js
```diff
@@ -452,9 +460,9 @@ Meteor.startup(() => {
         return;
       }
 
-      if (id === 'static') {
+      if (id === "static" || id === "identicon") {
```
kentonv: There's an argument to be made that we should differentiate these by path rather than by host: When a browser sees a new hostname, it is generally forced to create a new HTTP/HTTPS connection to that host rather than reuse an existing one, hurting performance.

dwrensha:
I've split the return value of `StaticAsset.getUrl()` into a `protocol` and a `hostPath` field, and I've moved the identicon server to our existing "static" host, under the "/identicon/" path.

Ready for re-review.

PR #2028, opened 2016-06-03
install.sh: Make error reporting enabled by default

paulproteus:
Details:

- Stop using curl args from the future, e.g. use data-ascii

- Text change: Clarify that failure is failure.

- Logic change: When asking if the user wants to report failures to us, insist
  on prompting them.

- IMPORTANT CHANGE: Pass ENVVARS when running script from current working
  directory. The lack of this was prevent the CURL_USER_AGENT environment
  variable from being passed into the script when it re-exec'd itself as
  root, if and only if the script decided it was being run from a local
  filename. However, this presumably does affect the test suite.

- While at it, adjust prepare-for-tests.sh to request a few more dependencies.

- Adjust existing installer-tests to not report errors, for simplicity.

- Adjust install.sh reference docs to remark on REPORT=no.

### DIFF
--- a/docs/administering/install-script.md
+++ b/docs/administering/install-script.md
@@ -89,6 +89,10 @@ free HTTPS certificate service.
 - `OVERRIDE_SANDCATS_BASE_DOMAIN`: If you run a different instance of the sandcats.io software,
   adjust this variable.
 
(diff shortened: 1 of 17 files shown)


paulproteus:
Self-merging because 100% installer & docs & installer-tests. Let's see if Jenkins likes it.

PR #2029, opened 2016-06-03
install.sh & installer-tests: Remove unnecessary "$@" and dependency check

paulproteus:


paulproteus:
Self-merging because 100% installer-tests & install.sh.

PR #2034, opened 2016-06-06
Fix bug in permissions computation.

dwrensha:
Fixes a bug found by @zarvox when trying out the new powerbox flow: https://github.com/zarvox/sandstorm-test-python/pull/4#issuecomment-223857013. Adds a testcase that fails before the fix and succeeds afterwards.

 `activateRelevantTokens()` is supposed to return `true` if any new activations may have made more progress possible. Currently it incorrectly returns false in the case where there are no new relevant tokens, but there are some new "grain owner edges".


dwrensha:
(The effect of the bug is that in some situations a user is prevented from opening a grain that they ought to have permission to open.)

PR #2042, opened 2016-06-08
install.sh: Verify that numeric inputs are numeric

### LINKED ISSUE #2036: install.sh: If people answer "y" to a numeric input prompt, the installer does something berserk
Steps to reproduce:

- Try to install Sandstorm on a VPS somewhere

- For the question about which install mode you want, answer `y` not `1` or `2` or just `<enter>`

- Watch in horror as the installer picks dev defaults

Proposed fix:

- Installer should reject input it doesn't understand, e.g.

> You entered: "y". Please enter a number.

paulproteus:
- Rationale: We discovered in user-testing that a person typed "y" rather than
  pressing enter (or typing "1"). This resulted in install.sh choosing mode 2,
  which is development mode, which resulted in a server bound to localhost.

- This commit also adds a test to verify the behavior.

Close #2036

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -199,6 +199,20 @@ prompt() {
   echo "$VALUE"
 }
 
+prompt-numeric() {
+  local NUMERIC_REGEX="^[0-9]+$"
+  while true; do
+    local VALUE=$(prompt "$@")
+
+    if ! [[ "$VALUE" =~ $NUMERIC_REGEX ]] ; then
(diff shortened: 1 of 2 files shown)


paulproteus:
Self-merging because this is 100% install.sh and installer-tests.

I've run this test on my laptop and it passes.

neynah:
Great! :+1:

PR #2056, opened 2016-06-08
Fix large downloads from apps buffering in memory.

### LINKED ISSUE #1054: sandstorm-http-bridge buffering uploads?
A user reports that during a large upload (actually, a git push in his case) sandstorm-http-bridge consumes increasing amounts of memory until finally being killed (presumably by the kernel OOM killer).

I thought that we supported streaming uploads, but this sounds like buffering. Need to investigate.

### LINKED ISSUE #1982: There is some sort of memory leak in the shell
In production lately we've seen a few cases where the node process has consumed gigabytes of RAM. It's unclear whether it is actively using said memory, or if it allocated memory for a transient operation then never released it back to the OS (common behavior for some memory allocators).

Things to investigate:
* Can this be caused by buffering large file downloads?
* Could the new permissions computation be using runaway RAM in some cases? (I think the problem predates the new permissions computation, though.)

kentonv:
This was causing several processes to allocate excessive memory as a result of any large download:
- sandstorm-http-bridge
- the backend coordinator
- the front-end

This change also reduces the total number of write() calls performed which improves performance. Previously, a write() RPC was occurring for every "chunk" in the HTTP chunking sense.

Unfortunately, an app update is required. But, only a few apps do large downloads.

Fixes #1982
Fixes #1054

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -1590,8 +1590,12 @@ class Proxy {
             response.resolveResponseStream(new Capnp.Capability(
                 new ResponseStream(response, code.id, code.title, resolve, reject),
                 ByteStream));
-          }).then(() => { streamHandle.close(); },
-                  (err) => { streamHandle.close(); throw err; });
+          }).then(() => {
+            streamHandle.close();
+          }, (err) => {
+            streamHandle.close();
+            throw err;
+          });
           promise.streamHandle = streamHandle;
           return promise;
         }
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -2247,9 +2247,12 @@ ResponseStream = class ResponseStream {
     this.reject = reject;
 
     this.ended = false;
-    this.writeResolvers = {};
-    this.counter = 0;
-    this.countInFlight = 0;
+    this.waiting = [];
+
+    response.on("drain", () => {
+      this.waiting.forEach(f => f());
+      this.waiting = [];
+    });
 
     response.on("close", () => {
       this.aborted = true;
@@ -2258,12 +2261,8 @@ ResponseStream = class ResponseStream {
       // an incomplete download is the client's problem, not the server's. The important thing is
       // to cancel the stream if it is still being generated; to that end, the next new write()
       // call after this point will throw.
-      for (const i in this.writeResolvers) {
-        console.log(i);
-        this.writeResolvers[i]();
-      }
-
-      this.writeResolvers = {};
+      this.waiting.forEach(f => f());
+      this.waiting = [];
     });
   }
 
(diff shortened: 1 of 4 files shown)


dwrensha:
Also pertains to https://github.com/sandstorm-io/sandstorm/issues/1054.

kentonv:
Good point, added that.

(Test failure with earlier version was because there was a one-line Cap'n Proto change I forgot to push. Fixed now, and updating the commit message to include the second issue number already triggered retesting.)

paulproteus:
Since https://github.com/sandstorm-io/sandstorm/issues/1054 is about uploads, @kentonv I'm curious if this fix probably solves that problem, or only downloads.

shell/server/proxy.js
```diff
@@ -2224,15 +2224,78 @@ const errorCodes = {
...
+
+    response.on("close", () => {
+      this.aborted = true;
+
+      // Resolve all outstanding writes. This is OK even if they didn't actually go through because
+      // an incomplete download is the client's problem, not the server's. The important thing is
+      // to cancel the stream if it is still being generated; to that end, the next new write()
+      // call after this point will throw.
+      for (const i in this.writeResolvers) {
+        console.log(i);
```
zarvox: stray console.log, or did you want to leave this in?

src/sandstorm/sandstorm-http-bridge.c++
```diff
@@ -456,22 +452,114 @@ private:
...
+
+    nextWriteSize = initData.size();
+    if (initData.size() > 0) {
+      memcpy(nextWriteData.get().begin(), initData.begin(), initData.size());
+    }
+
+    if (readStalled) {
+      // Start reading again.
+      readStalled = false;
+      taskSet.add(pumpStreamInternal());
```
dwrensha: Ok, so you've got these two concurrent tasks than can mutually wake each other up: `pumpWrites()` and `pumpStreamInternal()`. The latter wakes up the former using the `writeReady()` fulfiller, and former wakes up the latter with this `readStalled` conditional statement here. I think it be somewhat clearer if, instead, there was a `readReady()` fulfiller, so that `pumpStreamInternal()` would only ever be called by itself or by `pumpStream()`.

dwrensha: It also seems to me that it should be possible to fold the `pumpWrites()` stuff into the `pumpStreamInternal()` task. Perhaps that would simplify things a bit?

shell/server/proxy.js
kentonv: oops, thanks

kentonv:
@paulproteus If you read the thread of #1054, it turned out to be about downloads, not uploads. I apparently misinterpreted when filing the bug initially.

shell/server/proxy.js
```diff
@@ -2224,15 +2224,78 @@ const errorCodes = {
...
+      // It looks like the client is not doing flow control. Close out the RPC immediately in an
+      // (probably futile) attempt to avoid exploding Cap'n Proto call tables.
+      this.response.write(data);
+    } else {
+      // Try to do proper flow control.
+      return new Promise((resolve, reject) => {
+        const index = this.counter++;
+        this.writeResolvers[index] = resolve;
+
+        this.response.write(data, () => {
```
dwrensha: I tried testing this patch with the gitweb app. Buffering of large `git pull`s works as expected, but simply opening the grain through the Sandstorm shell doesn't work. It seems that the problem has to do with the callback argument to `ServerResponse.write()` here, which apparently doesn't get called in all situations that you might hope it would. Note that the callback argument does not show up in the [node 0.10 docs](https://nodejs.org/docs/latest-v0.10.x/api/http.html#http_response_write_chunk_encoding).

I tried the following and it seemed to work better:

```javascript
      // Try to do proper flow control.
      const flushed = this.response.write(data);
      if (flushed) {
        --this.countInFlight;
        console.log("already flushed");
        return;
      } else {
        console.log("need to wait for drain");
        return new Promise((resolve, reject) => {
          const index = this.counter++;
          this.writeResolvers[index] = resolve;
          this.response.on('drain', () => {
            console.log("drained");
            --this.countInFlight;
            delete this.writeResolvers[index];
            resolve();
          });
        });
      }
```

kentonv:
OK, now using "drain". Also added propagation of Content-Length so that large file downloads have known sizes, and reused this code for static file downloads. Yay.

dwrensha:
I'm no longer observing problems in GitWeb!

dwrensha:
Looks good to me.

dwrensha:
(I would prefer to see further simplifications in sandstorm-http-bridge.c++, but I could live with this as-is.)

PR #2057, opened 2016-06-08
accounts-email-token: save/restore URL in email login flow

zarvox:
One problem with the email login workflow is that clicking the link in the
email loses the URL you were at before starting the login flow (which in turn,
might have referenced a sharing token or a grain ID).

This changeset saves the current location to localStorage when requesting email
login, and consults and removes that URI from localStorage when following the
link from the login email.  This helps smooth out that login flow.


kentonv:
Hmm, wouldn't it be better to encode the target URL into the link send in the email? That way it works even if you cross over to a different browser.

(Or, alternatively, the redirect URL could be stored on the server along with the token.)

zarvox:
The former approach has a security problem: it might leak secret URLs/long-lived sharing tokens over email.  While the login token dies after 15 minutes or one use, the redirect URL doesn't.  So I think we shouldn't take that approach.

I tried doing your second suggestion and gave up because Meteor's accounts system doesn't pass the result on to `userCallback`.  In writing up why I gave up and reading the code a few more times, I came up with a new idea that I think will actually work here, so I'm going to give that a shot!

zarvox:
Here's an implementation that saves the current location's path serverside and redirects to that path on successful usage of the link.

kentonv:
Cool. One thing to think about, though: If the URL is a sharing link then storing it in Mongo means we're storing a leak-able token (whereas normally we only store hashes). These are deleted when the token expires, right? If so, not a huge deal for now. But ideally we would actually encrypt the URL using the token (and we don't store the token, only its hash). Maybe add a TODO about this -- or implement it if you're feeling adventurous. :)

paulproteus:
One other thing to think about is that if someone:

- Initiates login flow in browser1

- Clicks login link in email, which (due to system config) launches browser2 to do the login

Now browser2 has the tabs open, and there's no obvious way to get browser1 logged in with the right state. Actually, I guess the user could click "Send login email" again in browser1. But it's a little odd. I'd personally like to see a screenshot tour of this new flow you've created @zarvox and to have you get @neynah's approval of the UX it creates, which I pain to say because I don't want to _slow down_ work on email token usability, but I do want to make sure it does improve usability.

(That or I can install this build somewhere, e.g. rose.sandcats.io , and let someone e.g. @neynah test it to make sure it is what she would want.)

I guess to some extent, my real question here is - sure @kentonv is the code reviewer here - is he also the UX reviewer? Does it need a UX reviewer?

kentonv:
@paulproteus This change merely ensures that if you choose to click the login link, you ultimately get redirected back to the same path from which you initiated login, rather than being redirected to `/`. This is not really a "new flow" or any kind of fundamental redesign -- it's a narrow change that seems obviously to be an improvement.

paulproteus:
In this scenario:

- User has grain1 open in browser tab 1

- User clicks "Sign in" in top-right corner, and gets a login link

- User clicks link

- User gets a second browser tab (call it tab2) opened

tab2 now has all the same grains open as tab1, is that accurate? Or will it just get the first grain open, but not the other grains that are also open in tab1?

It seems to me that tab2 would have the first grain of tab1, and no other tabs, but I might misunderstand how this works.

kentonv:
@paulproteus It is currently impossible to have multiple grain tabs open when not logged in, so there's seemingly no issue here, at least for now.

paulproteus:
Interesting, OK! And does this do something sane when adding a new
email-address-based-identity to an account that's already logged-in, while
viewing account settings?

zarvox:
> And does this do something sane when adding a new email-address-based-identity to an account that's already logged-in, while viewing account settings?

This patchset will send you to the page you initiated the add-an-identity request from, which AFAIK can only happen on `/account`, which was the previous behavior anyway.  So yes.

@kentonv I'll look into sealing the path away from a theoretical attacker with database access by:

* using either a keyed MAC or a KDF to produce a symmetric key from the secret token
* using that symmetric key to encrypt/decrypt the path when saving or consuming the login token

zarvox:
Garply, retest this please

zarvox:
@kentonv This is ready for your re-review.

shell/packages/accounts-email-token/token_server.js
```diff
@@ -81,15 +90,34 @@ Accounts.registerLoginHandler("email", function (options) {
...
       error: new Meteor.Error(403, "Invalid authentication code"),
     };
   }
 
+  // Attempt to decrypt the resumePath, if provided.
+  let resumePath = undefined;
+  if (maybeToken.secureBox) {
+    const box = maybeToken.secureBox;
+    if (box.version === 1) {
+      const key = crypto.pbkdf2Sync(tokenString, box.salt, V1_ROUNDS, V1_KEYSIZE, V1_HASHFUNC);
```
kentonv: Have we verified that this function doesn't start out by taking a sha256 of tokenString? Since if so, then it would of course be possible to get the same results starting from the token hash as stored in the database.

zarvox: Hmm.  It's difficult to tell for sure - pbkdf2 uses an HMAC based on the given hash algorithm, but then HMAC-SHA256 will either pad the input with zeros if shorter than the block size, or...hash it if it's longer than the block size.

Perhaps we're better off switching V1_HASHFUNC to sha512 instead?

kentonv:
After this change, trying to create a new account with email login does not work. Reverting.

kentonv:
To be clear: In my testing, logging in with a pre-existing account worked, but entering a never-before-seen email address led to the error "Match failed". It looks like there is some over-aggressive type checking in the account creation code, but I felt it would be best to revert rather than try to debug in a hurry.

PR #2079, opened 2016-06-11
Finish moving enter-grain docs to debugging/

paulproteus:
I noticed a warning from mkdocs so I thought I'd fix it.


paulproteus:

Issue #2045, opened 2016-06-08
Make a mockup for the "(?)" on the /grains page


paulproteus:
We plan to add a "(?)" button on /grains.

This page tracks the creation of mockups (or other high-clarity information) to answer the following questions.

- On the /grains page, how will the (?) thing look before it is clicked
- On the /grains page, how will the (?) thing look after it is clicked, i.e. where will the text go
- On the /grains page, how will the (?) thing look when it is being hovered-over

- Both desktop & mobile for the above

neynah:
Mock-ups also show proposed text change "You don't have any grains yet..." for when there are 0 grains.

![grains1](https://cloud.githubusercontent.com/assets/855341/15914559/ce31e3de-2d96-11e6-8f75-9130c8a23a0a.png)
![grains2](https://cloud.githubusercontent.com/assets/855341/15914563/d0ff029a-2d96-11e6-911f-b601ffc2b0ef.png)
![grains-hover](https://cloud.githubusercontent.com/assets/855341/15914722/25695b18-2d98-11e6-8b9f-3d110cf7aada.png)

paulproteus:
@neynah you have an interesting question-mark here! How should I create it?

neynah:
I'll open a PR for the icon once I export it. :)

paulproteus:
Question: Should I remove the table header text in the case of no grains, then? I'll assume yes for now.

neynah:
Yes. I think the labeling is distracting especially since there is no relevant data displayed in the columns. Thanks for clarifying, I should have specified. :)

paulproteus:
Is it OK if I remove the gray header? Turns out that's easier than blanking
out the labels.​

neynah:
Yes, I think that's OK as long as the white box is still the same width (to give the impression that there will be some sort of table/list here once you have grains).

neynah:
Question mark icon PR: #2064

paulproteus:
@neynah you have chosen new text for this. Do you want me to copy your new text, or would you prefer we go with the text in the Etherpad, which is:

> Every document, chat room, mail box, notebook, blog, and everything else you create is a grain. Grains are private until they are shared.

I prefer the text in the Etherpad, but I'll use whatever you request.

paulproteus:
Also arguably the text should be:

> Each document, chat room, mail box, notebook, blog, and everything else you create is a grain. Grains are private until they are shared.

[...]

PR #2084, opened 2016-06-14
Add explanation of what a grain is

paulproteus:
@neynah requesting your UX review by manually testing at:

https://rose.sandcats.io/

BTW I can already see one issue:

- If you are a "Visitor" and you navigate to "Grains", then you see this text: "You don't have any grains yet. Install an app to create your first grain." but actually you can't install an app. We can declare that in-scope for this PR, or out-of-scope; I'm OK either way.


paulproteus:
BTW others are welcome to test on rose.sandcats.io, though you'd have to ask me if you want me to make you a User. Also, @zarvox your thoughts on my `exitAndRemoveOverlayNow()` function would be appreciated.

shell/packages/sandstorm-ui-grainlist/grainlist-client.js
```diff
@@ -244,6 +244,62 @@ Template.sandstormGrainListPage.onRendered(function () {
 });
 
 Template.sandstormGrainListPage.events({
+  "click .grain-list .question-mark": function (event) {
+    const templateData = Template.instance().data;
+
+    const exitAndremoveOverlayNow = () => {
```
zarvox: fix your camelcase?

paulproteus: literal LOL. Fixing now. (-:

paulproteus: Thanks for catching that @zarvox.

shell/packages/sandstorm-ui-grainlist/grainlist-client.js
```diff
@@ -244,6 +244,62 @@ Template.sandstormGrainListPage.onRendered(function () {
...
+      const overlay = document.querySelector(".introjs-overlay");
+      if (overlay) {
+        overlay.remove();
+      }
+    };
+
+    if (templateData.intro) {
+      // In this case, the intro is currently active, and the user clicked on the question mark. The
+      // sensible thing to do is to dismiss the intro and stop processing the click.
+      return exitAndremoveOverlayNow();
```
zarvox: The intent here would be slightly clearer if you separate this into a void function call and an empty return.  Otherwise, it looks like the function call is expected to return something.

Also fix the camelcase.

paulproteus: +1, fixed in a commit on top

shell/packages/sandstorm-ui-grainlist/grainlist-client.js
```diff
@@ -244,6 +244,62 @@ Template.sandstormGrainListPage.onRendered(function () {
...
+    }
+
+    const intro = Template.instance().data.intro = introJs();
+    intro.setOptions({
+      steps: [
+        {
+          element: document.querySelector(".grain-list .question-mark"),
+          intro: "A grain is every document, chat room, mail box, notebook, blog, and everything else you create. Grains are private until they are shared.",
```
zarvox: This verbiage seems strained and confusing.  A grain is not every document; at best it's one, not the whole.  Maybe:

Each document, chat room, mail box, notebook, blog, or anything else you create is a grain.  Grains are private until shared.

shell/packages/sandstorm-ui-grainlist/grainlist.html
```diff
@@ -4,7 +4,7 @@
   <div class="grain-list">
     <div class="top-row">
     <div class="non-usage-info">
-    <h1>Grains</h1>
+    <h1>Grains <a href="#" class="question-mark"></a></h1>
```
zarvox: Can this be a button rather than an `<a>`?  I more-or-less believe than any `<a href="#">` is semantically wrong.

shell/packages/sandstorm-ui-grainlist/grainlist-client.js
paulproteus: @zarvox, I prefer your verbiage, too. Nena and I were disagreeing about the
specific verbiage to use, and I delegate a final decision to her, but I'm
happy to express a preference for your verbiage over what I have here.​

shell/packages/sandstorm-ui-grainlist/grainlist.html
paulproteus: I agree with you. Can I do custom styles on buttons, like this icon? If so,
great; I vaguely recall that didn't actually work, but I'll test it now.​

paulproteus: I'll follow the example of e.g. `button.restore-button`!

shell/packages/sandstorm-ui-grainlist/grainlist-client.js
zarvox: Okay.  I will remark that what we have right now is both grammatically incorrect and misleading in terms of what a single grain is (all of the things vs. an instance of one of the things).

I am happy to seek alternatives that might preserve qualities of the current phrasing, like starting with "A grain is" or something like that.  I would be fairly unhappy shipping the exact string as is.

paulproteus:
rose.sandcats.io is up-to-date with the latest pushes to this branch, fwiw.

paulproteus:
@neynah let's make sure we end up with a string in the question-mark that we all like. Maybe we should set up a Loomio grain to vote, or something. (-;

I'll avoid switching what we have in this PR until there's consensus because I don't want to push a bajillion updates to this PR as our best ideas evolve.

paulproteus:
Oh I see on #2045 that @neynah did in fact delegate deciding this to me! So I'll go with the new string. Pushing to this branch now; will update rose later.

paulproteus:
rose updated.

Makefile
```diff
@@ -316,6 +318,14 @@ shell/public/email-494949.svg: icons/email.svg
 	@$(call color,custom color $<)
 	@sed -e 's/#111111/#494949/g' < $< > $@
 
+shell/public/question-a9a9a9.svg: icons/question.svg
```
zarvox: Should we be adding this to the fonticons set?  I seem to remember us wanting to move away from manually recoloring a bunch of SVGs and using CSS `color:` instead.

OTOH, it's easier to get exact pixel sizes right with `background-image`.  So don't consider this a blocker.

shell/packages/sandstorm-ui-grainlist/grainlist.html
```diff
@@ -4,7 +4,7 @@
   <div class="grain-list">
     <div class="top-row">
     <div class="non-usage-info">
-    <h1>Grains</h1>
+    <h1>Grains <button class="question-mark"></button></h1>
```
zarvox: Please add a title or some other accessible label to this button. :)

shell/packages/sandstorm-ui-grainlist/grainlist.html
```diff
@@ -55,10 +55,8 @@
             {{#if showTrash}}
               <p><strong>Trash is empty.</strong></p>
             {{else}}
-              <p><strong>You have no grains.</strong></p>
-              <p>A grain is an instance of a Sandstorm app. A grain might be a document, a chat room,
-                a mail box, a notebook, a blog, or more.</p>
-              <p><a href="{{ pathFor route='apps' }}">Install apps now to start creating.</a></p>
+          <p>You don't have any grains yet.
+            <a href="{{ pathFor route='apps' }}">Install an app to create your first grain.</a></p>
```
zarvox: style nit: drop the `</p>` to the next line to align with the opening `<p>`

zarvox:
Requesting an accessible label on the question mark -- perhaps "more info"?

Otherwise, looks good; thanks!

shell/packages/sandstorm-ui-grainlist/grainlist.html
paulproteus: +1 also fixed other indentation inconsistency

shell/packages/sandstorm-ui-grainlist/grainlist.html
paulproteus: +1, I went with "What is a grain?" which is verbose but I think pretty OK. Happy to hear other ideas.

paulproteus:
Thanks for your thoughtful review @zarvox!

Makefile
kentonv: I think we should be using fonticons for all newly-introduced icons, otherwise we're just creating additional work for ourselves later when we finally switch all the SVG's over.

shell/packages/sandstorm-ui-grainlist/grainlist-client.js
```diff
@@ -244,6 +244,63 @@ Template.sandstormGrainListPage.onRendered(function () {
...
+      exitAndRemoveOverlayNow();
+      return;
+    }
+
+    const intro = Template.instance().data.intro = introJs();
+    intro.setOptions({
+      steps: [
+        {
+          element: document.querySelector(".grain-list .question-mark"),
+          intro: "Each document, chat room, mail box, notebook, blog, or anything else you create is a grain. Grains are private until they are shared.",
```
kentonv: "Grains are private until they are shared" sounds like a tautology to me. How about: "Your grains are private until you share them."

neynah:
The box for the text is a bit too narrow; should probably adjust it so there are approximately 8-11ish words per line. Other than that, the behaviour is right & it looks good to me.

shell/packages/sandstorm-ui-grainlist/grainlist-client.js
paulproteus: I don't love "Your grains are private until you share them" because it's true of all grains, and the phrasing doesn't (IMHO) make that clear. Maybe:

- Grains are private by default.
- All data in Sandstorm is private by default.
- Grains start out 100% private.
- Grains are private until you share them.
- Your creations are private until you share them.

Hmm I don't know.

paulproteus:
@neynah I've updated rose.sandcats.io to have the new, wider box for the text. Let me know if you approve.

paulproteus:
@kentonv requesting your review/merge.

If the "Grains are private until they are shared" phrasing is too tautological-sounding, then I can switch to your proposed phrasing or anything else. I do like it as-is, but I accept that it is not perfect.

shell/packages/sandstorm-ui-grainlist/grainlist-client.js
kentonv: To me the missing information in the original phrasing is how they become "shared". What I want to express is that it is the grain owner's choice to share a grain. Sure, to most users this should be obvious but spelling it out means putting less cognitive load on them.

> I don't love "Your grains are private until you share them" because it's true of all grains,

I don't understand what you're saying here. Only your grains are private until you share them. Other people's grains are private until _they_ share them.

paulproteus: Spelling it out could put less cognitive load on them. Alternatively, spelling it out could put more cognitive load on them, as they mentally construct imagery of a sharing interface.

How about: "All your grains are private until you share them." ? For some reason I prefer that over "Your grains are private until you share them."

paulproteus:
@neynah some questions for you: the way I implemented this, it does also disable the other UI elements, e.g. the ability to click on a grain title and navigate to it. Is that a showstopper? Should I change the way it functions? Have you noticed that in your manual testing?

neynah:
@paulproteus I like it as is (since clicking anywhere outside also dismisses the box). Nice work!

shell/packages/sandstorm-ui-grainlist/grainlist-client.js
kentonv: > How about: "All your grains are private until you share them." ?

SGTM

paulproteus: +1 I will adjust & request merge

paulproteus:
Closing temporarily until I switch to being based on `icon-question`.

paulproteus:
Garply, retest this please

PR #2086, opened 2016-06-14
Dismiss offers and end after IP networking tests

zarvox:
Dismissing the offer prompts is something we should be doing explicitly anyway.

We should end the browser session at the end of the IP networking testsuite to avoid leaking a browser window when running tests.

### DIFF
--- a/tests/apps/ip-networking.js
+++ b/tests/apps/ip-networking.js
@@ -70,5 +70,6 @@ module.exports["Test Ip Interface"] = function (browser) {
       .click("form.test-ip-interface button")
     .frameParent()
     .assertTcpConnection(IP_INTERFACE_TEST_PORT, "tcptest")
+    .end()
 };
 
--- a/tests/apps/powerbox.js
+++ b/tests/apps/powerbox.js
@@ -47,6 +47,7 @@ module.exports["Test Powerbox"] = function (browser) {
     .waitForElementVisible("#powerbox-offer-url", short_wait)
     .getText("#powerbox-offer-url", function (result) {
         browser
+          .click(".popup.offer .frame button.dismiss")
           .frame("grain-frame")
             .click("#request")
           .frameParent()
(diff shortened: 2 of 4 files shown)


zarvox:
Garply, retest this please

zarvox:
Seemingly spurious EADDRINUSE.  Not sure what's up.

zarvox:
Same failure, but I'm still disbelieving that these changes are responsible for the failure.

zarvox:
Garply, retest this please

zarvox:
I watched this last failure from a shell on the build box.  A `netstat -autnp` around the time of the failure revealed nothing helpful - there seemed to be no other process appearing to use the randomly-chosen `SMTP_OUTGOING_PORT` (36989 for that run) at or near the time of the failure.

I'm really not sure what to make of this.

zarvox:
I attached strace during the relevant portion of the test run on the build server, and it's not lying about that EADDRINUSE:

```
[pid 11539] socket(PF_INET, SOCK_STREAM|SOCK_CLOEXEC|SOCK_NONBLOCK, IPPROTO_IP) = 12
[pid 11539] setsockopt(12, SOL_SOCKET, SO_REUSEADDR, [1], 4) = 0
[pid 11539] bind(12, {sa_family=AF_INET, sin_port=htons(38406), sin_addr=inet_addr("0.0.0.0")}, 16) = -1 EADDRINUSE (Address already in use)
```

I still have no idea *what* was supposedly using that address, since nothing relevant showed up in the output of `netstat -lntp` right around then.  I guess it's possible something has bound a socket, but not started listening, so it doesn't appear in the output with -l?  :/

That said, I can conclude that this issue isn't caused by any of these patches -- it's also appeared on tests running against master a couple days ago: https://build.sandstorm.io/job/Sandstorm/1243/ .

zarvox:
Well, it passed this time after dropping the port-probing and using fixed port assignments.

Maybe the answer is to just always run the testsuite with fixed ports, and (if we want to support parallel test runs) to have the Jenkins test runner unshare the network namespace or something. *shrug*

PR #2088, opened 2016-06-15
vagrant-spk install docs: Mention that Windows users do not need git

paulproteus:
Also make this page easier to skim, I hope.


paulproteus:

PR #2090, opened 2016-06-15
fix resumable sessions

dwrensha:
Currently, if a client loses connectivity for more than a few minutes, when it reconnects it will automatically refresh any open grains, causing them to get new sessions with new hostIds. This is because the GrainView's `sessionObserver` calls `GrainView.reset()` when it observes that its session has been removed from the `Sessions` table, and `GrainView.reset()` clears the session salt. The fix is to not call `GrainView.reset()` in this case. We still want to call reset in the case where there permissions have changed, so we add an `invalidated` field to signal that case.


dwrensha:
closing in favor of #2105

PR #2097, opened 2016-06-17
docs/admin/email: Say "All you need" less

paulproteus:
Also:

- Explain why Gmail won't work

- Remove mention of Mandrill, who dropped their free tier


paulproteus:

PR #2098, opened 2016-06-17
Fewer npm wrapper packages

zarvox:
We have a number of Meteor packages that exist to wrap NPM packages.

Meteor 1.3 has much better NPM integration, and ES6 imports let us import things precisely where we use them, rather than having globals running around all over.


meteor-bundle-main.js
```diff
@@ -14,8 +14,8 @@ var https = require('https');
 var net = require('net');
 var url = require('url');
 
-// Borrow the "node-forge" dependency of our meteor-node-forge package.
-var forge = require('./programs/server/npm/node_modules/meteor/meteor-node-forge/node_modules/node-forge');
+// Use the "node-forge" package from npm
```
paulproteus: I'm the one who created this wrapper package, and I'm very happy to see it go away.

PR #2099, opened 2016-06-17
Use new Cap'n-Proto-connection-level flow control in supervisor.

kentonv:
We limit an app to having 8MiB of outgoing calls in-flight at once. This prevents even apps that have not yet received last week's sandstorm-http-bridge update from inundating the front-end with queued writes. (However, apps with old sandstorm-http-bridge will still have the problem that the writes queue up inside sandstorm-http-bridge. Consuming RAM there is still bad, but should do less damage at least on Oasis.)

Note that we now no longer want the front-end to start prematurely closing out write()s when it detects a lack of flow control from the app, because that would prevent the connection-level flow control logic from doing its job.

This depends on a recent Cap'n Proto commit: https://github.com/sandstorm-io/capnproto/commit/538a767e04f11b8135fc086314c0250724cac148

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -2305,14 +2305,6 @@ ResponseStream = class ResponseStream {
         return;
       } else {
         // Write buffer is full. Don't complete until it has more space.
-        if (this.waiting.length >= 16) {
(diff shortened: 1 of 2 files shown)


dwrensha:
Looks good to me.

PR #2106, opened 2016-06-17
fix parameter order of DevPackages changed() observer

dwrensha:
http://docs.meteor.com/api/collections.html#Mongo-Cursor-observe

### DIFF
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -2047,7 +2047,7 @@ if (Meteor.isServer) {
         refPackage(newAction.packageId);
       },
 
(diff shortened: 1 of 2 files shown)


zarvox:
1. Wow, that's an 8-month old bug
2. There appear to be similar bugs in db.js, though their impact is low: [one](https://github.com/sandstorm-io/sandstorm/blob/master/shell/packages/sandstorm-db/db.js#L2050) [two](https://github.com/sandstorm-io/sandstorm/blob/master/shell/packages/sandstorm-db/db.js#L2062)

Can I bother you to add the other two fixes to this branch?

Issue #2102, opened 2016-06-17
Things to fix in end-user onboarding so far


paulproteus:
- [x] The question mark is not in a circle anymore. Wonder what's up with that.

- [x] Need to fix the grain-view question mark on mobile.

- [x] Bring back the top table header on the app details page, and leave the top table header absent on the grains page.

- [x] zarvox reports that the tests fail.

paulproteus:
- Bring back the top table header on the app details page, and leave the top table header absent on the grains page.

paulproteus:
- @zarvox reports that the tests fail, so I should look into that.

zarvox:
specifically: https://github.com/sandstorm-io/sandstorm/pull/2103 failed, I think: https://build.sandstorm.io/job/Sandstorm%20PRs%20-%20build/587/

paulproteus:
The question mark is in a circle now.

I'm going to call this done.

neynah:
Great, thank you @paulproteus

PR #2108, opened 2016-06-18
Disable more of the guided tour, for the test suite

paulproteus:
Refs #2102


paulproteus:
Thanks @zarvox for providing such detailed information about the test breakage I caused, so I could speedily fix it.

PR #2110, opened 2016-06-18
Always show table headers on app details page

paulproteus:
After:

screenshot from 2016-06-17 18 33 37

screenshot from 2016-06-17 18 33 43

This addresses feedback from @kentonv that it's preferable to always show the table headers on the app details page, even if the user has 0 grains with that app, while staying consistent with @neynah 's request that on the overall grains list page, there is no table header.

It also adds code documentation for an existing but undocumented parameter of the grain list template.


paulproteus:
Refs https://github.com/sandstorm-io/sandstorm/issues/2102

paulproteus:
BTW I noticed that the :small_red_triangle_down: down triangle seems to be in the wrong place. Maybe caused by the fontforge thing? Not sure yet; will investigate.

PR #2111, opened 2016-06-18
Revert "Use FontForge to convert SVGs to TTF"

paulproteus:
This reverts commit c9f0899ed0004152442b8084046ea6f957a65361.

Rationale:

- It introduces an issue where the down-arrow is in a weird place,
  among other bad icon issues. See https://github.com/sandstorm-io/sandstorm/pull/2110
  for one example.

- It means the question mark will have no circular outline. I personally
  believe this should not be a show-stopper for the user onboarding changes
  being improved.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -120,13 +120,12 @@ Please install the following:
 * `curl`
 * discount (markdown parser)
 * [Clang compiler](http://clang.llvm.org/) version 3.4 or better
-* FontForge, used by `grunt` to convert SVGs to WebFonts
 * [Meteor](http://meteor.com)
 
 On Debian or Ubuntu, you should be able to get all these with:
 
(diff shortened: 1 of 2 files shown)


paulproteus:
@kentonv requesting your merge/review.

paulproteus:
Kenton - is it OK to ship the question mark without the circle, or should I
go back to the method where I use the SVG as background? I don't think
there's time for me to root-cause it before the release, but if we want to
cause zero user-facing disruption, I can go back to the method where I use
the SVG as an image background (and all the Makefile stuff).

I am warming up to that idea, even though it comes at some code complexity
cost, but that way the user won't see any difference once we figure out how
to make it work the way we want to.

Penny for your thoughts. Thanks for your feedback on these changes so far.

PR #2114, opened 2016-06-20
Switch from icon font to background-image

paulproteus:
Context:

- The circle around the question mark in `icons/question.svg` is not rendered
  properly by the `node` backend in `icons/Gruntfile`.

- Switching to FontForge does allow the circle to show up, but it has the downside
  that other icons are rendered differently.

- We wanted to use an icon font for all icons going forward, but even after
  @neynah and I tried a few variations of her SVG file, the `node` renderer could
  not draw the circle around the question mark.

- Asheesh will be away for about 2.5 weeks starting very soon, so does not have
  time right now to root-cause the issue with the nodejs-backed svg2ttf package
  that our conversion pipeline depends on.

Therefore, this partially reverts:

- commit 561c2f4bce8dcd8b52bcf44faf1d0dd9089bea9b
- commit 3fef0cd03496d46946caf807d5fa40eaaba8af3c

Refs #2102


paulproteus:
@neynah and others - you can test this PR at https://rose.sandcats.io/grain , where I have deployed this change.

@kentonv I know this is 2 steps forward, 1 step back, but I hope you'll consider merging & reviewing this PR.

kentonv:
I'll merge this but we should treat it as temporary until we can figure out exactly why the circle disappears from the font.

PR #2119, opened 2016-06-20
Fix resource leaks in proxy.js and http-bridge

kentonv:
Due to an odd cyclic reference between proxy.js and sandstorm-http-bridge, various objects were not being cleaned up after a streaming download. The recent refactoring of sandstorm-http-bridge to support flow control made things worse: now sandstorm-http-bridge is leaking sockets, too, which eventually causes it to hit the FD limit. The underlying problem, though, was the pre-existing leak.

This change fixes both ends. The proxy.js-side fix solves the problem in the case of completed streams (probably 99.99% of cases). The sandstorm-http-bridge fix solves the problem in error cases.

See individual commits for more info.

@mnutt FYI

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -1467,10 +1467,14 @@ class Proxy {
     response.end();
   }
 
-  makeContext(request, response) {
+  makeContext(request, response, callback) {
     // Parses the cookies from the request, checks that the session ID is present and valid, then
-    // returns the request context which contains the other cookies.  Throws an exception if the
-    // session ID is missing or invalid.
+    // synchronously calls callback() passing it the request context object which contains the
+    // other cookies. Throws an exception if the session ID is missing or invalid. The callback
+    // must immediately use the context in an RPC. Once the callback returns, the context object
+    // will be torn down.
+    //
+    // `makeContext()` returns the result of `callback()`.
 
     const context = {};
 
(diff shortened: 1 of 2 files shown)


shell/server/proxy.js
```diff
@@ -1309,38 +1307,37 @@ class Proxy {
...
       } else {
-        // (_this.identityId might be null; this is fine)
-        vertex = { grain: { _id: _this.grainId, identityId: _this.identityId } };
+        // (this.identityId might be null; this is fine)
+        vertex = { grain: { _id: this.grainId, identityId: this.identityId } };
       }
 
-      let onInvalidated = function () {
+      let onInvalidated = () => {
         Sessions.remove({ _id: _this.sessionId });
```
zarvox: s/_this/this/ on this line

zarvox:
I think the trash test failure is caused by the _this/this bug.  I'm not sure what happened with the roundcube test.

shell/server/proxy.js
kentonv: Confused how I missed that. Fixed!

dwrensha:
I tried this out on Groove Basin. I noticed that I could get sandstorm-http-bridge to consume an unbounded amount of memory simply by repeatedly pressing the play/pause button. Apparently we don't yet handle the case where the stream is closed by the client! I implemented a fix here; https://github.com/dwrensha/sandstorm/commit/4ddcee59a671d66f93f0d040079b16643758b6b6 and after applying it, Groove Basin seems to work well!

dwrensha:
Aside from that issue, and the fact that git does a terrible job at lining up the diff https://github.com/sandstorm-io/sandstorm/pull/2119/commits/cfd226133e4d493ea0051cd34f9661e0053d6c89, which is mostly just indentation changes, this looks good to me.

kentonv:
@dwrensha FWIW `git diff -b` is great for reviewing indentation changes. Too bad Github doesn't have an option (AFAIK)...

kentonv:
@dwrensha FWIW cancellation is actually correctly handled if you update GrooveBasin to the latest sandstorm-http-bridge: It will stop downloading after the first time `write()` throws an exception. But old sandstorm-http-bridge would just keep going. I suppose also dropping the handle makes sense here.

dwrensha:
> cancellation is actually correctly handled if you update GrooveBasin to the latest sandstorm-http-bridge

I was testing it with the sandstorm-http-bridge from this pull request and from v0.167. In both, the `HttpParser` objects leaked.

dwrensha:
Hm... maybe the explanation is that Groove Basin is smart enough to stop writing to the responses that the client has cancelled. So it never actually triggers an exception.

PR #2120, opened 2016-06-21
Add guided-tour hint explaining how to share access

### LINKED ISSUE #2049: Add intro.js guided-tour highlight when user is on their first grain, to teach renaming & sharing
While the grain loads, display an intro.js modal guided-tour thing that says:

> Congrats on your first grain! You can rename it and share it with others.
> (Thanks!)

The word "Thanks!" would be a button the user clicks to dismiss this guided tour modal.

Implementation notes:

- Probably skip on mobile. @neynah are you OK with that?

To implement this without using any server-side storage, some ideas.

- We can store in e.g. the Meteor.Session when the user sees the intro.js box on /apps/{{appId}}.
- If the user saw that during this session, at a time when they had 0 grains, then store a note-to-self in the Meteor.Session saying that the first time we see a grain window, show this guided-tour popup thing.

No mockup needed, I believe!

I will also submit a screenshot as part of a PR that claims to fix this, and probably deploy a test build somewhere so people can play with it.

paulproteus:
Close #2049

Implementation notes:

- We only show this to appdemo users and to users who see
  the hint about installing apps. Therefore, the test suite
  should be unaffected.

- We check the viewport width, and if it is less than 900px
  (aka "$mobile" for Sandstorm) then we don't show this note.

### DIFF
--- a/shell/client/grain-client.js
+++ b/shell/client/grain-client.js
@@ -454,7 +454,7 @@ Template.grainShareButton.onRendered(() => {
   const activeGrain = globalGrains.getActive();
   let unsafeCurrentAppTitle = (activeGrain && activeGrain.appTitle()) || "";
 
-  /* Use DOM to escape HTML, so it is safe to pass to intro.js */
+  // Use DOM to escape HTML, so it is safe to pass to intro.js.
   let div = document.createElement("div");
   div.appendChild(document.createTextNode(unsafeCurrentAppTitle));
   let escapedCurrentAppTitle = div.innerHTML;
--- a/shell/client/grain-client.js
+++ b/shell/client/grain-client.js
@@ -16,6 +16,8 @@
 
 // This file implements /grain, i.e. the main view into an app.
 
+import { introJs } from "intro.js";
+
 import downloadFile from "/imports/client/download-file.js";
 
 // Pseudo-collections.
@@ -436,6 +438,58 @@ Template.shareableLinkTab.events({
     instance.find("form").reset();
     instance.find("form option[data-default-selected=true]").selected = true;
   },
+});
+
+Template.grainShareButton.onRendered(() => {
+  if (! Meteor._localStorage.getItem("userNeedsShareAccessHint")) {
+    return;
+  }
+
+  // Don't show hint on $mobile because the top bar isn't really a top bar in that case.
+  if (window.innerWidth <= 900) {
+    return;
+  }
+
+  let templateData = Template.instance();
+  const activeGrain = globalGrains.getActive();
+  var unsafeCurrentAppTitle = (activeGrain && activeGrain.appTitle()) || "";
+
+  /* Use DOM to escape HTML, so it is safe to pass to intro.js */
+  var div = document.createElement("div");
+  div.appendChild(document.createTextNode(unsafeCurrentAppTitle));
+  var escapedCurrentAppTitle = div.innerHTML;
+
+  const intro = templateData.intro = introJs();
+  let introOptions = {
+    steps: [
+      {
+        element: document.querySelector(".share"),
+        intro: "You've created your first " + escapedCurrentAppTitle + " grain. When you're ready, you can share it with others.",
+      },
+    ],
+    highlightClass: 'introjs-black-helperLayer',
+    tooltipPosition: "bottom",
+    positionPrecedence: ["bottom", "top", "left", "right"],
+    showStepNumbers: false,
+    exitOnOverlayClick: true,
+    overlayOpacity: 0.7,
+    showBullets: false,
+    disableInteraction: false,
+    doneLabel: "Thanks",
+  };
+
+  intro.setOptions(introOptions);
+  const dismissHint = () => {
+    Meteor._localStorage.removeItem("userNeedsShareAccessHint");
+  };
+
+  intro.oncomplete(dismissHint);
+  intro.onexit(dismissHint);
+
+  intro.start();
+
+  // HACK: Resize after 2 seconds, in case the grain size arrived late and caused the UI to reflow.
+  Meteor.setTimeout(() => window.dispatchEvent(new Event("resize")), 2000);
 });
 
 Template.grainPowerboxOfferPopup.events({
--- a/shell/client/grain-client.js
+++ b/shell/client/grain-client.js
@@ -441,7 +441,7 @@ Template.shareableLinkTab.events({
 });
 
 Template.grainShareButton.onRendered(() => {
-  if (! Meteor._localStorage.getItem("userNeedsShareAccessHint")) {
+  if (!Meteor._localStorage.getItem("userNeedsShareAccessHint")) {
     return;
   }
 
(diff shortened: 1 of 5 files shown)


paulproteus:
Screenshot!

screenshot from 2016-06-20 17 06 09

Gonna fix my style issues now, one sec.

kentonv:
FWIW, we probably have enough information to say "Wekan board" rather than "Wekan grain". Do you feel that it's better to use the word "grain" here for educational purposes?

Why "Thanks" instead of "Got it"? I've always felt like "Got it" was a surprisingly good button label for these things -- it somehow communicates to me quickly that this is an educational dialog with no side effects.

shell/client/grain-client.js
```diff
@@ -438,6 +440,58 @@ Template.shareableLinkTab.events({
...
+  // Don't show hint on $mobile because the top bar isn't really a top bar in that case.
+  if (window.innerWidth <= 900) {
+    return;
+  }
+
+  let templateData = Template.instance();
+  const activeGrain = globalGrains.getActive();
+  let unsafeCurrentAppTitle = (activeGrain && activeGrain.appTitle()) || "";
+
+  /* Use DOM to escape HTML, so it is safe to pass to intro.js */
```
kentonv: Nitpick: use // comments

paulproteus:
**re: board vs. grain:** We probably do have enough information to say "Wekan board" (though I haven't checked; let's assume we do for now).

I'm ambivalent; I'd be happy to try saying "Wekan board"; it didn't occur to me as an option earlier!

I think @neynah liked saying "grain" here to tie it to the "grains" UI element on the left, but I feel that saying "Wekan board" is gently better.

**re: Thanks vs. Got it:** @neynah wanted to convey that this is the last of the guided tour things you'll see. I think that, having interacted with it, I prefer "Got it", but then I don't know how to convey what @neynah wanted to convey.

I imagine we should be able to get her feedback by very-soon today, and I'm happy to change any of these to whatever or keep them at whatever.

I think she did say that I'm the authoritative chooser for the strings, so in that case, I'll express my preference:

- "Wekan board" is better than "Wekan grain"
- "Got it" is better than "Thanks"
- If there's a different way to succinctly express that this is the last guided-tour hint, other than saying "Thanks", then I'd be +1 on that I think.

shell/client/grain-client.js
```diff
@@ -438,6 +440,58 @@ Template.shareableLinkTab.events({
   },
 });
 
+Template.grainShareButton.onRendered(() => {
+  if (!Meteor._localStorage.getItem("userNeedsShareAccessHint")) {
+    return;
+  }
+
+  // Don't show hint on $mobile because the top bar isn't really a top bar in that case.
+  if (window.innerWidth <= 900) {
```
paulproteus: But in general, I agree that's a risk.

On the upside, if you `git grep` for `$mobile` you'll find this.

To brainstorm options:

- We could ask some behavioral question of the DOM that'll only ever be true in "desktop" mode, like if some element is visible.
- We could, in one standard place, set a `className` on the `body` element via JS, to label the body as being in `mobile` mode, and then depend on that CSS class having been applied to the body.
- We could, in the SCSS files next to the definition of `$mobile`, mention that anyone modifying the breakpoint should `git grep -F "$mobile"` to make sure that they understand other places that the value was hard-coded.

The last of these is my favorite option. I'm open to a variety of options, though. Curious for your input @kentonv .

shell/client/grain-client.js
paulproteus: +1, will revise

also one day I'd like to see this in `make stylecheck`

neynah:
It's just the last one that I think should have a different phrase than the other guided tour things. "Got it" doesn't feel as satisfying and doesn't indicate that you've completed the guided tour. What about "Finish", or "Done"?

paulproteus:
Since the button is an action the user is agreeing to take, by clicking on it, to me "Finish" should be related to the text they see, so maybe "Finish" would be interpreted as "Finish sharing". Similar for "Done". Maybe "Done with tour" would help with that, or "End the tour".

In general, a difficulty is that changing the button text doesn't necessarily convey that we've changed the button text intentionally.  I'd personally prefer to add a sentence; some options, to brainstorm.

> "You've created your first Wekan board. When you're ready, you can share it with others. This is your last hint!"

> "You've created your first Wekan board. When you're ready, you can share it with others. This concludes your Sandstorm tour!"

> "You've created your first Wekan board. When you're ready, you can share it with others. Thanks for reading all these hints!"

> "You've created your first Wekan board. When you're ready, you can share it with others. Thanks for reading our hints!"

> "You've created your first Wekan board. When you're ready, you can share it with others. Thanks and enjoy being productive!"

neynah:
Or "You've created your first Wekan board. When you're ready, you can share it with others. Enjoy Sandstorm!" [Got it]

I do think "grain" is slightly better and helps introduce the concept. Especially if people won't end up clicking the question mark or viewing /grains before they have any grains. But, I'm actually OK with either as I'll be user testing this anyway. :)

You are the authoritative chooser for the strings!

paulproteus:
> "You've created your first Wekan board. When you're ready, you can share it with others. This is your last hint; enjoy!"

> "You've created your first Wekan board. When you're ready, you can share it with others. This is your last hint; enjoy Sandstorm!"

> "You've created your first Wekan board. When you're ready, you can share it with others. This is your last hint!"

neynah:
I think "Enjoy Sandstorm" is probably enough to indicate that they're done with the tour?

paulproteus:
> You are the authoritative chooser for the strings!

OK!

**re: board vs. grain:** I'll go with "Wekan board" for now, and we'll see what happens.

**re: Got it vs. Thanks:** I'm tempted to avoid the word "Sandstorm" here, since there's already one proper noun (the app name). For now I think I'll try this even-shorter string.

> You've created your first Wekan board. When you're ready, you can share it with others. Enjoy!

I'll go make these changes now.

paulproteus:
I've made the changes that were requested. @kentonv if you can review/merge, that would be splendid.

kentonv:
Hmm, it's not clear to me why we want to convey that this is the "last" message. The word "thanks" definitely didn't communicate that to me, and I don't really think "Enjoy!" would either (hard to tell now that I know this is the goal), but it seems to me that this doesn't really matter.

Also, what if we add another one later (seems likely)? I worry we'd forget to update this one.

paulproteus:
I believe we want to convey this is the last message so that users have a sense that we will stop distracting them with these messages.

I will remove "Enjoy!" for now, and we can see if this concern shows up during user tests.

neynah:
1. To provide a sense of accomplishment which is important in displayed after users complete onboarding/other sequential tasks.

2. Indicate that they're now fully equipped to succeed.

Similar to how our Admin UI has a success message once you're "ready to roll".

paulproteus:
Hmm, now that I've heard from @neynah again, I like saying _something_ here. 

@kentonv I'm not super worried that we'll forget to remove the word "Enjoy!" if we add more guided-tour steps. What if we go with this as-is for now, and see what happens in user tests?

paulproteus:
Quoth Kenton elsewhere re: "Enjoy!": "I don't think it will have the desired effect, but it's not a showstopper". So I'll presume this will be merged whenever @kentonv has time to click the button, and @neynah let's make sure to see if it does have the desired effect when you user-test it.

PR #2123, opened 2016-06-21
Add explanatory text to /admin/users/invite

### LINKED ISSUE #2076: Admin Panel: Invite User does not communicate role & permissions granted
We're missing some important copy to communicate what role users will be invited as and what their default permissions will be. Additionally, since admin's can only invite users(and not Admins) at this time we should communicate that users can be promoted to admin's after they've accepted their invite.

Perhaps something like this but better:

"By default, users invited will assume the User role and will be able to install apps and create grains. To invite an Admin you must invite them as a user first and then promote them to Admin."

image

paulproteus:
Close #2076

I changed the proposed text here so that it:

- Avoids the word "just".

- Avoids all-caps "NOT".

- Avoids the word "magic", preferring a phrase specific to what kind of magic.

### DIFF
--- a/shell/client/admin/user-invite.html
+++ b/shell/client/admin/user-invite.html
@@ -1,4 +1,10 @@
 <template name="newAdminUserInviteLink">
+
+  <p>You can invite others to create accounts on your server by giving them an account creation
+    link. People you invite will be granted User permission and be able to install apps and create
+    grains of their own. Alternatively, if you want to share a specific grain with someone, you can
+    do that within the grain's "Share Access" menu.</p>
+
   <h2>Create a link</h2>
 
   {{#if hasSuccess}}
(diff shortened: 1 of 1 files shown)


paulproteus:
The explanatory text is now:

>   <p>You can invite others to create accounts on your server by giving them an account creation    link. People you invite will be granted User permission and be able to install apps and create     grains of their own. Alternatively, if you want to share a specific grain with someone, you can     do that within the grain's "Share Access" menu.</p>

paulproteus:
@neynah @zarvox requesting your +1/-1 on the text.

paulproteus:
Screenshot, for context:

screenshot from 2016-06-20 17 31 05

zarvox:
:+1: from me

PR #2128, opened 2016-06-21
docs: Give advice about outbound email and inbound email on sandcats

paulproteus:


paulproteus:
Self-merging as this is 100% docs.

Issue #2053, opened 2016-06-08
Adjust most-used-apps calculation to limit repetitiveness


paulproteus:
Context:

- Right now, if you create a new account on a Sandstorm server, and create 1 grain from each of 3 apps, you will see a "Most used" that contains 3 apps, right above your list of all apps, which contains the very same 3 apps.

Plan:

- If the user has installed fewer than 6 apps, don't show the most-used section at all. Rationale: It is unneeded visual duplication of what apps are present in the All Apps area.
- If each app has only 1 grain (max), then don't show the most-used section at all. Rationale: They're all equally-used.

My original text, which was sort of hard to comprehend: 

> Show most-used if the user has installed more than 6 apps, and the most used apps have at least 2 grains, and only show apps with >1 grain created with them as a "Most-used app", and only show "Most used" if there at least 6 apps with a grain created.

I'm OK revising this with other people's input etc; wanted to file an issue to indicate that it's being thought-through.

paulproteus:
I'm going to close this, since IMHO #2129 is enough.

PR #2129, opened 2016-06-21
Only show most-used if >=6 apps installed

paulproteus:
Partially-implements #2053


paulproteus:
IMHO if we have this, we might not need the full implementation of #2053, so I'd like to see this get merged.

paulproteus:
Note that a fuller implementation would be costlier to execute, so I'm not sure it's worth it.

PR #2131, opened 2016-06-22
Test that we are on a terminal first

### LINKED ISSUE #2081: install.sh: Reports bad file descriptor, when instead it should exit with a kernel-version-too-old error message
When I try to install sandstorm on a fresh debian 8 system it just shows:

```
install.sh: line 195: read: 3: invalid file descriptor: Bad file descriptor
install.sh: line 196: VALUE: unbound variable
*** Please answer "yes" or "no".
install.sh: line 192: 3: Bad file descriptor
install.sh: line 193: 3: Bad file descriptor
install.sh: line 194: 3: Bad file descriptor
```

over and over and over...

paulproteus:
Rationale:

- By running this assertion first, we cause the side-effect that FD#3 is
  ready.

- This happens to be important because since I enabled error reporting
  by default, and since the error reporting code assumes that FD#3 is
  ready, a horrifying infinite loop occurs instead.

- Since the installer-test suite runs with REPORT=no, it does not catch
  this error, but it is true that the "precise64" base box can be used
  to reproduce the error.

Close #2081

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -2100,13 +2100,13 @@ sandcats_generate_keys() {
 
 # Now that the steps exist as functions, run them in an order that
 # would result in a working install.
+assert_on_terminal
 assert_linux_x86_64
 assert_usable_kernel
 detect_current_uid
 detect_userns_clone
 assert_userns_clone_works_or_can_be_made_to_work
 handle_args "$@"
-assert_on_terminal
 assert_dependencies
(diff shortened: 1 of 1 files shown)


paulproteus:
Self-merging since I own install.sh. I will run installer-tests on Jenkins right after merging this.

PR #2132, opened 2016-06-22
install.sh: Handle args before asserting on a terminal

paulproteus:
Context: assert_on_terminal needs to know if USE_DEFAULTS is set.


paulproteus:
I will self-merge this once I verify that it makes the general tests pass.

While waiting on that, I'll run installer-tests.

paulproteus:
installer-tests pass locally; self-merging now.

PR #2135, opened 2016-06-23
email config: propagate information about errors users might hit

### LINKED ISSUE #2091: Admin wizard: improve send a test email errors
A number of users attempting to configure SMTP saw "Internal server error [500]" and did not know what they configured incorrectly. We should improve these e-mail test errors to be a little more specific where possible.

image

zarvox:
This provides more useful error messages when an exception is thrown while
attempting to send a test email.  Specific cases tested:

* hostname doesn't resolve
* hostname resolves but server isn't listening on that port
* server is listening but rejects your credentials

Fixes #2091.

### DIFF
--- a/shell/client/admin/email-config-client.js
+++ b/shell/client/admin/email-config-client.js
@@ -153,7 +153,7 @@ Template.newAdminEmailConfig.events({
     const formData = instance.getSmtpConfig();
     Meteor.call("setSmtpConfig", undefined, formData, (err) => {
       if (err) {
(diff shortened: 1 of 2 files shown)


zarvox:
Invalid hostname:

invalid-hostname

Not listening on port:
closed-port

Bad credentials:
bad-credentials

Other classes of failures will dump the whole exception as JSON and drop it in the UI, which is still better for debugging than giving a generic 500.

neynah:
Wow! Great.

PR #2140, opened 2016-06-24
Unclutter account view

### LINKED ISSUE #2065: "Confirm your profile" UI looks congested
1. The current width causes too many line-breaks in the text and makes the text more difficult to scan.

2. There's an overload of information on this page. Perhaps we can make adjustments to a few of the strings to reduce their length and improve/maintain the same level of comprehension.

3. Sections are pretty close together; some padding adjustments would help.

1st draft with the same copy to begin with (I'll update the mock-up to include the bottom items as well):

image

zarvox:
* Same total width as previous layout.
* Profile picture is no longer a separate column.
* The remaining content has more horizontal space.
* The form uses our standard form layout styles.
* Icons were removed, as suggested by @neynah.
* Page subheadings no longer have a distracting white background.
* Copy about handles was made more terse.

Longer-term, I think the account settings page should probably get split out
into multiple pages - we've already got a number of things that deserve a place
in the hierarchy:

* identities
* email addresses
* referral program
* billing

And I expect we'll wind up adding 2FA and notifications and maybe even OAuth
tokens some day.  So, we should plan to organize content and find a good way
to navigate between all these "my account" data.

Revised accounts page:
accounts-revised

Confirm your profile modal:
confirm-profile-modal

Closes #2065

### DIFF
--- a/shell/client/styles/_account.scss
+++ b/shell/client/styles/_account.scss
@@ -468,11 +468,10 @@ body>.main-content>.account {
   }
 
   h3.title-bar {
-    background-color: white;
     width: 660px;
     margin-bottom: 5px;
     border: none;
-    font-size: 12pt;
+    font-size: 14pt;
     padding: 5px;
   }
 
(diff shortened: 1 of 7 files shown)


kentonv:
The "upload new picture" button does not work. Reverting.

PR #2149, opened 2016-06-25
Email token check

zarvox:
A second try at #2057.  Same first commit, second commit fixes the check() issue in the account creation code.


shell/packages/accounts-email-token/token_server.js
```diff
@@ -218,7 +273,8 @@ Meteor.methods({
       throw new Meteor.Error(403, "Invalid authentication code.");
     }
 
-    if (!consumeToken(identity, token)) {
+    const maybeToken = consumeToken(user, options.token.trim());
```
kentonv: `user` is not defined.

kentonv: `options` is also not defined.

PR #2152, opened 2016-06-26
Fix regression in static web publishing.

kentonv:
This turned 404 errors into 500's. It may also have broken redirects to add a trailing '/' when the specified file is a directory.

The root of the problem is that I wrote some really terrible code in pre-meteor.js which made assumptions about ResponseStream that would not be obvious to someone changing ResponseStream, and sure enough a later change to ResponseStream broke it. So, now I did the right thing and added a method to ResponseStream for this specific case so that it's obvious that this is a requirement.


kentonv:
Merging now for release.

PR #2153, opened 2016-06-26
Add facility for admin to take heap dumps for memory leak debugging.

kentonv:


kentonv:
Merging now for release.

zarvox:
Combined with #2154, LGTM

zarvox:
Actually, it looks like `npm-shrinkwrap.json` needs to be regenerated whenever you add a package to `package.json`.  You can regenerate it with:

```bash
cd shell/
meteor npm shrinkwrap
```

zarvox:
I tried doing what I said, and that led me to hit a bug in npm, so I worked around it in #2156.

PR #2154, opened 2016-06-26
Fix JSCS issues and also configure JSCS to allow empty blocks that contain a comment.

kentonv:
I frequently write things like:

    if (condition) {
      // Nothing to do here because blah blah.
    } else {
      doStuff();
    }


kentonv:
Merging now to fix build.

PR #2155, opened 2016-06-26
tests: disable more of the new guided tour stuff

zarvox:
These were causing issues for me locally, because the guided tour backdrop
prevents click events from reaching anything else on the page.


zarvox:
Failed because of a lack of `heapdump` in the `npm-shrinkwrap.json`.  Should be fixed by #2156, so we can retest this once that's merged.

PR #2157, opened 2016-06-28
Route/proxy non-main-host requests before sockjs's handler swallows them

### LINKED ISSUE #1186: Grains fail to load when SockJS falls back to long polling
Meteor apps use SockJS which falls back to long polling when websocket connections fail.

However when running in Sandstorm on a server without websocket support (or being proxied through a webserver without support) grains get stuck in a reload loop.

But Sandstorm Shell does work, successfully falling back to long polling.

This is quite easily reproducible by setting up nginx as a reverse_proxy to Sandstorm using the config here: https://github.com/iflowfor8hours/sandcastle/blob/master/templates/reverse_proxy.conf.j2 but ripping out the websocket configuration.

I've attached tcpdump captures from the shell and wekan without websocket forwarding configured that I was looking at to attempt to debug this.

[shell.txt](https://github.com/sandstorm-io/sandstorm/files/40717/shell.txt)
[wekan.txt](https://github.com/sandstorm-io/sandstorm/files/40718/wekan.txt)

zarvox:
This fixes the longstanding "Meteor app refresh loop" issue.

Meteor's sockjs inserts itself as the first request listener on the Meteor HTTP
handler object.  Normally, this is fine, and keeps DDP requests from
accidentally getting routed to the application handlers.  This is problematic
for programs that happen to implement a proxy, and which need to proxy requests
that use the same paths and overall protocol as Meteor does internally.

As a result of the sockjs middleware being first in the chain, XHRs to an
Meteor app's sockjs endpoint would wind up getting answered by Sandstorm's
instance of sockjs and communicating over DDP to Sandstorm itself.  So any app
that fell back to long-polling would speak DDP to Sandstorm, rather than to the
intended app's backend.

Meteor hashes all the code that goes into an app to determine if a reload is
needed to have the latest code running on the client.  It publishes the current
version of the client code in a collection over DDP.  This lead the code
running in the browser to believe that there had been a code change, since it
was now getting the hash for **Sandstorm's** shell code over DDP, which did not
match the hash for the app's code.  So the app would refresh the page, hoping
to load the new code.  Browsers would cache that websockets had failed for that
host, and so the app would load the code, learn about Sandstorm's code version
over sockjs long-polling DDP, and attempt to refresh again.

The solution here is to do Sandstorm's proxying earlier in the request dispatch
chain.  We splice our own handler in after ddp-server has installed its sockjs
handler, and now everything works as intended.

Fixes #1186.

### DIFF
--- a/shell/server/pre-meteor.js
+++ b/shell/server/pre-meteor.js
@@ -357,12 +357,6 @@ const canonicalizeShellOrWildcardUrl = (hostname, url) => {
 };
 
 const dispatchToMeteorOrStaticPublishing = (req, res, next, redirectRatherThanServeShell) => {
-  if (!req.headers.host) {
-    res.writeHead(400, { 'Content-Type': 'text/plain' });
-    res.end('Missing Host header');
-    return;
-  }
-
   const hostname = req.headers.host.split(':')[0];
   if (isSandstormShell(hostname)) {
     // Go on to Meteor, or serve a redirect.
@@ -450,6 +444,12 @@ const dispatchToMeteorOrStaticPublishing = (req, res, next, redirectRatherThanSe
 };
 
 const redirectToMeteorOrServeStaticPublishing = (req, res, next) => {
+  if (!req.headers.host) {
+    res.writeHead(400, { 'Content-Type': 'text/plain' });
+    res.end('Missing Host header');
+    return;
+  }
+
   return dispatchToMeteorOrStaticPublishing(req, res, next, true);
 };
 
@@ -493,7 +493,9 @@ Meteor.startup(() => {
   WebApp.httpServer.removeAllListeners("request");
   WebApp.httpServer.on("request", (req, res) => {
     if (!req.headers.host) {
-      throw new Meteor.Error(400, 'Missing Host header');
+      res.writeHead(400, { 'Content-Type': 'text/plain' });
+      res.end('Missing Host header');
+      return;
     }
 
     const hostname = req.headers.host.split(":")[0];
@@ -540,7 +542,7 @@ Meteor.startup(() => {
     });
   });
 
-  listenOnAlternatePorts()
+  listenOnAlternatePorts();
 });
 
 const errorTxtMapping = {};
--- a/shell/server/pre-meteor.js
+++ b/shell/server/pre-meteor.js
@@ -433,7 +433,7 @@ const handleNonMeteorRequest = (req, res, next, redirectIfInWildcard) => {
 // This function serves responses on Sandstorm's main HTTP/HTTPS
 // port.
 const handleNonMeteorRequestDirectly = (req, res, next) => {
-  return handleNonMeteorRequest(req, res, next, false);
+  handleNonMeteorRequest(req, res, next, false);
 };
 
 // "Alternate ports" are ports other than the main HTTP or HTTPS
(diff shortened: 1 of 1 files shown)


shell/server/pre-meteor.js
```diff
@@ -341,180 +341,206 @@ function serveStaticAsset(req, res) {
...
+
+  nonMeteorRequestHandler.use(serveMeteorOrStaticPublishing);
+
+  WebApp.httpServer.removeAllListeners("request");
+  WebApp.httpServer.on("request", (req, res) => {
     if (!req.headers.host) {
-      res.writeHead(400, { 'Content-Type': 'text/plain' });
-      res.end('Missing Host header');
-      return;
+      throw new Meteor.Error(400, 'Missing Host header');
```
dwrensha: There is no exception handler to catch this error, so the frontend dies when it is triggered! I think you should instead do something like:

```javascript
      res.writeHead(400, { 'Content-Type': 'text/plain' });
      res.end('Missing Host header');
      return;
```

Note that you can test what happens with an empty Host header by doing the following, assuming you are not in `run-dev.sh` mode and your server is not behind a reverse proxy:
```
curl -v -H 'Host:' http://sandstorm.example.com/
```

shell/server/pre-meteor.js
```diff
@@ -341,180 +341,206 @@ function serveStaticAsset(req, res) {
...
+  targetUrl.host = hostname + ':' + targetUrl.port;
+
+  // The following allows to avoid decoding + re-encoding query
+  // string parameters, if provided.
+  targetUrl = Url.resolve(Url.format(targetUrl), url);
+  return targetUrl;
+};
+
+const dispatchToMeteorOrStaticPublishing = (req, res, next, redirectRatherThanServeShell) => {
+  if (!req.headers.host) {
```
dwrensha: The new `req.headers.host` check in the "request" handler makes this check redundant. Is there a good reason to keep this around?

zarvox: It is still needed for the `redirectToMeteorOrServeStaticPublishing` codepath (though perhaps that suggests that the check should be moved out to there).

shell/server/pre-meteor.js
zarvox: Yikes.

Yeah, I should double-check what code is responsible for catching exceptions/failed promises further down too.

zarvox: It looks like the rest of the Sandstorm-proxy request handlers properly catch anything they throw, and Meteor of course catches anything that is dispatched through its handlers.

The upgrade handler itself uses a Promise and simply destroys the socket if any exception is raised.

zarvox:
Garply, retest this please

zarvox:
Previous failure was caused by the fix from sandstorm-io/capnproto#344 causing the randomized schema-evolution test to fail.  It's good that the capnproto tests catch that particular evolution issue, but we can safely ignore it here.

shell/server/pre-meteor.js
```diff
@@ -341,180 +341,208 @@ function serveStaticAsset(req, res) {
...
 
-  const dispatchToMeteorOrStaticPublishing = (req, res, next, redirectRatherThanServeShell) => {
+  // Construct the middleware chain for requests to non-DDP, non-shell hosts.
+  const nonMeteorRequestHandler = Connect();
+  // BlackrockPayments is only defined in the Blackrock build of Sandstorm.
+  if (global.BlackrockPayments) { // Have to check with global, because it could be undefined.
+    nonMeteorRequestHandler.use(BlackrockPayments.makeConnectHandler(globalDb));
+  }
+
+  nonMeteorRequestHandler.use(serveMeteorOrStaticPublishing);
```
dwrensha: The names `nonMeteorRequestHandler` and `serveMeteorOrStaticPublishing` are confusing when taken together like this. Are we considering requests to grains to be `Meteor` or `nonMeteor`?

shell/server/pre-meteor.js
```diff
@@ -341,180 +341,208 @@ function serveStaticAsset(req, res) {
...
+
+  WebApp.httpServer.removeAllListeners("request");
+  WebApp.httpServer.on("request", (req, res) => {
```
dwrensha: I'm still a bit nervous about the fact that this stretch of code does not have an exception handler. Sure, maybe we can verify by inspection now that nothing bad is likely to happen, but who knows how the code will evolve? Especially because this code gets hit on *every* request, it seems like going out of our way for robustness would be worthwhile.

shell/server/pre-meteor.js
zarvox: Good point, the names of these functions and their parameters became quite confusing.

I've reworked the dispatch order and the naming, so hopefully it's clearer now.

shell/server/pre-meteor.js
zarvox: I'm pretty sure all the other things that can fail already catch their exceptions, but I suppose the real point of exceptions is to guard against programmer error, so I can't really argue with that. :)

Added top-level try/catch blocks here and for `handleNonMainPortRequest`.

shell/server/pre-meteor.js
```diff
@@ -422,45 +415,29 @@ const dispatchToMeteorOrStaticPublishing = (req, res, next, redirectRatherThanSe
...
 // port.
-const serveMeteorOrStaticPublishing = (req, res, next) => {
-  return dispatchToMeteorOrStaticPublishing(req, res, next, false);
+const handleNonMeteorRequestDirectly = (req, res, next) => {
+  return handleNonMeteorRequest(req, res, next, false);
 };
 
 // "Alternate ports" are ports other than the main HTTP or HTTPS
-// port. For requests to the shell & grains, we redirect to the main
+// port. For requests to the shell, grains, we redirect to the main
```
dwrensha: Weird punctuation edit here.

dwrensha:
Looks good to me.

shell/server/pre-meteor.js
zarvox: Rewrote that whole comment to make clearer/more accurate.

zarvox:
I think the test failure is the IpInterface test being racy.  It passed in the previous commit, and the only difference is comments.

zarvox:
Garply, retest this please

PR #2159, opened 2016-06-29
somewhat cleaner stream management for getWwwFileHack()

dwrensha:
This is an iteration on the bugfix in https://github.com/sandstorm-io/sandstorm/pull/2152. The idea is to provide the `stream` argument of `getWwwFileHack()` as a promise that's only fulfilled once we know it's actually needed.

### DIFF
--- a/shell/server/pre-meteor.js
+++ b/shell/server/pre-meteor.js
@@ -93,22 +93,23 @@ function wwwHandlerForGrain(grainId) {
       response.setHeader("Access-Control-Allow-Origin", "*");
     }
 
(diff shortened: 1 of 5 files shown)


kentonv:
Note that a supervisor can carry over from a previous version on upgrade, and this PR changes the contract of `getWwwFileHack()`. In theory we should be considering backwards-compatibility with these changes.

Does this change have a benefit other than making the code clearer?

zenhack:
I think it probably makes sense to close this -- it's a cleanup patch for code that doesn't exist anymore, so I don't think it's applicable anymore.

PR #2162, opened 2016-06-30
Refactor accounts-identity UI

zarvox:
This refactor makes explicit all of the states that the UI can be in.  We also
handle any Meteor method call failures, displaying the error message in the UI.
This will allow us to add UI to explain why login with identity fails when a user limit is reached, and provides a reasonable place for us to prompt for two-factor auth in the more distant future.

In the non-login-identity view, we:

* split the identity list out by account with a faint horizontal rule
* retain only one "unlink" button per account, since unlinking is an action
  taken on the account, rather than on the identity pictured
* avoid using window.confirm() in favor of an in-app modal dialog
* visibly handle success/failure of any of the Meteor method calls with
  the focusing success/error boxes

Additional stylistic changes include avoiding using the `event` global,
using `evt.currentTarget` rather than `evt.target`, and preferring ES6 fat-arrow
functions in most places.

non-login-identity

no-window-confirm

### DIFF
--- a/shell/packages/accounts-identity/accounts-identity-client.js
+++ b/shell/packages/accounts-identity/accounts-identity-client.js
@@ -33,21 +33,24 @@ Template.identityLoginInterstitial.onCreated(function () {
   // An exhaustive list of the top-level states this interstitial may be in.
   // Each state may include an object with additional context-specific information.
   //
-  //   * loading                 Waiting on subscriptions to finish loading
-  //   * creatingAccount         We have sent a createAccountForIdentity call and are awaiting the
-  //                             reply
-  //   * accountCreationFailed   The createAccountForIdentity call failed.  Value is the Error.
-  //   * loggingInWithIdentity   We have sent a loginWithIdentity call and are awaiting the reply
-  //   * loginWithIdentityFailed The loginWithIdentity call failed.  Value is the Error.
-  //   * loggingInWithToken      We have started logging in with a token and are awaiting the reply
-  //   * loginWithTokenFailed    The loginWithToken call failed.  Value is the Error.
-  //   * linkingIdentity         We have sent a linkIdentityToAccount call and are awaiting the
-  //                             reply
-  //   * noLoginIdentities       Subscriptions loaded, but we can neither login, create, nor link an
-  //                             account.  We require input.
+  //   * loading                    Waiting on subscriptions to finish loading
+  //   * creatingAccount            We have sent a createAccountForIdentity call and are awaiting
+  //                                the reply
+  //   * accountCreationFailed      The createAccountForIdentity call failed.  Value is the Error.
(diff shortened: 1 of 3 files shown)


zarvox:
Looks like JSCS can't handle object spreads, which is actually somewhat reasonable because I think they're officially an ES7 thing.

kentonv:
This seems like a good one for @dwrensha to review.

dwrensha:
> retain only one "unlink" button per account, since unlinking is an action taken on the account, rather than on the identity pictured

nice!

kentonv:
(@dwrensha let me know when you think this is ready for merge.)

shell/packages/accounts-identity/accounts-identity-client.js
```diff
@@ -30,14 +30,61 @@ Accounts.isLinkingNewIdentity = function () {
...
+  //
+  //   * loading                 Waiting on subscriptions to finish loading
+  //   * creatingAccount         We have sent a createAccountForIdentity call and are awaiting the reply
+  //   * accountCreationFailed   The createAccountForIdentity call failed.  Value is the Error.
+  //   * loggingInWithIdentity   We have sent a loginWithIdentity call and are awaiting the reply
+  //   * loginWithIdentityFailed The loginWithIdentity call failed.  Value is the Error.
+  //   * loggingInWithToken      We have started logging in with a token and are awaiting the reply
+  //   * loginWithTokenFailed    The loginWithToken call failed.  Value is the Error.
+  //   * linkingIdentity         We have sent a linkIdentityToAccount call and are awaiting the reply
+  //   * noLoginIdentities       Subscriptions loaded, but we can neither login, create, nor link an account.  We require input.
```
dwrensha: Would `identityCannotLogin` be a more appropriate name for this state?

shell/packages/accounts-identity/accounts-identity-client.js
```diff
@@ -30,14 +30,61 @@ Accounts.isLinkingNewIdentity = function () {
...
+  //
+  // The noLoginIdentities state also tracks substates directly:
```
dwrensha: s/noLoginIdentities/unlinkIdentityState/`?

shell/packages/accounts-identity/accounts-identity-client.js
zarvox: I might find that a bit ambiguous with `loginWithIdentityFailed`.

Maybe `identityIsNotLoginIdentity`?

shell/packages/accounts-identity/accounts-identity-client.js
zarvox: Clarified in c3a8014

zarvox:
Updated comments and such.

dwrensha:
Looks good to me.

PR #2167, opened 2016-07-01
admin user accounts list: allow sorting

zarvox:
The ability to sort by last active is an important affordance for allowing
admins to audit which accounts are active.

Blaze's "eventually rendered" semantics make the page look downright goofy
immediately after you click on the "Created" or "Last Active" header -- those
two columns repaint with the new data immediately, whereas the "Account" column
doesn't update until a future frame.  This seems to be unavoidable given
Blaze's design - unlike React, there doesn't appear to be a `key={_id}` prop
you can pass in explicitly to help the UI layer deal better with reorderings.
At least, I'm not finding a solution for this in the docs.

sort-by-last-active

### DIFF
--- a/shell/client/admin/user-accounts-client.js
+++ b/shell/client/admin/user-accounts-client.js
@@ -192,6 +192,7 @@ Template.newAdminUsers.onCreated(function () {
         return identity;
       });
       return {
+        _id: account._id,
         account,
         identities,
       };
--- a/shell/client/admin/user-accounts-client.js
+++ b/shell/client/admin/user-accounts-client.js
@@ -68,6 +68,84 @@ Template.newAdminUserTableRow.events({
   },
 });
 
+Template.newAdminUserTable.onCreated(function () {
+  this.sortOrder = new ReactiveVar({
+    key: "createdAt",
+    order: "ascending",
+  });
+});
+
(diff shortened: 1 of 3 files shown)


kentonv:
Maybe what I was remembering about Blaze is something like: If you do `#each` on a Cursor, it knows to track the elements by _id?

Is it possible to arrange the `#each` to be on a cursor in this case, and delay further computation on the user record to happen inside the loop?

zarvox:
Maybe.  Not with the exact sort and filter logic I've got here now due to the sort involving joins in a fallback case (which minimongo won't do), but you could probably rewrite things to just look at the account's values for the sort order rather than additionally filling from associated identities, move the `sortOrder` ReactiveVar out to the top-level scope, and then do the account-to-identity lookup in the per-account branch.  You'd also need to move the search and checkbox filtering down there, at which point the logic for the page is now split in half between the js and the templates, which hurts readability.

Even if it works (and it's not clear to me from any public documentation that such a change can be expected to fix the observed issue) I'm not sure it's worth the cost in readability.  I'll try it out if you expect it to be worth my time, though.

zarvox:
Figured out that if I put an `_id` field in the thing that I'm `{{#each}}`-ing over, Blaze will Do The Right Thing.

PR #2169, opened 2016-07-02
Move claimRequest() to SessionContext.

dwrensha:
Shortly after #2027 was merged, @kentonv noticed that it probably makes more sense for `claimRequest()` to go on `SessionContext` rather than `SandstormApi`.

This makes it so that you can only claim a request token that was created with your current session.


dwrensha:
I've also updated my pull requests to the test apps:
https://github.com/zarvox/sandstorm-test-python/pull/4
https://github.com/jparyani/sandstorm-test-app/pull/2

kentonv:
Awesome, thanks!

PR #2176, opened 2016-07-05
Fix bug where app update notifications were being merged with an arbi…

kentonv:
…trary notification belonging to the user, with strange results.


jparyani:
LGTM

Issue #1885, opened 2016-04-21
Title text seems to be ignored in dev mode


mitar:
The [action text](https://github.com/mitar/council-app/blob/master/sandstorm-pkgdef.capnp#L30) is not displayed in the dev mode for my Meteor app.

![screen shot 2016-04-20 at 8 42 33 pm](https://cloud.githubusercontent.com/assets/585279/14697428/a6893a28-0738-11e6-93b6-c6ac66c1d984.png)

Shouldn't it say "New Council App" there?

paulproteus:
Hi Mitar! This is almost definitely a bug. Code-wise, we treat
apps-in-development different from apps that are fully-installed, and so
this is probably related to that different ​treatment. Thank you for
reporting it!

zarvox:
Argh, this one's probably my fault.  The boilerplate generated by `spk init` is wrong.  You should provide a `nounPhrase` rather than a `title`: https://github.com/sandstorm-io/sandstorm/blob/master/src/sandstorm/package.capnp#L146..L152

zarvox:
Specifically, try changing "title" to "nounPhrase" at https://github.com/mitar/council-app/blob/master/sandstorm-pkgdef.capnp#L30

mitar:
Should it be lower case?

mitar:
OK, this fixed it. Should this ticket be closed?

kentonv:
Issue should stay open -- it's still a bug that `title` is being ignored in dev mode.

mitar:
`title` has been obsoleted, it seems.

Maybe the ticket should stay open for template to be updated?

paulproteus:
@zarvox does it make sense for you to look into how to best fix this, perhaps merely by changing the `spk init` template to account for your `nounPhrase` changes? If so, and you're wiling to do so in the next 0-2 weeks, I'd love for you to assign it to yourself.

If not, then I can see what else we can do.

dwrensha:
> it's still a bug that title is being ignored in dev mode.

[...]

PR #2177, opened 2016-07-05
update spk init template to use Action.nounPhrase rather than Action.title

dwrensha:
The `title` field [is deprecated](https://github.com/sandstorm-io/sandstorm/blob/v0.170/src/sandstorm/package.capnp#L147-L153).

See also #1885.


zarvox:
IIRC this was blocked on updating the allow/deny rules for `UserActions` to not always require a `title`.  I'm not sure if that ever got fixed; check that you can package and install such an .spk

zarvox:
Oh, I see you already fixed this in #2178.  Great!

PR #2178, opened 2016-07-05
allow insertion of UserActions that lack a title

dwrensha:
The `Action.title` field is deprecated, so we shouldn't require that it be present.

See also #2177.

### DIFF
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -1257,42 +1257,6 @@ _.extend(SandstormDb.prototype, {
     return value;
   },
 
-  addUserActions: function (packageId) {
-    //TODO(cleanup): implement this with meteor methods rather than client-side inserts/removes.
-    const pack = Packages.findOne(packageId);
-    if (pack) {
-      // Remove old versions.
-      UserActions.find({ userId: Meteor.userId(), appId: pack.appId })
-          .forEach(function (action) {
-        UserActions.remove(action._id);
-      });
-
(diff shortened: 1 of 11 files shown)


zarvox:
Do we want to require that at least one of (title, nounPhrase) is present?  Can that be expressed?

dwrensha:
I think having neither is fine because we fall back to "grain".

zarvox:
:shipit:

kentonv:
Maybe we should go ahead and introduce a method that takes a package ID as an argument? It feels kind of scary to be round-tripping this structure through the client, even though I don't think it's exploitable (currently).

kentonv:
Also I believe this is the only `allow()` rule we have in the whole codebase. Would be nice to have none.

dwrensha:
Garply, retest this please.

dwrensha:
The [transient failure on Jenkins](https://build.sandstorm.io/job/Sandstorm%20PRs%20-%20build/668/) seems to indicate some kind of race condition. I was unable to reproduce it locally. I'm not sure whether or not it is related to my changes in this pull request.

shell/packages/sandstorm-db/db.js
```diff
@@ -1257,42 +1257,6 @@ _.extend(SandstormDb.prototype, {
     return value;
   },
 
-  addUserActions: function (packageId) {
```
kentonv: This method is still in-use for auto-updates, which now seem to be broken.

PR #2181, opened 2016-07-07
Hide 'shared with me' grains that already show up as 'mine'.

### LINKED ISSUE #1284: a grain can show up twice in the grains list
If I own a grain and also share it with an identity that's linked to my account, the grain shows up twice in my grains list, which is confusing.

dwrensha:
Fixes #1284

### DIFF
--- a/shell/packages/sandstorm-ui-grainlist/grainlist-client.js
+++ b/shell/packages/sandstorm-ui-grainlist/grainlist-client.js
@@ -90,8 +90,13 @@ const filteredSortedGrains = function (showTrash) {
   const ref = Template.instance().data;
   const db = ref._db;
   const grains = db.currentUserGrains(showTrash).fetch();
+  const grainIdSet = {};
+  grains.map((g) => grainIdSet[g._id] = true);
+
   const itemsFromGrains = SandstormGrainListPage.mapGrainsToTemplateObject(grains, db);
-  const apiTokens = db.currentUserApiTokens(showTrash).fetch();
+  const apiTokens = db.currentUserApiTokens(showTrash).fetch()
+        .filter((t) => !(t.grainId in grainIdSet));
+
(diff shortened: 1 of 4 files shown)


shell/packages/sandstorm-ui-grainlist/grainlist-client.js
```diff
@@ -123,12 +128,14 @@ SandstormGrainListPage.bulkActionButtons = function (showTrash) {
         },
 
         onClicked: function (ownedGrainIds, sharedGrainIds) {
-          ownedGrainIds.forEach((grainId) => {
+          const grainIds = ownedGrainIds.concat(sharedGrainIds);
+
+          grainIds.forEach((grainId) => {
```
zarvox: Is the idea here that we'll call both `deleteGrain` and `forgetGrain` on each grain to delete, regardless of whether we own the grain or have ApiTokens for it?  I find it a bit weird that we're sending requests over the wire that we know a priori will fail here.

kentonv: I agree with this.

Maybe the `deleteGrain` method should be extended to automatically delete any owner-owned API tokens? Then you call `deleteGrain` if the grain is owned by the user and `forgetGrain` otherwise, but don't have to worry about the "both" case.

Or maybe these two methods should be merged into one with the server figuring out which one applies.

dwrensha:
> Maybe the deleteGrain method should be extended to automatically delete any owner-owned API tokens?

Done.

We could conceivably replace all instances of `forgetGrain(grainId, identityId)` with `deleteGrain(grainId)`, but I think it might prove useful to keep around the finer-grained API.

dwrensha:
Added another commit that tightens the semantics of `deleteGrain()` so that it could not actually be used as a substitute for `forgetGrain()`. Also cleaned up some cruft from `SandstormDb.deleteGrains()`. Rebased to avoid conflict with #2178.

PR #2183, opened 2016-07-07
include the package ID in the session ID hash

### LINKED ISSUE #2175: Sessions do not refresh when apps update
When updating the app associated with a grain, it seems that already-open sessions do not refresh. We need them to refresh to get the new client code.

dwrensha:
Fixes #2175.

@kentonv was wondering what would happen to live sessions when this update hits a Sandstorm server. Fortunately, in the common case such sessions will remain active. If there still exists an entry in the `Sessions` collection, then the server reconstructs the proxy [here](https://github.com/sandstorm-io/sandstorm/blob/v0.170/shell/server/proxy.js#L621-L622), avoiding the need to call `openSession()` and therefore avoiding generation of a new session ID. A new session ID will only be generated in the case where a timed-out session is resumed.

Another possible concern about this change is that it leaks some information about package IDs that I don't think we were previously leaking. If Alice shares a grain to Bob, then currently, as far as I am aware, Bob has no way to know the app ID or package ID of the grain. With this change, Bob will effectively be able to ask "does the grain have package ID X?". My feeling is that this leak is not a huge deal, and if we want to plug it, we could provide a way for apps to specify a dummy value other than `packageId` to put in the session ID hash.

### DIFF
--- a/shell/packages/sandstorm-backend/sandstorm-backend.js
+++ b/shell/packages/sandstorm-backend/sandstorm-backend.js
@@ -153,8 +153,10 @@ SandstormBackend.prototype.continueGrain = function (grainId) {
                            "Package ID: " + packageId);
   }
 
-  return this.startGrainInternal(
+  const result = this.startGrainInternal(
       packageId, grainId, grain.userId, manifest.continueCommand, false, isDev);
+  result.packageId = packageId;
+  return result;
 };
 
 SandstormBackend.prototype.startGrainInternal = function (packageId, grainId, ownerId, command, isNew, isDev) {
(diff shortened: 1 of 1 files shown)


zarvox:
Another option to block the `packageId` confirmation attack would be for `cachedSalt` to encode an encrypted blob which is opaque to the client, to prevent the client from being able to reproduce the hash computation directly.

kentonv:
Having the app be able to declare a fake `packageId` would probably lead to problems where the app doing the faking now has the same upgrading-doesn't-reset-sessions bug since the faked `packageId` doesn't change (unless they carefully make sure to use a different packageId from the real app each time, or something).

Hashing in a server secret would prevent the client from asking these kinds of questions, but then requires maintaining said server secret -- it needs to be stored somewhere, needs to have a rotation story, etc.

I don't particularly like the idea that users can probe the package ID but I'm not seeing a good solution here...

dwrensha:
Closing because [this suggestion](https://github.com/sandstorm-io/sandstorm/issues/2175#issuecomment-231426266) seems like a better idea.

PR #2185, opened 2016-07-08
Avoid using decrepit versions of Connect.

### LINKED ISSUE #2172: res.headersSent deprecation warning?
Saw this in the logs, not sure if we should be worried. Someone should investigate.

> connect deprecated res.headerSent: use standard res.headersSent at npm/node_modules/meteor/connect/node_modules/connect/lib/proto.js:125:22

kentonv:
Over two years ago, when I was the only person working on Sandstorm, in commit 3ab8575, I introduced a Meteor package wrapping Connect so that we could build some Connect handlers. I wanted to use exactly the version Meteor was using, but alas Meteor's package system offered me no way to do that. So, I wrote in package.js:

    // Please keep this updated to match the version used by Meteor. Ugh.
    Npm.depends({'connect': '2.9.0'});

We never, ever updated this. Not once.

Meteor is now on Connect 2.30.2.

This mostly never came up because over time our use of Connect had been reduced only to using its file-extension-to-MIME-type table.

That is, until last week, when #2157 went ahead and used the Connect imported from this package to try to reproduce the handler chain that Meteor had previously been setting up.

When this very old Connect interacted with the much-newer Connect used by Meteor, the result was... well, just one deprecation warning. Surprisingly, nothing really broke?

In any case, it appears that the WebApp package now exports its own copy of Connect as `WebAppInternals.NpmModules.connect.module`. So we can now use that instead, and not import our own. This change does that.

Separately, it turns out the SAML package was also importing some arbitrary ancient version of Connect. I updated that too. But it turns out the specific module that package was using is deprecated! Fortunately, there is a replacement that does what we need. (The old "bodyParser" module was deprecated because it handled mime/multipart encoding in addition to www-form-urlencoded, and apparently people decided that multipart decoding is hard enough that there are now several competing libraries to do it, with the one in Connect being deprecated in favor of the others. But SAML actually only uses urlencoded, which is still supported.)

Thus, this change removes two whole copies of Connect (and all its dependencies) from the code we ship, reducing the size of our release bundle by 10.8% (~2.6MiB).

Fixes #2172


zarvox:
:shipit:

PR #2187, opened 2016-07-08
Add desktop notifications, triggerable from the server

zarvox:
Tabs coordinate over localStorage to make desktop notifications spawn
from the most useful tab to handle them - one of the ones that already
has that grain open, or if none do, whichever one gets there first.

### DIFF
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -648,7 +648,7 @@ const DesktopNotifications = new Mongo.Collection("desktopNotifications");
 //
 //   _id: String.  Used as the tag to coordinate notification merging between browser tabs.
 //   creationDate: Date object. indicating when this notification was posted.
-//   accountId: String. Account id to which this notification was published.
+//   userId: String. Account id to which this notification was published.
 //   title: String. Primary label of the notification.
 //   body: String (optional).  Additional context information to place in the notification.  May be
 //                             elided by the browser or by Sandstorm.
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -183,7 +183,7 @@ UserActions = new Mongo.Collection("userActions");
 //
 // Each contains:
 //   _id:  random
-//   userId:  User who has installed this action.
+//   userId:  Account ID of the user who has installed this action.
 //   packageId:  Package used to run this action.
 //   appId:  Same as Packages.findOne(packageId).appId; denormalized for searchability.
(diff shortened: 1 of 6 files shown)


shell/packages/sandstorm-db/db.js
```diff
@@ -639,6 +639,30 @@ SetupSession = new Mongo.Collection("setupSession");
...
+const DesktopNotifications = new Mongo.Collection("desktopNotifications");
+// Responsible for very short-lived queueing of desktop notification information.
+// Entries are removed when they are ~30 seconds old.  This collection is a bit
+// odd in that it is intended primarily for edge-triggered communications, but
+// Meteor's collections aren't really designed to support that organization.
+// Fields for each :
+//
+//   _id: String.  Used as the tag to coordinate notification merging between browser tabs.
+//   creationDate: Date object. indicating when this notification was posted.
+//   accountId: String. Account id to which this notification was published.
```
kentonv: All of our other collections that describe items owned by specific accounts name the field `userId` rather than `accountId`. I know that `accountId` is less ambiguous, but the inconsistency can lead to mistakes when writing Mongo queries where I can't remember which field name to use. Maybe we should stick with `userId`?

zarvox: With the exception of ApiTokens, which is actually probably the most widely-used and complicated of our schemas, and which also uses both `accountId` and `identityId`.  And we `check()` the shape of the struct where it's inserted.

If Mongo let us rename fields across a collection cleanly and atomically, I'd say we should do that and drop `userId` entirely.  Given that it doesn't, and you probably don't want me writing a big risky migration to run against the other collections, I guess I'll rename the field...

It would probably be wise to update the other comments where `userId` is used to say "the `accountId` of the user" rather than "the user"

kentonv: > With the exception of ApiTokens

`ApiTokens` does not use `accountId` to identify to whom the token belongs -- and in fact, ApiTokens never belong to accounts. So my original statement is correct. :)

> It would probably be wise to update the other comments where userId is used to say "the accountId of the user" rather than "the user"

I agree.

shell/server/desktop-notifications-server.js
```diff
@@ -0,0 +1,68 @@
...
+
+    removed: (doc) => {
+      this.removed("desktopNotifications", doc._id);
+    },
+  };
+
+  const sub = db.collections.desktopNotifications.find({
+    userId: this.userId,
+    creationDate: { $gt: subscribeTime },
+  }).observe(callbacks);
```
kentonv: Hmm, couldn't this whole observe now be removed and instead you could just return the cursor?

PR #2191, opened 2016-07-08
Make SAML xml metadata always available

jparyani:


zarvox:
Garply, retest this please

zarvox:
Looks like we hit the IpInterface test race condition again.

PR #2192, opened 2016-07-08
Clarify that free keys are for charities / volunteers / co-ops / etc., not non-profits in general.

kentonv:
Our intent from the beginning was that free keys would be available to volunteer groups, or other groups that have little money and where the participants are unpaid or barely paid. It turns out "non-profit" wasn't the right term for these groups. For example, most educational institutions are non-profit even if they charge huge tuitions, as is any part of the government or any government-backed organization.


kentonv:
Merging now so it can go into the release.

paulproteus:
Seems OK to me. Thanks for requesting my review.

PR #2193, opened 2016-07-08
Move sandstorm accounts ui out of package

zarvox:
To fix the circular dependencies between packages and the shell, we need certain things which currently live statically in packages to get loaded later, so that the globals they depend on can be defined before they run.

One good way to incrementally improve things is to move packages that have no other packages (besides the shell itself) depending on them into the shell.  We can make use of ES6 imports to control when things are defined separately from where they are instantiated, and to make previously-implicit dependencies explicit.

Once we have the ability to control load order in the shell through imports, we can adjust instantiation order.

accounts-settings-ui is currently such a "leaf" package, and marks a relatively safe place to start.


zarvox:
It also turns out that sandstorm-accounts-ui has a decent amount of dead code, which we can remove.  The entire .less stylesheet, the code dealing with configuring login services, and so on have been replaced.

kentonv:
We should probably retain some sort of copyright notice indicating that this is a derived work of the Meteor code.

PR #2195, opened 2016-07-09
Validate quote marks

zarvox:
A pretty straightforward change, and I don't think anyone has patches against `proxy.js` or `pre-meteor.js` at the moment


kentonv:
Odd that this affected comments. Was this a JSCS auto-fix or manual?

zarvox:
Manual.

PR #2201, opened 2016-07-12
Add `deleteAccount` method

jparyani:
Currently only available to admins under the user details page. I'd appreciate suggestions on design and where to put it for users.

image

### DIFF
--- a/shell/client/admin/user-details-client.js
+++ b/shell/client/admin/user-details-client.js
@@ -230,4 +230,18 @@ Template.newAdminUserDetails.events({
       isAdmin: false,
     });
   },
+
+  "click .delete-account"(evt) {
+    const instance = Template.instance();
+    const confirm = window.confirm("Really delete this account forever? This operation is permament and irreversible.");
+    if (confirm) {
+      Meteor.call("deleteAccount", instance.userId, function (err) {
+        if (err) {
+          window.alert(err);
+        } else {
+          Router.go("newAdminUsers");
+        }
+      });
+    }
+  },
 });
--- a/shell/client/admin/user-details-client.js
+++ b/shell/client/admin/user-details-client.js
@@ -267,10 +267,10 @@ Template.newAdminUserDetails.events({
 
   "click [name=\"delete-account-real\"]"(evt, instance) {
     instance.deleteSubmitting.set(true);
-    Meteor.call("deleteAccount", instance.userId, function (err) {
+    Meteor.call("deleteAccount", instance.userId, (err) => {
       instance.deleteSubmitting.set(false);
       if (err) {
-        instance.deleteError.set(err);
+        instance.deleteError.set(err.message);
       } else {
         Router.go("newAdminUsers");
       }
--- a/shell/client/admin/user-details-client.js
+++ b/shell/client/admin/user-details-client.js
@@ -268,6 +268,7 @@ Template.newAdminUserDetails.events({
   "click [name=\"delete-account-real\"]"(evt, instance) {
     instance.deleteSubmitting.set(true);
     Meteor.call("deleteAccount", instance.userId, function (err) {
+      instance.deleteSubmitting.set(false);
       if (err) {
         instance.deleteError.set(err);
       } else {
--- a/shell/client/admin/user-details-client.js
+++ b/shell/client/admin/user-details-client.js
@@ -258,6 +258,7 @@ Template.newAdminUserDetails.events({
 
   "click [name=\"delete-account\"]"(evt, instance) {
     instance.showDeletePopup.set(true);
+    instance.deleteError.set(null);
   },
 
   "click [name=\"cancel-delete-account\"]"(evt, instance) {
--- a/shell/client/admin/user-details-client.js
+++ b/shell/client/admin/user-details-client.js
@@ -70,6 +70,9 @@ Template.newAdminUserDetails.onCreated(function () {
     state: "default", // Also: submitting, error, success
     message: undefined,
   });
+  this.deleteSubmitting = new ReactiveVar(false);
+  this.showDeletePopup = new ReactiveVar(false);
+  this.deleteError = new ReactiveVar(null);
 
   this.isReady = () => {
     // We guard on Router.current().params.userId existing because Iron Router and Blaze
@@ -204,6 +207,28 @@ Template.newAdminUserDetails.helpers({
     const formState = instance.formState.get();
     return formState.state === "submitting";
   },
+
+  showDeletePopup() {
+    const instance = Template.instance();
+    return instance.showDeletePopup.get();
+  },
+
+  deleteError() {
+    const instance = Template.instance();
+    return instance.deleteError.get();
+  },
+
+  deleteSubmitting() {
+    const instance = Template.instance();
+    return instance.deleteSubmitting.get();
+  },
+
+  cancelDelete() {
+    const instance = Template.instance();
+    return () => {
+      instance.showDeletePopup.set(false);
+    };
+  },
 });
 
 Template.newAdminUserDetails.events({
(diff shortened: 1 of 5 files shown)


zarvox:
This button probably belongs under a subheading of its own, where a "Deactivate" button will join it later.

Rather than using window.confirm/window.alert for error handling, please add a ReactiveVar on the template instance to hold a boolean, and have the initial delete button show a confirmation modal dialog.

Prefer binding events by something unique and orthogonal to styles e.g. `button[name=deleteAccount]` rather than by class.  We should be able to change CSS without it breaking event dispatch.

Place the `{{#modalDialogWithBackdrop}}` immediately after the section it should take effect for, and before the following section.

Add a ReactiveVar to capture a potential deletion error message, and provide its value in a helper "deleteError".

Add a ReactiveVar to track whether you are currently submitting the "delete" RPC so you can disable the button while the request is away.

```html
<h2>Status</h2>

<div style="display: flex; flex-direction: row;">
  Active
</div>
<form class="standard-form">
  <div class="button-row">
    <button type="button">Delete...</button>
  </div>
</form>

{{#if showDeletePopup}}
  {{#modalDialogWithBackdrop onDismiss=cancelDelete}}
    {{#if deleteError}}
      {{#focusingErrorBox}}
        {{deleteError}}
      {{/focusingErrorBox}}
    {{/if}}

    <p>Are you sure?  other copy about what will actually happen</p>
    <form class="standard-form">
      <div class="button-row">
        <button class="danger" type="button" disabled={{deleteSubmitting}}>Delete account</button>
        <button type="button">Cancel</button>
      </div>
    </form>
  {{/modalDialogWithBackdrop}}
{{/if}}
```

zarvox:
As for the "delete own account" workflow -- I think we need to break the account page into several subpages.  There's too much content there already, especially on Oasis, and we're likely to need to add a decent bit more, as we want to support notifications and other features.  Referral program stuff probably actually belongs under `/account/referrals`, billing under `/account/billing`, notifications under `/account/notifications`, identities under `/account/identities`, and so on.

I guess if we need to put this function in front of users before that can happen, we can put a similar section at the very bottom of the account settings page, with the same kind of modal prompt before actually deleting.

I would probably like the modal for end-users self-deleting accounts to require said users to type "Yes, permanently delete my account" verbatim into an input box before letting them click the final button-of-doom.  I would consider doing the same for admin account deletion to prevent tragic accidents.

shell/packages/sandstorm-db/db.js
```diff
@@ -2250,6 +2250,59 @@ if (Meteor.isServer) {
   };
 }
 
+if (Meteor.isServer) {
+  SandstormDb.prototype.deleteIdentity = function (identityId) {
+    this.removeApiTokens({ identityId: identityId });
```
zarvox: So, when we do this, what happens to:

* ApiTokens with `membraneRequirements` that reference that `identityId` (e.g. `permissionsHeld`)
* ApiTokens with `owner.user.identityId` referencing this identity

...and what do you expect the UI to show in these circumstances?

shell/packages/sandstorm-db/db.js
```diff
@@ -2250,6 +2250,59 @@ if (Meteor.isServer) {
...
+    this.collections.contacts.remove({ identityId: identityId });
+    Meteor.users.remove({ _id: identityId });
```
zarvox: If I have a save()'d identity capability, after this I should expect it to simply fail to restore()?

shell/client/admin/user-details-client.js
```diff
@@ -231,17 +256,22 @@ Template.newAdminUserDetails.events({
...
-      Meteor.call("deleteAccount", instance.userId, function (err) {
-        if (err) {
-          window.alert(err);
-        } else {
-          Router.go("newAdminUsers");
-        }
-      });
-    }
+  "click [name=\"delete-account\"]"(evt, instance) {
+    instance.showDeletePopup.set(true);
```
zarvox: Also clear the delete error here, lest you see the one from the dialog you previously dismissed?

jparyani:
After discussion in person, I think the change in fccb786 addresses your concerns. We will leave all tokens owned by other users alone, and let them fail to restore so that they can see some form of audit trail (albeit with the user's name deleted).

Testing now to make sure this displays correctly everywhere.

shell/client/admin/user-details.html
```diff
@@ -284,7 +284,36 @@
...
+
+    {{#if showDeletePopup}}
+      {{#modalDialogWithBackdrop onDismiss=cancelDelete}}
+        {{#if deleteError}}
+          {{#focusingErrorBox}}
+            {{deleteError}}
+          {{/focusingErrorBox}}
+        {{/if}}
+
+        <p>Are you sure? This operation is permament and irreversible.</p>
```
zarvox: s/permament/permanent/

shell/client/admin/user-details-client.js
```diff
@@ -231,17 +256,22 @@ Template.newAdminUserDetails.events({
...
+
+  "click [name=\"cancel-delete-account\"]"(evt, instance) {
+    instance.showDeletePopup.set(false);
+  },
+
+  "click [name=\"delete-account-real\"]"(evt, instance) {
+    instance.deleteSubmitting.set(true);
+    Meteor.call("deleteAccount", instance.userId, function (err) {
+      if (err) {
+        instance.deleteError.set(err);
```
zarvox: Set this to `err.message` or an appropriate field, since you appear to just be passing it through to the template.   Otherwise it'll render as `[object Object]`

shell/client/admin/user-details-client.js
```diff
@@ -231,17 +256,22 @@ Template.newAdminUserDetails.events({
...
+    instance.showDeletePopup.set(true);
+  },
+
+  "click [name=\"cancel-delete-account\"]"(evt, instance) {
+    instance.showDeletePopup.set(false);
+  },
+
+  "click [name=\"delete-account-real\"]"(evt, instance) {
+    instance.deleteSubmitting.set(true);
+    Meteor.call("deleteAccount", instance.userId, function (err) {
```
zarvox: prefer fat-arrow function here

kentonv:
* Let's update the demo user deletion code to use the new method, and meanwhile verify that the logic is the same (or that the new logic is superior).

* I put @zarvox in charge of reviewing the UI changes.

* This method call does seem frightening now that I look at it. One alternative idea would be to go ahead and implement "suspend", then have "delete" actually suspend the user while placing them in a deletion queue, with a notification email sent to the admins and some time window in which they can revert the action.

shell/client/admin/user-details.html
```diff
@@ -285,6 +285,36 @@
...
+        {{#if deleteError}}
+          {{#focusingErrorBox}}
+            {{deleteError}}
+          {{/focusingErrorBox}}
+        {{/if}}
+
+        <p>Are you sure? This operation is permanent and irreversible.</p>
+        <form class="standard-form">
+          <div class="button-row">
+            <button class="danger" type="button"
```
zarvox: Hmmm, this isn't showing up in the alarming-red I was expecting, which is quite likely my fault.

Mind adding the following patch to the standard-form in `shell/client/styles/_partials.scss`?

```diff
diff --git a/shell/client/styles/_partials.scss b/shell/client/styles/_partials.scss
index 9b20e73..5aec645 100644
--- a/shell/client/styles/_partials.scss
+++ b/shell/client/styles/_partials.scss
@@ -194,6 +194,9 @@
   button[type=button] {
     @extend %button-base;
     @extend %button-secondary;
+    &.danger {
+      @extend %button-danger;
+    }
   }
 
   &.disabled {
```

shell/client/admin/user-details.html
```diff
@@ -285,6 +285,36 @@
...
+
+    {{#if showDeletePopup}}
+      {{#modalDialogWithBackdrop onDismiss=cancelDelete}}
+        {{#if deleteError}}
+          {{#focusingErrorBox}}
+            {{deleteError}}
+          {{/focusingErrorBox}}
+        {{/if}}
+
+        <p>Are you sure? This operation is permanent and irreversible.</p>
```
zarvox: Can we adjust this copy?  "operation" seems needlessly technical, and yet we may also want to explain precise *what* will happen when this account is deleted.  Maybe:

```html
<h2>Delete {{guessUserName}}</h2>
<p>Deleting this user account will:</p>
<ul>
  <li>Permanently delete any grains created by {{guessUserName}}</li>
  <li>Revoke any sharing links created by {{guessUserName}}</li>
  <li>Remove {{guessUserName}} from all users' contact lists</li>
  <li>Delete all identities owned by this account (except identities associated with another account)</li>
</ul>
<p><strong>Account deletion is permanent and irreversible.</strong>  Please be certain.</p>
```

jparyani:
Closing in favor of #2201

PR #2203, opened 2016-07-12
add ETag header support for NO_CONTENT responses

### LINKED ISSUE #2160: sandstorm-http-bridge should support ETags in the case of 204 responses
Setting ETags on a 204 response can be useful in [certain situations](http://fideloper.com/etags-and-optimistic-concurrency-control).

Currently, when the app's http server gives us a 204 we [always](https://github.com/sandstorm-io/sandstorm/blob/432be37489b210b19d9677f29befcd5c910a0f76/src/sandstorm/sandstorm-http-bridge.c%2B%2B#L337) return the [`Response.noContent`](https://github.com/sandstorm-io/sandstorm/blob/v0.169/src/sandstorm/web-session.capnp#L308-L314) variant, which does not allow ETags to be set. 

Perhaps sandstorm-http-bridge should, in the 204 case, check to see if any etags are present, and if so then use the `Response.content` variant with [`content.statusCode`](https://github.com/sandstorm-io/sandstorm/blob/v0.169/src/sandstorm/web-session.capnp#L283) set to 204.

koivunej:
This resolves #2160 pretty much as hinted in the issue comments:

 * add new field `eTag` to the WebSession::noContent
 * copypaste eTag from-http-to-rpc in sandstorm-http-bridge.c++
 * copypaste eTag from-rpc-to-http-response in proxy.js

With this commit, my currently unpublished TiddlyWiki5 package started to work. TiddlyWiki5 is the only way I tested this change, in addition to building and updating my `vagrant-spk` vm to it.

I also tested by mistake using a package created by older sandstorm version with the sandstorm of this version. In that case, there were no obvious problems, but because of the old `sandstorm-http-bridge` version, the `proxy.js` always saw `noContent.eTag` as `undefined`. I believe this is to be expected feature of capnproto backwards compatibility and confirms that I managed to use correct field index 19 for the property.

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -1643,7 +1643,7 @@ class Proxy {
       const noContent = rpcResponse.noContent;
       const noContentCode = noContentSuccessCodes[noContent.shouldResetForm * 1];
       if (noContent.eTag) {
-        response.setHeader('ETag', composeETag(noContent.eTag));
+        response.setHeader("ETag", composeETag(noContent.eTag));
       }
 
       response.writeHead(noContentCode.id, noContentCode.title);
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -1645,6 +1645,7 @@ class Proxy {
       if (noContent.eTag) {
         response.setHeader('ETag', composeETag(noContent.eTag));
       }
+
       response.writeHead(noContentCode.id, noContentCode.title);
       response.end();
     } else if ('preconditionFailed' in rpcResponse) {
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -1642,6 +1642,9 @@ class Proxy {
     } else if ('noContent' in rpcResponse) {
       const noContent = rpcResponse.noContent;
       const noContentCode = noContentSuccessCodes[noContent.shouldResetForm * 1];
+      if (noContent.eTag) {
+        response.setHeader('ETag', composeETag(noContent.eTag));
+      }
       response.writeHead(noContentCode.id, noContentCode.title);
       response.end();
     } else if ('preconditionFailed' in rpcResponse) {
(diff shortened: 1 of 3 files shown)


garply:

dwrensha:
Garply, ok to test

shell/server/proxy.js
```diff
@@ -1642,6 +1642,9 @@ class Proxy {
     } else if ('noContent' in rpcResponse) {
       const noContent = rpcResponse.noContent;
       const noContentCode = noContentSuccessCodes[noContent.shouldResetForm * 1];
+      if (noContent.eTag) {
+        response.setHeader('ETag', composeETag(noContent.eTag));
+      }
```
dwrensha: Our javascript linter is requesting that you put a blank line after this closing brace. You can run the linter locally by doing `make stylecheck`.

dwrensha:
Looks good to me, though maybe there should be a documentation comment under the `web-session.capnp` declaration of `noContent.eTag`, explaining that the field is optional and, for bonus points, what it might be used for.`

koivunej:
> maybe there should be a documentation comment under the `web-session.capnp` declaration of `noContent.eTag`

Agreed and added.

There seems to be a merged stylecheck change causing the build to fail. I will update the code to pass these new rules even though they are not present in my branch.

dwrensha:
The double quotes change must be due to https://github.com/sandstorm-io/sandstorm/commit/992f50b0a847cbc6a8946de9ad347e7beaf7d281

kentonv:
Hi @koivunej,

This looks good. Can I ask you to sign our CLA? (It's the same CLA used by the Apache Foundation.)

https://sandstorm.io/cla/

Thanks!

koivunej:
@kentonv sorry must've missed the CLA link. Signed and submitted just now.

kentonv:
@koivunej Thanks! You didn't miss anything; we don't do a great job of publicizing our CLA currently.

PR #2204, opened 2016-07-12
test if openssl is available

Finkregh:
this tests if openssl is available, which may be not ;-)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -614,6 +614,7 @@ assert_dependencies() {
 
   which tar > /dev/null || fail "E_TAR_MISSING" "Please install tar(1)."
   which xz > /dev/null || fail "E_XZ_MISSING" "Please install xz(1). (Package may be called 'xz-utils'.)"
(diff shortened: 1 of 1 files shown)


garply:

paulproteus:
Hi!​ Thanks for this pull request!

For what it's worth, we attempt to check for openssl's existence a few
lines above that, but we only do it in the case that the user is enabling
sandcats HTTPS and/or dyndns.

Do you know if the check that exists is enough? I figure it's not since
you're submitting this PR, but I don't see where the install would fail at
the moment, so I'm a little confused, and hoping you can explain!

Thanks,

Asheesh.

paulproteus:
Oh, I see... we do the dependencies check before we know if the user is
going to pick sandcats!​

Finkregh:
oh, my; i must have played around too much with my alpine linux-testbed...

paulproteus:
Okay @Finkregh (-:

FWIW if it isn't enough, then do let me know! I am not convinced that install.sh is perfect; I know it's far from perfect!

PR #2207, opened 2016-07-12
Fix app auto-update, which relied on SandstormDb.addUserActions(), which was recently removed.

kentonv:
Also tighten up method inputs for updateApps().


jparyani:
LGTM

PR #2208, opened 2016-07-13
Honor resumePath when linking email identity to existing account

### LINKED ISSUE #2196: E-mail login by link for demo upgrade doesn't lead back to where the user was
1. Start a demo.
2. Initiate email login.
3. Log in by following the link in the email.

Expected result: After login, you return to the URL from which login was initiated.

Actual result: You end up at `/account`.

My guess is this is because upgrading a demo account "looks like" adding a new identity to an existing account, which is normally initiated from `/accounts`.

@zarvox

zarvox:
Previously, we only returned the `resumePath` to the Accounts.login
codepath; now we also provide it in the return value of the
`linkEmailIdentityToAccount` method.

Fixes #2196.

### DIFF
--- a/shell/packages/accounts-email-token/token_server.js
+++ b/shell/packages/accounts-email-token/token_server.js
@@ -46,7 +46,7 @@ const checkToken = function (tokens, token) {
   return foundToken;
 };
 
-function consumeToken(user, token) {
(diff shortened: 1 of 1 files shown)


jparyani:
Garply, retest this please

PR #2211, opened 2016-07-13
Fix loading anything at all under IE10

zarvox:
A package was using "const" without declaring api.use("ecmascript")
in its package.js.  This resulted in it not getting passed through
the transpilation pipeline, and having its contents placed in the
bundle directly.

IE10 does not support the "const" keyword, and parse errors when it
sees this code, and stops attempting to parse JS.

### DIFF
--- a/shell/packages/sandstorm-accounts-packages/package.js
+++ b/shell/packages/sandstorm-accounts-packages/package.js
@@ -4,6 +4,7 @@ Package.describe({
 });
 
 Package.onUse(function (api) {
+  api.use("ecmascript");
   api.use(["underscore", "random"]);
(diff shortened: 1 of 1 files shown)


paulproteus:
Perhaps we should have a lint check that all our local packages call `api.use("ecmascript")` or something like that? It's too bad that a linter didn't catch this since I don't think many of us use IE10 much.

paulproteus:
Either way, +1

zarvox:
FWIW, after discovering this, I manually audited the rest of our packages, and everything else except our simplesmtp NPM wrapper package calls `api.use("ecmascript")` (and that one is safe not doing so).

PR #2213, opened 2016-07-13
install.sh: Fixes related to running within Docker

paulproteus:
- Run the unprivileged userns test program always, to find out if unprivileged
  userns works properly; if so, skip warnings related to read-only /proc/sys.

- If the sysctl for Debian unprivileged userns is already set to 1, then
  skip warnings.

- Add tests for these.

- Also rearrange the warning to Docker users about the sysctl for easier
  readability.


paulproteus:
I will self-merge once installer-tests/ finishes, but feedback is very welcome!

paulproteus:
Self-merging now that installer-tests passed for me locally.

Issue #2190, opened 2016-07-08
grains should be allowed to offer capabilities to anonymous users


dwrensha:
`SessionContext.offer()` [currently fails](https://github.com/sandstorm-io/sandstorm/blob/3d882d1cc5a2dd33b1c1594b0fd0e809f531af96/shell/server/hack-session.js#L82-L85) for non-logged-in users. This implies that when an anonymous user opens a Collections grain, they have no way to actually navigate to the grains in the collection.

zarvox:
There's an interesting usability tradeoff here, in that by making it possible to interact with large sets of objects anonymously, many users may not bother logging in, which works against them and their teams longer-term, since:

* they'll likely lose track of the grain, since it can't be added to their grain list when they're anonymous
* they may make changes, which apps will not be able to display any useful authorship information around
* we lose audit trail information - all we can say is that a capability passed through a token
* it's unclear to me what notifications, if any, anonymous users should receive (right now, I think the answer is "none")

It's not immediately clear to me precisely what anonymous users should be able to do.  There's a bit of prisoner's dilemma going on with logging in - nobody wants to do it if they're not required to, but everyone else benefits from the additional context when you do.

kentonv:
I think it's pretty clear at least that anonymous users should be able to read a collection and open the grains within, which seems to be the subject of this issue. Whether they should be able to modify the collection is more debatable.

PR #2217, opened 2016-07-13
Deduplicate offered UiView capabilities.

dwrensha:
Immediately after saving, and before returning the cap to the user, check to see if we actually needed to this new cap. If not, then remove it.

### DIFF
--- a/shell/server/hack-session.js
+++ b/shell/server/hack-session.js
@@ -118,18 +118,34 @@ SessionContextImpl = class SessionContextImpl {
       const save = castedCap.save(apiTokenOwner);
       const sturdyRef = waitPromise(save).sturdyRef;
       ApiTokens.update({ _id: hashSturdyRef(sturdyRef) }, { $push: { requirements: requirement } });
-      const powerboxView = isUiView ? {
-        offer: {
-          uiView: {
-            token: sturdyRef.toString(),
-            tokenId: hashSturdyRef(sturdyRef.toString()),
+
+      let powerboxView;
+      if (isUiView) {
+        // Deduplicate.
+        let tokenId = hashSturdyRef(sturdyRef.toString());
+        const newApiToken = ApiTokens.findOne({ _id: tokenId });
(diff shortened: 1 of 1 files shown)


zarvox:
Ahh, I see.  This deletes the newly-generated ApiToken shortly after creating it, if it's the shell save()-ing the UiView, since the shell doesn't need the sturdyref to open a session.  This approach will only work for owner.user tokens, though, so we'll have to come up with something else if we want to support #2190 and also dedupe the tokens produced by the offer() to anonymous sessions.

I guess even with this patch, there'll be a brief window where the dupe `ApiToken` could be created and published but hasn't yet been deleted that the shell would still need to handle?

dwrensha:
>there'll be a brief window where the dupe ApiToken could be created and published but hasn't yet been deleted that the shell would still need to handle?

What special handling might be needed? If there's a dupe, clients will already be deduping.

zarvox:
> What special handling might be needed? If there's a dupe, clients will already be deduping.

No additional handling, but we don't get to drop the client-side deduping code as a result of this change.

kentonv:
I wonder if save() itself should do this deduplification when it sees `sealFor = (user = (...))`. I guess it's not clear what it would return in that case.

Regarding anonymous users, it would be nice not to generate a new sharing link for every click. One possibility here is that the inner grain's webkey could be deterministically generated based on a hash of the webkey that the user is using to view the collection together with the collection grain's SturdyRef to the inner grain.

dwrensha:
> I wonder if save() itself should do this deduplification when it sees sealFor = (user = (...)). I guess it's not clear what it would return in that case.

Maybe we could adjust `SystemPersistent` like so:

```capnp
interface SystemPersistent extends(Persistent(SystemObjectId, ApiTokenOwner)) {
  // ...
}

struct SystemObjectId {
  union {
    token @0 :Data;
    # New token.

    redundant @1 :Text;
   # Hash of existing token that's "as good" as the capability requested to be saved.
  }
}
```

I've shied away from this approach because it seems like overkill.

kentonv:
Let's go with this for now.

PR #2219, opened 2016-07-14
Refactor sturdyref handling code

kentonv:
A few things going on here:
* Creating a registry for FrontendRef types, so that it's easy to add new ones in a self-contained way.
* Getting rid of the ugly `persistentMethods` hack and instead having capability classes inherit from a common PersistentImpl class.
* Simplifying all the different code paths around save(). These turned out to have way more interlocking branches than were really needed. Note that save() no longer delegates to SandstormPermissions.createNewApiToken() -- it seemed that most of the complication of that function doesn't actually apply to the child token case anyhow. It may also be possible to simplify that function now that it's not used for this code ptah.
* Moving messy code into clean imports. @zarvox will like this.

Note that there is some unfinished work here:
* FrontendRefRegistry can register powerbox query handlers in addition to regular handlers, but I'm not actually using that functionality yet. This change only uses the registry for restore handlers.
* The various pseudo-driver implementations ought to be moved into import modules, where the main export is a function which registers the driver with some given FrontendRefRegistry. Then, 00-startup should call all the registration functions.
* core.js still contains a lot of stuff that doesn't really belong there. Again, should be refactored into imports.

I recommend reading the commits separately, though at present the middle commit contains most of the changes.

### DIFF
--- a/shell/imports/server/frontend-ref.js
+++ b/shell/imports/server/frontend-ref.js
@@ -18,9 +18,8 @@ import { checkRequirements } from "./persistent.js";
 
 class FrontendRefRegistry {
   constructor() {
-    this._restoreHandlers = {};
-    this._queryHandlers = {};
-    this._validateHandlers = {};
+    this._frontendRefHandlers = {};
+    this._typeIdHandlers = {};
   }
 
   create(db, frontendRef, requirements) {
@@ -45,20 +44,20 @@ class FrontendRefRegistry {
     }
 
     const key = keys[0];
-    const handler = this._restoreHandlers[key];
+    const handler = this._frontendRefHandlers[key];
     if (!handler) {
       throw new Error("invalid frontendRef: " + JSON.stringify(frontendRef));
     }
 
-    return handler(db, saveTemplate, frontendRef[key]);
+    return handler.restore(db, saveTemplate, frontendRef[key]);
   }
 
   query(db, userAccountId, tag) {
     // Performs a powerbox query using the appropriate registered handler.
 
-    const handler = this._queryHandlers[tag.id];
+    const handler = this._typeIdHandlers[tag.id];
     if (handler) {
-      return handler(db, userAccountId, tag.value);
+      return handler.query(db, userAccountId, tag.value);
     } else {
       return [];  // no matches
     }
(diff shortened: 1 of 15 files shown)


kentonv:
I had tried using Flow-style type annotations in some of the files (Meteor ignores them, so basically I was using them as documentation), but apparently JSCS did not like that so much. Darn. Removed now.

shell/server/core.js
```diff
@@ -501,16 +518,27 @@ restoreInternal = (originalToken, ownerPattern, requirements, tokenId, saveByCop
...
+    } else if (token.grainId) {
+      // It's a UiView.
+
+      // If a grain is attempting to restore a UiView, it gets a UiView which filters out all
+      // the method calls.  In the future, we may allow grains to restore UiViews that pass along the
+      // "is human" pseudopermission (say, to allow an app to proxy all requests to some grain and
+      // do some transformation), which will return a different capability.
+      return { cap: makePersistentUiView(globalDb, saveTemplate, token.grainId) };
+    } else {
+      console.log(token);
```
dwrensha: Stray debug `console.log()`?

shell/server/core.js
```diff
@@ -501,16 +518,27 @@ restoreInternal = (originalToken, ownerPattern, requirements, tokenId, saveByCop
       // discard persistentMethods.
```
dwrensha: `persistentMethods` is no longer a thing, so this comment doesn't make sense. It should probably refer to `PersistentImpl` instead.

shell/packages/sandstorm-ui-powerbox/powerbox-server.js
```diff
@@ -316,9 +321,11 @@ function getVerifiedEmails(db, userId, verifierId) {
   let services = null;
 
   if (verifierId) {
-    const verifier = db.collections.apiTokens.findOne(verifierId);
-    const verifierInfo = ((verifier || {}).frontendRef || {}).emailVerifier;
-    if (!verifierInfo) return [];  // invalid verifier
+    console.log(verifierId);
```
dwrensha: stray debug print?

shell/server/drivers/ip.js
```diff
@@ -121,9 +122,14 @@ IpInterfaceImpl = class IpInterfaceImpl {
   }
 };
 
-makeIpInterface = (persistentMethods) => {
-  return new Capnp.Capability(new IpInterfaceImpl(persistentMethods), IpRpc.PersistentIpInterface);
-};
+// TODO(cleanup): Meteor.startup() needed because 00-startup.js runs *after* code in subdirectories
+//   (ugh).
```
dwrensha: Should server/00-startup.js be moved to server/lib/00-startup.js?

https://guide.meteor.com/structure.html#load-order

shell/imports/server/frontend-ref.js
```diff
@@ -0,0 +1,101 @@
...
+
+  addRestoreHandler(fieldName: string,
+                    callback: (db: SandstormDb, saveTemplate: ApiToken, value: any)
+                              => Capability) {
+    // Registers a callback to use to restore frontendRef capabilities of this type.
+    //
+    // `fieldName` is the name of the field of `ApiTokens.frontendRef` which is filled in for this
+    // ref type.
+    //
+    // `callback` is of type `(db, value, saveTemplate) -> capability`, where
```
dwrensha: Looks like the parameter order is actually `(db, saveTemplate, value`).

shell/server/core.js
```diff
@@ -320,6 +316,7 @@ Meteor.methods({
   },
 });
 
+// TODO(now): Get rid of this maybe?
```
dwrensha: What do you think the best approach would be for removing `saveFrontendRef()`? Replace all uses with `globalFrontendRegistry.()` followed by a `save()`?

Looks like there are only three uses of `saveFrontendRef()`, two of which are left over from the old admin interface and are no longer accessible.

dwrensha:
Overall, I think this pull request makes the `frontendRef` code significantly clearer.

shell/server/core.js
```diff
@@ -140,11 +141,13 @@ const makeSandstormCore = (grainId) => {
...
-  constructor(notificationId, saved, persistentMethods) {
+class NotificationHandle extends PersistentImpl {
+  // TODO(cleanup): Move to a different file.
+
+  constructor(db, saveTemplate, notificationId) {
+    super(db, saveTemplate);
     this.notificationId = notificationId;
-    this.saved = saved;
-    _.extend(this, persistentMethods);
+    this.saved = !!saveTemplate.parentToken;  // TODO(cleanup): ugly
```
dwrensha: I'm confused about what's going on here. Could you add a comment about what the `saved` field actually means and what `parentToken` has to do with it?

shell/server/drivers/ip.js
kentonv: Maybe, but also I intend to move these pseudo-drivers to imports, at which point this particular ugliness will go away.

jparyani:
Garply, retest this please

shell/server/core.js
```diff
@@ -140,17 +141,17 @@ const makeSandstormCore = (grainId) => {
...
+  // TODO(cleanup): Move to a different file.
+
+  constructor(db, saveTemplate, notificationId) {
+    super(db, saveTemplate);
     this.notificationId = notificationId;
-    this.saved = saved;
-    _.extend(this, persistentMethods);
   }
 
   close() {
```
dwrensha: I tried injecting a `KJ_FAIL_REQUIRE()` in the `stayAwake()` implementation in supervisor.c++ and I noticed that this `close()` method never actually gets called! Apparently wrapping the capability with `new Capnp.Capability()` before returning it causes this `close()` to be shadowed or something. If I remove that wrapper, `close()` gets called as expected.

(Actually, you don't even need to inject a `KJ_FAIL_REQUIRE()`; `stayWake()` drops the handle in any case.)

dwrensha: Moreover, I think this is the wrong behavior for `close()`. I believe the logic here should look more like what's in `dropInternal()`. I.e. you should check in `ApiTokens` to see if there's anything left with `frontendRef.notificationHandle` referring to this notification ID, and if not, only then dismiss the notification.

kentonv: > Apparently wrapping the capability with new Capnp.Capability() before returning it causes this close() to be shadowed or something. If I remove that wrapper, close() gets called as expected.

I don't understand. How does the code work at all if you don't covert the object to a capability?

> Moreover, I think this is the wrong behavior for close().

I see what you're getting at here, but conversely, the logic in `dropInternal()` is arguably wrong too: it should not dismiss the notification if any live capabilities still exist. But, detecting live capabilities is hard. Since it doesn't detect live capabilities, if `close()` were implemented as you suggest, then the `dismissNotification()` call could end up happening twice, if the capability were saved and then dropped before the live capability was closed. This seemed to me to be worse than the current behavior.

kentonv:
Added a bunch more refactoring, so now powerbox queries also go through the registry.

shell/server/core.js
```diff
@@ -140,17 +142,17 @@ const makeSandstormCore = (grainId) => {
...
-    const _this = this;
     return inMeteor(() => {
-      if (!_this.saved) {
+      if (!this.isSaved()) {
         dismissNotification(_this.notificationId);
```
dwrensha: `_this` is undefined here.

shell/server/core.js
dwrensha: > How does the code work at all if you don't convert the object to a capability?

node-capnp must be doing the conversion automatically, when the object is passed in as a return value to `addOngoing()`. Note that currently that's the only place that notification handles are constructed. They are never actually restored after being saved, as far as I can tell. Rather, the point of saving them is so that the supervisor can later call `SandstormCore.drop()` on them.

If I edit the notification handle registration to this:

```javascript
globalFrontendRefRegistry.register({
  frontendRefField: "notificationHandle",

  restore(db, saveTemplate, notificationId) {
    return new NotificationHandle(db, saveTemplate, notificationId);
  },
});
```

and fix the dangling `_this` reference below, then wakelock notifications immediately dismiss themselves!

kentonv: Oh, I think I know why `close()` doesn't necessarily get called if the capability wrapper is constructed: The capability wrapper counts as a refcound and thus close() wouldn't be called until it is GC'd. Ouch. I think to fix this, node-capnp has to be adjusted to assume ownership of all returned capability wrappers (requiring the returner to use .clone() if they intended to keep a reference themselves).

kentonv: Also, I would guess that the reason you're seeing the handle dismiss itself if you don't wrap it is because save() doesn't work because you haven't specified that the object implements the Persistent interface.

kentonv: I just verified that _after a while_, `close()` does get called through the wrapper.

There's a bug here regarding capability lifecycle but the bug predates this PR.

dwrensha: The value of `isSaved()` is immutable and determined when the capability is constructed. Whether or not it gets saved, it will call `dismissNotification()` when it is dropped.

dwrensha: > The value of isSaved() is immutable and determined when the capability is constructed

Oh, oops. I just read `save()` and saw that it does the right thing.

kentonv:
I'm gonna merge this! (Tests passed locally.)

PR #2221, opened 2016-07-14
Fix getViewInfo for devPackages

jparyani:


shell/server/core.js
```diff
@@ -198,7 +198,11 @@ class PersistentUiViewImpl {
         throw new Error("grain no longer exists");
       }
 
-      const pkg = Packages.findOne({ _id: grain.packageId });
+      let pkg = Packages.findOne({ _id: grain.packageId });
+      if (!pkg) {
+        pkg = DevPackages.findOne({ appId: grain.appId });
```
dwrensha: This can still fail to find a package, e.g. if the grain was created in dev mode but then added to a collection while `spk dev` was no longer running. So I suppose we need to add a `|| {}` here.

kentonv: I agree with David.

Also note that in theory if a DevPackage is present it should take priority over installed packages since starting the grain will actually open the DevPackage, not the installed package... however, I'm not sure I'd actually act on that theory since it would mean needlessly searching the DevPackages collection even on production servers, so I guess ignore this.

PR #2228, opened 2016-07-16
Add test for Collections app.

### LINKED ISSUE #2225: id="grain-frame" attribute is not unique if there are multiple grains open
I don't know if this actually causes any buggy behavior, but it did catch me by surprise when I was writing some tests!

dwrensha:
Fixes #2225. Adds `newGrain()` and `grainFrame()` testing commands. Decouples `loginDevAccount()` from `installApp()`, because sometimes it's useful to install an app on an existing account.

### DIFF
--- a/docs/administering/self-signed.md
+++ b/docs/administering/self-signed.md
@@ -107,7 +107,7 @@ You can attempt to find out if this is the problem by taking the following steps
 
 - Right-click on the grain area, where the page is not loading, and click ***Inspect** in your browser.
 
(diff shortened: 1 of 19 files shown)


jparyani:
I'm not sure how much I like splitting `loginDevAccount` out of `installApp`. The idea at one point was to remove as much boilerplate as possible and wrap it up in `installApp`.

That being said, the alternative of adding yet another argument/flag to `installApp` is probably wrong as well, so I'm ok with this change I guess.

PR #2232, opened 2016-07-18
Fill in userInfo.pictureUrl with identicon if no other picture is available

### LINKED ISSUE #1866: Default avatar is not passed to the app

dwrensha:
Closes #1866.

### DIFF
--- a/docs/developing/auth.md
+++ b/docs/developing/auth.md
@@ -71,18 +71,13 @@ related to user identity and permissions:
 
 * `X-Sandstorm-User-Picture`: The URL of the user's profile picture.
   The exact resolution of the picture is not specified, but assume
-  it is optimized for a 256x256 or smaller viewport (i.e. the actual
-  size is around 512x512 for high-DPI displays). Although profile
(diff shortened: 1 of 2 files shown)


paulproteus:
Does this merit a docs change? Presumably we should write that we always provide the pictureUrl now, to dispel any old information that we don't. If you agree @dwrensha I'm hopeful you might be willing to add that to this PR. (-:

paulproteus:
Wowee! +1 to the docs change.

dwrensha:
@paulproteus: done. Also, docs.sandstorm.io currently claims that `pictureUrl` is optimized for display at 256x256 pixels, but that runs contrary to the documentation in [identity.capnp](https://github.com/sandstorm-io/sandstorm/blob/v0.172/src/sandstorm/identity.capnp#L63-L68). I've updated the former to match the latter.

PR #2240, opened 2016-07-20
Fix bug in grainlist deduplication on app details page.

dwrensha:
#2181 didn't get this quite right.


zarvox:
:+1:

"my kingdom for a typechecker"

(which should actually become possible/maintainable once Meteor 1.4 drops with node 4 and we switch from JSCS to ESLint, which can handle Flow type signatures)

kentonv:
The more I use dynamically-typed languages, the more I do not understand the arguments that dynamically-typed languages make people more productive.

PR #2244, opened 2016-07-20
Display warning when backing up a grain that holds capabilities.

dwrensha:
One potentially confusing aspect of the Collections app is that `UiView` capabilites held by it do not survive a grain backup/restore cycle. We need some way to teach the user about what to expect in such situations.

This patch proposes to display a warning when the user requests a backup of a grain that holds any capabilities other than the "user identity" capabilities saved by sandstorm-http-bridge. The warning looks like this:

backup-modal

To get the warning to display properly, I needed to give the topbar a `z-index` greater than 0, because otherwise the grain iframe gets drawn above the modal.

### DIFF
--- a/shell/client/grain.html
+++ b/shell/client/grain.html
@@ -148,9 +148,30 @@
 <template name="grainDebugLogButton">
   <button class="grain-button" title="Show Debug Log" id="openDebugLog">Log</button>
 </template>
+
 <template name="grainBackupButton">
   <button class="grain-button {{#if isLoading}}loading{{/if}}" title="Download Backup" id="backupGrain" disabled="{{#if isLoading}}true{{/if}}">Backup</button>
+  {{#if state.confirming }}
+    {{#modalDialogWithBackdrop onDismiss=cancelBackup}}
+      <form name="confirm">
+        <p>
+         Warning: this grain has access to the following objects.
+         A restored backup might not work correctly until you re-grant it access to these objects.
+        </p>
+        <ul class="token-list">
+        {{#each state.confirming}}
+          <li>{{.}}</li>
+        {{/each}}
+       </ul>
+       <div class="align-buttons">
+         <button name="cancel">Cancel</button>
+         <button name="confirm">Download backup</button>
+       </div>
+      </form>
+    {{/modalDialogWithBackdrop}}
+  {{/if}}
 </template>
+
 <template name="grainRestartButton">
   <button class="grain-button" title="Restart App" id="restartGrain">Restart</button>
 </template>
--- a/shell/client/styles/_topbar.scss
+++ b/shell/client/styles/_topbar.scss
@@ -75,8 +75,7 @@ body>.topbar {
   background-color: $topbar-background-color;
   color: #ccc;
 
-  // Using 0 rather than CSS default ("auto") allows introjs to highlight elements in the topbar.
-  z-index: 0;
+  z-index: 1;
 
   @media #{$mobile} {
     line-height: 48px;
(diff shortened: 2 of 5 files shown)


shell/client/grain.html
```diff
@@ -148,9 +148,30 @@
...
   <button class="grain-button" title="Show Debug Log" id="openDebugLog">Log</button>
 </template>
+
 <template name="grainBackupButton">
   <button class="grain-button {{#if isLoading}}loading{{/if}}" title="Download Backup" id="backupGrain" disabled="{{#if isLoading}}true{{/if}}">Backup</button>
+  {{#if state.confirming }}
+    {{#modalDialogWithBackdrop onDismiss=cancelBackup}}
+      <form name="confirm">
+        <p>
+         Warning: this grain has access to the following objects.
```
zarvox: I think the copy here is simultaneously a bit technical ("objects") and handwavey ('might not work correctly").  Can we be more specific or less-jargony?

Can we try putting the warning at the beginning, and adding context after the warning?  I worry users will see the technical line and ignore the rest.  Maybe something like:

```html
<p class="warning-message">
  Warning: restoring this backup might not work how you expect it to
</p>
<p>
  When you restore this backup, you'll need to add each of the following to the new grain:
</p>
<ul class="token-list">...</ul>
```

(I'm not thrilled with my proposed text either, especially once we have additional capability types, and knowing that the user making the backup may not have the authority to add the grains to the collection on their own.)

dwrensha:
Alternate proposal: we add a field `Manifest.backupWarning: LocalizedText` that allows a package to specify a warning message if backup/restore could present surprises.

shell/client/styles/_topbar.scss
```diff
@@ -75,8 +75,7 @@ body>.topbar {
   background-color: $topbar-background-color;
   color: #ccc;
 
-  // Using 0 rather than CSS default ("auto") allows introjs to highlight elements in the topbar.
-  z-index: 0;
+  z-index: 1;
```
zarvox: I'm slightly nervous about this because I think by establishing a separate stacking context it might prevent other modals that aren't direct descendants of the topbar from being able to cover the topbar/sidebar.

Can you double-check that this doesn't inadvertently cover, say, the modal that you'd see at http://local.sandstorm.io:6080/admin/identity when configuring a login provider?  If it does, can you try switching this back to `z-index: 0;` and push the grain-frames further back (potentially negative z-index) instead?

dwrensha: I tried giving `main-content` a `z-index` of -1, but then it didn't accept clicks at all. I don't fully understand what was happening, but it looked like the outer `<body>` tag was in front of it.

dwrensha: Indeed, with this patch as it currently stands, the topbar gets displayed on top of the login provider modals. It doesn't *look* that bad, but when you click the topbar the modal is not dismissed.

dwrensha: This highlights one convenient property of `window.confirm()`: you don't have to figure where to root it in the DOM.

zarvox: Argh, stacking contexts strike again.  We need the `.topbar` and the `.main-content` to be in the same stacking context so that either can have its children stack modals atop the other.  But then within a stacking context, things are sorted in reverse DOM order, which means that since `main-content` comes after `.topbar`, its children will stack on top of `.topbar`s children.

More on stacking contexts: https://philipwalton.com/articles/what-no-one-told-you-about-z-index/

Maybe what we should be doing with `#modalDialogWithBackdrop` is something far more complicated where we transport whatever you put in that part of the template to actually live somewhere else in the DOM where we can guarantee that it lives in the appropriate stacking context.  But then we also need to do a better job of managing focus for keyboard users, since the popup will no longer naturally be the next thing in the DOM to tab to, and when you dismiss the dialog, we'd want to return focus to wherever it was before the dialog was shown ("popping the focus stack")

dwrensha: In this particular case, I think I can work around the problem by using a topbar `popupTemplate` instead of a `modalDialogWithBackdrop`.

zarvox: That sounds like the right choice for now, then.

Someday we can figure out either transcluding `modalDialogWithBackdrop` or making topbar `popupTemplate` have a way to do a better job of isolating the popup from whatever is in the background.  For instance, today the popups block scroll events in the grain-frame, but don't give any visual indication that that should be the case.  And we've wanted the login-nag to be auto-displayed, but to not be application-modal either...

I suspect we'll eventually want a unified way to support at least two behaviors - modal dialog with backdrop (for all the prompt/confirm/share/etc transactional interactions), and toggleable dropdown (for notifications and the account/login menu).

kentonv: I haven't read this thread it complete detail, but note that the `z-index: 0` was added by Asheesh yesterday. Setting `z-index: 0` _also_ creates a new stacking context, which is something I was concerned about when merging that change. Are the problem you describe present for `z-index: 0` (as checked into master), or only for `z-index: 1`?

dwrensha: Only for `z-index: 1`, as far as I can tell.

zarvox: Correct.  It works at `z-index: 0` because the first ancestor with a stacking context of the other element that needs to stack on top of the navbar is also at the same depth in the DOM and `z-index: 0`, so it's order is computed in the same stacking context (and stacks on top because it's further down in the DOM), until you set `z-index: 1`, in which case it doesn't any more.

dwrensha:
Superceded by #2258.

PR #2250, opened 2016-07-21
Change restoreBackup to request a token up front

### LINKED ISSUE #2249: Visitors can upload backup files
Currently a visitor can log in and upload files as grain backups. This will lead to a `403 Unauthorized` page after the file is uploaded, but the actual file will remain in the disk, effectively allowing someone to create visitor accounts and attack the server by uploading enough files to exhaust disk space.

jparyani:
Fixes #2249

This fixes a possible DOS by the "visitor" class of users.

### DIFF
--- a/shell/server/backup.js
+++ b/shell/server/backup.js
@@ -24,10 +24,18 @@ const GrainInfo = Capnp.importSystem("sandstorm/grain.capnp").GrainInfo;
 const TOKEN_CLEANUP_MINUTES = 15;
 const TOKEN_CLEANUP_TIMER = TOKEN_CLEANUP_MINUTES * 60 * 1000;
 
+const uploadTokens = {};
+
 function cleanupToken(tokenId) {
   check(tokenId, String);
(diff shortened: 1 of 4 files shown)


shell/server/backup.js
```diff
@@ -67,10 +75,28 @@ Meteor.methods({
     return token._id;
   },
 
+  newRestoreToken() {
+    if (!isSignedUp()) {
```
kentonv: OrDemo?

shell/server/backup.js
```diff
@@ -24,10 +24,18 @@ const GrainInfo = Capnp.importSystem("sandstorm/grain.capnp").GrainInfo;
 const TOKEN_CLEANUP_MINUTES = 15;
 const TOKEN_CLEANUP_TIMER = TOKEN_CLEANUP_MINUTES * 60 * 1000;
 
+const uploadTokens = {};
```
kentonv: Why switch to an in-memory map? While technically this probably works with Blackrock due to the use of sticky sessions, I would rather not make such assumptions if we can avoid it.

PR #2258, opened 2016-07-21
Add backup warning for Collections app grains.

dwrensha:
This allows the Collections app to alert the user that backup/restore do not work as might be expected. 

backupwarning

After this change, a topbar popup always appears when you download a backup. If the app does not specify a warning, it goes straight to the "in progress" state, where there is a new "cancel" button. 

downloading


dwrensha:
(This is an alternative to #2244.)

kentonv:
Since this is a temporary hack, I'd prefer to trigger it on the app ID rather than add something to the manifest that we potentially have to support forever.

If we observe that other apps need a similar warning, we can reconsider extending the manifest.

dwrensha:
Fair enough. I'm sad that this means the Collections app will in some sense not be "just another app", but as you say, it'll be easy enough to revisit this decision in the future.

PR #2267, opened 2016-07-23
better centering in identicon rendering

dwrensha:
Currently, our identicons can be up to two pixels off-center, due to sloppy rounding in identicon.js. With this patch, we reduce the error to at most 1/2 pixel, which is the best we can do if we want the cells to have constant size.

Before:
old-identicon

After:
new-identicon

### DIFF
--- a/shell/packages/sandstorm-identicons/identicon.js
+++ b/shell/packages/sandstorm-identicons/identicon.js
@@ -27,8 +27,8 @@ Identicon.prototype = {
     render: function(){
         var hash    = this.hash,
             size    = this.size,
-            margin  = Math.floor(size * this.margin),
-            cell    = Math.floor((size - (margin * 2)) / 5),
+            cell    = Math.floor((size - (size * this.margin * 2)) / 5),
+            margin  = Math.floor((size - 5 * cell) / 2);
             image   = new PNGlib(size, size, 256);
 
         // light-grey background


paulproteus:
The results look great. I'm not sure I understand the rationale behind the particular arithmetic though. I wonder if you can explain that, perhaps even in a comment!

Also presumably it's smart to ask upstream to make the same change. Any chance you'd be willing to do that? https://github.com/stewartlord/identicon.js/blob/master/identicon.js#L52 seems to be upstream's version of this file.

dwrensha:
Oh, interesting. Looks like upstream has already put in a similar fix. 

The upstream logic is:
```javascript
                baseMargin  = Math.floor(size * this.margin),
                cell    = Math.floor((size - (baseMargin * 2)) / 5),
                margin = Math.floor((size - cell * 5) / 2),
```

while my logic is:
```javascript 
            cell    = Math.floor((size - (size * this.margin * 2)) / 5),
            margin  = Math.floor((size - 5 * cell) / 2);
```

In my version, `this.margin` is an upper bound on the margin. In the upstream version, its meaning is less clear.

paulproteus:
https://github.com/stewartlord/identicon.js/issues/17 upstream seems to want a SVG implementation, too!

kentonv:
Superseded by #2268

PR #2268, opened 2016-07-23
Switch identicons from PNG to SVG.

### LINKED ISSUE #2161: Contact autocomplete in sharing is slow on first load
Using into the contact autocomplete input is slow and freezes the UI when typing the first character into the box.

It turns out this is due to `makeIdenticon`. It will be cached for subsequent uses, but the first time it will take upwards of 500ms to generate even just a handful of portraits.

dwrensha:
Warning: I have not yet tested this patch with any Microsoft browsers.

This fixes the problem addressed by #2267, and should also fix #2161.

It would be simpler to drop the `size` parameter of `Identicon`, but I've kept it in the hope that maybe it will help Internet Explorer render these things properly.


paulproteus:
https://github.com/stewartlord/identicon.js/issues/17 upstream seems to want a SVG implementation, too!

(Also I accidentally left this comment in the wrong place a moment ago.)

ocdtrekkie:
Yeah, please do check the Microsoft browsers, they have wonky SVG handling.

kentonv:
Whoa awesome! Agreed needs careful testing in IE / Edge.

dwrensha:
Update: I've verified that this works with Edge and IE 11 in Windows 10.

dwrensha:
I tried IE 10 in a Windows 7 testing VM, and the SVG identicons don't appear to cause any trouble there, either. I did notice, however, some unrelated rendering problems:

ie10-1
ie10-2

kentonv:
Yeah a lot of the icons are broken on IE10, but IE10 is no longer supported by Microsoft, so...

stewartlord:
Hi folks, I added a PR for SVG support upstream. Wrote it before I saw this actually. I wonder if it meets your requirements? One significant difference is that I keep support for PNGs (thinking of BC).

If it does work for you, maybe it could save you having to fork?

stewartlord:
https://github.com/stewartlord/identicon.js/pull/18

dwrensha:
@stewartlord Thanks for the heads-up!

PR #2269, opened 2016-07-23
installer-tests: disable https tests for now

paulproteus:
They're failing, so keeping them running isn't doing anyone any good.


paulproteus:

PR #2271, opened 2016-07-23
Response header whitelist and X-Sandstorm-App- prefix

kentonv:
This is mostly @sprin's https://github.com/sandstorm-io/sandstorm/pull/1202 with some additional changes:

* While @sprin was working on the change, @zombiezen independently added header prefix whitelisting using a trailing '*' to indicate that the whitelist entry is a prefix. @sprin reconciled this with his changes by instead creating a separate list for prefixes. I feel, though, that having one list with wildcards is easier to read and understand, so I went back to this model. (This involved removing a couple of @sprin's commits that weren't relevant, and then adding some code of my own.)
* I prematurely optimized the header matching by introducing a `HeaderWhitelist` class in both C++ and Javascript which sort of "precompiles" the whitelist for efficiency. This probably wasn't necessary -- linear scans would have been fine since the whitelist is not huge.
* I changed the approach to preventing HTTP injections. Instead of applying regexes on the additionalHeaders only, I changed the code at a higher level to simply detect if any of the HTTP lines ended up containing newline characters. This approach solves not only injections in additionalHeaders but also across other headers that are represented as arbitrary strings in the capnp.

I will probably merge this later today for release.

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -1582,6 +1582,26 @@ class Proxy {
       response.setHeader("Content-Security-Policy", "default-src 'none'; sandbox");
     }
 
+    if (rpcResponse.additionalHeaders && rpcResponse.additionalHeaders.length > 0) {
+      // Build object for whitelisted header lookup
+      var whitelisted = {};
+      WebSession.Response.headerWhitelist.forEach(function(header) {
+        whitelisted[header] = 1;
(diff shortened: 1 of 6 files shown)


mnutt:
Tested this branch with Davros / `X-OC-Mtime` and it works just fine as well.

PR #2273, opened 2016-07-24
installer-tests: Wait a little bit before asserting it has bound the …

paulproteus:
…ports

This should be fixed by adding a wait-loop to install.sh rather than doing the
wait-loop in a test, but life is short for now.


paulproteus:

PR #2274, opened 2016-07-24
docs: Clarify install landing area

paulproteus:
In the /administering/ page of the docs, we (before this PR) emphasize the https://sandstorm.io/install page. But the https://sandstorm.io/install page's main point is to lead people to https://docs.sandstorm.io/en/latest/install/ in my opinion. So I now emphasize https://docs.sandstorm.io/en/latest/install/ .


paulproteus:

PR #2277, opened 2016-07-25
install.sh: fix specifying an admin token

paulproteus:


paulproteus:
Testing done:

- The first commit contains a test; I ran this test, and verified that it failed.
- The second commit contains some code changes. I ran the test after the code changes and verified that the test now passes.

Self-merging since 100% installer-tests & install.sh.

PR #2283, opened 2016-07-25
install.sh: Avoid putting :80 in BASE_URL

### LINKED ISSUE #2252: GitHub login redirect loop if ":80" appears in BASE_URL
I'm currently experiencing a redirect loop where on http://tpcikmmyjjapslvl.sandcats-dev.sandstorm.io/ (a test Sandstorm install I created), whenever I try to log in as the first GitHub user, I get sent to `/`, and then immediately redirected to `/setup`. Additionally, login does not seem to be succeeding.

I'll attempt to diagnose but wanted to leave a note that I'm experiencing something weird.

paulproteus:
See also: #2252

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -958,8 +958,17 @@ configure_hostnames() {
     fi
   fi
 
-  # A typical server's DEFAULT_BASE_URL is its hostname plus port over HTTP.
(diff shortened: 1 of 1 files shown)


paulproteus:
It's really more like see-also-#2252 because there could be servers in the wild with this problem. So I'll try not to close #2252 for now.

PR #2286, opened 2016-07-26
Mention that moving a feature-key is OK

paulproteus:


paulproteus:
I'm excited to get feedback from @mrdomino if this change would have helped him. For context, the change would appear here: https://docs.sandstorm.io/en/latest/administering/for-work/

In the absence of feedback I'll merge tonight, which will update https://docs.sandstorm.io/en/latest/administering/for-work/ .

mrdomino:
It might've. I definitely wouldn't have checked that page before asking, but I might've read that back when I was learning about Sandstorm for Work and retained it.

paulproteus:
Good enough for me for now. I'll merge it! Thanks @mrdomino

PR #2287, opened 2016-07-26
Add preinstalled apps

jparyani:
Currently it is only a new route in the admin panel. A subsequent commit
will add it to the setup wizard as well.

### DIFF
--- a/shell/client/admin/admin.html
+++ b/shell/client/admin/admin.html
@@ -86,13 +86,13 @@
             <div class="item-subtext">Customize a splash page, terms of service, privacy policy,
                 etc.</div>
           {{/adminNavItem}}
-          {{#adminNavItem routeName="newAdminAppSources"}}
-            <div class="item-name">App sources</div>
-            <div class="item-subtext">Where to look for apps and app updates.</div>
-          {{/adminNavItem}}
           {{#adminNavItem routeName="newAdminPreinstalledApps"}}
             <div class="item-name">Pre-installed apps</div>
             <div class="item-subtext">Choose apps to pre-install for new users.</div>
+          {{/adminNavItem}}
+          {{#adminNavItem routeName="newAdminAppSources"}}
+            <div class="item-name">App sources</div>
+            <div class="item-subtext">Where to look for apps and app updates.</div>
           {{/adminNavItem}}
         </ul>
       </li>
--- a/shell/client/admin/admin.html
+++ b/shell/client/admin/admin.html
@@ -90,6 +90,10 @@
             <div class="item-name">App sources</div>
             <div class="item-subtext">Where to look for apps and app updates.</div>
           {{/adminNavItem}}
+          {{#adminNavItem routeName="newAdminPreinstalledApps"}}
+            <div class="item-name">Pre-installed Apps</div>
+            <div class="item-subtext">Configure pre-installed apps for new users.</div>
+          {{/adminNavItem}}
         </ul>
       </li>
       <li>
--- a/shell/client/admin/admin.html
+++ b/shell/client/admin/admin.html
@@ -91,8 +91,8 @@
             <div class="item-subtext">Where to look for apps and app updates.</div>
           {{/adminNavItem}}
           {{#adminNavItem routeName="newAdminPreinstalledApps"}}
-            <div class="item-name">Pre-installed Apps</div>
-            <div class="item-subtext">Configure pre-installed apps for new users.</div>
+            <div class="item-name">Pre-installed apps</div>
+            <div class="item-subtext">Choose apps to pre-install for new users.</div>
           {{/adminNavItem}}
         </ul>
       </li>
--- a/shell/client/admin/preinstalled-apps-client.js
+++ b/shell/client/admin/preinstalled-apps-client.js
@@ -13,8 +13,12 @@ Template.newAdminPreinstalledApps.onCreated(function () {
     state: "edit", // Other allowed states: "submitting", "success", and "error"
     message: undefined,
   });
-  const preinstalledApps = globalDb.getSettingWithFallback("preinstalledApps", {});
-  this.preinstalledAppIds = new ReactiveVar(_.pluck(preinstalledApps, "appId"));
+  const preinstalledApps = globalDb.getSettingWithFallback("preinstalledApps", []);
+  const preinstalledAppAndPackgeIds = {};
+  preinstalledApps.forEach((row) => {
+    preinstalledAppAndPackgeIds[row.appId] = row.packageId;
+  });
+  this.preinstalledAppAndPackgeIds = new ReactiveVar(preinstalledAppAndPackgeIds);
   this.showAllApps = new ReactiveVar(false);
   this.appIndexSubscription = this.subscribe("appIndexAdmin");
   this.packageSubscription = this.subscribe("allPackages");
@@ -70,7 +74,7 @@ Template.newAdminPreinstalledApps.helpers({
 
   getRowData() {
     const instance = Template.instance();
-    this.preinstalledAppIds = instance.preinstalledAppIds;
+    this.preinstalledAppAndPackgeIds = instance.preinstalledAppAndPackgeIds;
     this.formChanged = instance.formChanged;
     return this;
   },
@@ -82,7 +86,7 @@ Template.newAdminPreinstalledApps.helpers({
 
 Template._appRow.helpers({
   isAppPreinstalled() {
-    return _.contains(this.preinstalledAppIds.get(), this.appId);
+    return !!this.preinstalledAppAndPackgeIds.get()[this.appId];
   },
 
   isAppDownloaded() {
@@ -120,13 +124,19 @@ Template.newAdminPreinstalledApps.events({
 
   "click .save"(evt) {
     const instance = Template.instance();
-    const preinstalledAppIds = instance.preinstalledAppIds.get();
+    const preinstalledAppAndPackgeIds = instance.preinstalledAppAndPackgeIds.get();
 
     instance.formState.set({
       state: "submitting",
       message: "",
     });
-    Meteor.call("setPreinstalledApps", preinstalledAppIds, (err) => {
+    const appAndPackageIdList = _.map(preinstalledAppAndPackgeIds, (val, key) => {
+      return {
+        appId: key,
+        packageId: val,
+      };
+    });
+    Meteor.call("setPreinstalledApps", appAndPackageIdList, (err) => {
       if (err) {
         instance.formState.set({
           state: "error",
@@ -134,8 +144,7 @@ Template.newAdminPreinstalledApps.events({
         });
       } else {
         let notYetInstalled = [];
-        preinstalledAppIds.forEach((appId) => {
-          const packageId = globalDb.collections.appIndex.findOne({ appId: appId }).packageId;
+        _.each(preinstalledAppAndPackgeIds, (packageId) => {
           const pack = globalDb.collections.packages.findOne({ _id: packageId });
           if (!pack || pack.status !== "ready") {
             notYetInstalled.push(packageId);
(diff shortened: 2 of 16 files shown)


shell/client/admin/admin.html
```diff
@@ -90,6 +90,10 @@
             <div class="item-name">App sources</div>
             <div class="item-subtext">Where to look for apps and app updates.</div>
           {{/adminNavItem}}
+          {{#adminNavItem routeName="newAdminPreinstalledApps"}}
+            <div class="item-name">Pre-installed Apps</div>
```
zarvox: fix capitalization: "Pre-installed apps"

shell/client/admin/admin.html
```diff
@@ -90,6 +90,10 @@
...
+            <div class="item-subtext">Configure pre-installed apps for new users.</div>
```
zarvox: maybe "Choose apps to pre-install for new users."?

shell/client/admin/preinstalled-apps.html
```diff
@@ -0,0 +1,108 @@
...
+  </p>
+  <form class="admin-preinstalled-apps">
+    <div class="form-group">
+      <label>
+        Productivity Suite
+      </label>
+      <span class="form-subtext">
+        We recommend pre-installing these core apps that are useful to most users.
+      </span>
+      <div class="productivity-suite" role="grid">
```
zarvox: suggest renaming `productivity-suite` to `suite` if you're going to reuse it for system apps

shell/client/admin/preinstalled-apps.html
```diff
@@ -0,0 +1,108 @@
...
+      </div>
+    </div>
+
+    <div class="button-row">
+      <button type="submit" class="save" disabled="{{saveDisabled}}">Save</button>
+    </div>
+  </form>
+</template>
+
+<template name="_appRow">
```
zarvox: Mind documenting what a data context is expected to look like for this template?  e.g. which of the things below are expected to come from the data context, and which are actually template helpers?

shell/client/admin/preinstalled-apps.html
```diff
@@ -0,0 +1,108 @@
...
+      <div class="all-apps" role="grid">
+        <div role="rowgroup">
+          {{#each allApps}}
+            {{> _appRow getRowData}}
+          {{/each}}
+        </div>
+        {{#if paginateApps}}
+          <div role="row" class="toggle-more">
+            <div role="gridcell">
+              <div role="button">
```
zarvox: Does it work for this to just be a `<button>` proper?

shell/server/installer.js
```diff
@@ -56,7 +56,8 @@ const deletePackageInternal = (pkg) => {
     const grain = Grains.findOne({ packageId: packageId });
     const notificationQuery = {};
     notificationQuery["appUpdates." + pkg.appId] = { $exists: true };
-    if (!grain && !action && !(pkg.isAutoUpdated && Notifications.findOne(notificationQuery))) {
+    if (!grain && !action && !(pkg.isAutoUpdated && Notifications.findOne(notificationQuery))
+        && !globalDb.isPackagePreinstalled(ppackageId)) {
```
zarvox: is `ppackageId` a typo here?

shell/client/admin/preinstalled-apps-client.js
```diff
@@ -13,8 +13,12 @@ Template.newAdminPreinstalledApps.onCreated(function () {
     state: "edit", // Other allowed states: "submitting", "success", and "error"
     message: undefined,
   });
-  const preinstalledApps = globalDb.getSettingWithFallback("preinstalledApps", {});
-  this.preinstalledAppIds = new ReactiveVar(_.pluck(preinstalledApps, "appId"));
+  const preinstalledApps = globalDb.getSettingWithFallback("preinstalledApps", []);
+  const preinstalledAppAndPackgeIds = {};
```
paulproteus: Typo: `preinstalledAppAndPackgeIds` != `preinstalledAppAndPackageIds`

Not necessarily a showstopper, but it's a bit odd, so I thought I should mention it.

shell/packages/sandstorm-db/db.js
```diff
@@ -1651,27 +1661,29 @@ _.extend(SandstormDb.prototype, {
...
+    // This function both sets the appId as ready and updates the packageId for the given appId
+    // Setting the packageId is especially useful in installer.js, as it always ensures the
+    // latest installed package will be set as ready.
+    this.collections.settings.update(
+      { _id: "preinstalledApps", "value.appId": appId },
+      { $set: { "value.$.status": "ready", "value.$.packageId": packageId } });
   },
 
-  ensureAppPreinstall: function (appId) {
+  ensureAppPreinstall: function (packageId, appId) {
```
paulproteus: Weird that this function takes the args in `packageId, appId` order then passes them to another function that expects them in `appId, packageId` order (`setPreinstallAppAsReady()`).

shell/server/installer.js
```diff
@@ -385,6 +388,10 @@ AppInstaller = class AppInstaller {
             (manifest.appTitle && manifest.appTitle.defaultText), manifest.appVersion,
             (manifest.appMarketingVersion && manifest.appMarketingVersion.defaultText));
         }
+
+        if (_this.isPreinstalled) {
```
paulproteus: I guess this means that any user accepting an update notification for a pre-installed app updates it for all users who have the app pre-installed?

ocdtrekkie: What if someone upload-installs a non-public version of an update to an app marked as a preinstall app?

What if I have preinstalled apps, but don't want to update them on my account from where I'm at?

jparyani: Sorry, I discussed this with @paulproteus in person after he posted this note, but the quick answer is that this doesn't actually update pre-existing users. This function is called when a package has been successfully downloaded to the server, but it does not add UserActions (ie. add it to a user's app list).

To your first question, the code now handles this correctly. The only way the packageId of a preinstalled app is ever changed is either when an admin clicks "Save" on the pre-installed apps admin panel, or when an autoupdate comes over the wire. Local versions are ignored.

ocdtrekkie: Sounds good.

kentonv:
* Would it make sense to merge "admin/app-sources" and "admin/preinstalled-apps" into a single "admin/apps" route? IIRC we are expecting to extend this table with other app settings not related to pre-installation.
* Would it make sense to show the "downloaded" indicator even for apps that aren't set to auto-update? It seems like useful information.

Also interested in @zarvox's and @neynah's opinions on these questions.

shell/client/setup-wizard/wizard.html
```diff
@@ -81,6 +90,7 @@
     <div class="{{#if currentStepAtOrPast "organization"}}complete{{else}}incomplete{{/if}}"></div>
     {{/if}}
     <div class="{{#if currentStepAtOrPast "email"       }}complete{{else}}incomplete{{/if}}"></div>
+    <div class="{{#if currentStepAtOrPast "preinstalled"       }}complete{{else}}incomplete{{/if}}"></div>
```
kentonv: Line up braces with surroundings?

shell/packages/sandstorm-db/migrations.js
```diff
@@ -695,12 +726,14 @@ function migrateToLatest() {
...
   } else {
     const applied = Migrations.findOne({ _id: "migrations_applied" });
+    let isNewServer = false;
     let start;
     if (!applied) {
       // Migrations table is not yet seeded with a value.  This means it has
       // applied 0 migrations.  Persist this.
       Migrations.insert({ _id: "migrations_applied", value: 0 });
       start = 0;
+      isNewServer = true;
```
kentonv: If the server crashes in the middle of running migrations but before running new-server setup then when it comes back up it'll think it is not a new server and won't do the setup.

zarvox:
Re: merging app sources/preinstalled apps: in due time, yes, but not right now.

My vision for this would actually be making the "app list and things you can do with the apps" be the primary content on `Admin / Apps`, but to add a link on that page to `Admin / Apps / Sources` or something that links to a page that contains what `Admin / App sources` has today.  I expect that approximately zero users have actually modified any of the settings on `Admin / App sources`, and I don't feel like that content should distract from the rest of the "Apps" page.  We'd also have to either have two separate save buttons, or apply all the settings at once...the workflows become less clear.

I think that when we have the full-on "admin-managed app lifecycle", the time to rename/restructure the sections to just `admin/apps` will be upon us, but for now, let's keep them separate.

All that said, it's probably worth putting preinstalled apps above app sources in the list?

shell/packages/sandstorm-db/db.js
```diff
@@ -153,6 +153,7 @@ Packages = new Mongo.Collection("packages");
 //   url:  When status is "download", the URL from which the SPK can be obtained, if provided.
 //   isAutoUpdated: This package was downloaded as part of an auto-update. We shouldn't clean it up
 //     even if it has no users.
+//   isPreinstalled: This package was downloaded in order to be pre-installed by users.
```
kentonv: This field is somewhat counter-intuitive as it really only flags why the package was originally downloaded, not whether the package is currently chosen for pre-installation. Would it make sense to check the admin setting dynamically inside the installer code, rather than check for this field on the package record?

shell/packages/sandstorm-db/db.js
```diff
@@ -1614,6 +1615,212 @@ _.extend(SandstormDb.prototype, {
...
+          }
+        } else {
+          this.startInstall(app.packageId, url, false, true, isAppPreinstalled);
+        }
+      }
+    });
+  },
+
+  getAppIdForPreinstalledPackage: function (packageId) {
+    const setting = Settings.findOne({ _id: "preinstalledApps", "value.packageId": packageId },
```
kentonv: Please document the `preinstalledApps` setting structure in the documentation for the `Settings` collection.

shell/packages/sandstorm-db/db.js
kentonv: Note that the dynamic post-install check would be: "If app ID is marked for pre-install *and* the package ID is the most recent one listed in the app index"

jparyani: Actually I believe the check should only be "Is this packageId marked for pre-install". I purposely denormalized packageId into the setting for situations like this, so that it's more feasible to implement non-app store apps as preinstalled (someday).

jparyani: On no, scratch that, it doesn't work with autoupdates.

neynah:
> Would it make sense to merge "admin/app-sources" and "admin/preinstalled-apps" into a single "admin/apps" route? IIRC we are expecting to extend this table with other app settings not related to pre-installation.

I feel these configurations are quite separate and I'm happy having them split for now.

> All that said, it's probably worth putting preinstalled apps above app sources in the list?

I agree. +1

shell/packages/sandstorm-db/migrations.js
```diff
@@ -664,6 +664,12 @@ function startPreinstallingApps() {
   Meteor.setTimeout(startPreinstallingAppsHelper, 0);
 }
 
+function setNewServer() {
+  if (!Migrations.findOne({ _id: "new_server_migrations_applied" })) {
+    Migrations.insert({ _id: "new_server_migrations_applied", value: true });
```
kentonv: Shouldn't you be setting the value to `false` here?

shell/packages/sandstorm-db/db.js
```diff
@@ -1704,16 +1705,22 @@ _.extend(SandstormDb.prototype, {
 
   getPreinstalledAppIds: function () {
```
kentonv: Maybe this should change to `getReadyPreinstalledAppIds()` to avoid confusion.

shell/packages/sandstorm-db/db.js
```diff
@@ -1755,23 +1762,19 @@ _.extend(SandstormDb.prototype, {
       if (pack.status === "ready") {
         this.setPreinstallAppAsReady(appId, packageId);
       } else {
-        const newPack = this.collections.packages.findAndModify({
```
kentonv: You are removing the declaration of `newPack`, but it is still used below?

shell/server/installer.js
```diff
@@ -389,7 +387,12 @@ AppInstaller = class AppInstaller {
             (manifest.appMarketingVersion && manifest.appMarketingVersion.defaultText));
         }
 
-        if (_this.isPreinstalled) {
+        if (globalDb.getPackageIdForPreinstalledApp(_this.appId) &&
+            globalDb.collections.appIndex.findOne({ "packageId": _this.packageId })) {
```
kentonv: Nitpick: This query would be more efficient if you searched by _id as well. Otherwise it's going to table-scan.

jparyani:
All comments have been addressed.

PR #2289, opened 2016-07-26
Move some packages to shell

zarvox:
## Changes made

This commit rearranges a number of things that previously lived in `shell/packages` to live in combinations of `shell/client`, `shell/server`, and `shell/imports` (with the long-term goal of fixing our dependency cycles and avoiding packages which are loaded before the shell depending on globals instantiated by the shell).

Most accounts-related code moved to `client/accounts/<identity-provider>` and `server/accounts/<identity-provider>` for the respective IDPs.  These seemed like information-scentful places to go looking for that sort of code if you were new to the Sandstorm codebase.

I did various cleanups to code inside packages, like making more things locally-scoped, fixing declaration orders (avoid relying on JS's scope-hoisting), extracting code to `imports`, and so on.  A common theme is avoiding hanging functions off the `Meteor` or `Accounts` globals.

There were a number of strange NPM dependencies specified by packages - `packages/sandstorm-email` and `packages/simplesmtp` pulled in and used different versions of simplesmtp and mailcomposer.  I adopted the newest version of each package into `shell/package.json` and tested email login, which worked fine.

I removed the `typ:ldapjs` dependency in favor of pulling ldapjs straight from NPM (at the previously-frozen version).  Upgrades can happen later.

## Testing done

I tested all five identity providers with the as-refactored code, and all behaved as expected.


zarvox:
FWIW, as of this writing, "Extract account-saml from packages/" is the last patch in the series, but due to internal rebase shenanigans the date of "Eliminate use of SandstormEmail global" is later than the aforementioned commit, and GitHub appears to sort the series by date, rather than by commit order in the branch.  Which is why the CI status dot appears to be on the second-to-last commit, when it's actually HEAD of the branch in question.

kentonv:
Removing SandstormEmail from the global scope broke Oasis invoices. I guess this is not super-urgent since they are all $0 anyhow, but...

PR #2291, opened 2016-07-29
Create uninstall.sh

benyanke:
uninstaller for sandstorm.

### DIFF
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+rm /opt/sandstorm -rf
+rm /usr/local/bin/* -rf
+rm /etc/init.d/sandstorm -rf


garply:

ocdtrekkie:
I would be -1 on this file as it's currently written. (My opinion has no significant weight, mind you.)

This command, without any additional warning, would delete all user data for the server, which is potentially very, very bad.

A proper uninstaller should probably ask if you want to preserve user data or not. If you choose to delete user data, it should ask you to type "I UNDERSTAND" or some similar string, after a large red text warning that all Sandstorm user data will be deleted.

benyanke:
Good call. Perhaps I should add a prompt giving the option "Remove User
Data" or "Remove Sandstorm Only."


In Christ,
Ben Yanke

*benyanke.com <http://benyanke.com/>*
C: 608-318-3642

ocdtrekkie:
Yeah, I mean, the biggest issue I've seen is a lot of people uninstalling Sandstorm because they want to reinstall it. So the challenge is to ensure the uninstaller can keep Sandstorm grains and the data needed to tell Sandstorm it still has those grains if you reinstall it. Presumably.

benyanke:
That makes sense.

I'll retool it to give those options to the user and give another pull
request. Feel free to dump this one.

In Christ,
Ben Yanke

c: 608-318-3642

-sent from my mobile

zarvox:
Also note that while `/opt/sandstorm` is the default install path, the installer script allows installation to a user-selected directory - an uninstall script should probably also take its installed location into account when doing an uninstallation.

benyanke:
Is that notated somewhere? Or is there a reliable way to find it?

In Christ,
Ben Yanke

c: 608-318-3642

-sent from my mobile

zarvox:
We don't explicitly save the install root anywhere.

I would imagine that you could infer it from either the target of the `/usr/local/bin/sandstorm` symlink, or from the absolute path of the `uninstall.sh` if the uninstall script were part of the package.

I think I like the second option better - placing the uninstall script in the `sandstorm-$VERSION` means that the uninstallation logic can be updated over time if we wind up changing data layout, or need to fix a bug in the uninstaller, or anything like that.  So I'd imagine in a standard install, you'd have:

* Most things installed under `/opt/sandstorm/`
* Inside `/opt/sandstorm`, `latest` is a symlink to the latest bundle (at the time of writing, `sandstorm-173`)
* Inside `sandstorm-173`, there's a file `uninstall.sh`
* `/opt/sandstorm/uninstall.sh` is a symlink to `/opt/sandstorm/latest/uninstall.sh`

Then your script could find its canonical path (`/opt/sandstorm/sandstorm-173/uninstall.sh`), and look up its containing folder (`/opt/sandstorm/sandstorm-173`) and then the parent of that folder (`/opt/sandstorm`) to find the installation root.  You'd probably want to make sure that the directory layout appears as expected before doing anything destructive.  Otherwise, a user might move the uninstall script somewhere else, or try to run it out of the sandstorm git checkout, and inadvertently wipe out their home directory or something.

kentonv:
IMO, we should have a `sandstorm uninstall` command for this, rather than a separately-distributed script.

benyanke:
Also agreed. I was hoping to full the gap with a script, but a command is
much better.

In Christ,
Ben Yanke

c: 608-318-3642

-sent from my mobile

kentonv:
OK, here's `sandstorm uninstall`: https://github.com/sandstorm-io/sandstorm/pull/2383

PR #2296, opened 2016-07-30
Drop invalid ETag data

### LINKED ISSUE #2295: sandstorm-http-bridge crashes if ETag header has no quotation marks
Context:

- @ocdtrekkie has a Sandstorm app (EtherDraw) that "used to work" with old versions of Sandstorm's sandstorm-http-bridge. With new versions of sandstorm-http-bridge, the app "doesn't work".
- Further review indicates that EtherDraw is sending an `ETag` of `1.0.6` for the `socket.io.js` file, which is invalid per the HTTP RFC. The HTTP RFC specifies that an ETag MUST have a double-quote mark around the actual value, plus an optional `W/` prefix.
- Our code crashes in the case of an invalid ETag; see: https://github.com/sandstorm-io/sandstorm/blob/master/src/sandstorm/sandstorm-http-bridge.c++#L786
- I've reproduced the issue with a PHP script that starts with: `<?php header("ETag: invalid"); ?>`

Options:

We can:

- 1.  Drop ETag data that is invalid, and stop crashing sandstorm-http-bridge. This will result in more HTTP traffic as the browser doesn't know as much about how to cache the data. It would also be undiscoverable for app authors, unless we log it somewhere (perhaps stderr so it appears in the grain log, though if it appears in the grain log, then we're wasting user storage as well).
- 2. Silently upgrade ETag data that lacks quotes into a version with quotes. This also has impact on caching because browsers will send an `If-None-Match` header that the app backend won't understand as the same ETag. See also [this issue in Play Framework](https://play.lighthouseapp.com/projects/57987-play-framework/tickets/1522-etags-header-is-not-being-set-in-line-with-specification-standard). We probably should write something to stderr each time we do this from sandstorm-http-bridge asking the packager to fix the app.
- 3. Continue to crash in the face of invalid ETag data.
- 4. Have further discussion.

I'll note that we already have workarounds for other app bugs in sandstorm-http-bridge; consider the parsing of the `Date:` header.  Note also that this is a blocker for @ocdtrekkie reuploading the EtherDraw app.

My plan:

- Implement option 1 above and submit a pull request with it, awaiting review/merge by @kentonv.

paulproteus:
Close #2295

Requesting preliminary review from everyone, and requesting final review/merge/rejection from @kentonv.

### DIFF
--- a/docs/developing/troubleshooting.md
+++ b/docs/developing/troubleshooting.md
@@ -66,10 +66,12 @@ applications however.
 
 ## ETag data
 
-Sandstorm drops invalid ETag data. This typically has no impact on app functionality. When Sandstorm
-drops invalid ETag data, it stores a warning in the grain log that it did so, limiting the log
-messages to at most one per launch of a grain to avoid filling up the log file. See also [issue
-#2295.](https://github.com/sandstorm-io/sandstorm/issues/2295)
+Sandstorm normally allows valid `ETag` headers to pass through to the client. However, the HTTP
+standard requires `ETag` values to take a specific form -- for example, they must be quoted. Many
+applications incorrectly provide non-quoted values for this header. Sandstorm does not permit
+invalid header values because ambiguity can lead to security problems. Therefore, Sandstorm will
+drop invalid `ETag` headers and will write a warning to the grain's debug log. Usually, this does
+not affect app functionality, other than to reduce the effectiveness of the browser's cache.
 
 In versions v0.176 and earlier, the grain would crash, rather than log an error message, if an
 invalid ETag header was emitted by the app.  In more-recent versions, Sandstorm drops the header and
--- a/docs/developing/troubleshooting.md
+++ b/docs/developing/troubleshooting.md
@@ -63,3 +63,27 @@ If your app wants to write to locations other than `/var`, and it's not easy to
 change the app's code, one work-around is to create a symlink from the location
 you app wants to modify to a location under `/var`.  This won't work for all
 applications however.
+
+## ETag data
+
+Sandstorm drops invalid ETag data. This typically has no impact on app functionality. When Sandstorm
+drops invalid ETag data, it stores a warning in the grain log that it did so, limiting the log
+messages to at most one per launch of a grain to avoid filling up the log file. See also [issue
+#2295.](https://github.com/sandstorm-io/sandstorm/issues/2295)
+
+In versions v0.176 and earlier, the grain would crash, rather than log an error message, if an
+invalid ETag header was emitted by the app.  In more-recent versions, Sandstorm drops the header and
+logs a message.
+
+Web apps often use the `ETag` HTTP header to control caching. In Sandstorm, this header is permitted
+so long as it follows the HTTP specification's requirement that the header values are formatted with
+double-quote marks, like: `ETag: "value"` or `ETag: W/"value"`.
+
+Some apps use strings without quotation marks in their ETag headers, such as `ETag: value`, which
+violates the HTTP standard. Sandstorm drops these headers because it is unable to understand the
+app's intent.  The headers are dropped within the grain at the
+[sandstorm-http-bridge](../using/how-it-works.md) layer.
+
+If an ETag header is dropped, typically the worst-case behavior is that web browsers and other HTTP
+clients will refuse to cache a resource and will therefore download the resource more often. This
+can waste bandwidth but typically has no impact on app functionality.
--- a/docs/developing/troubleshooting.md
+++ b/docs/developing/troubleshooting.md
@@ -78,5 +78,5 @@ reduce the effectiveness of the browser's cache.
 
 HTTP headers are processed within the grain at the [sandstorm-http-bridge](../using/how-it-works.md)
 layer. Recent versions of `sandstorm-http-bridge` drop the header and log a message. In versions
-v0.176 and earlier of `sandstorm-http-bridge`, invalid ETag data would trigger an exception,
+v0.177 and earlier of `sandstorm-http-bridge`, invalid ETag data would trigger an exception,
 causing the request to fail.
--- a/docs/developing/troubleshooting.md
+++ b/docs/developing/troubleshooting.md
@@ -66,26 +66,17 @@ applications however.
 
 ## ETag data
 
-Sandstorm normally allows valid `ETag` headers to pass through to the client. However, the HTTP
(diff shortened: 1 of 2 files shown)


kentonv:
Could we design this so that it only prints the warning once per run? I worry that some app will be logging this error on every single request and it will waste space.

paulproteus:
One downside of once per run is that it's less useful for debugging. If you have two places that have invalid ETag data in your app, then you'll only see the warning once per time your app starts, which means you'd never see the second invalid ETag info until you fix the first one.

However, the purpose of this message (to be frank) is not mostly to enable debugging the specific issue; it's mostly to give the developer something to read that indicates to them the reason we deleted their ETag data. So I think I'm OK with printing the warning once per run, and I'll work on that.

paulproteus:
@kentonv requesting your fresh review on this! Thanks for answering my C++ style question earlier today.

paulproteus:
er, needs me to fix one thing. Closing temporarily.

paulproteus:
@kentonv ready for your review & merge.

ocdtrekkie:
Do you suspect I'll be able to update my package after this PR?

paulproteus:
@ocdtrekkie Yup.

docs/developing/troubleshooting.md
```diff
@@ -63,3 +63,27 @@ If your app wants to write to locations other than `/var`, and it's not easy to
 change the app's code, one work-around is to create a symlink from the location
 you app wants to modify to a location under `/var`.  This won't work for all
 applications however.
+
+## ETag data
+
+Sandstorm drops invalid ETag data. This typically has no impact on app functionality. When Sandstorm
```
kentonv: It's surprising that this section would open by talking about an error case. Because of this, especially when skimming, it's easy for people to miss the word "invalid" and think that this line is saying that Sandstorm drops all ETags.

Suggestion:

> Sandstorm normally allows valid `ETag` headers to pass through to the client. However, the HTTP standard requires `ETag` values to take a specific form -- for example, they must be quoted. Many applications incorrectly provide non-quoted values for this header. Sandstorm does not permit invalid header values because ambiguity can lead to security problems. Therefore, Sandstorm will drop invalid `ETag` headers and will write a warning to the grain's debug log. Usually, this does not affect app functionality, other than to reduce the effectiveness of the browser's cache.

paulproteus: +1 to your proposed text

paulproteus:
@kentonv I've added a followup commit w/ your suggested docs change.

docs/developing/troubleshooting.md
```diff
@@ -63,3 +63,27 @@ If your app wants to write to locations other than `/var`, and it's not easy to
...
+drops invalid ETag data, it stores a warning in the grain log that it did so, limiting the log
+messages to at most one per launch of a grain to avoid filling up the log file. See also [issue
+#2295.](https://github.com/sandstorm-io/sandstorm/issues/2295)
+
+In versions v0.176 and earlier, the grain would crash, rather than log an error message, if an
```
kentonv: IMO the text from here down should be deleted and maybe replaced with a link to Wikipedia or the HTTP standard about ETags. Otherwise people who already know what ETags are will spend time reading this text and not get anything useful out of it.

src/sandstorm/sandstorm-http-bridge.c++
```diff
@@ -784,23 +784,41 @@ private:
 #undef ON_EVENT
 
   static void parseETag(kj::StringPtr input, WebSession::ETag::Builder builder) {
+    auto maybePrintInvalidEtagWarning = [](kj::StringPtr input) -> void {
```
kentonv: It's pretty unusual to use a lambda in this way, and it has a (small) runtime cost since you're actually creating a stack-allocated value that holds this lambda. I would instead make this a static method, sibling to `parseETag()`.

docs/developing/troubleshooting.md
paulproteus: My main purpose for this docs section existing is troubleshooting an app. If ocdtrekkie were to visit this page, and didn't really know what ETags are, and came here via a link that was emitted by sandstorm-http-bridge, I want him to have enough information to understand why the app's ETag header is problematic, and have enough info to file a bug against the app to have them fix the app. See also the docs URL that I added to `sandstorm-http-bridge` to provide "information scent" to allow app authors to understand and possibly fix the problem.

Proposed new text for the section:

## ETag data

Web apps often use the `ETag` HTTP response header to control caching. In Sandstorm, this header is
permitted so long as it follows the HTTP specification's requirement that the header values are
formatted with double-quote marks, like: `ETag: "value"` or `ETag: W/"value"`.

Some apps use strings without quotation marks in their ETag headers, such as `ETag: value`, which
violates the HTTP standard.  Sandstorm does not permit invalid header values because ambiguity can
lead to security problems. Therefore, Sandstorm will drop invalid `ETag` headers and will write a
warning to the grain's debug log. Usually, this does not affect app functionality, other than to
reduce the effectiveness of the browser's cache.

The headers are dropped within the grain at the
[sandstorm-http-bridge](../using/how-it-works.md) layer. Recent versions of `sandstorm-http-bridge` drop the header and
log a message. In versions v0.176 and earlier of `sandstorm-http-bridge`, invalid ETag data would cause the grain to crash rather than log an error message.

src/sandstorm/sandstorm-http-bridge.c++
paulproteus: Fair enough. I didn't want to add clutter to the namespace that `parseETag()` is in, but I am OK doing it if that's how you want it!

docs/developing/troubleshooting.md
kentonv: That seems reasonable to me, except I think the last sentence is incorrect: I don't think we ever crashed the grain on invalid ETag headers, we only threw an exception, which caused the request to fail.

paulproteus: I didn't realize that the exception got handled! I think you're right.

paulproteus: I'll also replace

> The headers are dropped within the grain at the

with

> Headers are processed within the grain at the

since I think that explains the paragraph better. I imagine you don't care strongly about that rewording.

paulproteus:
I've extracted the function and I've adjusted the docs.

I also freshly manually-tested this, and confirm that the change causes invalid-ETag HTTP response to stop failing, and that the message only gets logged once.

ocdtrekkie:
Docs bug: 0.177 already exists.

PR #2302, opened 2016-08-01
Makefile: force usage of node/npm from the proper meteor bundle

zarvox:
Invoking "meteor npm" appears to sometimes still have issues with calling
node-gyp with the wrong node version.  In particular, if you have a more
up-to-date meteor-tool on your system, it will attempt to compile against
that tool, even if you have an older .meteor/release file.  I was getting
compile failures for bignum, and node-gyp was trying to build with node 4.
Which seems wrong.

Setting PATH and the precise npm invoked appears to fix this.

Separately: we shouldn't be touching tmp/.shell-env until after we've finished
the work for which that file marks successful completion.


dwrensha:
👍 I currently see node-gyp errors when I build from a new git clone on my Fedora box. This patch eliminates the problem.

zarvox:
Covered by 216f58344ab717d761a545175e1bca6d0e44dbab which landed in #2289

PR #2303, opened 2016-08-02
docs: Correctly state the default MONGO_PORT

paulproteus:
Thanks @dwrensha for noticing!


paulproteus:
100% docs; self-merging

PR #2304, opened 2016-08-02
pseudo-img-tag: prefer contain to cover

zarvox:


paulproteus:
I'm +1 on this. If you're willing to provide before and after screenshots, I'll even hit "merge".

paulproteus:
Upon doing further research, I'm happy to +1 this as-is. I would merge per the temporary rule that merges are OK during Kenton's and Jade's vacation, to be post-commit reviewed when they return, but those rules also require a 24-hour cooling-off period so I'd have to wait until tomorrow.

Issue #2134, opened 2016-06-22
In mobile view, users can't switch between grains


neynah:
I imagine we'll want to make other fixes(#2126, etc.) to the nav menu first but some form of grain switching is important for navigation as well. 

![image](https://cloud.githubusercontent.com/assets/855341/16286186/aba92292-388f-11e6-9166-25267d1bb638.png)

dwrensha:
Fixed by #2311.

zarvox:
Well, mostly.  If you have more things than fit on the screen at once, you still can't scroll the list in the mobile view.

PR #2311, opened 2016-08-02
Allow anonymous users to be offered capabilities.

### LINKED ISSUE #2190: grains should be allowed to offer capabilities to anonymous users

dwrensha:
The two big changes here are:
  1. `MembraneRequirement.permissionsHeld.identityId` is now part of a union, the other variant being `MembraneRequirement.permissionsHeld.tokenId`.
  2. `ApiTokenOwner` has a new variant `clientPowerboxOffer` whose purpose is to allow `SessionContext.offer()` to push a webkey to the client without ever actually writing the webkey to the database (as we currently do in the copy/paste powerbox). In principle, we should be able to avoid the need for this owner variant (probably by using a pseudo-collection), but that seems at least somewhat tricky, and we already have machinery in place for doing it the way proposed here.

Fixes #2190.

This does *not* dedupe webkeys offered to anonymous users, as had been suggested in #2190. Deduping seems less important in this case than in the owned-by-a-logged-in-user case, because:
  1. There are no grain-renaming issues.
  2. The "duplicate" tokens are likely to be pruned early in the permissions computation; they will typically be dead-ends in the sharing graph and easy to prove to be irrelevant to most computations.

### DIFF
--- a/shell/server/hack-session.js
+++ b/shell/server/hack-session.js
@@ -80,13 +80,15 @@ SessionContextImpl = class SessionContextImpl {
 
   offer(cap, requiredPermissions, descriptor, displayInfo) {
     return inMeteor(() => {
-      if (!this.identityId) {
-        // TODO(soon): allow non logged in users?
-        throw new Meteor.Error(400, "Only logged in users can offer capabilities.");
+
+      const session = Sessions.findOne({ _id: this.sessionId });
+
+      if (!session.identityId && !session.hashedToken) {
+        throw new Meteor.Error(400, "");
       }
 
       const castedCap = cap.castAs(SystemPersistent);
-      let apiTokenOwner = { webkey: null };
+      let apiTokenOwner = { clientPowerboxOffer: { sessionId: this.sessionId, }, };
       const isUiView = descriptor && descriptor.tags && descriptor.tags.length === 1 &&
           descriptor.tags[0] && descriptor.tags[0].id &&
           descriptor.tags[0].id === Grain.UiView.typeId;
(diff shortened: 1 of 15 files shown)


dwrensha:
The CSS change in https://github.com/sandstorm-io/sandstorm/pull/2311/commits/9f10153ef0f4b41fd87f57565bd77c3d1ccfa4c6 also fixes a problem where open grains were not visible in the mobile view of the navbar. (This problem is present on master.)

dwrensha:
The mobile navbar issue is https://github.com/sandstorm-io/sandstorm/issues/2134.

shell/server/hack-session.js
```diff
@@ -80,13 +80,15 @@ SessionContextImpl = class SessionContextImpl {
...
   offer(cap, requiredPermissions, descriptor, displayInfo) {
     return inMeteor(() => {
-      if (!this.identityId) {
-        // TODO(soon): allow non logged in users?
-        throw new Meteor.Error(400, "Only logged in users can offer capabilities.");
+
+      const session = Sessions.findOne({ _id: this.sessionId });
+
+      if (!session.identityId && !session.hashedToken) {
+        throw new Meteor.Error(400, "");
```
kentonv: error message?

kentonv:
Good idea with `clientPowerboxOffer`.

I'm "not there yet" on skipping deduplification of the webkeys. It seems that this means every time an anonymous user clicks on a collection item, we add a new permanent database entry. It seems like that could lead to a huge amount of bloat if, say, someone with a lot of followers tweeted a collection. While it's true that the permissions algorithm would largely ignore the new leaves, we are likely to have other kinds of operations we do on ApiTokens which may have a harder time.

Couldn't `acceptPowerboxOffer()` take the collection's token as a third argument and use it to deterministically compute a child token?

kentonv:
(Hmm, I think you'd also want to mix in material from the grain-owned SturdyRef. Imagine a "password-protected" collection where the collection doesn't reveal its contents to anonymous users until they type in the correct password. You woludn't want someone to be able to derive the inner tokens without knowing the password. But this should be easy: Take some sort of a hash of the grain-owned SturdyRef and store it as `clientPowerboxOffer.secretMaterial`.)

dwrensha:
> Good idea with clientPowerboxOffer.

It was @jparyani's suggestion. :)

dwrensha:
> Couldn't acceptPowerboxOffer() take the collection's token as a third argument and use it to deterministically compute a child token?

Done.

>  I think you'd also want to mix in material from the grain-owned SturdyRef.

Ideally, yes. But I don't see a way to get the raw unhashed sturdyref data to the `acceptPowerboxOffer` method body without writing it to the database, unless we go ahead and figure out a way to avoid the need for `clientPowerboxOffer`. So I've opted to use the ID of the parent token. Please see my comment in the body of `acceptPowerOffer()`.

shell/server/hack-session.js
dwrensha: got it

dwrensha:
Garply, retest this please

PR #2312, opened 2016-08-02
defineMutationMethods: Meteor.isClient for all Mongo collections

dwrensha:
The [`defineMutationMethods` option](http://docs.meteor.com/api/collections.html#Mongo-Collection) was introduced in https://github.com/meteor/meteor/commit/2ddf14575a8b45a2899e6e6c5760fc0b8f6b6828 . Since all of our client-initiated database writes go through Meteor methods, we can take advantage of it.

### DIFF
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -144,7 +144,7 @@ Meteor.users.ensureIndexOnServer("suspended.willDelete", { sparse: 1 });
 // TODO(cleanup): This index is obsolete; delete it.
 Meteor.users.ensureIndexOnServer("identities.id", { unique: 1, sparse: 1 });
 
-Packages = new Mongo.Collection("packages");
+Packages = new Mongo.Collection("packages", { defineMutationMethods: false });
 // Packages which are installed or downloading.
 //
 // Each contains:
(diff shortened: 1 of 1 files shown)


dwrensha:
Oh, right. Some Meteor methods depend on these mutation methods being defined for clientside simulation...

kentonv:
Odd. In theory whether or not client-side simulation of mutation methods is available should be independent of whether or not the mutation methods are exported over DDP. But I guess they tied these things together?

paulproteus:
We could choose to only set it to `false` server-side.

kentonv:
@paulproteus Brilliant!

dwrensha:
This might actually be a bug in Meteor. There error that you get when you try to call e.g. `Grains.insert({})` is something like "failed to call method undefinedinsert", which seems fishy to me.

kentonv:
@dwrensha Does it work to pass the flag only on the server side? I like the idea of not having these endpoints exposed at all, since they will never work anyway.

dwrensha:
Yeah, it seems to work if we set the flag to false only on the server side. I've pushed a commit making that change.

dwrensha:
Garply, retest this please.

PR #2314, opened 2016-08-02
Display "Loading..." instead of "(incognito)" on first page load.

### LINKED ISSUE #2275: Flash of erroneous "incognito" login state
When I visit a Sandstorm page, the account info in the upper right starts out saying "Sign in", then goes to "(incognito)", then my actual profile name.

There are two problems here. The one that I'm not actually filing this issue about is that I don't know what "incognito" means here: the term is not mentioned on any account-related page I've seen, or in the documentation.

The other one is that the fact that it's a brief flash suggests that the transition from "(incognito)" to the proper name indicates some kind of network request is occurring. Therefore, if the network is slow or disconnected at an unfortunate time, the user interface could _erroneously_ display "(incognito)" (whatever that means) for a significant amount of time, confusing the user.

Please make sure that the UI always displays either correct information or obviously partial/missing/'loading…' information.

dwrensha:
Fixes #2275.

### DIFF
--- a/shell/client/accounts/login-buttons.js
+++ b/shell/client/accounts/login-buttons.js
@@ -65,7 +65,7 @@ function getActiveIdentityId(grains) {
 
 Template.accountButtons.helpers({
   displayName: function () {
-    if (Meteor.user() && !Meteor.user().loginIdentities) {
+    if (Meteor.user() && !Meteor.user().loginIdentities && !Meteor.user().profile) {
       // Need to wait for resume token to complete login. For some reason, `Meteor.loggingIn()`
       // is still false in this case.
       return "Loading...";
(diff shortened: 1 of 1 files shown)


shell/client/accounts/login-buttons.js
```diff
@@ -65,8 +65,19 @@ function getActiveIdentityId(grains) {
 
 Template.accountButtons.helpers({
   displayName: function () {
+    if (Meteor.user() && !Meteor.user().loginIdentities) {
```
zarvox: Would be nice if we had an explicit `.ready()` on a subscription handle here.

I imagine this will show "Loading..." during the login-with-identity interstitial if you login as a non-login identity?

dwrensha: I guess we could try to check the whether the main `identityProfile` subscription is ready? Seems somewhat tricky due to the [reactivity logic](https://github.com/sandstorm-io/sandstorm/blob/v0.173/shell/client/shell-client.js#L34-L53).

I've updated this so that it won't immediately return "loading..." in the case that the user is an identity user. Note, however, that when we display the login-with-identity interstitial, the login button is hidden.

PR #2317, opened 2016-08-02
Fix PYTHONPATH to use Meteor's bundled node-gyp modules in shell build

zarvox:
Meteor doesn't store anything in $HOME/.meteor/tools/ any more, so there
is no way the existing code here is correct.  This change makes the Makefile
do what run-dev.sh does.


kentonv:
I think this was papering over a bug in Meteor and/or npm that I also reported upstream at the time (like two years ago). It's possible the env var isn't needed at all anymore and we should just remove it.

zarvox:
Okay, closing in favor of #2329

PR #2319, opened 2016-08-03
drop unneeded es6-promise polyfill

dwrensha:
Now that Meteor understand ES6, we don't need to include this es6-promises polyfill on the client side.

We also have a bunch of instances of `Npm.require("es6-promise")` on the server side. I get weird errors if I try to remove those, probably because node-capnp depends on the polyfill. Maybe once we upgrade to Meteor 1.4 we'll be able to use the native `Promise` in those cases as well.


zarvox:
:shipit:

PR #2320, opened 2016-08-03
Prevent error that could happen on reloading account-settings page.

dwrensha:
Currently, if I visit account-settings and then refresh the page, my debug console display a bunch of errors like: `Exception in template helper: TypeError: Cannot read property 'privateIntrinsicName' of undefined`. This patch prevents those errors.

### DIFF
--- a/shell/client/accounts/account-settings.js
+++ b/shell/client/accounts/account-settings.js
@@ -59,15 +59,15 @@ const helpers = {
   },
 
   identities: function () {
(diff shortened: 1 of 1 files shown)


kentonv:
What actually changed here? The code looks functionally equivalent AFAICT.

dwrensha:
` .filter((identity) => !!identity)` removes nulls.

kentonv:
Oh right, I was thinking the `if (identity)` also accomplished that, but I suppose if it's all inside a `.map()` then the result will have `undefined` elements.

PR #2323, opened 2016-08-05
Numerous install.sh fixes

paulproteus:
See the commits for details.

Note that on my testing environment, the test on the precise64 base box fails, but I think my testing environment is weird.


paulproteus:
Self-merging because 100% install.sh & installer-test & docs.

paulproteus:
Seemingly this broke the installer tests:

https://build.sandstorm.io/job/Sandstorm-installer/

I'll look into it tomorrow, which is enough time before our typical release cycle (weekends).

PR #2325, opened 2016-08-05
installer-tests: Look for a more reliable string

paulproteus:
I wonder if inconsistent timing of reads from buffers is ruining my day. Who knows.


paulproteus:

PR #2326, opened 2016-08-05
Add single-quote marks to get identicons to work in powerbox cards.

dwrensha:
If these quotes are not present, then our data-url svgs fail to render. Other similar usages of `background-image` in our code also use single quotes, e.g. https://github.com/sandstorm-io/sandstorm/blob/7ead9c71c9bbd2932df6b1ba31f3397d96a571bf/shell/client/grain/grainlist.html#L150


zarvox:
:shipit:

paulproteus:
Garply, retest this please

PR #2331, opened 2016-08-05
Fix packaging tutorial where we replaced actions.title with nounPhrase

zarvox:
@paulproteus docs-only change

Fixes sandstorm-io/vagrant-spk#183


paulproteus:
Yippee!!!

PR #2332, opened 2016-08-05
use a locale category that works on glibc 2.24

dwrensha:


dwrensha:
In glibc 2.24, apparently only three values will be accepted for the category: https://github.com/bminor/glibc/blob/900f59f084bfe35cb389bbe0dc464413a1a38e90/locale/programs/ld-identification.c#L187-L189

"i18n:2012" is the newest, so let's use that one.

Note that I don't actually know what this stuff means, but this change fixes the build for me on my Arch box (glibc 2.24), and it does not break the build on my Fedora box (glibc 2.23).

kentonv:
Do you happen to have a Debian box handy? If not I'll test here, but not until next week probably.

zarvox:
If it passes on the build server, that's Debian.

PR #2341, opened 2016-08-08
Temporarily restore SandstormEmail to global scope to fix Blackrock invoices.

kentonv:
We can come up with a cleaner fix later.


zarvox:
:shipit:

(looks like `blackrock-payments` never declared its dependency on `sandstorm-email`)

PR #2342, opened 2016-08-08
Make sure the referral program notification's 'dismiss' button does not send the user to the referral program page.

kentonv:


zarvox:
:shipit:

PR #2343, opened 2016-08-08
Close the powerbox popup when an error occurs.

kentonv:


zarvox:
Hmmm, depending on what kind of failure this is, it might make more sense for the powerbox popup to display the error.  Otherwise, it seems likely that the user might lose whatever powerbox-selection context they had produced.

Are there situations where failures here are something the app can do something intelligent about, or is Sandstorm in a better position to display/message the failure, since it's likely a Sandstorm bug causing the failure case?  I'd rather only dismiss the dialog if passing the failure on to the app is more likely to produce a better user experience, which I currently believe is unlikely to be the case.

kentonv:
Well, currently the error is not displayed by the powerbox at all. I ran into the problem when writing the payments API that if I dismissed the Stripe checkout.js popup, the powerbox didn't close. Either this method needs to close the powerbox, or I need to add a new method for specifically that purpose. I'm not sure what other kinds of errors are possible right now, making it hard for me to say what exactly the behavior should be.

zarvox:
Ahh.  Sorry about errors-not-being-displayed, that was probably annoying.  https://github.com/sandstorm-io/sandstorm/pull/2348 should fix that.

Arguably, closing the Stripe checkout.js popup only indicates your desire to pop that last piece of the action stack - adding a card.  It does not necessarily mean you've given up on completing that powerbox request, and perhaps you should still be able to click the "Add new card" button or any of your existing cards still.

But if you think that cancelling the Stripe checkout.js popup *really* should cancel the whole powerbox request, regardless of other actions you took on the way there, then here's my proposal:

Let's add a new method `cancelRequest` to `SandstormPowerboxRequest` which calls `this.finalize()` and then `this._requestInfo.onCompleted()`.  When the Stripe checkout.js popup is dismissed without completing the flow, call `this.data.powerboxRequest.cancelRequest()` rather than `failRequest()`.  Continue to call `failRequest()` in the branch where the Meteor method call to add the card fails.

The Sandstorm-side half of that is implemented in https://github.com/sandstorm-io/sandstorm/pull/2349 ; if you want the dismiss-request behavior, consider merging it.  If I've convinced you that popping the interaction stack is sufficient, feel free to close it instead. :)

kentonv:
I merged #2349 before reading your comment here. I think popping the interaction stack makes sense, and we could get there by changing the behavior of `cancelRequest()` to pop the stack instead of close the window altogether.

In any case, I'm closing this PR since it is superseded by yours.

PR #2358, opened 2016-08-09
add some docs about developing an app using the powerbox

dwrensha:
Adding these docs because the blog post https://github.com/sandstorm-io/sandstorm-website/pull/212 want to link to them.

### DIFF
--- a/docs/developing/powerbox.md
+++ b/docs/developing/powerbox.md
@@ -1,0 +1,30 @@
+Grains begin life completely isolated from the outside world.
+To gain access to external capabilities, they need to go through the *powerbox*,
+which allows users to mediate and audit any connections that are made.
+
+The definitive reference for the powerbox's interfaces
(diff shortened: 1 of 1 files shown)


paulproteus:
This doc looks great for a start; later on, we can enhance the doc as-needed.

To show up in the sidebar you see at docs.sandstorm.io, it must be added to `mkdocs.yml`. I can do that in a follow-up commit if you like.

I would link to the the code of the collections app in a way where the URL isn't visible to the user (e.g. "A collection [makes a powerbox request for a UiView capability](http://example.com)"). If you're +1 on that, you can make the change, or you can ask me to in a follow-up commit.

Thanks so much for writing this! Let me know if you're OK making the above changes, or you're bored of the `docs/` system and want me to do it (which is OK), or you have other feedback.

dwrensha:
I am +1 on those changes, and would prefer that you do them because running mkdocs is not super convenient for me.

paulproteus:
+1

PR #2359, opened 2016-08-09
Powerbox doc 

paulproteus:
This supercedes #2358


paulproteus:

PR #2361, opened 2016-08-10
Pin bundled mongo to 2.6 to avoid in-place upgrade woes

zarvox:
sandstorm-84 and earlier shipped mongo 2.4, which creates its auth schema in a
different way than mongo 2.6 and above.  While mongo 2.6 is able to read said
schema, mongo 3.2 is not.

To avoid breaking servers that were originally installed with mongo 2.4, we
can't jump to mongo 3.2 just yet.  Mongo 2.6 is still fine, and Meteor remains
backwards-compatible with Mongo 2.6, so we can continue shipping this for a while.

This change makes make-bundle.sh cache a copy of sandstorm-171 (the latest
release as of this writing) and copy its mongo/mongod into the new bundle,
rather than pulling those binaries from the current Meteor dev bundle.

Additional notes on mongo upgrades:

* https://jira.mongodb.org/browse/SERVER-22004
* https://docs.mongodb.com/manual/release-notes/2.6-upgrade-authorization/
* https://docs.mongodb.com/manual/release-notes/3.2-upgrade/

### DIFF
--- a/hack/.gitignore
+++ b/hack/.gitignore
@@ -1,0 +1,1 @@
+sandstorm-171.tar.xz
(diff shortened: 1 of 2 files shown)


dwrensha:
I tried this out on my server. It appears to work.

PR #2362, opened 2016-08-10
Reimplement email support atop nodemailer

### LINKED ISSUE #2350: simplesmtp can corrupt messages, occasionally breaking links and email delivery
`simplesmtp` is responsible for taking a `mailcomposer` object and sending the email message it represents to the server.  To achieve this, it needs to wrap lines that exceed 76 columns, and to make sure that lines that start with a `.` are escaped (since a `.` on a line of its own indicates end-of-message).

It fails to do the second part correctly, causing the `.` to be swallowed, which (if you're unlucky) can mean several different failure modes, depending on the length and contents of the personal message you wrote:

1. Messages can be truncated and users may receive message fragments. 
2. Messages can lose any number of `.` in the message, including the one in the link back to the grain, making links break.  For instance, I received a link to https://oasis.sandstormio/shared/..., which is missing the . between "sandstorm" and "io".  sandstormio is not a valid TLD.

Options appear to be:

1. Upgrade to node 4 and use `smtp-connection`, which is actively maintained.  This involves fixing the `node-capnp` crashes and the usual amount of cautious testing involved when doing major upgrades.
2. Fork/patch simplesmtp now despite wanting to eventually do 1.
3. Do nothing.

@jparyani feel free to correct anything above!

zarvox:
`simplesmtp` is unmaintained and has major problems.  We are dropping it.

`nodemailer` (by the same author) is a collection of packages which together
form the most widely used JS mail library.  It provides a number of transports,
including speaking SMTP directly, to actually deliver mail.  It's possible
we'll want to support transports other than just plain SMTP; this change paves
the way to do so neatly.  We use the `nodemailer-smtp-pool` transport to
deliver mail, which is comparable to the `simplesmtp` pool.

`mailcomposer` is an implementation detail of nodemailer, so we no longer need
to pull it in explicitly.  Mail is expressed as simply a plain JS object.

`mimelib` is no longer used, so I have removed it from our package.json.

`smtp-server` is a package which replaces simplesmtp's server support.  It was
pretty simple to drop in.  We configure it to not require authentication,
because we can't require auth for receiving email.

`mailparser` (used by the email driver to parse inbound email) has changed its
API a bit to separate envelope session information from message data, so we
change how we process mail in the inbound mail driver a bit.

I have manually tested:

* Email self-test in setup/admin panel works as expected
* Email sent by Roundcube gets delivered correctly (modulo the extra trailing .
  that the sandstorm-smtp-bridge inside the app package injects for text/plain mail)
* Email sent to grain gets delivered to Roundcube.

And this should pass the testsuite.

Fixes #2350.


shell/server/drivers/mail.js
```diff
@@ -53,132 +52,142 @@ if (!Meteor.settings.replicaNumber) {  // only first replica
...
+  const server = new SMTPServer({
+    banner: "Sandstorm Mail Server",
+    authOptional: true,
+    disabledCommands: ["STARTTLS", "AUTH"],
+    onData: (req, session, callback) => {
+      const mailparser = new MailParser();
+      req.pipe(mailparser);
+
+      mailparser.on("end", (mail) => {
+        console.log(mail);
```
jparyani: Stray console.log

kentonv:
@jparyani please review.

jparyani:
LGTM

PR #2371, opened 2016-08-11
Save inotify-derived grain sizes to database

kentonv:
After this change:

* Storage quota can be enforced on single-machine Sandstorm. (However, the UI needs to be detangled from the Oasis payments code -- @jparyani is working on that.)
* The size of each grain is shown in the grain list. (However, preexisting grains won't show a size until they are next opened.)

### DIFF
--- a/shell/client/grain/grainlist-client.js
+++ b/shell/client/grain/grainlist-client.js
@@ -23,6 +23,7 @@ SandstormGrainListPage.mapGrainsToTemplateObject = function (grains, db) {
       isOwnedByMe: true,
       trashed: grain.trashed,
       unread: !grain.ownerSeenAllActivity,
+      size: grain.size,
     };
   });
 };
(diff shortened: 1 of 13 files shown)


ocdtrekkie:
> The size of each grain is shown in the grain list. (However, preexisting grains won't show a size until they are next opened.)

I am super excited about this.

shell/server/core.js
```diff
@@ -137,6 +137,31 @@ class SandstormCoreImpl {
       logActivity(this.grainId, null, event);
     });
   }
+
+  reportGrainSize(bytes) {
+    bytes = parseInt(bytes);  // int64s are stringified but precision isn't critical here
```
zarvox: nit: please don't reassign method arguments.  Bind a new name, so that each variable binding always has exactly one type (and where possible, exactly one value).

(I should get ESLint up and running some time Soon™ now that we have node 4...)

shell/client/grain/grainlist-client.js
```diff
@@ -463,6 +464,12 @@ Template.sandstormGrainTable.helpers({
     return !!(Template.instance().data.alwaysShowTableHeaders ||
               Template.instance().data.grains.length);
   },
+
+  grainSize: function () {
+    // TODO(cleanup): extract prettySize and other similar helpers from globals into a package
+    // TODO(cleanup): access Meteor.user() through db object
```
zarvox: This second TODO seems to be unrelated to the code here

PR #2372, opened 2016-08-12
Backfill grain sizes

kentonv:
Also improve the grain size computation slightly.

### DIFF
--- a/shell/packages/sandstorm-db/migrations.js
+++ b/shell/packages/sandstorm-db/migrations.js
@@ -672,6 +672,31 @@ function setNewServer() {
   }
 }
 
+function backgroundFillInGrainSizes() {
+  // Fill in sizes for all grains that don't have them. Since computing a grain size requires a
+  // directory walk, we don't want to do them all at once. Instead, we compute one a second until
+  // all grains have sizes.
+
+  try {
+    const grain = Grains.findOne({ size: { $exists: false } }, { fields: { _id: 1, userId: 1 } });
+
+    if (grain) {
+      // Compute size!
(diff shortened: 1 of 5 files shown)


kentonv:
@jparyani Added a commit to improve robustness and remove the debug logging.

shell/packages/sandstorm-db/migrations.js
```diff
@@ -672,6 +672,43 @@ function setNewServer() {
   }
 }
 
+function backgroundFillInGrainSizes() {
```
zarvox: Can you move this out to `server/00-startup.js` or somewhere similar?  The fact that it reaches into `globalBackend` makes me believe that it doesn't belong in the `sandstorm-db` module.

shell/packages/sandstorm-db/migrations.js
```diff
@@ -784,6 +821,9 @@ function migrateToLatest() {
 
       Migrations.update({ _id: "new_server_migrations_applied" }, { $set: { value: true } });
     }
+
+    // Start background migrations.
+    backgroundFillInGrainSizes();
```
zarvox: Ditto - how about this goes in the `Meteor.startup()` block in `server/00-startup.js`?

jparyani:
LGTM, although I'm +0 on @zarvox suggestion to move the size updating code to startup.js instead of migrations.js.

shell/packages/sandstorm-db/migrations.js
kentonv: We have other migrations that use `globalBackend`. I felt the code made the most sense here because it is a migration. Arguably migrations.js should be moved somewhere where it can more easily depend on other APIs since it commonly needs them.

PR #2374, opened 2016-08-12
Desktop notifications for activity events

zarvox:


zarvox:
Tested with Rocket.Chat and Etherpad.  Etherpad does not provide a `caption` for comments (which are as far as I can tell the only way to trigger getting notified in Etherpad?).

@kentonv can you make Etherpad provide the comment text in the caption, or add a body field, or whatever would be appropriate here?

PR #2376, opened 2016-08-12
Fix Sandcats TLS setup on Node 4.

kentonv:
Sandcats servers now get an A from Qualys.

screenshot from 2016-08-11 20-57-27

Previously they got an A-, due to lack of forward secrecy. Now that Meteor supports modern Node versions and we've upgraded, FS is supported.

Qualys still dings us for requiring SNI and for not supporting forward secrecy on older browsers that lack ECDH. We could support classic DH, but it requires generating dhparams, which is kind of a pain.


paulproteus:
+1'd in person

PR #2379, opened 2016-08-12
Add customizable quota management

jparyani:
There is a new route, `/admin/hosting-management`, that is not linked to from the admin panel that allows configuring this.

It's a bit convoluted at the moment, but you must have both SAML and LDAP identity providers configured (although LDAP can be disabled as long as it has working settings). The email from your SAML identity will be used to query the LDAP directory, and fetch a field that provides the user's quota (in bytes).


shell/packages/sandstorm-db/db.js
```diff
@@ -1797,6 +1789,50 @@ _.extend(SandstormDb.prototype, {
     return setting && setting.value && this.isFeatureKeyValid();
   },
 
+  updateUserQuotaFromLdap: function (user) {
```
kentonv: I don't think this method belongs in `db.js`. It's specific to LDAP, it depends on `LDAP` being present, etc.

It seems like you're putting it here because `getUserQuota()` needs to call it, but do we really want to be pinging the LDAP server synchronously every time we need to look up a user's quota? If the LDAP server goes down temporarily, Sandstorm would be essentially inoperable.

shell/packages/sandstorm-db/db.js
```diff
@@ -1797,6 +1789,50 @@ _.extend(SandstormDb.prototype, {
...
+
+    const newStorageQuota = +ldapUser.searchResults[setting.value];
+    Meteor.users.update({ _id: user._id }, { $set: { cachedStorageQuota: newStorageQuota } });
+    // TODO(someday): cache timestamp as well and only check/update if greater than 60s ago
+    return newStorageQuota || fallback;
+  },
+
+  getUserQuota: function (user) {
+    if (this.isReferralEnabled()) {
+      // TODO(someday): isReferralEnabled isn't quite the right check
```
kentonv: I think what you want to be checking is "is LDAP quota enabled?", not "is Oasis?"

shell/server/accounts/ldap/ldap-server.js
```diff
@@ -23,7 +23,7 @@ LDAP_DEFAULTS = {
  @class LDAP
  @constructor
  */
-let LDAP = function () {
+LDAP = function () {
```
kentonv: Please find a way to design this code such that we don't make db.js depend on this global.

jparyani: I'm open to suggestions on what to do then. We've tried to move away from using packages, which is the only other reasonable way I know how to do this.

shell/client/admin/hosting-management.html
```diff
@@ -0,0 +1,67 @@
...
+    {{/focusingErrorBox}}
+  {{/if}}
+  <form class="admin-hosting-management-form {{#if formDisabled}}disabled{{/if}}">
+    <div class="form-group">
+      <label>
+        <input type="checkbox" name="quotaEnabled" checked="{{quotaEnabled}}" />
+        Enable quota for users
+      </label>
+      <span class="form-subtext">
+        Updates to apps will be automatically downloaded by the Sandstorm server daily and
```
kentonv: This text seems misplaced.

shell/imports/server/accounts/ldap.js
```diff
@@ -0,0 +1,240 @@
...
+    ldapUser = this.ldapCheck(db, { searchUsername: email, });
+  } catch (err) {
+    console.error("Error looking up quota from LDAP");
+    return fallback;
+  }
+
+  if (!ldapUser || !ldapUser.searchResults) return fallback;
+
+  const newStorageQuota = +ldapUser.searchResults[setting.value];
+  Meteor.users.update({ _id: user._id }, { $set: { cachedStorageQuota: newStorageQuota } });
```
kentonv: Skip this write if the value hasn't changed?

shell/imports/server/accounts/ldap.js
```diff
@@ -0,0 +1,240 @@
...
+    return ldapAsyncFut.wait();
+
+  } else {
+    throw new Meteor.Error(403, "Missing LDAP Auth Parameter");
+  }
+
+};
+
+LDAP.prototype.updateUserQuota = function (db, user) {
+  const fallback = user.cachedStorageQuota || 0;
```
kentonv: The fallback needs to be an object of {storage, grains, compute} to match the non-fallback return value, right?

shell/client/admin/hosting-management.html
```diff
@@ -35,8 +35,8 @@
         Enable quota for users
       </label>
       <span class="form-subtext">
-        Updates to apps will be automatically downloaded by the Sandstorm server daily and
-        notifications sent to users asking if they want to update their installed apps.
+        Enable Quota management on this server. This requires both SAML and LDAP to be configured, and email addresses will be used to match between the two.
+        </address>
```
kentonv: What's this tag?

PR #2380, opened 2016-08-12
Call Notification.requestPermission() to make notifications work

zarvox:
I had tested with a browser that had cached the permission grant, and incognito windows blanket-deny desktop notifications entirely, and I mistakenly thought that creating the first `new Notification()` would trigger the browser to prompt you to show notifications.  Oops.

This change requests permissions to show notifications the first time Sandstorm would want to show a desktop notification.  This obviously isn't perfect, but this at least avoids the problem of "user is asked for permission when they don't have the context to understand why that permission is being asked for" and makes notifications available to users that would clearly benefit from them.  I am very strongly against dropping a permission request on users first thing when they e.g. log in or refresh their browser window.

With this PR, you'll see the bell notification go red and increment the number at the same time your browser asks for permission to show notifications, so hopefully users will connect the two and make the right choice.

I'm open to alternate/additional design proposals, but I want to close the feature gap this release if at all possible, especially now that Rocket.Chat is exclusively using the activity events API.

### DIFF
--- a/shell/client/desktop-notifications-client.js
+++ b/shell/client/desktop-notifications-client.js
@@ -173,7 +173,11 @@ const showActivityDesktopNotification = (notif) => {
 };
 
 Template.desktopNotifications.onCreated(function () {
+  // Don't bother with any of this if we don't have both Notification support and localStorage.
   if (!localStorageWorks) return;
+  if (!window.Notification) return;
+  this.enabled = true;
+
   // There's some tricky logic here to try to make sure notifications are preferentially handled by
   // a tab that already has the associated grain open, rather than just by the first or last tab to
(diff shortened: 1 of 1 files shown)


kentonv:
Can you also check that the notification API exists (`"Notification" in window`) before trying to use it?

kpreid:
> With this PR, you'll see the bell notification go red and increment the number at the same time your browser asks for permission to show notifications, so hopefully users will connect the two and make the right choice.

This will still pop up a semi-modal thing that is unrelated to _what the user's doing right that second_. Proposed alternative: Put a thing _in the notification menu_ for “Show desktop notifications” that if they click it requests permission.

zarvox:
Yeah, that sounds imminently reasonable.

PR #2383, opened 2016-08-12
Implement `sandstorm uninstall`

kentonv:
Closes #2291

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -349,62 +349,18 @@ virtualization generally, please email support@sandstorm.io.
 
 ## Uninstall
 
+To uninstall Sandstorm, run:
+
+    sandstorm uninstall
+
+This will remove all files installed by Sandstorm, but will not delete your user data. If you
+wish to remove user data as well, do:
+
+    sandstorm uninstall --delete-user-data
+
 If you installed Sandstorm with default options, the following actions will fully remove
 Sandstorm. If you customized the install, you'll need to change these commands accordingly.
 
 If you want to _change settings_, you can edit `/opt/sandstorm/sandstorm.conf`.
-
-- Stop the service.
-
-```bash
-sudo sandstorm stop
-sudo service sandstorm stop
-```
-
-- Remove the service file(s).
-
-For systemd:
-
-```bash
-sudo systemctl disable sandstorm.service
-sudo rm -f /etc/systemd/system/sandstorm.service
-sudo systemctl daemon-reload
-```
-
-For sysvinit:
-
-```bash
-sudo update-rc.d sandstorm remove
-sudo rm -f /etc/init.d/sandstorm
-```
-
-- Remove the two symlinks Sandstorm creates.
-
-```bash
-sudo rm -f /usr/local/bin/spk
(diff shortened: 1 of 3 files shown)


paulproteus:
I almost love this!

**Feedback that is important to me that you address**

- Part of the reason the docs explain how to uninstall Sandstorm is to indicate to people where Sandstorm puts its files. It'd be nice if we could retain the information about that, so that people have a reference to look at when we say that Sandstorm is very tidy.

**Feedback that is lower-priority**

- The install process creates a systemd symlink. On my Debian system, that is: `/etc/systemd/system/multi-user.target.wants/sandstorm.service` Presumably smart to remove that too. @zarvox is that typical and/or are there other places we should look for a symlink to ourselves? Alternatively (smarter) `sandstorm uninstall` could run `sudo systemd disable sandstorm` or something like that.

- I find italics hard to read in our docs, so I recommend using bold, like this:

> To **change settings**

- or not using italics/bold for emphasis in this case.

- I wonder if this would suit the needs of people who need to "just transfer the Sandstorm data to a new server", but I'm just thinking aloud at this point.

The diff seems generally exceedingly reasonable.


**Other notes**

I can't seem to quickly find any info about `systemd disable`:

```bash
paulproteus@slittingmill:~$ systemd disable -h
Excess arguments.
paulproteus@slittingmill:~$ systemd disable --help
Excess arguments.
paulproteus@slittingmill:~$ man systemd-disable
No manual entry for systemd-disable
```

zarvox:
> I can't seem to quickly find any info about systemd disable:

It's `systemctl disable`, not `systemd disable`.

zarvox:
> The install process creates a systemd symlink. On my Debian system, that is: /etc/systemd/system/multi-user.target.wants/sandstorm.service Presumably smart to remove that too. @zarvox is that typical and/or are there other places we should look for a symlink to ourselves? Alternatively (smarter) sandstorm uninstall could run sudo systemd disable sandstorm or something like that.

Running `systemctl disable sandstorm.service` removes this symlink, and this code does so [here](https://github.com/sandstorm-io/sandstorm/pull/2383/files#diff-3d39d92daed96aa7386c4f61fe8782dbR1019)

paulproteus:
Ignore my remark above about `systemd disable`!

kentonv:
> I find italics hard to read in our docs, so I recommend using bold, like this:

I did not write that line -- it was there and emphasized that way before this change.

> Part of the reason the docs explain how to uninstall Sandstorm is to indicate to people where Sandstorm puts its files. It'd be nice if we could retain the information about that, so that people have a reference to look at when we say that Sandstorm is very tidy.

I am not sure I understand what you want here. To avoid "wrong rock syndrome", would you like to take over the docs changes? I could merge just the first two commits for now.

paulproteus:
I'm +1 on you merging this as-is, and me eventually making the changes I suggest, then!

TIL http://workingsmarter.typepad.com/my_weblog/2008/12/bring-me-a-different-rock.html , pretty cool concept.

paulproteus:
That is - you should merge your docs change as-is, too, and I can make add-on commits as I feel is important, or not.

PR #2384, opened 2016-08-13
Add Collections app/System suite to preinstalled apps

jparyani:


shell/client/admin/preinstalled-apps.html
```diff
@@ -36,7 +36,7 @@
       </div>
     </div>
 
-    <!-- <div class="form-group">
```
paulproteus: +1

PR #2386, opened 2016-08-15
docs: Use vagrant-spk vm provision

paulproteus:
In some places, we used to write:

```bash
cd .sandstorm
vagrant provision
cd ..
```

But this is no longer needed due to https://github.com/sandstorm-io/vagrant-spk/pull/145
by @zarvox. Therefore, we say this instead:

```bash
vagrant-spk vm provision
```


paulproteus:

PR #2388, opened 2016-08-15
Fix grain title in desktop notifications on shared grains (sometimes?)

zarvox:
The `.title` field canonically lives on `ApiTokenOwner`, not on
`.denormalizedGrainMetadata.title`.  Ditto `.upstreamTitle`.

While debugging this with @jparyani, we discovered that sometimes you can wind up with a stub object in your `Grains` collection with only the fields `_id` and `title`, even if you don't own that grain.  This causes the grain lookup logic to get confused and think that you own the grain, which breaks my expectations.

### DIFF
--- a/shell/client/desktop-notifications-client.js
+++ b/shell/client/desktop-notifications-client.js
@@ -14,6 +14,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import { computeTitleFromTokenOwnerUser } from "/imports/client/model-helpers.js";
+
 // Test if localStorage is usable.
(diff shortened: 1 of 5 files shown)


dwrensha:
> we discovered that sometimes you can wind up with a stub object in your Grains collection with only the fields _id and title,

To be clear: you mean just on the client side, right? There aren't broken database rows on the server, are there?

zarvox:
Yes, just on the client side as far as I know (though to be honest, I haven't inspected the server because we're observing this on Oasis)

zarvox:
Added a commit that also fixes the Grains collection having `{_id, title}` entries and DRYs up title computation.

PR #2389, opened 2016-08-15
docs: Explain not 1.4-compatible yet

paulproteus:


paulproteus:

PR #2390, opened 2016-08-16
Don't check that mail server certs are signed by any CA in particular

zarvox:
This is restoring the behavior that existed before #2362.  Several users have
reported that the most recent upgrade has broken mail delivery on their systems
due to attempting to perform STARTTLS, then aborting when the cert provided by
the mailserver didn't validate.  This appears to be a fairly common configuration.

Node 0.10 did not validate certs at all before now, so I'm disabling cert
validation to undo the breakage.  Longer-term, we should add UI for admins to
specify one of the following behaviors:

* "none" use STARTTLS if available, but don't validate the cert
* "starttls" require STARTTLS, and validate the certificate
* "smtps" connect on port 465 and validate the cert

### DIFF
--- a/shell/imports/server/email.js
+++ b/shell/imports/server/email.js
@@ -15,12 +15,20 @@ const makePool = function (mailConfig) {
     auth = mailConfig.auth;
   }
 
+  const secure = (mailConfig.port === 465);
+  const tlsOptions = {
+    // Previously, node 0.10 did not attempt to validate certificates received when connecting
+    // with STARTTLS, so to avoid regressing we need to preserve that behavior here for now.
+    rejectUnauthorized: false,
+  };
+
   const pool = nodemailer.createTransport(smtpPool({
     host: mailConfig.hostname,
     port: mailConfig.port,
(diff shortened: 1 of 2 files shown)


dwrensha:
> then aborting when the cert provided by the mailserver didn't validate

How does such a case arise? I suppose the cert is being used for an external address of the mailserver, and Sandstorm is connecting to it over a different, internal address?

zarvox:
> How does such a case arise? I suppose the cert is being used for an external address of the mailserver, and Sandstorm is connecting to it over a different, internal address?

That's one way.  Another is if the server is using a self-signed "snakeoil" cert, which due to node 0.10 not validating certs by default, and simplesmtp not overriding that default for [smtps](https://github.com/sandstorm-io/simplesmtp/blob/master/lib/client.js#L187..L201) nor [starttls](https://github.com/sandstorm-io/simplesmtp/blob/master/lib/client.js#L240..L250) unless you explicitly requested it (which we didn't).

PR #2394, opened 2016-08-16
Avoid using variables named "event" throughout the codebase.

zarvox:
Background: you may have noticed your text editor colors `event` the same as it does `window`.  This is because, like `window`, `event` is a global.

An unfortunate piece of legacy behavior from IE is the `window.event` global
variable, accessible via "event".  The precise behavior of this global varies
between the browsers, and it's easy to have code that accidentally accesses the
global when it means to access a local that is shadowing the global.

For cleanliness, prefer to use a different name (I like `evt` and `ev`) for
variables you might otherwise name `event`.  This way you can't accidentally
write code that reads the global and happens to work anyway in Chrome and
IE but breaks in Firefox.

In particular, an unintended dependency on `event` caused the "Unsubscribe" button to
break on Oasis.  I filed the appropriate fix against blackrock, but it seemed
worthwhile to ensure we're not making the same mistake elsewhere.


paulproteus:
This is amazing, @zarvox .

Is there a linter rule that can help us avoid this in the future, too?

zarvox:
http://eslint.org/docs/rules/no-shadow-restricted-names is probably what we want, but we fail too many ESLint rules for actual cause that I can't reasonably make that part of the CI process yet.  Improving on that front is one of my background tasks, though.

kentonv:
I think this change makes the code worse. I could be convinced that it makes the code better if we somehow enforced that no one ever uses the identifier "event" going forward. Otherwise this change is actually un-shadowing the dangerous API, which makes it easier to use accidentally.

But I think the best possible solution would be a linter rule that simply detects uses of the global "event". In theory this should be easy for a JS parser to detect statically.

zarvox:
That's a valid position to have, and I would certainly prefer to have a linter rule that detects this, and if I can get the shell refactored to the point where we can actually run ESLint, I will. :)

That said, I think we'd do better to agree not to use `event` in general.  It's easier to enforce a dumb rule like "don't use `event`" through code review than a complicated rule like "don't use `event` unless you can prove that it's actually being overridden to what you think it is in local scope".  And editors syntax-highlight `event` as a global like `window`, so it ought to be fairly hard to accidentally use `event`.

It also means that botched copy-paste jobs, or new contributors attempting to follow the same style as exists in the codebase, are less likely to make a subtle functionality-affecting error.

It's easy for me to see `event` in a diff.  It's much more work for me to prove that it's safe to shadow the variable.  It's easiest to have a linter do the check, but we can't do that particularly easily right now.

I think while we're relying on human-enforced correctness, it's best to just steer clear of potentially-problematic names.  But I can live with either outcome, so long as the bug on the blackrock side is fixed.

kentonv:
> It's easier to enforce a dumb rule like "don't use event" through code review

I dunno, I think I would be terrible at enforcing this kind of rule when I review code. I don't even notice when people use keywords as variable names, how am I expected to remember all members of `window`?

> And editors syntax-highlight event as a global like window

None of mine do. TBH I don't think I'd want them to, unless they have parsed the code and determined that that identifier is actually referencing the global, in which case renaming to "evt" again isn't needed.

zarvox:
> I dunno, I think I would be terrible at enforcing this kind of rule when I review code. I don't even notice when people use keywords as variable names, how am I expected to remember all members of window?

I can hardly imagine you'd be *worse* at that than at checking if variable names match where they're declared and where they're later used. ;)

But fine, we'll agree to disagree on this.  For what it's worth, I will still write any new event handlers with names other than `event`, and I encourage everyone else to pick names that definitely won't conflict with globals rather than ones that might be dangerous under certain circumstances.

PR #2395, opened 2016-08-16
docs: Clarify org management features

paulproteus:
@zarvox I humbly request your +1/-1 on this, since the details are subtle.

I do think this is a vast improvement over the earlier text, but I want someone to make sure I'm not saying anything misleading or false.

### DIFF
--- a/docs/administering/for-work.md
+++ b/docs/administering/for-work.md
@@ -15,9 +15,9 @@ Many features of Sandstorm for Work depend on a server administrator specifying 
 be working together using Sandstorm. We call this **organization management.** You can enable and disable
 organization-related features via the `/admin/organization` settings area.
 
-To apply settings to all users within an organization, you must configure at least one login provider so
-that it knows which people are part of the organization. Users within the organization will automatically
-be able to log in, install apps, and create grains. 
+To apply settings to all users within an organization, you must use organization management settings to
+configure the boundary of your organization with regard to at least one login provider. Users within the
+organization will automatically be able to log in, install apps, and create grains.
 
 To enable this feature:
 
--- a/docs/administering/for-work.md
+++ b/docs/administering/for-work.md
@@ -11,7 +11,7 @@ Work, and it gives us time to finish testing the payments system.
 
 ### Defining an organization, and its impact on permissions
 
-Many features of Sandstorm for Work  depend on a server administrator specifying the group of users that will
+Many features of Sandstorm for Work depend on a server administrator specifying the group of users that will
 be working together using Sandstorm. We call this **organization management.** You can enable and disable
 organization-related features via the `/admin/organization` settings area.
 
--- a/docs/administering/for-work.md
+++ b/docs/administering/for-work.md
@@ -63,8 +63,13 @@ grains can only be seen by users logged-in as a member of your organization. Thi
 
 - Unauthenticated (aka anonymous) users cannot view grains, even if they have a sharing link.
 
-- Visitors (logged-in users who have no special permissions on your server) cannot view grains, even
-  if they have a sharing link.
+- When a user attempts to create an account or sign in, Sandstorm validates that they are part
+  of your organization. If not, then they may not create an account or sign in. This prevents
+  [Visitors](guide.md) from using their accounts.
+
+- Security note: at the moment, this setting only applies to new logins. If a user was already signed
+  in when you first enabled this option, the user would be able to continue to use their account even
+  if they're not a member of your organization.
 
 You can **make all organization users visible to each other.** This setting automatically adds
 users within the organization to each other's contact list so that they can share grains with each
--- a/docs/administering/for-work.md
+++ b/docs/administering/for-work.md
@@ -9,6 +9,60 @@ Work, and it gives us time to finish testing the payments system.
 
 ## Sandstorm for Work in depth
 
+## Organization management
+
+Sandstorm for Work allows you to define an organization. You can automatically apply some settings to all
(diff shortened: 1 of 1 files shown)


docs/administering/for-work.md
```diff
@@ -9,6 +9,68 @@ Work, and it gives us time to finish testing the payments system.
...
 ## Sandstorm for Work in depth
 
+### Defining an organization, and its impact on permissions
+
+Many features of Sandstorm for Work depend on a server administrator specifying the group of users that will
+be working together using Sandstorm. We call this **organization management.** You can enable and disable
+organization-related features via the `/admin/organization` settings area.
+
+To apply settings to all users within an organization, you must configure at least one login provider so
+that it knows which people are part of the organization. Users within the organization will automatically
```
zarvox: what is "it" here?

docs/administering/for-work.md
```diff
@@ -9,6 +9,68 @@ Work, and it gives us time to finish testing the payments system.
...
+- Log in to your server as an administrator.
+
+- Click your name at the top right, then **Admin panel**; this should take you to `/admin`.
+
+- Under the "Configuration" header, click "Organization settings"
+
+- Enable and configure your organization membership on a per-login-provider basis.
+
+Login providers have different settings that are used to define your organization. A user is
+considered a member of your organization if the settings for **at least one** login provider
```
zarvox: Putting my multi-identity-naive hat on, it's not clear to me what "at least one" login provider could mean.  Perhaps this deserves a link to wherever our reference on multiple identities lives?

docs/administering/for-work.md
paulproteus: "it" is "the login provider"

### alternate proposal

> To apply settings to all users within an organization, you must use organization management settings to configure the boundary of your organization with regard to at least one login provider. Users within the organization will automatically be able to log in, install apps, and create grains.

docs/administering/for-work.md
zarvox: I tried coming up with a better phrasing, but I struggled to find one.  Maybe this is fine as it stands.

paulproteus: I wasn't thinking of multi-identity accounts! I was thinking of how:

- If you configure SAML to provide organization-member status, and

- If you define GApps to provide organization-member status,

- Then either one will do.

i.e. it's a disjunction, not a conjunction.

I'm at a loss for other ways to phrase this at the moment, having just tried for a couple of minutes; I'm hoping the current text can stand and/or you can provide a suggested alternative here.

docs/administering/for-work.md
paulproteus: I'll make this change now.

paulproteus:
@zarvox I made one change (and I introduced a typo in a commit message; oops).

+1/-1/wait further?

docs/administering/for-work.md
```diff
@@ -9,6 +9,68 @@ Work, and it gives us time to finish testing the payments system.
...
+
+Once you have defined your organization, you can optionally configure system-wide rules based on
+organization membership. This can be done from `/admin/organization` within your Sandstorm server.
+
+You can **disallow collaboration with users outside the organization.** If you enable this option,
+grains can only be seen by users logged-in as a member of your organization. This means:
+
+- Unauthenticated (aka anonymous) users cannot view grains, even if they have a sharing link.
+
+- Visitors (logged-in users who have no special permissions on your server) cannot view grains, even
```
zarvox: This isn't *exactly* right.  The precise behavior is:

* you can neither log in nor create a new account unless your identity provider marks your identity as in the organization
* this means that non-org-members cannot open grains or do anything else with their accounts
* except that if you have an account already and were already logged in, you remain logged in until your resumeToken expires and can still open grains that you already had access to, and (I think) create new grains, and share grains with other users who *are* in the org.  We should probably add an org membership check to `openSession` and/or the grain-creation codepath, or a way to forcibly terminate sessions/expire `resumeToken`s across the deployment.

So in that context, there's not supposed to be such a thing as a "logged-in visitor".

paulproteus: > there's not supposed to be such a thing as a "logged-in visitor".

but there _is_ in general, right? Just not when this option is enabled?

Assuming the answer to the above is "yes", here is some proposed alternate text.

### proposed alternate text

[unchanged]

> - Unauthenticated (aka anonymous) users cannot view grains, even if they have a sharing link.

[changed]

> - When a user attempts to create an account or sign in, Sandstorm validates that they are part of your organization. If not, then they may not create an account or sign in. This prevents [Visitors](guide.md#accounts-authentication-and-access-control) from using their accounts.

[newly added]

> - Security note: at the moment, this setting only applies to new logins. If a user was already signed in when you first enabled this option, the user would be able to continue to use their account even if they're not a member of your organization.

docs/administering/for-work.md
zarvox: Yeah, let's leave this text for now.

docs/administering/for-work.md
zarvox: :+1:

Consider filing an issue which, when closed, would make that Security note obsolete.  I'd really rather have the feature just do the thing the user needs rather than have weird caveats that we document.

paulproteus:
OK! @zarvox I responded to all comments you left. +1/-1/wait?

In the absence of further feedback I'll merge as-is, in about 1h. More feedback welcome post-merge.

zarvox:
:shipit:

paulproteus:

PR #2399, opened 2016-08-17
Relocate migrations

zarvox:
This changeset extracts migrations out of sandstorm-db and into a separate file.

I also extracted some accounts-related network-touching code that previously lived in `sandstorm-db` because it needed to be executed during migrations.

`migrateToLatest` now receives two arguments, an instance of `SandstormDb` and an instance of `SandstormBackend`.  It no longer reaches into global state.

This is one of several steps involved in my ongoing effort to clean up initialization order.

### DIFF
--- a/shell/imports/server/accounts/picture.js
+++ b/shell/imports/server/accounts/picture.js
@@ -1,0 +1,50 @@
+import { HTTP } from "meteor/http";
+
+const userPictureUrl = function (user) {
+  if (user.services && !(user.profile && user.profile.picture)) {
+    // Try to determine user's avatar URL from login service.
+
+    const google = user.services.google;
+    if (google && google.picture) {
+      return google.picture;
+    }
+
+    const github = user.services.github;
+    if (github && github.id) {
+      return "https://avatars.githubusercontent.com/u/" + github.id;
+    }
+
+    // Note that we do NOT support Gravatar for email addresses because pinging Gravatar would be
+    // a data leak, revealing that the user has logged into this Sandstorm server. Google and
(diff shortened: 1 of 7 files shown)


kentonv:
Why do you keep changing:

    function foo() {

to:

    const foo = function() {

?

AFAIK, in the top-level scope, this has no effect except to make the code harder to read.

kentonv:
I guess that's not a merge-blocker to me, but I'm curious to know if there's some effect I'm not aware of.

zarvox:
This is my understanding:

In Meteor packages, the global scope is by default hidden.

In the Meteor root project, function declarations show up in the global scope.  In imports, function declarations do not show up in the global scope.

If you use `const foo = function() {` you are guaranteed that the declaration does not leak into global scope, regardless of where the file containing that code lives.  This means the actual location of the declaration can't change the semantics of the declaration, which is helpful when lifting code out of packages which may or may not immediately wind up in `imports/`.

This is also a convenient way to mark to myself "yes, I have audited the rest of the codebase, and nowhere else depends on this function being present in scope without explicitly being imported from somewhere".

dwrensha:
> In the Meteor root project, function declarations show up in the global scope.

This is false, as far as I can tell.


I think the only advantage is that `const` prevents you from defining the same function twice in the same scope, whereas 

```javascript
function foo() {}
```

is equivalent to

```javascript
var foo = function() {}
```

kentonv:
@zarvox `function foo()` is always file-local in Meteor -- package or not. To declare a global function, you have to do `foo = function()`.

PR #2404, opened 2016-08-17
docs: add missing link

dwrensha:


paulproteus:
That's a big glaring problem that you fixed. Thanks for fixing it. Merging (which will cause a docs deploy) now!

PR #2413, opened 2016-08-18
appdemo: Use AppIndex to choose app version

### LINKED ISSUE #2397: appdemo should rely on app-index
Context:

- If an app author uploads a new version of an app to Oasis, this can break the `/appdemo/` functionality.

- You can see this with Davros right now.

Proposed fix:

- Modify the `appdemo` code so that, rather than querying Mongo for the most recent `Package` available on this Sandstorm install, it queries the app index for the version to use.
- If that version is not available on this server, then ... ??? TODO: Someone help me design this.

Side-effects:

- This will have the result that appdemo only works for apps in the app index. In my opinion, this change is desired behavior.

paulproteus:
Close #2397

### DIFF
--- a/shell/server/demo-server.js
+++ b/shell/server/demo-server.js
@@ -127,11 +127,11 @@ if (allowDemo) {
     // is typically 6-24 hours delayed from reality, so if an app is newly-available
     // in the app market, it won't be in the app index. Therefore we don't bail-out if the app
     // is missing from the AppIndex collection.
-    let appIndexData = globalDb.collections.appIndex.findOne({appId: appId});
+    let appIndexData = globalDb.collections.appIndex.findOne({ appId: appId });
 
     // Prepare a helper function we can use to transform a package cursor into an appropriate return
     // value for this function.
-    const packageCursorAndMaybeUserActions = function(userId, appId, packageCursor) {
+    const packageCursorAndMaybeUserActions = function (userId, appId, packageCursor) {
       // This allows us to avoid creating duplicate UserActions.
       if (userId) {
         return [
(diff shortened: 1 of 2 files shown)


paulproteus:
Once Jenkins expresses its satisfaction-or-not, then I'll ask Kenton to review/merge.

paulproteus:
@kentonv I believe the root cause is that, if you're doing an appdemo of a preinstalled app, the app preinstallation logic runs before the app demo logic, so you end up with a `UserAction` of the app market version of the app, rather than a `UserAction` of the crazy-latest-unpublished-version of the app. This results in the client-side prediction being wrong.

The client side doesn't really expect the server-side to be doing an install on app creation, and it's a race condition where the client wants to take actions as soon as the user logs in, and the server does too, so the client will make decisions based on not-up-to-date data because the latency is > 0.

My new code will do the right thing in the case of preinstalled apps, I believe, so we should be OK for this case with this patch. Specifically, if the app is a "preinstalled" one, there is _some_ version server-side, and we'll end up latest available version. There is one remaining race condition, I believe, if:

- User visits `/appdemo/{{appId}}`

- App index was updated 0 seconds ago

- App index begins downloading new version of app

- User's page successfully loads, with the client caching a packageId of yesterday's version of the app

- App index finishes downloading today's version of the app

- User clicks "Try {{appName}}"

- Client believes the right package is yesterday's version

- Server believes the right package is today's version

- Client calls `addUserActions` method on yesterday's packageId

- Result: mismatch

There is a very small time window, I believe, when this could happen, so I'm OK ignoring this problem for now. @kentonv curious for your thoughts.

I'm not even sure this race condition is a real problem. The client only calls `addUserActions` if the `appId` has no corresponding `UserActions`, and it could easily be the case that if preinstallation happens early enough, the `UserActions` will have reached the client by the time that `demo-client.js` does the query, in which case, no mismatch, since no client-driven call to the `addUserActions` method.

paulproteus:
Jenkins is happy, so ready for @kentonv 's review/merge/-1/etc.

shell/server/demo-server.js
```diff
@@ -115,36 +115,57 @@ if (allowDemo) {
...
+        return packageCursorAndMaybeUserActions(this.userId, appId, appIndexPackageQuery);
+      }
 
-    // This allows us to avoid creating duplicate UserActions.
-    if (this.userId) {
-      return [
-        packageCursor,
-        UserActions.find({ userId: this.userId, appId: appId }),
-      ];
+      packageQuery["manifest.appVersion"] = { $lte: appIndexData.versionNumber };
```
kentonv: Arguably should be `$lt`: If the *exact* match isn't found, then any other package with the same version definitely *isn't* from the app index and so is likely an experimental version uploaded by the app author.

shell/server/demo-server.js
```diff
@@ -115,36 +115,57 @@ if (allowDemo) {
...
     }
 
-    return packageCursor;
+    // If the specific package from the app index isn't installed, or the app isn't there at all, do
+    // our best.
+    packageQuery.appId = appId;
+    const packageCursor = Packages.find(
+      packageQuery,
+      { sort: { "manifest.appVersion": -1 } });
```
kentonv: Should we add `limit: 1`? Weird to expose the whole history of the app.

shell/server/demo-server.js
paulproteus: +1, though that's subtle and deserves a comment; I'll do it with `$lt` and add a comment!

shell/server/demo-server.js
paulproteus: +1

paulproteus:
:tada:

PR #2415, opened 2016-08-18
docs: I => We

### LINKED ISSUE #2015: docs: Decide what "I" vs. "we" means, then make the docs consistent
Quoth @zarvox 

> Do we use "I" or "we" in the docs? I think I've seen both; we should probably attempt to be consistent. (I prefer "we" in documentation maintained by multiple individuals, unless there's also clearly attributed authorship for each article/unit, so we know who is "I".)

See also: https://github.com/sandstorm-io/sandstorm/pull/1979

paulproteus:
Close #2015


paulproteus:

PR #2425, opened 2016-08-20
Allow sorting grain list by last used, title, and size

zarvox:
Now that we let you see per-grain size everywhere, it makes sense to also
allow sorting the grain list by size, in addition to the default sort order
of "last activity".  And we might as well let you sort by title too, if you want.

grain-list

app-details-page

### DIFF
--- a/shell/client/grain/grainlist-client.js
+++ b/shell/client/grain/grainlist-client.js
@@ -89,7 +89,84 @@ const compileMatchFilter = function (searchString) {
   };
 };
 
-const filteredSortedGrains = function (showTrash) {
+const sortGrains = function (grains, sortRules) {
+  // Overengineered sort logic.
+  //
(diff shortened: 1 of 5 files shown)


shell/client/grain/grainlist-client.js
```diff
@@ -470,9 +551,65 @@ Template.sandstormGrainTable.helpers({
...
+    return instance.sortOrder.get();
+  },
+
+  equal(a, b) {
+    return a === b;
+  },
 });
 
 Template.sandstormGrainTable.events({
+  "click thead td.grain-name"(evt) {
```
kentonv: Should we be wrapping each heading in `<button>` for accessibility and mobile touch target hinting?

kentonv:
Merging for now but I think we should address the accessibility issue soon.

PR #2438, opened 2016-08-23
docs: Explain where to put certificate files

paulproteus:


paulproteus:

PR #2439, opened 2016-08-24
Account buttons cleanup

zarvox:
* Fix bug where identity cards weren't looking up the current identity correctly
* Give identity cards some margin so they don't look weird when stacked
* Audit the codebase for use of `{{> someTemplate }}` and convert to either `{{> someTemplate . }}` (explicitly passing on data context) or `{{> someTemplate ""}}` (passing on an empty data context).
* Add a couple typechecks on the args to the login/account button templates

### DIFF
--- a/shell/client/accounts/account-settings.html
+++ b/shell/client/accounts/account-settings.html
@@ -106,7 +106,7 @@ limitations under the License.
     <div class="add-new">
       Add new e-mail address
       {{> emailAuthenticationForm emailLoginFormData}}
-      {{> _loginButtonsMessages}}
+      {{> _loginButtonsMessages "" }}
     </div>
   </div>
 
@@ -229,7 +229,7 @@ setActionCompleted: (optional) Function.  Callback triggered on profile saved or
             </li>
           {{/with}}
           {{#if isPaymentsEnabled}}
-            {{> billingOptins}}
+            {{> billingOptins "" }}
           {{/if}}
         {{/unless}}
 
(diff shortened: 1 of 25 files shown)


dwrensha:
Could you upload a screenshot of the new spacing?

zarvox:
I appear to have some bugs to fix, but:

spaced-identity-choices

zarvox:
Okay, this is in better shape now.

neynah:
The pure black border looks harsh to me. I think we should be consistent and use something closer to #b3b3b3. The spacing looks good though.

kentonv:
Personally I find `Template.currentData()` confusing because at a glance I tend to interpret it as `Template.instance().data`, because in my mind the current data context would have been accessed as `this` and not `Template.something`.

zarvox:
Yeah, I could see that.

I find it kind of annoying that `this` is the template instance in `Template.onCreated()` and `Template.onRendered()`, and if you want to reactively use the data context, then you need to use `Template.currentData()`, whereas in template helpers and event maps `this` is the data context, and you have to use `Template.instance()` to get a handle to the template instance.

I guess I'll try to leave these alone in future patchsets, then?

kentonv:
Hmm, in `onCreated()` you can use `this.data`, right?

But yeah, the inconsistency is annoying.

Anyway, no action needed at this time.

shell/client/accounts/account-settings.html
```diff
@@ -229,7 +229,7 @@ <h3 class="title-bar">Primary e-mail address for service-related notifications</
             </li>
           {{/with}}
           {{#if isPaymentsEnabled}}
-            {{> billingOptins}}
+            {{> billingOptins "" }}
```
kentonv: This broke the mailing list signup prompt, which needs `db` in order to display the user's current storage quota.

I've fixed this, but I'm worried that it's hard to know how many other things might be broken.

PR #2441, opened 2016-08-24
sandstorm-http-bridge: re-save identity caps if they fail to restore

### LINKED ISSUE #2440: identities saved through sandstorm-http-bridge break after backup/restore
1. Alice creates a new rocket.chat grain G.
  2. Alice shares G to Bob.
  3. Both Alice and Bob confirm their usernames in G.
  4. Alice downloads a backup of G and then restores it as H.
  5. Alice shares H with Bob.
  6. Bob visits H, then closes H.
  7. H reboots (perhaps through Alice clicking the "restart" button in the topbar).
  8. Alice writes a message in H's "#general" channel, @-mentioning Bob.

I expect Bob to receive a notification about the @-mention, but he does not!
The problem is that H's sturdyref for Bob's identity is only valid within G, and H never tries to refresh the sturdyref, even if calling `restore()` on it fails.

dwrensha:
Fixes #2440.

Adds some logic so that now, every time that we add a new entry to `identitiesMap`, if there is already a corresponding sturdyref on disk, we kick off a task that tries to restore that sturdyref, and we re-save it upon failure.

### DIFF
--- a/src/sandstorm/sandstorm-http-bridge.c++
+++ b/src/sandstorm/sandstorm-http-bridge.c++
@@ -1056,7 +1056,8 @@ class BridgeContext: private kj::TaskSet::ErrorHandler {
 public:
   BridgeContext(SandstormApi<>::Client apiCap, spk::BridgeConfig::Reader config)
       : apiCap(kj::mv(apiCap)), config(config),
-        identitiesDir(openIdentitiesDir(config)), tasks(*this) {}
+        identitiesDir(openIdentitiesDir(config)),
+        trashDir(openTrashDir(config)), tasks(*this) {}
 
   void saveIdentity(capnp::Data::Reader identityId, Identity::Client identity) {
     if (!config.getSaveIdentityCaps()) return;
@@ -1072,25 +1073,20 @@ public:
 
       if (faccessat(identitiesDir, textIdRef.cStr(), F_OK, AT_SYMLINK_NOFOLLOW) != 0) {
         // Not yet recorded to disk. Need to save a SturdyRef.
-        auto req = apiCap.saveRequest();
-        req.setCap(identity);
-        req.initLabel().setDefaultText("user identity");
-        tasks.add(req.send().then([this,textIdRef](auto result) -> void {
-          // Sandstorm tokens are primarily text but use percent-encoding to be safe.
-          auto tokenText = percentEncode(result.getToken());
-
-          // Store as a symlink. ext4 can store up to 60 bytes directly in the inode, avoiding
-          // allocating a block.
-          KJ_SYSCALL(symlinkat(tokenText.cStr(), identitiesDir, textIdRef.cStr()));
-
-          // Make sure it's really saved.
-          KJ_SYSCALL(fsync(identitiesDir));
+        saveIdentityInternal(textIdRef, kj::mv(identity));
+      } else {
+        // Try restoring the existing SturdyRef and re-save on failure.
+        tasks.add(loadIdentityFromDisk(textIdRef).ignoreResult().catch_(
+            [this, textIdRef, identity](auto error) mutable {
+          if (error.getType() == kj::Exception::Type::FAILED) {
+            saveIdentityInternal(textIdRef, kj::mv(identity));
+          }
         }));
       }
     }
   }
 
-  Identity::Client loadIdentity(kj::StringPtr origId) {
+  kj::Promise<Identity::Client> loadIdentity(kj::StringPtr origId) {
     // Obtain the identity capability for the given identity ID.
 
     KJ_REQUIRE(config.getSaveIdentityCaps(),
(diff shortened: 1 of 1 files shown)


src/sandstorm/sandstorm-http-bridge.c++
```diff
@@ -1195,7 +1201,59 @@ private:
     auto req = apiCap.restoreRequest();
     req.setToken(percentDecode(buf));
 
-    return req.send().getCap().castAs<Identity>();
+    // Don't directly call `getCap()` on the `RemotePromise` because we want to make sure
```
kentonv: This breaks pipelining in several cases, but it looks like there's only one call site that actually cares about catching exceptions. Can you instead go back to returning a pipelined capability, but call `.whenResolved()` on it in the case where you're actually testing the capability's validity? This should be equivalent.

src/sandstorm/sandstorm-http-bridge.c++
```diff
@@ -1195,7 +1201,59 @@ private:
...
+      // We name the trash file after the token, not the identity ID. This way, it's okay
+      // if we overwrite an existing entry of the trash directory.
+      auto trashSymlink = kj::heapString(buf);
+      KJ_SYSCALL(renameat(identitiesDir, symlink.cStr(), trashDir, trashSymlink.cStr()));
+
+      auto req = apiCap.dropRequest();
+      req.setToken(percentDecode(buf));
+      tasks.add(req.send().then([KJ_MVCAP(trashSymlink), this](auto response) -> void {
+        KJ_SYSCALL(unlinkat(trashDir, trashSymlink.cStr(), 0));
+        KJ_SYSCALL(fsync(trashDir));
```
kentonv: I think this fsync() isn't accomplishing anything except to slow things down, since nothing is actually gated on the completion of this operation. Normally you need to fsync() before returning a success code to the caller, so that the caller knows that success means durability, but there's no caller here.

With or without the fsync(), it's possible to be left in a state where there are tokens in the trash that were in fact already dropped. That's fine, since the worst case is that drop() will be harmlessly called again later.

src/sandstorm/sandstorm-http-bridge.c++
```diff
@@ -1195,7 +1201,59 @@ private:
...
+  void saveIdentityInternal(kj::StringPtr textId, Identity::Client identity) {
+    auto req = apiCap.saveRequest();
+    req.setCap(identity);
+    req.initLabel().setDefaultText("user identity");
+    tasks.add(req.send().then([this,textId](auto result) -> void {
+      // Sandstorm tokens are primarily text but use percent-encoding to be safe.
+      auto tokenText = percentEncode(result.getToken());
+
+      // Clean up any existing symlink.
+      dropIdentity(textId);
```
kentonv: Hmm, it seems there's a period of time here between dropIdentity() and symlinkat() during which a crash would lose the identity altogether.

However, presumably if we're saving over an existing token, it's because the existing token was already observed to be dead?

Maybe dropIdentity() should actually be called at the point where this observation was made, rather than here? And if, in the meantime, some other concurrent task has saved a copy of the identity, then we should not overwrite it?

src/sandstorm/sandstorm-http-bridge.c++
```diff
@@ -1072,25 +1073,20 @@ public:
...
-          // allocating a block.
-          KJ_SYSCALL(symlinkat(tokenText.cStr(), identitiesDir, textIdRef.cStr()));
-
-          // Make sure it's really saved.
-          KJ_SYSCALL(fsync(identitiesDir));
+        saveIdentityInternal(textIdRef, kj::mv(identity));
+      } else {
+        // Try restoring the existing SturdyRef and re-save on failure.
+        tasks.add(loadIdentityFromDisk(textIdRef).ignoreResult().catch_(
+            [this, textIdRef, identity](auto error) mutable {
```
kentonv: KJ_MVCAP(identity)? (saves a refcount)

src/sandstorm/sandstorm-http-bridge.c++
dwrensha: Ah, I hadn't realized that I could both pipeline and detect errors like that. Updated.

src/sandstorm/sandstorm-http-bridge.c++
dwrensha: Removed.

src/sandstorm/sandstorm-http-bridge.c++
dwrensha: Ah, I need to deal with the double-save case too. Doing it...

src/sandstorm/sandstorm-http-bridge.c++
dwrensha: Done.

src/sandstorm/sandstorm-http-bridge.c++
```diff
@@ -1129,14 +1130,15 @@ public:
                                       -> kj::Promise<Identity::Client> {
         if (e2.getType() == kj::Exception::Type::DISCONNECTED) {
           // Disconnected. We'll need to reload from disk.
-          Identity::Client newIdentity = loadIdentityFromDisk(textId);
-
-          // Save the new identity to the map so that we don't have to reload it again.
-          auto iter = liveIdentities.find(textId);
-          KJ_ASSERT(iter != liveIdentities.end());
-          iter->second.identity = newIdentity;
+          auto identity = loadIdentityFromDisk(textId);
```
kentonv: By renaming this `newIdentity` -> `identity`, you're now shadowing a captured variable, which is confusing.

dwrensha: OK. Unrenamed.

src/sandstorm/sandstorm-http-bridge.c++
dwrensha: > However, presumably if we're saving over an existing token, it's because the existing token was already observed to be dead?

Yeah, but there's also a nonzero chance that the failure was somehow transient, so we'd like to delay dropping the old sturdyref as long as we can.

We want to:
  1. Avoid leaking sturdyrefs.
  2. Not delete anything in the identities/ directory until we have something to replace it with. Ideally, the replacement would be atomic.
  3. Keep it simple.

> Maybe dropIdentity() should actually be called at the point where this observation was made, rather than here? And if, in the meantime, some other concurrent task has saved a copy of the identity, then we should not overwrite it?

This seems strictly worse for goals (2) and (3).

PR #2448, opened 2016-08-25
Better error message when calling SandstormApi.save() with cap = null.

dwrensha:
Currently if you call `SandstormApi.save()` and forget to set `cap`, you get this error: 

```
Failed: remote exception: Called null capability.
```

which is not super helpful. This patch changes the error to:

```
Failed: remote exception: expected args.hasCap(); Cannot save a null capability.
```


zarvox:
Garply, retest this please

PR #2453, opened 2016-08-26
Add link parameter to offer template

jparyani:
This allows you to make the whole offer template into a link. At the
moment, the only thing you can specify is a scheme, and the rest is
autogenerated for you.

This needs some input on what we're actually going to do about filtering dangerous schemes. Either add a whitelist or only allow them to start with "sandstorm-" (or automatically prefix it ourselves).

### DIFF
--- a/shell/client/grain-client.js
+++ b/shell/client/grain-client.js
@@ -1436,11 +1436,9 @@ Meteor.startup(function () {
         });
 
         rawLink = call.link;
-        // TODO(soon): check against whitelist
         // rfc3986 specifies schemes as:
         // scheme = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )
-        if (rawLink && (rawLink.scheme.search(/^[^a-zA-Z]/) !== -1 ||
-                        rawLink.scheme.search(/[^a-zA-Z0-9+-.]/) !== -1)) {
+        if (rawLink && (rawLink.scheme.search(/[^a-zA-Z0-9+-.]/) !== -1)) {
           throw new Meteor.Error(400, "rawlink.scheme has invalid characters.");
         }
       } catch (error) {
@@ -1528,7 +1526,7 @@ Meteor.startup(function () {
                                          .replace(/\$GRAIN_TITLE_SLUG/g, grainTitleSlug);
         let link = undefined;
         if (rawLink) {
-          link = `${rawLink.scheme}:${window.location.protocol}//${host}#${tokenId}`;
+          link = `clientapp-${rawLink.scheme}:${window.location.protocol}//${host}#${tokenId}`;
         }
 
         sessionStorage.setItem(key, JSON.stringify({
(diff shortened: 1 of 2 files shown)


shell/client/grain-client.js
```diff
@@ -1509,12 +1523,18 @@ Meteor.startup(function () {
         const renderedTemplate = template.replace(/\$API_TOKEN/g, tokenId)
                                          .replace(/\$API_HOST/g, host)
                                          .replace(/\$GRAIN_TITLE_SLUG/g, grainTitleSlug);
+        let link = undefined;
+        if (rawLink) {
+          link = `${rawLink.scheme}://${window.location.protocol}//${host}#${tokenId}`;
```
kentonv: Change the first `://` to just `:` -- the `//` doesn't make sense since it's not followed by a hostname.

kentonv:
How about `clientapp-*:` for the scheme?

I wonder if we should use basic-auth-like syntax instead of our webkey syntax...

    clientapp-foo:https://*:<token>@<host>/

Apps that know they support `Authorization: bearer` would of course be free to use the password as a token. (Note that the URL spec allows `*` as a username but this also seems like a good way to indicate "doesn't matter".)

jparyani:
Discussed in person with @kentonv, but using basic auth doesn't work on android. I've tracked it down to the "@" character causing problems. It results in an ERR_UNKNOWN_URL_SCHEME for the android browser.

kentonv:
The test failures appear to be legitimate, BTW.

jparyani:
Ready to merge

shell/client/grain-client.js
```diff
@@ -1428,7 +1430,17 @@ Meteor.startup(function () {
           unauthenticated: Match.Optional(Object),
           // Note: `unauthenticated` will be validated on the server. We just pass it through
           //   here.
+          link: Match.Optional({
```
kentonv: Let's call this "clientapp".

shell/public/offer-template.html
```diff
@@ -129,7 +140,14 @@
        } else {
           var errorText = "Error refreshing token. Reloading the page might help.";
           console.error(xhr.responseText);
-          document.getElementById("text").textContent = errorText;
+          var textElement = document.getElementById("text");
+          textElement.textContent = errorText;
+          var link = document.getElementById("href");
```
kentonv: It's the id "link"?

shell/client/grain-client.js
```diff
@@ -1428,7 +1430,17 @@ Meteor.startup(function () {
...
+            scheme: String,
+          }),
         });
+
+        rawLink = call.link;
+        // rfc3986 specifies schemes as:
+        // scheme = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )
+        if (rawLink && (rawLink.scheme.search(/[^a-zA-Z0-9+-.]/) !== -1)) {
```
kentonv: Use Match.Where

PR #2459, opened 2016-08-27
admin page: Remove bogus query parameter from link to Hacker CMS.

dwrensha:


patrickod:
:+1: thanks!

kentonv:
Should we actually substitute the server's URL in there?

dwrensha:
OK. I've also added target=_blank so that the link acts more like the "install app" button.

PR #2466, opened 2016-08-29
Dropping a token you don't own should not be an error.

dwrensha:
Currently, if you back up and restore a grain that uses a self-repairing identity database as in #2441, and the original grain has not been completely deleted, the restored grain ends up with identities accumulating in its trash directory, because `drop()` fails. This patch fixes the problem.

### DIFF
--- a/shell/server/core.js
+++ b/shell/server/core.js
@@ -567,7 +567,11 @@ function dropInternal(db, sturdyRef, ownerPattern) {
     return;
   }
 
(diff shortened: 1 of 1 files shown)


kentonv:
Can you clarify in the comment that the idea is to treat this condition as if the token was not found (which is documented as not an error)? That would have helped me convince myself that this is correct a little faster. :)

PR #2471, opened 2016-08-29
Add delete confirmation modals/inputs

jparyani:


shell/client/widgets/widgets.html
```diff
@@ -4,7 +4,7 @@
                          by clicking on a close button or outside the boundary of the modal.
    A content block of what to put inside the modal.
 --}}
-<div class="modal" role="dialog">
+<div class="modal {{class}}" role="dialog">
```
zarvox: Please document the new optional parameter in the comment above.

shell/client/admin/user-details.html
```diff
@@ -327,11 +327,16 @@
           <strong>Account deletion is permanent and irreversible.</strong>  Please be certain.
           The account will first be suspended for 7 days so that you have a chance to undo the
           deletion. At the end of the 7 days it will be deleted.
+
+        </p>
+        <p>
+          Please type "Yes, delete this account" below to confirm.
         </p>
+        <input class="confirm" type="text">
```
zarvox: I'd normally expect this input to be a child of the `<form>` tag so it gets the usual `display: block` and font size and label rules - is there a reason it's not?  avoiding having to `preventDefault()` the `onSubmit()` event?

jparyani: Talked in person with @zarvox, and decided it wasn't worth it to wrap it in a form.

kentonv:
Can you add some screenshots to this PR?

jparyani:
image

image

image

shell/client/accounts/account-settings.html
```diff
@@ -124,6 +124,42 @@ <h3 class="title-bar">Primary e-mail address for service-related notifications</
...
+      <h2>Delete your account</h2>
+      <p>Deleting this user account will:</p>
+      <ul>
+        <li>Permanently delete any grains created by you</li>
+        <li>Revoke any sharing links created by you</li>
+        <li>Remove you from all users' contact lists</li>
+        <li>Delete all identities owned by this account (except identities associated with another account)</li>
+      </ul>
+      <p>
+        <strong>Account deletion is permanent and irreversible.</strong>  Please be certain.
```
zarvox: This copy is kinda goofy - "permanent and irreversible", yet "will first be suspended for 7 days".  Maybe something like:

------

You are about to mark your account for deletion.  Once marked for deletion, your account will be suspended, and you will have 7 days to cancel the pending deletion by logging back into your account.

After 7 days, your account will be permanently deleted.  Deleting your account will:

* Permanently delete any grains you created
* Revoke all sharing links you created
* Remove you from all users' contact lists
* Delete all identities owned by your account (except identities also associated with another account).

<strong>Account deletion is permanent and irreversible.</strong>  Please be certain.

Type "Yes, delete my account" below to confirm.

------

A v2 of this might show precisely which identities would be deleted, and which ones would not, by using the `accountsOfIdentity` subscription.

shell/client/admin/user-details.html
```diff
@@ -327,11 +327,16 @@
           <strong>Account deletion is permanent and irreversible.</strong>  Please be certain.
```
zarvox: Probably the same copy adjustments suggested above make sense here, with changes as appropriate to refer to `{{guessUserName}}` rather than "you" and "your grains"

zarvox:
Fascinating seemingly-unrelated incoming mail test failure: https://build.sandstorm.io/job/Sandstorm%20PRs%20-%20build/1002/

zarvox:
Garply, retest this please

kentonv:
Making admins type some text to confirm deleting each account seems excessive. I can imagine an admin trying to delete a bunch of accounts and getting frustrated with this. Maybe they should just have to click the button to confirm? After all, they can undelete easily (since they don't get logged out).

For self-deletes, I think the phrase users have to type probably shouldn't be case-sensitive nor involve punctuation. I think non-technical users would find this surprising.

The text "Account deletion is permanent and irreversible. Please be certain." is misleading, since it's not irreversible at all. I think this text is not adding useful information and can probably just be deleted.

zarvox:
> Making admins type some text to confirm deleting each account seems excessive. I can imagine an admin trying to delete a bunch of accounts and getting frustrated with this. Maybe they should just have to click the button to confirm? After all, they can undelete easily (since they don't get logged out).

I can live with this either way, but we should probably preserve the confirmation text (or perhaps disallow administrative suspension entirely?) in the event the admin user is visiting their own user details page, e.g. `/admin/users/{{Meteor.userId()}}`.

> For self-deletes, I think the phrase users have to type probably shouldn't be case-sensitive nor involve punctuation. I think non-technical users would find this surprising.

Good idea.  Let's make the string to type be `delete my account`.

> The text "Account deletion is permanent and irreversible. Please be certain." is misleading, since it's totally reversible for the first 7 days. I think this text is not adding useful information and can probably just be deleted.

Yeah, I guess this makes sense now that we've built the cooling-off period.

jparyani:
Ready to merge

shell/server/account-suspension.js
```diff
@@ -86,6 +86,10 @@ Meteor.methods({
       throw new Meteor.Error(403, "Only admins can suspend other users.");
     }
 
+    if (userId === Meteor.userId()) {
+      throw new Meteor.Error(400, "Admins cannot suspend their own accounts from the admin page. Please go to your account setttings.");
```
zarvox: Hilariously, users in an organization can't suspend their own accounts from the account settings page either.  But I guess having to ask another admin to suspend your account for you is probably okay in this circumstance.

PR #2474, opened 2016-08-30
docs: Explain how file autodetection works

paulproteus:
I'll merge this right now to get it into the docs; post-commit feedback very welcome here.


paulproteus:

PR #2475, opened 2016-08-30
docs: Explain how to change Sandstorm server hostname

paulproteus:


paulproteus:

Also how did we go so long without having docs for this...?

PR #2477, opened 2016-08-30
Add total charges from Blackrock to stats collection

jparyani:


shell/server/stats-server.js
```diff
@@ -182,6 +182,10 @@ function recordStats() {
     plans: planStats,
   };
   record.computeTime = Date.now() - now;
+  if (SandstormDb.getTotalCharges) {
+    // This only exists under Blackrock
+    record.totalCharges = SandstormDb.getTotalCharges();
```
zarvox: Perhaps this function should hang off of `BlackrockPayments` rather than monkey-patching `SandstormDb` then, if it only exists on Blackrock?

jparyani: BlackrockPayments is weird in that it may or may not exist, and checking for its existence is annoying (you have to check in globals). Attaching Blackrock methods to SandstormDb is the pattern we've used elsewhere.

zarvox: It's weird, because it reflects the reality that the code may or may not exist.  It's also more precise about what it's doing.

IIRC, we likely only added those two methods to `SandstormDb` because they were called by migrations, which at the time of authorship lived inside the `sandstorm-db` module, so we had to hoist the code into scope for the migrations, because `blackrock-payments` depended on `sandstorm-db`.

Since those two calls are stubs now, and migrations have been lifted out of the `sandstorm-db` package, the two migration-supporting functions can move to `BlackrockPayments` too. ;)

PR #2479, opened 2016-08-30
Opt Sandstorm for Work trial users into stats reporting

jparyani:
The hard coded date assumes we update our terms/privacy policy for the
feature key vendor by tomorrow.

Also add their customer ID to the stats object

### DIFF
--- a/shell/client/admin/stats.html
+++ b/shell/client/admin/stats.html
@@ -117,6 +117,11 @@
     </ul>
   </h1>
 
+  {{#if featureKeyForceStats}}
+    <div class="flash-message info-message">
+      During your free trial of Sandstorm for Work, the stats on this page will be reported to Sandstorm.io so that we can help you get up and running. Once you purchase a key (or downgrade to standard Sandstorm by removing your trial key), these stats will stop being reported.
+    </div>
+  {{else}}
   <h2>Publish anonymously to Sandstorm dev team</h2>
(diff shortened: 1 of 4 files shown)


shell/client/admin/stats.html
```diff
@@ -159,6 +159,9 @@
       Any stats shared are submitted anonymously.
     </p>
     <p>
+      {{#if featureKeyForceStats}}
+        You have a Sandstorm for Work trial key, and as part of that agreement, you elected to opt into stats reporting. Please upgrade to a paid key if you wish to disable this.
```
kentonv: Let's change this to:

> During your free trial of Sandstorm for Work, the stats on this page will be reported to Sandstorm so that we can help you get up and running. Once you purchase a key (or downgrade to standard Sandstorm by removing your trial key), these stats will stop being reported.

Additionally, in trial mode we should not show the heading "Publish anonymously to Sandstorm dev team" since trial stats are not anonymous. Let's replace the whole section. Maybe we should put it in a highlighted box?

shell/server/stats-server.js
```diff
@@ -187,12 +187,18 @@ function recordStats() {
   const age = ActivityStats.find().count();
   if (age > 3) {
```
kentonv: We should report stats immediately for trial users.

PR #2480, opened 2016-08-30
Don't strip binaries in the Sandstorm bundle

zarvox:
This allows `perf(1)` and friends to provide a better profiling experience.

The bundle size difference is fairly small (about 8MB uncompressed, <1MB compressed) and
there is basically no runtime overhead because the sections won't get faulted
into memory until something tries to make use of them.

### DIFF
--- a/make-bundle.sh
+++ b/make-bundle.sh
@@ -201,9 +201,9 @@ touch bundle/dev/{null,zero,random,urandom,fuse}
 mkdir -p bundle/usr/lib/locale
 localedef --no-archive --inputfile=./localedata-C --charmap=UTF-8 bundle/usr/lib/locale/C.UTF-8
 
-# Make bundle smaller by stripping stuff.
-strip bundle/sandstorm bundle/bin/*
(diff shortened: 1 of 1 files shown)


paulproteus:
Yay!!!​ Out of curiosity, does this mean addr2line and/or KJ backtraces
will be more informative?

zarvox:
My understanding is that this change means `addr2line` can give useful information when run against a release version of Sandstorm.

The KJ backtraces will remain inscrutable in release bundles because:

1. we don't ship a copy of `addr2line` in the Sandstorm bundle, which the pretty KJ backtrace generator depends on, and
2. we don't compile Sandstorm releases with `KJ_DEBUG`, without which the backtrace generator does not attempt symbol lookup.

But on the upside, you can run `addr2line` yourself and get actual symbols out of a release build.

kentonv:
Note that `addr2line` will only give you anything useful with `-f`, which tells it to output function names. Normally it would output file names and line numbers, but that information does not exist in release builds (stripped or not).

Issue #371, opened 2015-04-30
Log accesses to /proc from `spk dev` flag, and create optional `--with-proc` mode


paulproteus:
Rust's debugging flag, `RUST_BACKTRACE=1`, seems to require a working `/proc`. Other debugging-type things probably want `/proc` too. This ticket tracks the work to make that happen.

@kentonv believes that making `/proc` available when apps are running normally within the sandbox is too dangerous, but having it available as a debugging option for `spk dev` is fine.

@zarvox remarks that having `spk dev` _log_ accesses to `/proc` would be helpful, which resulted in the following proposal: In `spk dev` mode, we should log all `/proc` accesses, and if any occur, print a note saying, e.g., "The app accesed `/proc/self/maps`. You can get access to `/proc` in dev mode via `spk dev --with-proc`, but note that when your users are running the app, they won't get `/proc`."

Also, @kentonv remarks on IRC:

> the sandstorm supervisor has a `--proc` option, so it's just a matter of threading through for that. It also has a `--dev` option which is already enabled specifically when using `spk dev`, but I don't think we want to mount /proc by default because it might confuse people.

Thanks to @oshepherd for bringing this up, with regard to how it interacts with `RUST_BACKTRACE=1`.

dwrensha:
Looks like this will involve adding an argument to Backend.StartGrain().

```capnp
  startGrain @0 (ownerId :Text, grainId :Text, packageId :Text,
                 command :Package.Manifest.Command, isNew :Bool, devMode :Bool = false)
             -> (supervisor :Supervisor);
```

```capnp
  startGrain @0 (ownerId :Text, grainId :Text, packageId :Text,
                 command :Package.Manifest.Command, isNew :Bool,
 :Bool = false)
             -> (supervisor :Supervisor);
```

kentonv:
OMG ALL THAT TIME I SPENT IMPLEMENTING AGGREGATE DEFAULT VALUES MIGHT BE USEFUL!!!1

kentonv:
But `SupervisorOptions` should probably actually put defaults directly on the field declarations, rather than declare the defaults in the method decl, so... maybe my time spent was not useful after all. :(

(It does still make sense for `supervisorOptions` the parameter to declare a default value of `()`, though.)

Issue #1582, opened 2016-02-24
Alternative for /proc/self/exe


pgrm:
I'm trying to create a stack for .net core and it was really easy. I based it on their docker image and everything went fine until I tried starting the app. I got the following error:

    dnx: src/dnx/pal.linux.cpp:14: std::string GetNativeBootstrapperDirectory(): Assertion `ret != -1' failed.
    /opt/app/.sandstorm/launcher.sh: line 32:     3 Aborted                 /opt/DNX_BRANCH/runtimes/default/bin/dnx -p project.json web

turns out the code failing is 

    ssize_t ret = readlink("/proc/self/exe", buffer, PATH_MAX);

from [aspnet/dnx/src/dnx/pal.linux.cpp](https://github.com/aspnet/dnx/blob/7ac7929aa575e17b3c271e4a7a0c164418de0395/src/dnx/pal.linux.cpp). AFAIK one of the differences between sandstorm and docker containers is the missing /proc in sandstorm. Also, I couldn't find an alternative to `/proc/self/exe`

unfortunately my C++ skills are limited at best so I'm stuck here a bit. I also know that MS is working on replacing dnx with dotnet cli, but that doesn't help since it also will want to use the same functionality: https://github.com/dotnet/cli/search?utf8=%E2%9C%93&q=get_own_executable_path

Any suggestions how I could get ASP.Net Core 1.0 running in sandstorm?

PS here are my changes: https://github.com/pgrm/vagrant-spk/commit/ec5625d131eefe4e00107e234c5f4cbd9a163222

kentonv:
I would call this a bug in .NET: good Linux apps have a fallback if /proc isn't available, and generally the right way to find the executable path is based on argv[0] in any case.

That said, that doesn't help you!

Perhaps as a hack you could map a "fake" /proc into your app. In sandstorm-pkgdef.capnp you can define a mapping from /proc to some alternate directory in which you set up self/exe to be a symlink to your executable. As long as your app only has one .NET executable then it'll work?

pgrm:
@kentonv thx - that sounds like an easy solution, unfortunately it isn't. I've tried to create the link like this:

	mkdir -p /proc/self/
	ln -s /opt/DNX_BRANCH/runtimes/default/bin/dnx /proc/self/exe

[...]

PR #2481, opened 2016-08-31
Implement `spk dev --proc`.

dwrensha:
This allows dev-mode apps to opt-in to getting a procfs. I've verified that it does indeed allow Rust programs to generate human-readable stack traces (see #371).

Since `--proc` already exists as a supervisor flag, most of the changes here are to propagate the flag from `spk dev` to the supervisor. This involves writing a "1\n" or a "0\n" to a socket, adding a field to `DevPackages`, and adding an argument to `Backend.startGrain()`.

### DIFF
--- a/shell/imports/server/backend.js
+++ b/shell/imports/server/backend.js
@@ -119,10 +119,12 @@ class SandstormBackend {
     // package, so that the grain runs using the dev app.
     const devPackage = DevPackages.findOne({ appId: grain.appId });
     let isDev;
+    let mountProc;
     let pkg;
     if (devPackage) {
       isDev = true;
       pkg = devPackage;
+      mountProc = pkg.mountProc;
     } else {
       pkg = Packages.findOne(grain.packageId);
       if (!pkg) {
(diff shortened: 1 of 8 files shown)


zarvox:
FWIW, it looks like it's just `/proc/self/exe` that libbacktrace relies on: https://github.com/rust-lang/rust/blob/7a187c39c79e21cd61d1688d449bdd52d7510281/src/libbacktrace/fileline.c#L97 , in case this turns out to be something we want to make this available by default for apps.

paulproteus:
As a random comment here: .NET also needs `/proc/self/exe` #1582 .

LibreOffice needed `/proc/self/exe` and (seemingly) all the following, see: https://github.com/mnutt/davros/commit/48a10de1d55eab743667cbda60df118ce3aa462a

- /proc/filesystems
- /proc/meminfo (this surprises me)
- /proc/mounts
- /proc/self/exe (mentioned above)
- /proc/version

Perhaps vagrant-spk (via sandstorm-pkgdef.capnp) and/or Sandstorm-itself should specify a standard way to mock those out, or expose them to the grain.

kentonv:
Unfortunately, I don't think it's possible to map in /proc/self/exe without mapping in all of proc. The problem is that `/proc/self` is a symlink to the specific pid. Even if it's possible to map that in meaningfully (not sure), we'd then also have to somehow map in the specific pid number before it actually exists and before we know what it is.

I think @mnutt's approach is the right one. Usually these things can be faked adequately for any particular app. I'm OK with our tools assisting in setting up these fakes.

I do realize that it would be convenient for our developers if we could simply map in all of proc, but I cannot agree to do so. proc is huge, and the fact that there haven't been any CVEs reported against it lately seems more likely to be because no one is looking (because there is lower-hanging fruit elsewhere) than that it is secure. If we start allowing proc in production we have no hope of disabling it again in the future.

With all that said, I don't mind adding this flag to `spk dev`. (Reviewing now.)

mnutt:
This approach looks good to me, some other things that might be useful if they're possible:

* have `spk dev` track stat calls for inclusion in sandstorm-files.list
* have `spk dev` catch reads to `/proc` that aren't in `sandstorm-files.list` and print a warning with a link to the docs for stubbing them out

zarvox:
One potential issue I forsee is that the approach you're recommending is that (in the case of `/proc/self/exe`) it will only work for one executable that cares about what /proc/ looks like, per grain.  If an app has two executables with this requirement, you'd probably have to resort to some `LD_PRELOAD` hackery or something.

PR #2485, opened 2016-09-01
docs: Explain that there is no ARM support

### LINKED ISSUE #2432: docs: Add "Why not on ARM?" to administering/faq.md
Request by @ocdtrekkie --

> @paulproteus: IMHO, there should be an FAQ question somewhere for "why doesn't Sandstorm currently support ARM?". It gets asked a lot. By definition, a frequently asked question.

in https://github.com/sandstorm-io/sandstorm/issues/2083#issuecomment-225474293

paulproteus:
Close #2432


paulproteus:

PR #2489, opened 2016-09-01
Make `db.userGrains()` and `db.currentUserGrains()` more flexible.

dwrensha:
We've noticed that the observer [here](https://github.com/sandstorm-io/sandstorm/blob/990debe376cf545931853664834e880c5e665aeb/shell/packages/sandstorm-db/db.js#L2662-L2677) can spawn a large number of fibers when a user moves a bunch of grains into their trash. It seems that the problem has to do with the observed cursor's query, which includes `{ trashed: { $exists: false } }`. In addition to to the performance problem, there is a bug. We actually want the query to simply be `Grains.find({userId: this.userId})`.

This patch makes it easier to ask for what you want in `db.userGrains()` and `db.currentUserGrains()`. In particular, if you want *all* of the grains, including the trashed ones, you don't provide any additional arguments. So this fixes the problem case described above. It also clarifies other uses of the function.

### DIFF
--- a/shell/client/apps/app-details-client.js
+++ b/shell/client/apps/app-details-client.js
@@ -40,7 +40,7 @@ const compileMatchFilter = function (searchString) {
 };
 
 const appGrains = function (db, appId, trashed) {
-  return _.filter(db.currentUserGrains({ trashed: { $exists: trashed } }).fetch(),
+  return _.filter(db.currentUserGrains({ includeTrashOnly: trashed }).fetch(),
                   function (grain) {return grain.appId === appId; });
 };
 
--- a/shell/client/apps/app-details-client.js
+++ b/shell/client/apps/app-details-client.js
@@ -40,7 +40,7 @@ const compileMatchFilter = function (searchString) {
 };
 
 const appGrains = function (db, appId, trashed) {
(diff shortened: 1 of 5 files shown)


kentonv:
This is somewhat sad in that it means we once again have to remember to add `{trashed: {$exists: false}}` in most places where we'd call this. I think it was nice that the previous version of the method would omit trashed grains by default, which is usually what you want.

I also would like to move towards `SandstormDb` encapsulating the details of the underlying query language, so that in theory some day when we want to move away from Mongo we can do so without editing anything outside of SandstormDb.

Finally, `SandstormDb`'s methods are supposed to type-check their parameters as an added defense against injections, but it's not really possible or meaningful to type-check a Mongo query.

Can we adjust the interface so that the caller can choose between only-non-trashed, only-trashed, and everything? Or maybe it makes sense to have multiple methods?

dwrensha:
OK. Adjusted so that `userGrains()` takes an options object with two fields: `includeTrash` and `includeTrashOnly`. If you pass no options, you get no trash, consistent with the current behavior.

Note that I've adjusted a few `hasGrains` computations to use `includeTrash` so that we don't display an introjs animation to users whose only grains are in the trash.

PR #2490, opened 2016-09-01
Add button to feature key page linking to key vendor.

kentonv:
Any admin will be able to click on the button and arrive at the key's info in the feature key vendor, where they can do things like convert a trial key to permanent, specify payment source, purchase more users, change contact info, etc.

Currently the button looks like this:

screenshot from 2016-09-01 16-56-04

@neynah, @zarvox: Any opinions on the button style and label? Maybe this should be promoted to the primary button, or otherwise colored differently?

### DIFF
--- a/docs/administering/for-work.md
+++ b/docs/administering/for-work.md
@@ -4,7 +4,7 @@ Sandstorm server running at their own organization. Special pricing, sometimes i
 feature keys, are available to [non-profits and community groups](#special-pricing). For pricing,
 please read the information on the [main website](https://sandstorm.io/business).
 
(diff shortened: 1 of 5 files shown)


zarvox:
(as discussed in person, but leaving for archive)

I'm pleased with this - it's good to be able to navigate from one place to the other!

Let's make the text on that button be "Manage billing", make it primary, demote the "Update feature key" button to secondary, and make the "Manage billing" button be all the way to the right (first button in the DOM for screenreaders).

neynah:
I recognize this is from 4 days ago but I wanted to say that Drew's suggestion sounds good to me.

PR #2491, opened 2016-09-02
Add temporary work-around for fiber slow-down problems: Abort if too …

kentonv:
…many fibers are allocated.


zarvox:
:+1: :shipit:

PR #2495, opened 2016-09-02
When there are too many fibers, log sample stacks creating fibers.

kentonv:


shell/imports/server/async-helpers.js
```diff
@@ -2,12 +2,21 @@ import { Meteor } from "meteor/meteor";
 
 const Future = Npm.require("fibers/future");
 
+let inMeteorListener = undefined;
+const onInMeteor = (callback) => {
+  inMeteorListener = callback;
+}
```
zarvox: missing `;` here

shell/server/00-startup.js
```diff
@@ -85,17 +86,51 @@ Meteor.startup(() => { migrateToLatest(globalDb, globalBackend); });
...
+  // Sample calls to inMeteor().
+  onInMeteor(sampleStackIfTooManyFibers);
+
+  // Also take samples from various Meteor infrastructure that schedules fibers.
+  function addFiberSampling(prototype, functionName) {
+    const old = prototype[functionName];
+    prototype[functionName] = function () {
+      sampleStackIfTooManyFibers();
+      return old.apply(this, arguments);
+    }
```
zarvox: missing `;` here

zarvox:
semicolons aside, LGTM

kentonv:
I added a commit which monkey-patches console.{log,warn,error} to indicate the replica number, which should make blackrock logs easier to follow.

zarvox:
Perhaps we should prepend timestamps to all `console.log()` invocations too (or maybe do that on the C++ sandstorm side, since Meteor does so in dev mode already)?

Either way, LGTM.  :shipit:

kentonv:
@zarvox Blackrock already adds timestamps to logs, so what you describe should probably be a Sandstorm run-bundle change. I agree it would make a lot of sense but not in this PR.

PR #2497, opened 2016-09-02
Call BlackrockPayments.delete/suspendAccount

jparyani:


kentonv:
I think we should call the payments methods *before* updating Mongo because it would be pretty bad if a poorly-timed crash meant the user remained subscribed and couldn't do anything about it.

kentonv:
I guess I'll make the changes.

PR #2498, opened 2016-09-02
Build and ship a patched node in the Sandstorm bundle

zarvox:
v8's performance in the face of many fibers is currently quite poor because
they use a linked list for all thread control block lookups.  We've observed
frontends on Oasis sitting with 65% CPU usage just navigating that linked
list.  See #2484 for screenshots.

Switching to a real data structure (a hash map) improves this situation
drastically.  @jparyani's test app [1] spends 46% of CPU time in
Isolate::FindOrAllocatePerThreadDataForThisThread when run with the default
meteor-bundled node.  That function doesn't even register in the perf report
when run against a node compiled with the hashmap patch.

I'm not thrilled about building node and carrying this patch, so if we can get
a proper fix upstream, that would be preferable.

[1] - https://github.com/jparyani/meteor-fiber-repro

Profiles from the test app:

Bad:
perf-bad

Good:
perf-good

### DIFF
--- a/make-bundle.sh
+++ b/make-bundle.sh
@@ -85,12 +85,23 @@ METEOR_DEV_BUNDLE=$(./find-meteor-dev-bundle.sh)
 # Build patched nodejs from source.  Our patches significantly improve performance in the
 # presence of many fibers.  See https://github.com/sandstorm-io/sandstorm/pull/2484 and
 # https://github.com/sandstorm-io/node/tree/std-unordered-map-for-thread-data
-echo "Building node"
-pushd deps/node
+#
+# We build node out of tree because Jenkins wants to run builds in folders with spaces, and GNU make
+# makes dealing with spaces in implicit rules exceedingly difficult.  So we build in a fixed path in
+# /var/tmp.  If we were to vary the path, we would lose the ability to cache the build artifacts,
+# which is also rather undesirable.
+NODE_BUILD_ROOT=/var/tmp/sandstorm-node-build-dir
+echo "Building node out-of-tree"
+rm -rf "$NODE_BUILD_ROOT"
+mkdir -p "$NODE_BUILD_ROOT"
+cp -a deps/node "$NODE_BUILD_ROOT"
+pushd "$NODE_BUILD_ROOT/node"
 # The rebuild here is fast if nothing has changed.
 ./configure --partly-static
 make -j$(nproc)
 popd
+rm -rf deps/node/out
+mv "$NODE_BUILD_ROOT/node/out" deps/node/
 
 # Start with the meteor bundle.
 cp -r shell-build/bundle bundle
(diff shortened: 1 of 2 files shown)


dwrensha:
The node build process assumes that `#!/usr/bin/env python` is Python 2, so this breaks the build on my Arch box.

dwrensha:
https://github.com/nodejs/node/blob/19c9567686781b6cfd4d8b3908182d0df09c0d95/configure#L1

dwrensha:
https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/nodejs#n27

dwrensha:
I got my build to succeed like so:

```
mkdir ~/fake-python
ln -s /usr/bin/python2 ~/fake-python/python
PATH=~/fake-python:$PATH make update
```

... but this is not excellent.

make-bundle.sh
```diff
@@ -85,12 +85,23 @@ METEOR_DEV_BUNDLE=$(./find-meteor-dev-bundle.sh)
...
+rm -rf "$NODE_BUILD_ROOT"
+mkdir -p "$NODE_BUILD_ROOT"
+cp -a deps/node "$NODE_BUILD_ROOT"
+pushd "$NODE_BUILD_ROOT/node"
 # The rebuild here is fast if nothing has changed.
 ./configure --partly-static
 make -j$(nproc)
 popd
+rm -rf deps/node/out
+mv "$NODE_BUILD_ROOT/node/out" deps/node/
```
dwrensha: This cases Make to believe that its `deps/node` target has been updated.

It used to be that if you did `make update` twice in a row, the second time finished very quickly because Make detected that no work needed to be done. Now the second `make update` takes as long as the first.

This often comes up for me when I do `make update` and then wander off, causing the final `sudo` step to time out. It used to be that I could then do `make update` again and instantaneously finish the update.

zarvox: Sorry this is such a pain. :(

Maybe a workaround is to carefully push the changes down one folder instead:

 ```bash
mkdir -p deps/node/out
rm -rf deps/node/out/*
mv "$NODE_BUILD_ROOT/node/out/"* deps/node/out/
```

jparyani:
This adds about 20 minutes to the total build time on build.sandstorm.io. We should probably add some form of caching, or a more intelligent make-rule.

zarvox:
My changes originally avoided cleaning the node build directory in `make clean` for this reason.  It looks like https://github.com/sandstorm-io/sandstorm/commit/8493a727d6b73f1456187fc3ba93fe58cd27d0e1#diff-b67911656ef5d18c4ae36cb6741b7965R135 added the line which wipes out that cache.

@kentonv Would you mind removing that line, or suggesting an alternative approach to keep our tests fast?

zarvox:
Oh, on second thought, reading the node `Makefile`, it looks like `make clean` doesn't actually touch the root folder, so https://github.com/sandstorm-io/sandstorm/pull/2620 might make the caching work again.

zarvox:
Disregard the previous comment; I'm going crazy.  Of course removing the artifacts of the build is going to require that you rebuild the artifacts again.

PR #2509, opened 2016-09-07
Accounts page: wait for subscriptions

zarvox:
Without this, you get some nasty flashes of partial content on the page if you refresh `/account`

### DIFF
--- a/shell/client/shell-client.js
+++ b/shell/client/shell-client.js
@@ -909,6 +909,10 @@ Router.map(function () {
   this.route("account", {
     path: "/account",
 
+    waitOn() {
+      return globalSubs;
+    },
+
     data: function () {
       // Don't allow logged-out or demo users to visit the accounts page. There should be no way
       // for them to get there except for typing the URL manually. In theory showing the accounts


shell/client/shell-client.js
```diff
@@ -909,6 +909,10 @@ Router.map(function () {
   this.route("account", {
     path: "/account",
 
+    waitOn() {
+      return globalSubs;
```
paulproteus: +1

It's almost too bad we don't have a GlobalSubsHelper of some kind specific to the Sandstorm codebase.

kentonv:
Arguably we could make the top-level template refuse to render anything until `globalSubs` are ready. Thoughts?

zarvox:
The tradeoff is that you might wind up with cascading subscriptions, where you wait for `globalSubs`, incur the roundtrip waiting for the subs to be ready, and only once the subs are ready would we attempt to render the `{{yield}}`ed template, which would then requests and waits for more subs.  I'm not sure exactly how iron-router handles this situation.

I guess if all of our routes need everything from `globalSubs`, and nothing more, and no new routes will ever need anything else, then that's okay.  Otherwise, I'd want to verify the behavior of iron-router w.r.t. cascading subs before making that kind of change, and I'd probably prefer to do that in a separate PR.

PR #2511, opened 2016-09-07
Assorted db.js cleanups

zarvox:
* use function shorthand so you get named functions in stacktraces
* remove references to `globalDb` from db.js
* use this.collections.* with all collections (except Meteor.users)

### DIFF
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -1928,7 +1928,7 @@ _.extend(SandstormDb.prototype, {
     check(byAdminUserId, Match.OneOf(String, null, undefined));
     check(willDelete, Boolean);
 
-    const user = globalDb.collections.users.findOne({ _id: userId });
+    const user = this.collections.users.findOne({ _id: userId });
     const suspension = {
       timestamp: new Date(),
       willDelete: willDelete || false,
@@ -1971,7 +1971,7 @@ _.extend(SandstormDb.prototype, {
   unsuspendAccount(userId) {
     check(userId, String);
 
-    const user = globalDb.collections.users.findOne({ _id: userId });
+    const user = this.collections.users.findOne({ _id: userId });
     this.collections.users.update({ _id: userId }, { $unset: { suspended: 1 } });
     this.collections.grains.update({ userId: userId }, { $unset: { suspended: 1 } }, { multi: true });
 
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -966,7 +966,7 @@ SandstormDb = function (quotaManager) {
     settings: Settings,
     migrations: Migrations,
     // Omitted: StaticAssets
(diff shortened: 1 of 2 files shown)


shell/packages/sandstorm-db/db.js
```diff
@@ -966,7 +966,7 @@ SandstormDb = function (quotaManager) {
     settings: Settings,
     migrations: Migrations,
     // Omitted: StaticAssets
-    // Omitted: AssetUploadTokens
+    assetUploadTokens: AssetUploadTokens,
```
kentonv: So, in my ideal world `db.collections` wouldn't be a thing, an instead all of the collections would be private with specific methods on `SandstormDb` providing narrow access. This is why some collections were omitted here -- because the methods of `SandstormDb` already provided all the necessary operations on them, so I was trying to keep them private.

It looks like you're adding `assetUpdloadTokens` to `db.collections` here because you don't like referring to a global in the method below, but you aren't adding any external references. Is there a way we can solve both goals at once?

zarvox: How about I make a `Symbol` and store `AssetUploadTokens` on the SandstormDb instance at that symbol?

Perhaps I should do that for the other hidden collections as well?  Do you prefer a single hidden object behind one Symbol, or several hidden objects, each behind its own Symbol?  (I should do the same for `StaticAssets` too, I imagine.)

That said, I have found it useful in the past to be able to navigate to collection objects when debugging (especially since I haven't yet found a way to extract them from Meteor's internals).  For example, when I was attempting to replicate Oasis's plans for a local blackrock development instance, I wanted to access the `Plans` collection from the console, which I could only get a handle to through `globalDb.collections.plans`.

shell/packages/sandstorm-db/db.js
```diff
@@ -2583,15 +2574,15 @@ if (Meteor.isServer) {
           proof.status = "unverified";
         });
 
-        KeybaseProfiles.update(keyFingerprint, { $set: record }, { upsert: true });
+        this.collections.keybaseProfiles.update(keyFingerprint, { $set: record }, { upsert: true });
```
kentonv: `this` does not refer to the SandstormDb here because we're in a function(). Luckily I coincidentally noticed because an app update landed on my machine while I had run-dev going in a visible console window.

I think we need to be more cautious about these mass refactorings. We are very frequently breaking things and introducing bugs. We really need to develop much better tests before we can do this kind of stuff.

shell/packages/sandstorm-db/db.js
```diff
@@ -2795,10 +2786,10 @@ Meteor.methods({
   removeUserAction(actionId) {
     check(actionId, String);
     if (this.isSimulation) {
-      UserActions.remove({ _id: actionId });
+      this.collections.userActions.remove({ _id: actionId });
```
kentonv: This isn't even inside a SandstormDb method!

PR #2512, opened 2016-09-07
Harden against clickjacking and CSRF vulnerabilities in apps

kentonv:
See individual commits for more info.

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -1146,7 +1146,7 @@ tryProxyRequest = (hostId, req, res) => {
           // Deny requests coming from an external referer, since there is no legitimate use case
           // for these. Note of course that an attacker can trivially suppress the referer header
           // in practice, so this check alone does not give us any security benefit if we accept
-          // requests that lack a referrer.
+          // requests that lack a referer.
           if (referer != expectedOrigin && !referer.startsWith(expectedOrigin + "/") &&
               referer != mainUrl && !referer.startsWith(mainUrl + "/")) {
             throw new Meteor.Error(403, "Blocked illegal cross-origin referral from: " + referer);
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -1118,9 +1118,57 @@ tryProxyRequest = (hostId, req, res) => {
 
     return Promise.resolve(true);
   } else {
-    const isAlreadyOpened = req.headers.cookie && req.headers.cookie.indexOf("sandstorm-sid=") !== -1;
+    const isAlreadyOpened = req.headers.cookie &&
+        req.headers.cookie.indexOf("sandstorm-sid=") !== -1;
+
     return getProxyForHostId(hostId, isAlreadyOpened).then((proxy) => {
       if (proxy) {
+        // Let's do some CSRF defense.
+        const expectedOrigin = PROTOCOL + "//" + req.headers.host;
+        const mainUrl = process.env.ROOT_URL;
+        const origin = req.headers.origin;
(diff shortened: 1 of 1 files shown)


paulproteus:
This looks nice. I also hope that one day we can ditch hostname unguessability in general, as it remains a pain-point for people trying to set up Sandstorm. Perhaps suborigins will help with that.

shell/server/proxy.js
```diff
@@ -1118,9 +1118,57 @@ tryProxyRequest = (hostId, req, res) => {
...
     return Promise.resolve(true);
   } else {
-    const isAlreadyOpened = req.headers.cookie && req.headers.cookie.indexOf("sandstorm-sid=") !== -1;
+    const isAlreadyOpened = req.headers.cookie &&
+        req.headers.cookie.indexOf("sandstorm-sid=") !== -1;
+
     return getProxyForHostId(hostId, isAlreadyOpened).then((proxy) => {
       if (proxy) {
+        // Let's do some CSRF defense.
+        const expectedOrigin = PROTOCOL + "//" + req.headers.host;
```
thejh: Why don't you have to permit requests coming from `BASE_URL`?

thejh: You permit `mainUrl` in the `referer` case, but not the `origin` case, right?

kentonv: The reason `mainUrl` has to be permitted in the `referer` case is because the initial page loaded in the iframe will have the main URL as its referer. No browser sends an `Origin` header in this case.

That said, now that I think of it, I'm not entirely sure if browsers are allowed to send `Origin` in this case. I know they are explicitly not supposed to send an `Origin` header for hyperlinks, but I don't know about iframes. Do you happen to know? Maybe I should allow `mainUrl` as an origin to be safe?

shell/server/proxy.js
```diff
@@ -1118,9 +1118,57 @@ tryProxyRequest = (hostId, req, res) => {
...
+          }
+        } else if (referer) {
+          // Mark that we've seed a Referer header on this host, which indicates that the user's
+          // browser is not suppressing them.
+          proxy.seenReferer = true;
+
+          // Deny requests coming from an external referer, since there is no legitimate use case
+          // for these. Note of course that an attacker can trivially suppress the referer header
+          // in practice, so this check alone does not give us any security benefit if we accept
+          // requests that lack a referrer.
```
dwrensha: I recommend the "referer" spelling here, to be consistent with your other comments and the variable names.

dwrensha:
Looks good to me.

shell/server/proxy.js
```diff
@@ -1549,6 +1597,10 @@ class Proxy {
         response.setHeader("Set-Cookie", rpcResponse.setCookies.map(makeSetCookieHeader));
       }
 
+      const mainUrl = process.env.ROOT_URL;
+      response.setHeader("Content-Security-Policy", "frame-ancestors " + mainUrl);
```
dwrensha: Firefox is warning me:
```
Content Security Policy: Directive ‘frame-src’ has been deprecated. Please use directive ‘child-src’ instead.
```
which agrees with https://content-security-policy.com/

dwrensha: Huh... nevermind. Looks we don't actually set the `frame-src` directive from our code. Maybe Meteor's doing it.

kentonv: We use Meteor's BrowserPolicy API to set it. The difficulty here is that `child-src` is unsupported by many browsers, though maybe the situation is better now than when this was written...

PR #2518, opened 2016-09-09
Implement feature key auto-renewal

kentonv:
The commits are arranged in bite-sized chunks for your reviewing pleasure.

Overview:
* Once a feature key is installed, it will be automatically renewed.
    * To avoid accidents, we no longer allow installing a feature key that is expired, since this would trigger immediate renewal, which could be a costly mistake for the user.
    * However, if the user tries to install an expired key, we will go check if a newer version of the same key is available, e.g. because it had been renewed already, and the user just happened to use the wrong version. If so, we'll go ahead and use the new version. This will not trigger a renewal.
* If renewal fails, we will record the error.
    * We display it on the feature key config page, in a red box.
    * The box has a "retry" button to immediately retry renewal.
    * We also deliver a bell notification (with a scary red background) and an email notification (labeled "URGENT:").

Note that this doesn't yet implement actual enforcement of expired keys or user limits, but those will come soon once everyone has received plenty of warning and opportunity to do something about it.

screenshot from 2016-09-09 01-31-06
screenshot from 2016-09-09 01-32-56

### DIFF
--- a/shell/client/admin/feature-key-client.js
+++ b/shell/client/admin/feature-key-client.js
@@ -127,6 +127,7 @@ Template.featureKeyUploadForm.helpers({
 
 Template.adminFeatureKeyModifyForm.onCreated(function () {
   this.showForm = new ReactiveVar(undefined);
+  this.renewInFlight = new ReactiveVar(false);
 });
 
 Template.adminFeatureKeyModifyForm.helpers({
@@ -166,6 +167,10 @@ Template.adminFeatureKeyModifyForm.helpers({
   renewalProblem() {
     return globalDb.currentFeatureKey().renewalProblem;
   },
+
+  renewInFlight() {
+    return Template.instance().renewInFlight.get();
+  },
 });
 
 Template.adminFeatureKeyModifyForm.events({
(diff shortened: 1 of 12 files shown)


shell/server/feature-key.js
```diff
@@ -73,3 +74,252 @@ loadSignedFeatureKey = function (buf) {
...
+          `Failed to send deletion email to admin (id=${user._id}) with error: ${err}`);
+      }
+    });
+  }
+}
+
+let keyRenewalTimeout = null;
+let keyObserver = null;
+let currentlyRenewing = false;
+keepFeatureKeyRenewed = function (db) {
```
jparyani: Consider moving this whole file to `/imports` and importing this function in 00-startup.js

jparyani: Nevermind, it seems `loadSignedFeatureKey` is a weird global that needs to be exported for use in `db.js`.

shell/client/admin/feature-key-client.js
```diff
@@ -172,6 +181,21 @@ Template.adminFeatureKeyModifyForm.events({
...
+    const instance = Template.instance();
+    const state = Iron.controller().state;
+    const token = state.get("token");
+    const renewInFlight = instance.renewInFlight;
+    renewInFlight.set(true);
+    Meteor.call("renewFeatureKey", token, (err) => {
+      renewInFlight.set(false);
+      if (err) {
+        // Note: Renewal failures aren't reported this way. If we get here there was a bug.
+        console.error(err);
```
jparyani: Consider adding better error reporting here (even if it's just an `alert`). Most users won't think to look in the console when an unexpected error occurs, and may even think it succeeded.

jparyani:
LGTM other than my above comments

neynah:
I think we should style "Manage billing" as secondary rather than primary in this context.

paulproteus:
@neynah as a clarifying question - style **Manage billing** as secondary if there is a billing problem, else keep it primary?

kentonv:
@neynah Well... "Manage billing" is probably exactly where the user needs to go to fix billing problems.

neynah:
>@neynah as a clarifying question - style Manage billing as secondary if there is a billing problem, else keep it primary?

> @neynah Well... "Manage billing" is probably exactly where the user needs to go to fix billing problems.

I see that my comment was confusing. For some reason I thought the button on the far right took you to the same place as "Manage billing". So on second thought let's not make "Manage billing" secondary. Perhaps "Retry" should be secondary then.

PR #2521, opened 2016-09-09
Always show new shares at the top of the grain list.

### LINKED ISSUE #2516: When receiving a new shared grain, it should show at the top of my grain list
Steps to reproduce:

- Jade creates an Etherpad

- Jade shares it with Asheesh by identity

- Asheesh logs into that Sandstorm server

Expected behavior:

- That grain appears at the top of Asheesh's grain list

Actual behavior:

- The grain has no "last modified" time and therefore appears at the bottom

kentonv:
Fixes #2516

### DIFF
--- a/shell/client/grain/grainlist-client.js
+++ b/shell/client/grain/grainlist-client.js
@@ -134,9 +134,9 @@ const sortGrains = function (grains, sortRules) {
     } else {
       // Sorting by date.
       return function (a, b) {
-        // Null propagation on dates does the thing we want by default.
-        const aRaw = a[key];
-        const bRaw = b[key];
+        // Items with no lastUsed are new shares which we always want to show on top.
+        const aRaw = a[key] || new Date();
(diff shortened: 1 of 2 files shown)


jparyani:
I know this was the old behavior, but it always kind of bugged me that grains shared with me that I have no intention of ever opening will remain at the top of my grain list until I click them. That behavior is also somewhat non-intuitive (clicking them gives them an actual lastUsed date that will eventually let it sort and move down the list).

That being said, the code changes look fine to me given this is the behavior you want.

paulproteus:
I would like it if the value in the table were the string "New" or
"Unopened" or "Newly shared" maybe. I think I'm also OK with this behavior
as-is but wanted to indicate there might be other ways to express things
that minimize confusion.

kentonv:
OK, now we sort new shares by the time of the sharing action, but instead of displaying that time we display it as "Unopened".

jparyani:
LGTM

PR #2522, opened 2016-09-09
Don't hide topbar trash button when user has some tokens in trash and some not.

dwrensha:
Fixes a bug where sometimes the trash button does not show up in the topbar. The problem is that we are checking whether the user has the grain in the trash while what we should really be checking is whether the user has any *non-trashed* references to the grain.

Also adds a test that fails before and passes after the fix.

### DIFF
--- a/shell/client/grain-client.js
+++ b/shell/client/grain-client.js
@@ -731,8 +731,23 @@ Template.grain.helpers({
   },
 
   displayTrashButton: function () {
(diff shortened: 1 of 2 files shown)


paulproteus:
The :hooray: is for the test!

dwrensha:
unrelated ECONNREFUSED error in TestIpInterface

dwrensha:
Garply, retest this please

PR #2523, opened 2016-09-10
Add feedback input to account deletion

### LINKED ISSUE #2513: Add a feedback form to account deletion
We should add a text box where people can optionally write some feedback. The feedback should be included in the notification e-mail sent to the admin.

Suggested copy:

> Optional: Why have you decided not to use [server name]? This feedback will be sent to the server admin.

jparyani:
Fixes #2513

image

### DIFF
--- a/shell/client/accounts/account-settings.js
+++ b/shell/client/accounts/account-settings.js
@@ -304,7 +304,7 @@ Template.sandstormAccountSettings.events({
 
   "click button.delete-account-real": function (ev, instance) {
     instance._deleteSubmitting.set(true);
-    Meteor.call("deleteOwnAccount", instance.$(".feedback").val(), function (err) {
+    Meteor.call("deleteOwnAccount", instance.find(".feedback").value, function (err) {
       if (err) {
         instance._deleteError.set(err.message);
       } else {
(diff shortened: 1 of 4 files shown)


kentonv:
Using a multi-line (even just 2-3 lines) textarea would help signal to the user that this is a freeform text input. Usually a single-line input suggests something else, so could be confusing.

shell/client/accounts/account-settings.js
```diff
@@ -305,7 +304,7 @@ Template.sandstormAccountSettings.events({
 
   "click button.delete-account-real": function (ev, instance) {
     instance._deleteSubmitting.set(true);
-    Meteor.call("deleteOwnAccount", instance._deleteFeedback.get(), function (err) {
+    Meteor.call("deleteOwnAccount", instance.$(".feedback").val(), function (err) {
```
kentonv: Whoa I think this is the only use of jQuery anywhere in our code. Any reason you chose this instead of `.find(".feedback").value`?

jparyani: No particular reason, just forgot that we prefer `.find`

zarvox: FWIW we also seem to use jQuery in [shell/client/grain-client.js](https://github.com/sandstorm-io/sandstorm/blob/master/shell/client/grain-client.js#L389..L391)

PR #2465, opened 2016-08-29
Log a message if grain is slow to start

paulproteus:
This is intended to help:

- App authors who are mistaken about which port number their app was
  supposed to bind to, e.g. an app author a few months ago.

- App authors who forgot to ever actually listen on the port, e.g. an app
  author who showed up to IRC today.

PR #2532, opened 2016-09-12
Add X-Phabricator-* to the HTTP header whitelist

zarvox:
Phabricator's CSRF tokens for AJAX requests are passed in this manner, and
there are a couple other headers they use like `X-Phabricator-Via` that it'd be
better to not remove from requests.

cc @liamzdenek

### DIFF
--- a/src/sandstorm/web-session.capnp
+++ b/src/sandstorm/web-session.capnp
@@ -159,6 +159,7 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
       "oc-fileid",             # Owncloud client
       "oc-chunked",            # Owncloud client
       "x-hgarg-*",             # Mercurial client
+      "x-phabricator-*",       # Phabricator
     ];
   }
 


liamzdenek:
👍

paulproteus:
@zarvox any chance you'd be willing to get proxy.js to log when these headers are dropped? I'm also curious how you figured this out. :)

zarvox:
I was looking at the Phabricator grain log and the server code complained loudly about missing CSRF tokens.  I looked at the network tab and observed that there were some headers that weren't making it to the backend.

I'm hesitant to log *every* header that we drop, since that could potentially be an awful lot.  Perhaps we'd only log headers that are neither whitelisted nor explicitly ignored?  Maybe we'd only log those for grains running dev packages, to prevent this from being a potential DoS avenue?

paulproteus:
> I looked at the network tab and observed that there were some headers that weren't making it to the backend.

I see - you observed that the headers were being sent by your browser, but you determined they weren't being observed by Phabricator due to its CSRF errors, yeah?

> I'm hesitant to log every header that we drop, since that could potentially be an awful lot. Perhaps we'd only log headers that are neither whitelisted nor explicitly ignored? Maybe we'd only log those for grains running dev packages, to prevent this from being a potential DoS avenue?

Gentle +1 here.

Big +1 to "log headers that are neither whitelisted nor explicitly ignored".

For logging things like this, one thing I did in the past was in https://github.com/sandstorm-io/sandstorm/pull/2465 I log to the grain log, but max once per app launch. I chose that approach because I couldn't find a way to determine easily if a grain is in dev mode from the perspective of `sandstorm-http-bridge`. You might have not have that problem with proxy.js.

OTOH in your case, once-per-app-launch is still a bit annoying if random Internet people can fill up a server's disk by spamming the grain log. So I think "in dev mode" is smarter than "once per unique header."

A different approach, which if it is easy is awesome, is to console.log() the bad headers. Now there's no grain storage problem and it's still possibly discoverable.

Anyway do whatever makes sense to you, including your original plan. :)

kentonv:
We don't currently have any list of "headers which we knowingly ignore", so it might take some effort to build that. Even then, we're likely to see lots of spam from random weird agents that like to send random headers. For example, Verizon's super-cookie (which hopefully they've discontinued by now) was a header that would be sent to every server even though neither the server nor the client asked for it. So my expectation is that logging ignored headers would be overly spammy.

Maybe it would make sense specifically in dev mode.

PR #2535, opened 2016-09-12
Zero-downtime self-hosted updates.

kentonv:
Currently, there is a few-second window during updates when the user might experience connection errors.

We can close this window by opening our listen sockets higher-up in run-bundle and actually passing them on to the successor version on updates. Thus, the listen sockets never close, they just temporarily pause accepting connections.

### DIFF
--- a/src/sandstorm/run-bundle.c++
+++ b/src/sandstorm/run-bundle.c++
@@ -1849,6 +1849,16 @@ private:
           KJ_UNREACHABLE;
         }
       } else if (siginfo.ssi_signo == SIGINT) {
+        // Frontend startup or shutdown request, used with run-dev.
+
+        if (!siginfo.ssi_int) {
+          // We have to close all the listen ports otherwise run-dev won't be able to open them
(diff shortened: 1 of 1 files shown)


paulproteus:
Not a tech review, just a product remark:

This is _really great_ because for people whose Sandstorm servers are behind reverse proxies, reverse proxies sometimes cache the outage, resulting in a bigger perceived outage.

paulproteus:
@kentonv out of curiosity, will this require me to rebase the app market deploy off the new Sandstorm bundle?

kentonv:
@paulproteus Only if you want app market deploys to be zero-downtime, which seems like it would be nice. Other than that, there's no immediate need to change anything because the interface between run-bundle.c++ and meteor-bundle-main.js hasn't changed (as you can tell since meteor-bundle-main.js is not modified in this PR).

Note that if you do update the bundle, you probably need to update the app market to Meteor 1.4. Test carefully before deploying.

paulproteus:
@kentonv Thanks!

jparyani:
LGTM

PR #2538, opened 2016-09-13
Don't hide "Upload app" button while searching on app list page

### LINKED ISSUE #2038: "Upload app..." button disappears during search
When typing in the search box the "Uninstall..." button remains but the "Upload app..." button disappears.

The behaviour is strange but I don't think this is super urgent.
image
image

zarvox:
Now that we've moved the two buttons up above the horizontal rule, they should
no longer be confused as being related to search results.

Fixes #2038.

### DIFF
--- a/shell/client/apps/applist.html
+++ b/shell/client/apps/applist.html
@@ -17,7 +17,7 @@
                     title="Uninstall an app"
             >Uninstall...</button>
           {{/if}}
-          <button type="button" class="upload-button {{#if searching}}hide{{/if}}">Upload app...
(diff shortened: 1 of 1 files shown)


paulproteus:
Humbly requesting screenshot; tentative +0. :)

zarvox:
I'm not sure a screenshot adds anything, but here you go.

screenshot

paulproteus:
I'm +1 now. I had misunderstood what UI this patch would change. The screenshot fixed my misunderstanding.

Thanks @zarvox for humoring me! :)

PR #2540, opened 2016-09-13
Locale aware grain and app sorting

### LINKED ISSUE #2529: Grain list uses naïve text sorting
If the grain list is sorted by name, text appears to be sorted by character code and not the proper Unicode algorithm, e.g. “É” will appear last rather than next to “E”, and “•” is not at the same end as “-”.

(As always, I wish this constituted a bug in An App That Does Grain Management rather than Sandstorm itself. Collections is progress there…)

zarvox:
If available, use [String.prototype.localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) when sorting app titles or grain titles.  (It's [not available in some versions of Safari](http://caniuse.com/#feat=internationalization), so we retain the fallback logic of comparing lowercased strings directly.)

Fixes #2529.

### DIFF
--- a/shell/client/apps/applist-client.js
+++ b/shell/client/apps/applist-client.js
@@ -114,11 +114,32 @@ Template.sandstormAppListPage.helpers({
   apps: function () {
     const ref = Template.instance().data;
     const apps = matchApps(ref._filter.get());
-    return _.chain(apps)
-            .map(appToTemplateObject)
-            .sortBy(function (appTemplateObj) { return appTemplateObj.appTitle.toLowerCase(); })
-            .sortBy(function (appTemplateObj) { return appTemplateObj.dev ? 0 : 1; })
-            .value();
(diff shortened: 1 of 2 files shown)


zarvox:
I looked into what it would take to make locale-aware search a thing; it seems it's much more involved due to our desire to match substrings.  (`localeCompare()` offers no way to look for a matching substring.)  We'll basically have to roll our own locale-aware `indexOf`, which I'm afraid will have negative perf implications.

An idea, if we find the "proper" implementation too slow: we may be able to approximate the desired behavior by:

* converting to NFKD
* applying [toLocaleLowerCase()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleLowerCase)
* removing any characters that are not alphanumeric
* then matching with `indexOf()`

I'd rather do that in a separate PR, if at all.

PR #2544, opened 2016-09-13
docs: Provide sample code for handling Handle

paulproteus:
This takes a recommendation from @kentonv 's recent conversation on sandstorm-dev and adds a link to @dwrensha 's GitLab code.

Review requested from at least one of:

- @dwrensha 

- @kentonv 

I'm willing to continue to make changes this, or use other people's preferred text, as needed.

### DIFF
--- a/docs/developing/auth.md
+++ b/docs/developing/auth.md
@@ -63,11 +63,17 @@ related to user identity and permissions:
   duplicates (such as prompting the user to choose a different one,
   or just appending some digits). Apps should strongly consider
   using display names (`X-Sandstorm-Username`) instead of handles.
-  **WARNING: A user can change their preferred handle at any time.
+  WARNING: A user can change their preferred handle at any time.
   Two users can have the same preferred handle. The preferred handle
   is just another form of display name. Do not use preferred handles
   as primary keys or for security; use `X-Sandstorm-User-Id`
-  instead.**
+  instead.
+
+    * **Our recommendation on how to use this:** Use the preferred handle as a hint when creating a
+      new account. If there is already an account for the user ID, don't change its handle. If
+      another account exists with the same handle, then append some numbers to make the new one
+      unique (this should be rare). This is for example how Gitlab works on Sandstorm. See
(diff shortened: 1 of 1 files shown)


kentonv:
This recommendation only applies when packaging apps that are already coded around the assumption of a username that is a primary key. For new code I would recommend avoiding usernames as a primary key.

paulproteus:
OK @kentonv how about now? :)

docs/developing/auth.md
```diff
@@ -63,11 +63,22 @@ related to user identity and permissions:
...
   Two users can have the same preferred handle. The preferred handle
   is just another form of display name. Do not use preferred handles
   as primary keys or for security; use `X-Sandstorm-User-Id`
-  instead.**
+  instead.
+
+    * **Our recommendation on how to use this:** If you need a unique identifier, use the
+      `X-Sandstorm-User-Id`. If your app requires that handles also be unique, then use the
+      following algorithm. If there is already an account for the user ID, don't change its
+      handle. If another account exists with the same handle, then append some numbers to make the
```
kentonv: I feel like this algorithm description is in the uncanny valley between precise and informal. How about:

    If your app requires that handles be unique -- for example, because the app is designed to use "usernames" as the primary user identifier -- then you will need to track and store those handles in your app's storage. We suggest implementing the following algorithm, for each request:
    1. If there is already a user account in your app's database associated with the request's `X-Sandstorm-User-Id`, use the handle associated with it.
    2. Otherwise, if no existing user account is already associated with `X-Sandstorm-Preferred-Handle`, create a new account for this user's ID and handle, and use that.
    3. Otherwise, this is a new user but their handle is already in use. Append a number to the handle such that it is unique, create a user account using the user's ID and the modified handle, and use that.

ocdtrekkie:
I'm going to close this one out, it is already the largest section for the most discouraged use part of this page (and this adds an entire paragraph in a block with no paragraphs), and is more guidance on how one could adapt this feature if you want to misuse it, depending on ancient sample code and specific use cases you might find when porting a legacy app. I don't really think it fits, and I don't feel it's necessary. It may add more confusion than simply expressing to people what they should not do here, which we already do.

Issue #2545, opened 2016-09-14
Update all self-description of Sandstorm to say it's a productivity suite


paulproteus:
Context:
- @neynah recently noticed that https://sandstorm.io/ says "Sandstorm is a self-hostable web productivity suite" but other places don't.

Here's a list of places that need updating:
- [x] The info page for the Sandstorm demo: https://oasis.sandstorm.io/demo
  - PR in progress: https://github.com/sandstorm-io/sandstorm/pull/2546
- [x] Our GitHub: https://github.com/sandstorm-io/sandstorm
  - PR in progress #2547
- [x] Our blog sidebar: https://sandstorm.io/news/
  - PR in progress https://github.com/sandstorm-io/sandstorm-website/pull/227
- [x] Oasis landing page: :https://oasis.sandstorm.io/ (view in incognito window)
  - Issue in progress https://github.com/sandstorm-io/sandstorm-website/issues/228
- [x] App market homepage: "Sandstorm.io is an open source platform for servers."
  - PR in progress: https://github.com/sandstorm-io/sandstorm-app-market/pull/104 ... **Done!**
- [x] GitHub README for this repo
  - PR in progress #2547
- [x] Docs index page
  - PR: https://github.com/sandstorm-io/sandstorm/pull/2554

@neynah anywhere else? :)

kpreid:
(I still think this is a bad idea, per the email thread, but:) the app market.

paulproteus:
@kpreid Thanks! I appreciate your willingness to help, doubly so given that you think it's a bad idea.

ocdtrekkie:
Closing this because 6 of 7 boxes were checked, the 7th referred to a PR that was, in fact, merged. So I checked that box too.

PR #2554, opened 2016-09-14
docs: Call it a productivity suite

paulproteus:
See also #2545


paulproteus:

PR #2559, opened 2016-09-15
fix new user validation to include 'pronoun' instead of 'pronouns' 

dwrensha:
"pronoun" is the correct field to be using here: https://github.com/sandstorm-io/sandstorm/blob/v0.185/shell/packages/sandstorm-db/db.js#L86

I don't think we ever actually fill in this field when creating a new user (usually it gets filled in lazily, with a default being inferred from the service-specific identity data), so I don't think there would be any point in writing a migration to clean up database entries that have a `profile.pronouns` field.


dwrensha:
Garply, retest this please

dwrensha:
transient ECONNREFUSED error, look like

PR #2560, opened 2016-09-15
fix broken 'this' reference in `removeApiTokens()`

dwrensha:
The bug was introduced in https://github.com/sandstorm-io/sandstorm/commit/d4f96cea87024305daa6f3ebcb50e1c69ddfe07d.


dwrensha:
A symptom of this bug is:

```
Exception in setTimeout callback: TypeError: Cannot read property 'apiHosts' of undefined
```

PR #2562, opened 2016-09-16
Implement powerbox requests for identities.

dwrensha:
Allows a user to introduce one of their contacts to a grain, giving the grain access to the contact's profile information while simultaneously granting the contact access to the grain.

I've updated the sandstorm-test-python app to test out this functionality. I also have a Wekan branch that uses it.

Looks like:
contacts

Future work is to figure out some nice way to include the identity-provider icon and the intrinsic name on these powerbox cards.

### DIFF
--- a/shell/server/identity.js
+++ b/shell/server/identity.js
@@ -79,8 +79,8 @@ globalFrontendRefRegistry.register({
                                 IdentityRpc.PersistentIdentity);
   },
 
-  validate(db, session, identityId, options) {
-    check(identityId, String);
+  validate(db, session, value) {
+    check(value, { id: String, roleAssignment: SandstormDb.prototype.roleAssignmentPattern, });
 
     if (!session.userId) {
       throw new Meteor.Error(403, "Not logged in.");
@@ -94,8 +94,8 @@ globalFrontendRefRegistry.register({
       },
       session.grainId,
       "petname",
-      options.roleAssignment,
-      { user: { identityId: identityId, title: grain.title, }, });
+      value.roleAssignment,
+      { user: { identityId: value.id, title: grain.title, }, });
 
     return {
       descriptor: {
@@ -104,19 +104,20 @@ globalFrontendRefRegistry.register({
             id: Identity.typeId,
             value: Capnp.serialize(
               Identity.PowerboxTag,
-              { identityId: new Buffer(identityId, "hex"), }),
+              { identityId: new Buffer(value.id, "hex"), }),
           },
         ],
       },
       requirements: [
         {
           permissionsHeld: {
-            identityId: identityId,
+            identityId: value.id,
             grainId: session.grainId,
             permissions: [],
           },
         },
       ],
+      frontendRef: value.id,
     };
   },
 
--- a/shell/server/identity.js
+++ b/shell/server/identity.js
@@ -97,6 +97,14 @@ globalFrontendRefRegistry.register({
       value.roleAssignment,
       { user: { identityId: value.id, title: grain.title, }, });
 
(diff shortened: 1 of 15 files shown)


src/sandstorm/identity.capnp
```diff
@@ -39,6 +39,13 @@ interface Identity {
 
   # TODO(someday): Public key info? Ability to seal messages / check signatures?
 
+  struct PowerboxTag {
+    # Tag to be used in a `PowerboxDescriptor` to describe an `Identity`.
+
+    identityId @0 :Data;
```
kentonv: Hmm, why in the tag and not returned by a method call?

shell/server/identity.js
```diff
@@ -57,9 +58,72 @@ makeIdentity = (identityId, requriements) => {
...
 globalFrontendRefRegistry.register({
   frontendRefField: "identity",
+  typeId: Identity.typeId,
 
   restore(db, saveTemplate, identityId) {
     return new Capnp.Capability(new IdentityImpl(db, saveTemplate, identityId),
                                 IdentityRpc.PersistentIdentity);
   },
+
+  validate(db, session, identityId, options) {
```
kentonv: I'm concerned that this `options` isn't being type-checked.

More generally, possibly validate() has the wrong signature. It should input a "frontendRef request", which may or may not look the same as a final frontendRef. It should then return -- in addition to `descriptor` and `requirements` -- the final ref.

This relates to why validate() is documented as being allowed to modify the frontendRef passed in -- some other ref types really send up more of a "request" which validate() munges into the final form. But you can't quite do that here since an identity ref is just a string, not an object, so it's not something you can modify in-place.

src/sandstorm/identity.capnp
dwrensha: There has been some talk about us eventually moving to a model where `identityId` is not global and stable.  I think adding something like `Identity.getId()` or `Profile.identityId` could make such a change trickier.

shell/server/identity.js
dwrensha: I've updated the signature of `validate()`. Now it takes in `frontendRefRequest`, which must be type-checked, and returns `frontendRef`.

dwrensha:
I've added a `permissions` field to `Identity.PowerboxTag`. This is important in Wekan because viewers that don't have the "edit" permission cannot be assigned a card, so when Wekan uses the Powerbox to request an identity to assign a card, it needs to be able to say "I only want identities that have the edit permission".

src/sandstorm/identity.capnp
kentonv: It seems equally tricky or perhaps more tricky if it's in the powerbox tag, though.

For a method call, we can intercept and map at the realm gateway pretty easily.

Another possibility is to have `SandstormApi.getIdentityId(identity)`.

To deal with PowerboxTag we'd need to add a new system for intercepting and transforming powerbox descriptors in queries and fulfillments, which so far we don't have any need to do.

(Though admittedly it's unclear when you would ever want an Identity to be fulfilled from another grain, in which case all of this may be moot?)

kentonv:
I think you may need to update BlackrockPayments as well for the validate() signature change.

src/sandstorm/identity.capnp
dwrensha: OK, I've removed `PowerboxTag.identityId` and added `SandstormApi.getIdentityId(identity)`, which is the option that I think has the easiest-to-understand transition path.

It occurs to me that `PowerboxTag.permissions` might eventually require us to do some kind of descriptor interception, at least on the fulfillment side. However, I'm not to worried about that, as it seems like it would be relatively straightforward to handle.

Issue #2414, opened 2016-08-18
I can receive notifications on grains I can no longer access


dwrensha:
1. Alice shares an Etherpad grain to Bob.
  2. Bob starts a comment thread.
  3. Alice revokes Bob's access to the grain.
  4. Alice replies to the comment thread.
  5. Bob receives a notification, but clicking on it fails to open the grain.

The grain should not be allowed to notify Bob once Bob has lost access. One way to enforce that would be to add a `permissionsHeld` membrane requirement in [`makeIdentity()`](https://github.com/sandstorm-io/sandstorm/blob/09a08c2dc215b20a3b12bc11c370429086a16825/shell/server/identity.js#L28-L32).

PR #2565, opened 2016-09-16
add membrane requirements to identities

dwrensha:
Our current `makeIdentity()` function ignores its `requirements` parameter. Therefore `Identity` capabilities passed to grains have no membrane requirements, and hence #2414.

This patch migrates existing saved identity capabilities to add the appropriate membrane requirement. It also removes the buggy `makeIdentity()` function in favor of `FrontendRefRegistry.create()`, which additionally validates the requirement. (A potential downside here is that this triggers another permissions computation. I don't think we've seen performance problems related to permissions computations yet, so maybe this isn't worth worrying about.)

This patch helps with https://github.com/sandstorm-io/sandstorm/issues/2414, but is not a complete fix; it does not kill any live identity capabilities and hence still requires grains to be rebooted before the revocation takes full effect.

### DIFF
--- a/shell/imports/server/migrations.js
+++ b/shell/imports/server/migrations.js
@@ -700,6 +700,31 @@ const setNewServer = function (db, backend) {
   }
 };
 
+const addMembraneRequirementsToIdentities = function (db, backend) {
+  const query = {
+    "frontendRef.identity": { $exists: true, },
+    "owner.grain.grainId": { $exists: true, },
+    "requirements.0": { $exists: false, },
+  };
+
+  db.collections.apiTokens.find(query).map((apiToken) => {
+    db.collections.apiTokens.update(
+      { _id: apiToken._id },
+      {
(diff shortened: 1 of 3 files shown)


kentonv:
I'd prefer to fix `makeIdentity()` by having it add `requirements` to the `saveTemplate`. Doubling the number of permissions checks we do makes me uncomfortable.

kentonv:
FWIW, makeIdentity() is intended to be an internal API. I don't know that type-checking `requirements` is too important here, particularly if it's going to make the code ugly.

dwrensha:
Well, I don't always trust myself to pass well-formed values to a function, and typechecking seems especially prudent here where the value will be written to the database.

kentonv:
Garply, retest this please.

kentonv:
Ugh I'm just going to merge this and run the tests locally.

PR #2566, opened 2016-09-16
Add standalone domain support

jparyani:
To test it out, from a `meteor shell` run `globalDb.collections.standaloneDomains.insert({_id: "test.local.sandstorm.io", token: "$SHARING_TOKEN"})

This still isn't completely done, and has some outstanding issues in no particular order:
* Add login overlay postmessage api
* Make all shell overlay popups be centered
* Supress Sandstorm autoupdate refresh banner?
* Remove old hostname from window.title
* Change email login to send the link for the correct domain
* Add admin page for setting up standalone domains?

Logging in with github has been tested working. To test, change the `isStandalone` helper in shell-client.js to return false.

### DIFF
--- a/shell/client/grain-client.js
+++ b/shell/client/grain-client.js
@@ -20,6 +20,7 @@ import { introJs } from "intro.js";
 
 import downloadFile from "/imports/client/download-file.js";
 import { ContactProfiles } from "/imports/client/contacts.js";
+import { isStandalone } from "/imports/client/standalone.js";
 
 // Pseudo-collections.
 TokenInfo = new Mongo.Collection("tokenInfo");
@@ -72,7 +73,7 @@ const mapGrainStateToTemplateData = function (grainState) {
     hasNotLoaded: !(grainState.hasLoaded()),
     sessionId: grainState.sessionId(),
     originalPath: grainState._originalPath,
-    interstitial: grainState.shouldShowInterstitial(),
+    interstitial: !isStandalone() && grainState.shouldShowInterstitial(),
     token: grainState.token(),
     viewInfo: grainState.viewInfo(),
     signinOverlay: grainState.signinOverlay(),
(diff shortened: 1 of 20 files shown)


shell/client/grain-client.js
```diff
@@ -20,6 +20,7 @@ import { introJs } from "intro.js";
 
 import downloadFile from "/imports/client/download-file.js";
 import { ContactProfiles } from "/imports/client/contacts.js";
+import { isStandalone } from "/imports/client/standalone.js";
```
kentonv: It looks like you forgot to add this file to the commit. :(

shell/server/proxy.js
```diff
@@ -1165,7 +1182,7 @@ tryProxyRequest = (hostId, req, res) => {
           // requests that lack a referer.
           if (referer != expectedOrigin && !referer.startsWith(expectedOrigin + "/") &&
               referer != mainUrl && !referer.startsWith(mainUrl + "/")) {
-            throw new Meteor.Error(403, "Blocked illegal cross-origin referral from: " + referer);
+            // throw new Meteor.Error(403, "Blocked illegal cross-origin referral from: " + referer);
```
kentonv: Un-comment please.

shell/server/proxy.js
```diff
@@ -1616,8 +1633,8 @@ class Proxy {
 
       const mainUrl = process.env.ROOT_URL;
       const grainHost = PROTOCOL + "//" + request.headers.host;
-      response.setHeader("Content-Security-Policy", "frame-ancestors " + mainUrl + " " + grainHost);
-      response.setHeader("X-Frame-Options", "ALLOW-FROM " + mainUrl);
+      // response.setHeader("Content-Security-Policy", "frame-ancestors " + mainUrl + " " + grainHost);
```
kentonv: Un-comment please.

jparyani: Sorry, I should've commented on the issue. This isn't ready for merge yet as it's blockking on this.

shell/client/grain-client.js
```diff
@@ -2024,3 +2025,112 @@ Router.map(function () {
...
+        this.state.set("beforeActionHookRan", true);
+
+        Tracker.nonreactive(() => {
+          const token = globalDb.collections.standaloneDomains.findOne({
+            _id: window.location.hostname, }).token;
+          const path = "/" + (this.params.path || "") + (this.originalUrl.match(/[#?].*$/) || "");
+          const hash = this.params.hash;
+
+          const tokenInfo = TokenInfo.findOne({ _id: token });
+          if (!tokenInfo) {
```
kentonv: There's a lot of code here that is copy/pasted from elsewhere. We should factor it out so that it can be shared.

shell/server/proxy.js
```diff
@@ -380,7 +381,23 @@ Meteor.methods({
       }
     }
 
-    if (this.userId && !incognito) {
+    if (this.userId && standalone) {
+      if (!SandstormPermissions.mayOpenGrain(globalDb, { token: apiToken })) {
```
kentonv: This block appears to be an exact copy of the anonymous case, below. Please just adjust the if() statement below so that these cases take the same code path, rather than copy/pasting a bunch of code.

kentonv:
Oops, apparently I'm reviewing code that isn't ready for merge yet. I'll stop. :)

jparyani:
This is now ready for review.

jparyani:
Merge conflicts with email code resolved. Otherwise this was unchanged.

shell/server/pre-meteor.js
```diff
@@ -548,32 +549,70 @@ Meteor.startup(() => {
...
+              const parsedRedirect = Url.parse(state.redirectUrl);
+              const redirectHostname = parsedRedirect.hostname;
+              return inMeteor(function () {
+                if (globalDb.hostIsStandalone(redirectHostname)) {
+                  res.writeHead(302, { "Location": parsedRedirect.protocol + "//" +
+                    parsedRedirect.host + req.url, });
+                  res.end();
+                  return;
+                } else {
+                  throw new Meteor.Error(400, "redirectUrl in OAuth was for an unknown host: " +
```
kentonv: You haven't handled the main domain here, so currently OAuth login doesn't work on the main domain.

shell/server/pre-meteor.js
```diff
@@ -548,32 +549,70 @@ Meteor.startup(() => {
...
         for (let i = 0; i < meteorRequestListeners.length; i++) {
           meteorRequestListeners[i](req, res);
         }
       } else {
-        // Otherwise, dispatch to our own middleware proxy chain.
-        nonMeteorRequestHandler(req, res);
-        // Adjust timeouts on proxied requests to allow apps to long-poll if needed.
-        WebApp._timeoutAdjustmentRequestCallback(req, res);
+        return inMeteor(function () {
+          if (globalDb.hostIsStandalone(hostname)) {
```
kentonv: Can we do some sort of caching to avoid this DB query on every HTTP request?

Something like:
- If it's a wildcard host and the host ID is in the Proxy map, we know it's an app host.
- We can similarly maintain a map of known grain public IDs, and avoid any lookups if it's in there.
- Finally, we can also cache known standalone hosts, so that only the first request to a host hits the DB.

shell/server/pre-meteor.js
```diff
@@ -591,12 +630,25 @@ Meteor.startup(() => {
...
         return true;
       } else {
-        const id = matchWildcardHost(req.headers.host);
-        if (id) {
-          return tryProxyUpgrade(id, req, socket, head);
-        } else {
-          return false;
-        }
+        return inMeteor(function () {
+          // TODO(someday): This is slightly sad that some requests will now need a Meteor fiber
```
kentonv: I'm more worried about the DB query.

Note that we could have all DDP WebSockets go to the main host rather than handle them on the standalone host. This will presumably already happen on Oasis since it sets DDP_DEFAULT_CONNECTION_URL to ddp.oasis.sandstorm.io.

shell/client/grain-client.js
```diff
@@ -2037,3 +2038,117 @@ Router.map(function () {
...
+            return this.next();
+          } else if (tokenInfo.invalidToken) {
+            this.state.set("invalidToken", true);
+          } else if (tokenInfo.revoked) {
+            this.state.set("revoked", true);
+          } else if (tokenInfo.identityOwner) {
+            this.state.set("invalidToken", true);
+          } else if (tokenInfo.alreadyRedeemed) {
```
jparyani: This is an incredibly weird code path to hit, since you would have to have visited the grain using exactly the same webkey as the standalone domain. Having been shared the grain in any other manner wouldn't trigger this.

That being said, it's obviously a bug. I'll go ahead and add a parameter to the tokenInfo sub.

shell/server/proxy.js
```diff
@@ -348,6 +353,7 @@ Meteor.methods({
 
     const token = params.token;
     const incognito = params.incognito;
+    const standalone = globalDb.getStandaloneDomainForToken(token);
```
jparyani: What's the purpose behind parentOrigin here? Do you mean for it to be passed to the Proxy and enforced there?

kentonv: Yes.

kentonv:
I plan to finish up the remaining comments tomorrow before pushing a release. Let me know before then if you're still working on anything.

PR #2568, opened 2016-09-17
Email notifications for app activity

zarvox:
For preliminary review.  Don't merge yet, see below.

We should probably figure out what else is needed for us to consider this feature fully-baked and something that we're willing to unleash on our users.  For instance, we probably don't want Rocket.Chat sending you an email for every single message sent while you're not looking at Sandstorm, which is probably what happens out of the box.

----

This changeset implements client-side activity tracking for detecting when a user is definitely present, and marking notifications as delivered if a user has been detected as present within the past two minutes.

If, after the desktop-notification delivery period, no browser tab has marked the notification as delivered to the user, this code generates an email with a link back to the grain/path where the activity took place, and sends it.  The email also has a capability link which mutes that user's subscription to that particular thread of that grain.


zarvox:
Closing to avoid accidentally sending way too much email to all our users.

zarvox:
I think my favorite possible outcome here is for this to move behind a per-user feature flag, and give users the ability to opt in/out of this feature through a route that we don't link to.  That way we can all enable it for ourselves, in development and on Oasis, and avoid bitrot, while buying the time to figure out notification policies and (if needed) get app updates to make this useful.

Releasing this should probably also depend on https://github.com/kentonv/etherpad-lite/issues/32

zeigerpuppy:
this would be fantastic, and an excellent solution for notifications.
for example: https://github.com/dwrensha/gitlab-sandstorm/issues/22

I agree that there should be an option, probably on a per-grain basis for allowing notifications to go out by email.

I think the most logical interface for this would be:
1) a checkbox in the list of grains
2) a checkbox in the "notifications" menu when viewing an app

Pressing the checkbox cycles through:
1) notification in Sandstorm only,
2) notification to Email only
3) notification to both
4) back to no notification

Or 2 checkboxes could be used

notification2
notification1

PR #2572, opened 2016-09-20
Setup wizard adjustments from user testing feedback

zarvox:
These are mostly just the super low-hanging fruit:

* Correct the price in the setup wizard to our sticker price of $10/user/month
* Show completed progress bar on the final page of the setup wizard
* Deemphasize steps other than the current one
* Improve language in step titles and use nouns more consistently

### DIFF
--- a/shell/client/setup-wizard/wizard.html
+++ b/shell/client/setup-wizard/wizard.html
@@ -16,7 +16,7 @@
 --}}
 <li class="setup-progress-step step-{{stepName}} {{className}}">
   {{#if isCurrentStep}}
-    <span>
+    <span class="setup-current-step">
       {{> Template.contentBlock}}
     </span>
   {{else}}
@@ -25,7 +25,7 @@
         {{> Template.contentBlock}}
       {{/linkTo}}
     {{else}}
-      <span>
+      <span class="setup-not-current-step">
         {{> Template.contentBlock}}
       </span>
     {{/if}}
--- a/shell/client/setup-wizard/wizard.html
+++ b/shell/client/setup-wizard/wizard.html
@@ -44,7 +44,7 @@
                                    mayJumpTo=(mayJumpTo "identity")
                                    route="setupWizardIdentity"
                                    className=itemClassName}}
-      Login provider(s)
+      Identity providers
     {{/setupWizardProgressBarItem}}
 
     {{#if hasFeatureKey}}
@@ -63,7 +63,7 @@
                                    route="setupWizardEmailConfig"
                                    className=itemClassName
                                    }}
-      Set up email
+      Email
     {{/setupWizardProgressBarItem}}
 
     {{#setupWizardProgressBarItem stepName="preinstalled"
@@ -72,7 +72,7 @@
                                    route="setupWizardPreinstalled"
                                    className=itemClassName
                                    }}
-      Pre-installed Apps
+      Pre-installed apps
     {{/setupWizardProgressBarItem}}
 
     {{#setupWizardProgressBarItem stepName="user"
@@ -262,7 +262,7 @@
 
 <template name="setupWizardIdentity">
 <div class="setup-page-content">
-  <h1 class="setup-page-title">Configure Identity Provider(s)</h1>
+  <h1 class="setup-page-title">Identity providers</h1>
   {{> setupWizardProgressBar currentStep="identity"}}
 
   <p>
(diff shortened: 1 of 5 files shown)


shell/client/setup-wizard/wizard.html
```diff
@@ -44,7 +44,7 @@
                                    mayJumpTo=(mayJumpTo "identity")
                                    route="setupWizardIdentity"
                                    className=itemClassName}}
-      Login provider(s)
+      Identity providers
```
paulproteus: Aw, I preferred the string "Login provider(s)" but I can accept whatever if we're consistent.

shell/client/setup-wizard/wizard.html
```diff
@@ -262,7 +262,7 @@ <h1 class="setup-page-title">Sandstorm for Work feature key</h1>
 
 <template name="setupWizardIdentity">
 <div class="setup-page-content">
-  <h1 class="setup-page-title">Configure Identity Provider(s)</h1>
+  <h1 class="setup-page-title">Identity providers</h1>
```
paulproteus: I see this says "Identity providers" so, well, OK. See above. Consistency is priority no. 1. In my little world, we'd call these Login Providers. I can live with either.

shell/client/setup-wizard/wizard.html
kentonv: I think I agree with @paulproteus -- while "identity providers" feels more technically correct, I think "login providers" is probably easier for people to grasp quickly. Of course, if user testing suggested otherwise I defer to that.

neynah: No definitive conclusions from user testing other than it's a somewhat unclear what "Identity providers" means. I would expect "Login providers" to be more clear to users.

kentonv: Sounds like we should go with "Login providers" then.

zarvox: The logical conclusion of this (switching from noun "identity" to noun "login") is that we wind up with "Linked logins" and "allow this login to log in to your account?" on the /account page, which I find more confusing and likely to be seen by a greater number of users.  I find this a worse overall outcome.

Another reason I shy away from "login" as a noun is that we already use it as a verb.

Is there some other word we can use that will work well in all the contexts?  FWIW I don't think we've received criticism of the term "identity" being problematic; only that inconsistency in copy is confusing.

paulproteus: For your specific question: I think calling them "Login providers" is OK even if they create "identities" for users.

In general: I'm +1 on separating this discussion into a separate ticket/PR. That way, we can have a conversation that isn't a blocker on an existing PR. For my part, I don't need the perfect to be the enemy of the good here.

neynah: I'm OK with either. Also happy to go back to each of the users and ask them about these specific strings (to see which one makes more sense to them) along with a screenshot for context.

paulproteus: One other thing:

> I don't think we've received criticism of the term "identity" being problematic; only that inconsistency in copy is confusing.

I think @neynah above says that we have received criticism of the term "identity" being confusing, unless I'm misunderstanding her.

Regardless I'm still +1 on making this discussion not be a blocker for this particular PR. Let's take the time to make a careful decision, and until then, carefully make no decision. The places where this PR changes the string are OK with me; they increase consistency.

I filed https://github.com/sandstorm-io/sandstorm/issues/2584 to track that conversation.

neynah: >I think @neynah above says that we have received criticism of the term "identity" being confusing, unless I'm misunderstanding her.

Not exactly! I just got that impression so I wouldn't take it as fact.

Also, I just asked 2 of the users right now and they both prefer "Identity providers"! So I'd say it's pretty safe to go with Drew's choice of string.

kentonv: > The logical conclusion of this (switching from noun "identity" to noun "login") is that we wind up with "Linked logins" and "allow this login to log in to your account?"

Our aim is to use the language that is easiest to understand, not the language which is the most consistent. We don't need to be pedantically consistent when it harms clarity.

> Another reason I shy away from "login" as a noun is that we already use it as a verb.

"Login" is always a noun. "Log in" is a verb. If we use "login" as a verb we should correct it.

If you mean that we shouldn't use both "login" and "log in"... I agree not in the same sentence (because it sounds silly), but I don't think there's a problem with using both of these terms.

> Also, I just asked 2 of the users right now and they both prefer "Identity providers"!

Surprising, but this is a good reason to use "identity provider".

I do want to bring up one other point, though: Our future plans around identity and accounts actually calls for us to disassociate identity from login providers. Will using "identity provider" now create additional confusion later, when we make that change?

PR #2577, opened 2016-09-21
docs: Remark that whitelabeling is in SfW

paulproteus:


paulproteus:

PR #2581, opened 2016-09-22
Combine the first two parameters of Proxy's contructor.

dwrensha:
https://github.com/sandstorm-io/sandstorm/commit/0084bc458a3c4b516af738ce94705f41fd5de09f introduced a bug where the second argument to `Proxy`'s constructor is the caller of the Meteor method `openSession()`, rather than the owner of the grain. This can lead to a situation where `Backed.getGrain()` gets called with the wrong `ownerId`. (Earlier I had mentioned that I thought this bug might explain some of our resumable-sessions problems, but I am less sure of that now.)

The present patch should make such mistakes more difficult to make. It consolidates the first two parameters of `Proxy.constructor` into a single parameter `grain`, intended to be filled in directly with an entry from `Grains`.

If and when we implement a by-package-ID CSRF whitelist, this change should also make it easy for a `Proxy` to know its package ID -- it can read it directly from the `grain` parameter.

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -320,7 +320,7 @@ Meteor.methods({
     const opened = globalBackend.openSessionInternal(grainId, this.userId, identityId,
                                                      null, null, cachedSalt);
     const result = opened.methodResult;
-    const proxy = new Proxy(grainId, this.userId, result.sessionId,
+    const proxy = new Proxy(grain, result.sessionId,
                             result.hostId, result.tabId, identityId, false, opened.supervisor);
     proxiesByHostId[result.hostId] = proxy;
     return result;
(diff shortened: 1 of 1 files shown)


paulproteus:
"Can we avoid this bug through code re-use?" was a question I meant to ask in the morning meeting. So -- +1 to this code unification, even if it doesn't necessarily fix the resumable session issues I'm experiencing. :)

kentonv:
From my reading of the code, if a supervisor capability is provided to Proxy's constructor, then ownerId is never actually used: in maybeRetryAfterError(), immediately after resetConnection() we call globalBackend.continueGrain() to get a new supervisor cap, passing only the grain ID. It seems that Proxy.getConnection() (which direcgtly calls globalBackaend.cap().getGrain() passing both ownerId and grainId) is only ever used if no supervisor cap is passed to the constructor.

@dwrensha is that your conclusion as well?

dwrensha:
Ah! That would explain why we didn't notice the problem until now. I agree that it looks like the current code never actually sends the wrong `ownerId` to `getGrain()`.

PR #2583, opened 2016-09-23
Fix that the preinstalled app package IDs could become outdated.

kentonv:
When an app update is discovered through an app index refresh, if the newer version of the app was already installed (for example, because someone had gone to the app market and installed it manually before the index update landed), then the preinstall package ID was never updated. This means the old version continued to be preinstalled and also broke appdemos of the particular app.


kentonv:
This fix is incorrect. 965470d is the correct fix. (I think. Hard to test.)

PR #2587, opened 2016-09-23
Fix bugs around deleteUser()

kentonv:
See individual commit messages.

### DIFF
--- a/shell/server/demo-server.js
+++ b/shell/server/demo-server.js
@@ -50,7 +50,7 @@ function cleanupExpiredUsers() {
   // Delete expired demo accounts and all their grains.
 
   const now = new Date(Date.now() - DEMO_GRACE_MS);
-  Meteor.users.find({ expires: { $lt: now }, loginIdentities: {$exists: true} },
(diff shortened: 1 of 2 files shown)


shell/server/demo-server.js
```diff
@@ -53,21 +53,15 @@ function cleanupExpiredUsers() {
   Meteor.users.find({ expires: { $lt: now } },
                     { fields: { _id: 1, loginIdentities: 1, lastActive: 1, appDemoId: 1 } })
               .forEach(function (user) {
+    console.log("delete demo user: " + user._id);
+    if (user.loginIdentities) {
+      globalDb.deleteAccount(user._id, globalBackend);
```
dwrensha: This can trigger deletion of identities which maybe already be queued up to be deleted in a later step of this `forEach()` loop. That shouldn't cause any correctness problems, but it does seem a bit awkward.

Perhaps it would be better make two queries on `Meteor.users()`; in the first, iterate through and delete expired accounts, and in the second delete any remaining expired identities.

kentonv: Done.

dwrensha:
Looks good to me.

PR #2588, opened 2016-09-23
re-enable Origin-based CSRF protection, whitelisting Gogs

dwrensha:
@kentonv mentioned that there were some apps other than Gogs that were apparently triggering the warning message, so those should be investigated before merging this.

The whitelist should probably also include older versions of Gogs, but I was unable to find them.

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -1136,20 +1136,17 @@ tryProxyRequest = (hostId, req, res) => {
           //     https://bugzilla.mozilla.org/show_bug.cgi?id=446344
           // I haven't found any information on IE/Edge's behavior.
           if (origin !== expectedOrigin) {
-            // TODO(security): Alas, it turns out we have apps that have:
-            //   <meta name="referrer" content="no-referrer">
-            // as Chrome sends "Origin: null" in these cases. :( These apps need to switch to:
-            //   <meta name="referrer" content="same-origin">
-            // It's important that we don't break apps, so we will accept null origins for now,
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
I can get you a full list of package IDs for any app when I get home in a few hours.

kentonv:
lol, thanks @ocdtrekkie, but we actually do have these records. :)

zenhack:
This logic has moved to the gateway; proxy.js no longer exists. The comments in the current version suggest that "The affected apps appear to be limited to Etherpad and Gogs." Are we confident enough in that assessment to merge an equivalent patch for the current implementation?

kentonv:
@zenhack Well, it's possible that browser behavior has changed over the years... We may have to re-analyze this.

Actually, at this point it may make more sense to use the new `SameSite=strict` cookie flag on UI session cookies. That would protect against CSRF attacks from separate domains, though I suppose a CSRF is still possible from an app hosted on the same sandstorm server?

zenhack:
Yeah, from what I can tell samesite=strict would treat everything under the sandstorm's wildcard domain as the same site.

zenhack:
Just checked, can confirm that chrome still behaves as described in the comment (as does Firefox now; looks like the issue linked in the comment finally got addressed).

So a version of this change would still make sense (probably _in addition to_ samesite=strict). But probably we should close this and open a separate issue for that, since this is a patch to the old js version of the code that doesn't exist anymore.

zenhack:
I opened #3358 for the issue this is trying to address (and #3357 for the samesite=strict idea). We can close this.

PR #2589, opened 2016-09-23
Correctly handle activity from anonymous users.

kentonv:


shell/packages/sandstorm-db/db.js
```diff
@@ -731,7 +733,11 @@ const DesktopNotifications = new Mongo.Collection("desktopNotifications", collec
 //   notificationId: String.  ID of the matching event in the Notifications table to dismiss if this
 //                            notification is activated.
 //   appActivity: Object with fields:
-//     user: {               The user whose action caused this notification.
+//     user: {               The user whose action caused this notification. Not present if this
+//                           notification wasn't generated by a user. {anonymous: true} if this
```
zarvox: This `{anonymous: true}` here caught me off-guard - I don't usually look for caveats of other allowable structures in the comment block on the right when reading over otherwise-clear shapes on the left.

Can you make this something like:

```
user: Optional Object.  Not present if this notification wasn't generated by a user.  If present, it will have one of the following shapes:
  { anonymous: true } if this notification was generated by an anonymous user.  Otherwise:
  {
     identityId: String  The user's identity ID
     name: String        The user's display name
     avatarUrl: String   The URL for the user's profile picture.
  }
```

zarvox:
Looks good, thanks!

kentonv:
> Are there any actual use cases where we think people will combine `backgroundActivity()` and `notifySubscribers: true`?

Lots of cases. Two off the top of my head:
* A build bot might use `backgroundActivity()` when a build fails.
* An RSS reader might use `backgroundActivity()` when a new item arrives.

PR #2590, opened 2016-09-23
Work around template rendering timing issue causing duplicate topbar registrations.

kentonv:
This was sometimes (non-deterministically) causing a blank page when the demo expired (rather than showing the login form).

### DIFF
--- a/shell/packages/sandstorm-ui-topbar/topbar.js
+++ b/shell/packages/sandstorm-ui-topbar/topbar.js
@@ -434,7 +434,8 @@ SandstormTopbar.prototype.addItem = function (item) {
     // Duplicate item. This can sometimes happen due to template redraw timing issues: the old
     // item was supposed to be removed before the new item was added, but things were scheduled
     // in the wrong order. So, we replace the old item with the new one, and also make sure to
-    // close the old item.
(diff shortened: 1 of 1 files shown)


shell/packages/sandstorm-ui-topbar/topbar.js
```diff
@@ -431,7 +431,14 @@ SandstormTopbar.prototype.addItem = function (item) {
   }
 
   if (item.name in this._items) {
-    throw new Error("duplicate top bar item name:", item.name);
+    // Duplicate item. This can sometimes happen due to template redraw timing issues: the old
+    // item was supposed to be removed before the new item was added, but things were scheduled
+    // in the wrong order. So, we replace the old item with the new one, and also make sure to
+    // close the old item.
```
zarvox: The comment about closing the old item is a little confusing, since you don't actually call the close method on that item.  I see that it's correct, but it took a couple minutes of rereading to realize that the item itself doesn't hold the `close()`, and there's no `dispose(item)` on `this` or anything like that available, just that closure at the end.

I guess the topbar could probably benefit from a refactor.

zarvox:
:shipit:

PR #2591, opened 2016-09-23
Reinstate installer fixes

paulproteus:


paulproteus:

kentonv:
Why did this have to be reverted originally, and why is it better now? Would be nice to have some context here.

paulproteus:
Hi @kentonv and thanks for your patience with my answering.

I believe that it was removed in the past due to breaking installer-tests, which I believe related to the fact that there was an issue with how we integrate sandcats-dev.sandstorm.io with GlobalSign's non-production certificate authority API. In particular, they want us to switch which non-production certificate authority API endpoints we're calling, but I haven't made that code change yet in sandcats. Therefore, the installer-tests that verify the HTTPS bringup behavior were failing. This made installer-tests red, which was a dangerous place to leave things.

My laptop was surprise-failing at the time, I believe, and I asked Jason to revert the installer changes I had made.

This change reinstates most of the changes, but it leaves the HTTPS-bringup tests disabled until I have time to fix the sandcats-dev.sandstorm.io service to use GlobalSign's new API endpoint.

Having said all that, I do _also_ vaguely recall there was some reason of our own that we wanted to revert these installer changes. Let me read Rocket.Chat scrollback...

I see the following notes after spending some time searching:

- @dwrensha mentioned that `$ ./install.sh -d -u sandstorm-0-fast.tar.xz ` was acting weird, but @zarvox worked with him and found that it was a problem relating to David's use of both 127.0.0.1 and 127.0.0.2 which confused the install script, which is at least not a very common situation. David later-on says he would not describe this as a regression.

I believe that's the only issue people raised with these changes when they were first made. That was on/around August 1.

I hope that helps. Thanks for asking, and for your patience with my answer.

PR #2596, opened 2016-09-27
Fix findIdentitiesByEmail()

kentonv:
This isn't called from the Sandstorm codebase, but is called by the code which receives mailing list subscribe/unsubscribe events from Mailchimp and updates bonuses accordingly. Probably, a few people who unsubscribed are still receiving the bonus. Oh well.

Also fixed it to handle LDAP correctly, even though this code is not currently used on any server that uses LDAP.


dwrensha:
Looks good to me. I hope someday we can statically catch such bugs.

PR #2598, opened 2016-09-27
Correctly serve utf8 filenames for backups

### LINKED ISSUE #2597: Download backup of grain with non-ascii-clean title can fail
To reproduce:

* Open https://oasis.sandstorm.io
* Create a new Wekan grain
* rename the grain "šos"
* click the download arrow to attempt to download a backup

Expected behavior: browser downloads a file named šos.zip

Actual behavior: browser says something like "/tmp/mozilla_zarvox0/tuPw8fkf.zip.part could not be saved, because the source file could not be read", and you get no backup.

----
This seems to not affect all apps equally.  Attempting the same with Etherpad successfully downloads an appropriately-named zip with correct content.

It's not immediately clear to me if this is a bug in the frontend, the backend, their interactions, or some combination thereof.

jparyani:
This follows RFC 6266, which all browsers have supported for a while now.

Fixes #2597

### DIFF
--- a/shell/server/backup.js
+++ b/shell/server/backup.js
@@ -183,7 +183,7 @@ Router.map(function () {
       }
 
       let started = false;
-      const filename = (token.name.replace(/["\n]/g, "") || "backup") + ".zip";
+      const filename = encodeURIComponent(token.name || "backup") + ".zip";
       let sawEnd = false;
 
       const stream = {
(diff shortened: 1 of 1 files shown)


shell/server/backup.js
```diff
@@ -183,7 +183,7 @@ Router.map(function () {
       }
 
       let started = false;
-      const filename = (token.name.replace(/["\n]/g, "") || "backup") + ".zip";
+      const filename = encodeURIComponent(token.name || "backup") + ".zip";
```
kpreid: Might I suggest naming the variable "encodedFilename" or similar, so the uses of it below look less suspicious?

PR #2599, opened 2016-09-28
Don't count trashed grains against grain quota.

kentonv:
Users find this confusing.

A better solution would be to automatically delete trashed grains as needed when the user is running up against quota limits, but that would be trickier to implement, and as a result we haven't done it. In the meantime, this stopgap is much better than nothing.

Technically a user could abuse this by swapping grains in and out of their trash, but whatever.

### DIFF
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -1896,7 +1896,7 @@ _.extend(SandstormDb.prototype, {
 
     const plan = this.getUserQuota(user);
     if (plan.grains < Infinity) {
-      const count = this.collections.grains.find({ userId: user._id },
+      const count = this.collections.grains.find({ userId: user._id, trashed: { $exists: false } },
         { fields: {}, limit: plan.grains }).count();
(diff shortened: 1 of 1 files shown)


dwrensha:
To be consistent, we should update the "Grains I own" count to not include trashed grains: https://github.com/sandstorm-io/sandstorm/blob/v0.187/shell/client/grain/grainlist-client.js#L286

dwrensha:
I guess "My grains' total size" needs to still count the sizes of grains in the trash? I can see how this might get confusing: you can have zero non-trashed grains but still be consuming disk space.

dwrensha:
Hm. It looks like it would be possible, though maybe a bit tricky, to update the disk quota check to account for trashedness. If such an update is out-of-scope for this pull request, then I suppose that the numbers we report for both "Grains I own" and "My grains' total size" ought to continue to include trashed grains. Because these numbers appear right next to each other on the page, I think it would be very confusing if one included trashed grains and one didn't.

kentonv:
Yeah, for this pull request, all I want to do is make it less likely that people get paywalled when they don't expect to be.

PR #2604, opened 2016-09-29
Add overlaySignin postMessage method

jparyani:
Adds a postmessage signin overlay method. Usable by doing:
`window.top.postMessage({overlaySignin: {top: 20, left: 20}}, '*');`

and:
`window.top.postMessage({overlaySignin: {disable: true}}, '*');`

This doesn't handle the case of not being inside a standalone domain, because it's a bit tricky with the topbar changing size at the mobile breakpoint.

### DIFF
--- a/shell/imports/client/grain/grainview.js
+++ b/shell/imports/client/grain/grainview.js
@@ -275,6 +275,10 @@ class GrainView {
     return this.fullTitle().title;
   }
 
+  signinOverlay() {
+    return this._signinOverlay;
+  }
+
   fullTitle() {
     // Returns the user's name for this grain, not the browser tab title.
     // Three cases:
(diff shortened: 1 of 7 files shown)


kentonv:
Hmm, does it work to make this part of the grain view template rather than part of the topbar template? That would presumably help with positioning in the presence of topbar/sidebar and would help with grain switching. We are likely to have more per-grain overlay content in the future (e.g. around inline powerbox) so it probably makes sense to think about where such overlays should go.

shell/imports/client/grain/grainview.js
```diff
@@ -275,6 +275,10 @@ class GrainView {
     return this.fullTitle().title;
   }
 
+  signinOverlay() {
+    return this._signinOverlay;
```
kentonv: Needs a `this._dep.depend()`.

PR #2605, opened 2016-09-29
Make system log accessible during setup wizard

kentonv:
This:
* Allows access to `/admin/status` with only a setup token (not logged in).
* Adds a button to the setup wizard footer to open the log.
* Improves some adjacent code.

This is important because often debugging login problems requires looking at the system log. #2603 in particular adds more debug logging around SAML and adds a link to `/admin/status` to the SAML error page; this isn't helpful if the user has to be logged in to view the log.

### DIFF
--- a/shell/client/setup-wizard/wizard.html
+++ b/shell/client/setup-wizard/wizard.html
@@ -100,6 +100,7 @@
 <template name="setupWizardHelpFooter">
 <div class="setup-help-row">
   <span class="setup-help-label">Need help?</span>
+  <a class="setup-help-link" target="_blank" href="/admin/status">System Log</a>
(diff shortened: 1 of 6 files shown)


shell/client/setup-wizard/wizard.html
```diff
@@ -100,6 +100,7 @@
 <template name="setupWizardHelpFooter">
 <div class="setup-help-row">
   <span class="setup-help-label">Need help?</span>
+  <a class="setup-help-link" target="_blank" href="/admin/status">System Log</a>
```
zarvox: Marginally weird that this is visible even if you're not authenticated, but I'm not convinced it's worth hiding.

PR #2606, opened 2016-09-30
Allow apps to display 'data:' and 'blob:' URLs in their iframe.

### LINKED ISSUE #2602: Content Security Policy blocks dynamically generated pdf from downloading
Hi I'm slowly porting [invoiceninja](https://www.invoiceninja.com/). It dynamically creates a pdf with [pdfmake](http://pdfmake.org/playground.html) and then allows the user to download it.

When I'm inside the grain URL, e.g. https://24ba72c5324e3dea1d5ae1bf73e7e448.xxx.sandcats.io/dashboard this works fine but when I'm inside the sandstorm UI e.g. https://xxx.sandcats.io/grain/NZosSy6HJnux4nCkuqeyJ7 i get the following content security policy message in the javascript console

`Content Security Policy: The page’s settings blocked the loading of a resource at blob:https://24ba72c5324e3dea1d5ae1bf73e7e448.xxx.sandcats.io/0e3d5466-ecc3-4b32-9234-0681604cdd0b (“frame-src https://*.xxx.sandcats.io”).`

Any help on how to fix this would be appreciated.
Thanks

kentonv:
The purposes of restricting frame-src is more to prevent weird behavior than it is security: We don't want the grain iframe to browse to an external site because that would be weird and confusing. I don't see any reason to prevent data or blob URLs, though, and they are very useful for some kinds of apps.

Fixes #2602

### DIFF
--- a/shell/server/shell-server.js
+++ b/shell/server/shell-server.js
@@ -40,6 +40,8 @@ Meteor.startup(() => {
   const frameSetter = () => {
     BrowserPolicy.content.disallowFrame(); // This clears all the old rules
     BrowserPolicy.content.allowFrameOrigin(getWildcardOrigin());
+    BrowserPolicy.content.allowFrameOrigin("data:");
+    BrowserPolicy.content.allowFrameOrigin("blob:");
     const billingPromptUrl = globalDb.getBillingPromptUrl();
     if (billingPromptUrl) {
       BrowserPolicy.content.allowFrameOrigin(billingPromptUrl);


kentonv:
This might not be safe. See discussion on #2602.

ocdtrekkie:
@zenhack Is this something that would be worth investigating the concerns of? A good number of app export/download features end up blocked by this.

zenhack:
Agree it's worth following up on, left a comment in the other issue; not sure I quite understand what the issue is.

zenhack:
We should close this; the relevant code no longer exists, since the gateway was written in C++. There's still discussion to be had about whether an analogous change makes sense, but it would have to be a different patch.

PR #2610, opened 2016-09-30
Allow for ldap self signed certs

jparyani:
It requires the user to copy/paste the pem key into the admin config page.

### DIFF
--- a/shell/imports/server/accounts/ldap.js
+++ b/shell/imports/server/accounts/ldap.js
@@ -73,11 +73,15 @@ LDAP.prototype.ldapCheck = function (db, options) {
     };
 
     if (fullUrl.indexOf("ldaps://") === 0) {
+      const tlsOptions = {};
+      const cert = db.getLdapCaCert();
+      if (cert) {
+        tlsOptions.ca = cert;
+      }
+
       client = ldapjs.createClient({
         url: fullUrl,
-        tlsOptions: {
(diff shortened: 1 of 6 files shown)


shell/imports/server/accounts/ldap.js
```diff
@@ -76,7 +76,7 @@ LDAP.prototype.ldapCheck = function (db, options) {
       client = ldapjs.createClient({
         url: fullUrl,
         tlsOptions: {
-          ca: [_this.options.ldapsCertificate],
+          ca: [db.getLdapCaCert() || false],
```
zarvox: Is `false` a valid input here?  https://nodejs.org/dist/latest-v4.x/docs/api/tls.html#tls_tls_connect_options_callback seems to indicate it expects either a string, buffer or array of either of those.

Depending on how the implementation details work out, it seems like this maybe has the potential to break non-self-signed certs, by specifying an invalid cacert here?

Maybe this is better written as:

```js
const cert = db.getLdapCaCert();
const tlsOptions = {};
if (cert) {
  tlsOptions.ca = cert;
}
client = ldapjs.createClient({
    ...
    tlsOptions,
});
```

jparyani: I was going off what was originally here (which was originally written by the meteor-accounts-ldap project that we forked from). In it, `_this.options.ldapsCertificate` defaults to `false`.

PR #2615, opened 2016-10-01
Automatically remove '---BEGIN CERTIFICATE---', etc. from SAML certs.

kentonv:
It's probably typical for users to include these when pasting.


jparyani:
LGTM

PR #2617, opened 2016-10-03
Make it possible to view the system log inline in the setup wizard

### LINKED ISSUE #1956: Make it possible to obtain server logs from within setup wizard
See #1955 for an instance of where this would have been helpful.

It's difficult to debug SMTP delivery failure, particularly in the setup wizard, where you can't see the server log without SSHing into the machine.

We should make it possible to download and/or view the server log (where an exception stacktrace might be recorded) from the setup wizard.  I think I prefer the "download", since it's less UI to build/let users get lost in, and if you're having a problem that you can't figure out yourself, it's helpful if you can upload or email the log somewhere.

I imagine this would be something like a button labeled "Download server log" attached to the failure message box on the email configuration/test email sending page.

zarvox:
The link at the bottom of the page is now turned into a button that shows the log inline on the page.

Closes #1956 

logged-in

unauthorized

### DIFF
--- a/shell/client/styles/_setup-wizard.scss
+++ b/shell/client/styles/_setup-wizard.scss
@@ -52,6 +52,22 @@
     margin: 4px;
     text-decoration: none;
   }
+
+  .setup-system-log {
+    color: #000000;
+  }
+
+  .setup-system-log-header {
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    justify-content: space-between;
+
+    .setup-system-log-header-text {
+      font-size: 24px;
+      font-weight: normal;
+    }
+  }
 }
 
 .setup-progress {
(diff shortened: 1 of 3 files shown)


shell/client/styles/_setup-wizard.scss
```diff
@@ -52,6 +52,22 @@
     margin: 4px;
     text-decoration: none;
   }
+
+  .setup-system-log {
+    color: #000000;
+  }
+
+  .setup-system-log-header {
+    display: flex;
```
paulproteus: Props for `flex`. I believe this means that if the window is not very tall, then it won't cause unnecessary scrolling.

Let me know if I'm misunderstanding though!

zarvox: You're misunderstanding - this makes it so the header text and the download button contained therein are horizontally spaced.

We always want a box, so we can:

1. jump the view to the end, where the most recent log lines were written and
2. tail the end if scrolled to the bottom, which is where interesting changes will appear.

That box should avoid making another box overflow though; we accomplish this with `.admin-log-box` in `shell/client/styles/_admin-status.scss`.

PR #2620, opened 2016-10-04
Avoid make thinking that deps/node has been modified

zarvox:
Creating or removing children of a directory causes the directory's mtime to change.  The node-building code previously rm'd `deps/node/out` and replaced it with a newly-built one.  This caused make to think that we had updated `deps/node`, which would force an ekam run and everything else that follows in the build process, when really a rebuild on unchanged source should be a noop.

@dwrensha This should make repeated `make update` calls fast, as they used to be.

### DIFF
--- a/make-bundle.sh
+++ b/make-bundle.sh
@@ -100,6 +100,9 @@ pushd "$NODE_BUILD_ROOT/node"
 ./configure --partly-static
 make -j$(nproc)
 popd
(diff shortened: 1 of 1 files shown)


dwrensha:
Wow, that's a simpler change than I was expecting!

I've verified that it acts as advertised on my system.

There should probably be a brief comment explaining that we do things this way.

PR #2621, opened 2016-10-04
docs: SfW: Explain what the expiry means

paulproteus:
@kentonv requesting your +1/-1 on the accuracy of this text.

I read the code and wrote this text based on how it works now, but I'm concerned that I might have misunderstood an aspect or am taking an implementation detail and turning it into a public promise, and I don't want to give customers any inaccurate info, so I am hopeful you can let me know if you're OK with this text.

### DIFF
--- a/docs/administering/for-work.md
+++ b/docs/administering/for-work.md
@@ -250,13 +250,14 @@ renewal/expiration date for your feature key. Under normal operation, your Sands
 contact our billing server when your feature key expires and bill you at your own rate for your
 Sandstorm for Work usage. Once that succeeds, your Sandstorm server will automatically download a
 new feature key with an expiration date in the future. For that reason, the "Expires" or "Next
-Renewal" date in the Sandstorm for Work admin page on your server will change over time. To account
-for the fact that reaching the billing server can take some time, Sandstorm for Work has a small
-grace period where an expired key will continue to work. For customers who chose monthly billing,
-the new feature key will have a next renewal date one month in the future. For customers who chose
-annual billing, the new feature key will have a next renewal date one year in the future. Customers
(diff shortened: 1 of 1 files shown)


kentonv:
The update actually happens exactly when the key expires, not just before.

If it fails, there is a grace period during which the server will continue to operate -- we'll probably set this as 30 days though I'm not sure to what extent we want to document that.

In the feature key vendor (err, SfW dashboard) I changed the term "expires" to something like "next renewal". Probably we should change the in-Sandstorm UI similarly, to avoid scaring people.

paulproteus:
Gotcha.

FWIW this is the line in the template:

```
shell/client/admin/feature-key.html:          {{#if featureKey.isTrial}}Expires{{else}}Next Renewal{{/if}}
```

I'll file a separate bug about deciding what we should change the strings to, and focus for now on making sure the docs accurately describe the current strings: #2622

I'll revise this docs change, and then push it.

paulproteus:
Re-requesting @kentonv 's +1/-1.

kentonv:
> To account for the fact that reaching the billing server can take some time, Sandstorm for Work has a small grace period where an expired key will continue to work.

Reading this, my natural interpretation was: "To account for the fact that communicating over the internet could take as long as a few seconds, Sandstorm for Work has a grace period of one whole minute."

Maybe try: "If Sandstorm is unable to renew your feature key automatically, it will notify all admins by email (if correctly configured) and by a notification delivered via the bell menu. Sandstorm for Work will continue to operate for at least one week to allow time for the problem to be fixed."

(And then we actually give them a whole month.)

paulproteus:
@kentonv thanks, that's solid. I totally agree with your "natural interpretation" being my natural interpretation, too!

I'll include that language, then push.

paulproteus:

PR #2624, opened 2016-10-04
Implement "add a new email" in verifiedEmail powerbox picker.

dwrensha:
If the verifier is configured to accept emails from the "email" identity provider, displays a powerbox option like this:
pickemail

Clicking the "Add a new email..." button leads to a prompt:

enteremail

If you enter an email that's already verified, then this is the last step, and the button automatically updates to reflect that fact:
alreadyverified

Otherwise, you need to get the token from your email:
entertoken

If you enter the wrong token, you get an error:
error

I opted to write these templates from scratch because our existing email login templates are kind of a mess.

By adding a `searchTerms` field, this also makes it so the powerbox search bar actually works in email requests.

### DIFF
--- a/shell/client/accounts/email-token/token-templates.html
+++ b/shell/client/accounts/email-token/token-templates.html
@@ -34,13 +34,13 @@
   {{/if}}
 
   {{#if state.enterToken }}
+    {{#if state.error}}
+      <button name="reset" type="button">Try a different address</button>
+    {{/if}}
     <p>{{enterTokenMessage}}</p>
     <form class="verified-email" name="enter-token">
       <input type="text" name="token" placeholder="enter your token" value={{token}}>
       <button class="connect-button" disabled={{#unless token}}true{{/unless}}>Confirm</button>
-      {{#if state.error}}
-        <button name="reset" type="button">Try a different address</button>
-      {{/if}}
     </form>
   {{/if}}
 
--- a/shell/client/accounts/email-token/token-templates.html
+++ b/shell/client/accounts/email-token/token-templates.html
@@ -23,14 +23,14 @@
         <button class="connect-button">Select</button>
       {{else}}
         <button class="connect-button" disabled={{#unless email}}true{{/unless}}>
-          Send verification email
+          Send confirmation email
         </button>
       {{/if}}
     </form>
   {{/if}}
 
   {{#if state.sendingEmail }}
-    <p>Sending e-mail to {{email}}...</p>
+    <p>Sending to {{email}}...</p>
   {{/if}}
 
   {{#if state.enterToken }}
--- a/shell/client/accounts/email-token/token-templates.html
+++ b/shell/client/accounts/email-token/token-templates.html
@@ -12,11 +12,13 @@
 <template name="addNewVerifiedEmailPowerboxConfiguration">
   {{#if state.enterEmail}}
     <form class="verified-email" name="enter-email">
-      <input type="text" name="email" placeholder="enter email address">
+      <input type="text" name="email" placeholder="enter email address" value={{email}}>
       {{#if alreadyVerified}}
-        <button class="connect-button">Select (already verified)</button>
+        <button class="connect-button">Select</button>
       {{else}}
-        <button class="connect-button">Send verification email</button>
+        <button class="connect-button" disabled={{#unless email}}true{{/unless}}>
+          Send verification email
+        </button>
       {{/if}}
     </form>
   {{/if}}
@@ -26,10 +28,10 @@
   {{/if}}
 
   {{#if state.enterToken }}
-    <p>{{state.enterToken}}</p>
+    <p>{{enterTokenMessage}}</p>
     <form class="verified-email" name="enter-token">
-      <input type="text" name="token" placeholder="enter your token">
-      <button class="connect-button">Confirm</button>
+      <input type="text" name="token" placeholder="enter your token" value={{token}}>
+      <button class="connect-button" disabled={{#unless token}}true{{/unless}}>Confirm</button>
     </form>
   {{/if}}
 
@@ -38,7 +40,6 @@
   {{/if}}
 
   {{#if state.error}}
-    <p>{{state.error}}</p>
-    <button name="reset">Start over</button>
+    <p class="error">{{state.error}}</p>
   {{/if}}
 </template>
--- a/shell/client/accounts/email-token/token-templates.html
+++ b/shell/client/accounts/email-token/token-templates.html
@@ -38,6 +38,9 @@
     <form class="verified-email" name="enter-token">
       <input type="text" name="token" placeholder="enter your token" value={{token}}>
       <button class="connect-button" disabled={{#unless token}}true{{/unless}}>Confirm</button>
+      {{#if state.error}}
+        <button name="reset" type="button">Try a different address</button>
+      {{/if}}
     </form>
   {{/if}}
 
--- a/shell/client/accounts/email-token/token-templates.html
+++ b/shell/client/accounts/email-token/token-templates.html
@@ -35,7 +35,9 @@
 
   {{#if state.enterToken }}
     {{#if state.error}}
+    <div class="reset-row">
       <button name="reset" type="button">Try a different address</button>
+    </div>
     {{/if}}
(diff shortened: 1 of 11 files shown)


neynah:
Some suggestions/thoughts:

General things:
- For the powerbox I think we should only display a scrollbar when it's required (as in there isn't sufficient vertical space to display the full list of things).
- The (currently) purple buttons should appear disabled by default and only become purple when we detect the user has entered something.

Screenshot 2:
- The text "Select one" doesn't really make sense anymore since the user has already decided to add a new email address and there is nothing to select. It should probably say "Add new email"
- Do we still have to display the "Add new email address..."button if the user has already clicked on it? I don't suspect the user should have the option to click on "Add new email address..." when they are already in the process of adding a new email address.
- I think the behaviour for adding an already-verified email is strange. Perhaps instead of having the button say "Select (already confirmed)" we should instead keep the "Send verification email" string and when the user attempts to send a verification email to an already-verified email address we should immediately show a success message indicating that this email has already been verified.

Screenshot 5:
- What does the user see after they click the "Start over" button? Do they have to re-add their email address and get another code sent to them?
- I don't think we want to get rid of the previous UI in screenshot 4. We should instead append an error message when the code is invalid and possibly give them options to verify a different email address or re-send the code.

dwrensha:
> I think the behaviour for adding an already-verified email is strange. Perhaps instead of having the button say "Select (already confirmed)" we should instead keep the "Send verification email" string and when the user attempts to send a verification email to an already-verified email address we should immediately show a success message indicating that this email has already been verified.

I can do this, but I'm curious to hear more about what makes the "Select (already confirmed)" button strange.

Note that there is no success message in any other case in the flow; once an email is successfully confirmed, we immediately close the powerbox popup.

dwrensha:
> What does the user see after they click the "Start over" button? Do they have to re-add their email address and get another code sent to them?

It takes them back to the [enter email address][Send verification email] step. Ah... I can see how that would be annoying if you made a typo in the code, and now you need to get sent a new email.

neynah:
>I can do this, but I'm curious to hear more about what makes the "Select (already confirmed)" button strange.

>Note that there is no success message in any other case in the flow; once an email is successfully confirmed, we immediately close the powerbox popup.

Hmm... I see. It's strange that the action the button performs is dependent on whether the email is already verified or not. However, since we aren't displaying any success messages then perhaps we should just have the "Select (already confirmed)" button instead say "Select". The user doesn't need to think about whether the email is already confirmed or not since they'll be able to succeed with selecting this email.

kentonv:
Can you use icons/add-email.svg instead of the purple plus? (It was actually created for this use case.)

dwrensha:
New screenshots.

Starting state:
verified-email-1

After clicking "Add a new email":
verified-email-2

Upon entering an already-linked email:
verified-email-3

After submitting a not-yet-linked email:
verified-email-4

After submitted an incorrect token:
verified-email-5

dwrensha:
> The text "Select one" doesn't really make sense anymore since the user has already decided to add a new email address and there is nothing to select. It should probably say "Add new email"

I've changed this to "selected". Does that make more sense?

> Do we still have to display the "Add new email address..."button if the user has already clicked on it? I don't suspect the user should have the option to click on "Add new email address..." when they are already in the process of adding a new email address.

At that point it's no longer a button. It's there to remind the user of the option they clicked in the previous step. We do the same thing for other powerbox flows. For example, when you add a grain to a collection, there is a second step where you choose the permissions to grant. During that second step, the card for the grain you selected is displayed.

zarvox:
>> Do we still have to display the "Add new email address..."button if the user has already clicked on it? I don't suspect the user should have the option to click on "Add new email address..." when they are already in the process of adding a new email address.

> At that point it's no longer a button. It's there to remind the user of the option they clicked in the previous step. We do the same thing for other powerbox flows. For example, when you add a grain to a collection, there is a second step where you choose the permissions to grant. During that second step, the card for the grain you selected is displayed.

Just wanted to throw in my +1 for keeping this unit of context in the powerbox in general.  It confirms to the user that their click was observed, lets them confirm that they selected the right item (or if they didn't, correct course), and that subsequent interactions in the window are contextualized by that item.

neynah:
+1. Although I would add a bit of padding on the "Invalid authentication code" error message. Similar to the padding we have on "Add a new email".

So it seems like if they have an invalid authentication code, retype it and still get the same error message they'll need to close the window and try again or click on "add a new email"? Maybe we need a "Try another email" button. Just a suggestion though- I'll leave it up to you.

zarvox:
Oh, general design comment, which happens to also apply here, now that I'm looking:

Put success or error messages at the *top* of the form to which the action applied (but below a top-level description of the form).  Compare with the admin email test form.

Rationale: this placement makes it more likely to be seen and interpreted as relevant by a user.  If the "submit" button is the bottom of the items they consider relevant to their current action, adding an error message below the submit button carries a chance that it will not be seen at all in scrollable areas.  On success or failure, the newly-focused success or error message marks the top of the form within which the error applies, and the button that they clicked to submit the form marks the bottom of the form.

Hmmm, I should probably add this to the styleguide!

shell/client/accounts/email-token/token-templates.html
```diff
@@ -8,3 +8,38 @@
...
+      <input type="text" name="token" placeholder="enter your token" value={{token}}>
+      <button class="connect-button" disabled={{#unless token}}true{{/unless}}>Confirm</button>
+    </form>
+  {{/if}}
+
+  {{#if state.sendingToken }}
+    <p>Verifying token...</p>
+  {{/if}}
+
+  {{#if state.error}}
```
zarvox: As noted: place this above the user input

shell/client/accounts/email-token/token-templates.html
```diff
@@ -8,3 +8,38 @@
...
+    <p class="error">{{state.error}}</p>
```
zarvox: Consider using:

```
{{#focusingErrorBox}}
  {{state.error}}
{{/focusingErrorBox}}
```

and eschewing the additional CSS rule

dwrensha: Cool! Errors now look like: 

verified-email-6

dwrensha:
> So it seems like if they have an invalid authentication code, retype it and still get the same error message they'll need to close the window and try again or click on "add a new email"? Maybe we need a "Try another email" button. Just a suggestion though- I'll leave it up to you.

OK, I've added a "Try a different address" button that shows up if you hit an error in the "enter a token" stage. Looks like: 
verified-email-8

dwrensha:
I adjusted some terminology for consistency. Now the card says "Add a new address", and the error after entering a wrong code looks like:
verified-email-9

kentonv:
@neynah and @zarvox, since you've had significant comments here, let me know when you feel this is ready.

neynah:
I would right align the "Try a different address" button" and round the edges of the light border around the "Invalid authentication code." error message. Then it's +1 from me.

dwrensha:
> I would right-align the "Try a different address" button" 

done: 
verified-email-10

> and round the edges of the light border around the "Invalid authentication code." error message

That border comes from @zarvox's `focusingErrorBox` component, which appears in lots of other places. Maybe the styling should be tweaked, but that feels to me like it would belong in a separate pull request.

zarvox:
Thanks for the changes! :D

If the "try a different address" button is contextually bound to the error, perhaps it belongs inside the error box?

The faint not-quite-rounded-correctly border is actually the browser vendor's focus ring outline.  This can hopefully be fixed (outline vs border is a weird case...), but definitely belongs in a separate PR.

neynah:
>That border comes from @zarvox's focusingErrorBox component, which appears in lots of other places. Maybe the styling should be tweaked, but that feels to me like it would belong in a separate pull request.

Ah I see. I have no other objections so I'm happy with what you & Drew decide on.

dwrensha:
> If the "try a different address" button is contextually bound to the error, perhaps it belongs inside the error box?

I actually did try that and I thought it didn't look quite right. It seemed to be indicating that the *preferred* way to resolve the error would be to click "try a different address", whereas I suspect usually there will have just been a copy/paste error and the user should try entering the token again.

zarvox:
:shipit:

PR #2627, opened 2016-10-05
install.sh: Permit installing on an empty /opt/sandstorm

### LINKED ISSUE #2626: install.sh: Allow user to install to an empty-but-existing /opt/sandstorm
Context:

- A user wanted to install into an empty but existent /opt/sandstorm created by their storage management system: https://groups.google.com/forum/#!topic/sandstorm-dev/M9ehImegggo

- install.sh refused to allow the install to proceed since that directory existed: https://github.com/sandstorm-io/sandstorm/blob/master/install.sh#L1087

- install.sh does so to avoid confusion with users attempting to install Sandstorm over an existing install, which generally does not work properly.

We need to come up with some reasonable way to allow sysadmins to install to an existing-but-empty /opt/sandstorm. Some options include:

- Allowing the user to override this check.

    - I don't _love_ this option because it defeats the main purpose of the check - helping users who don't know what they're doing recognize that their needs will be best-served by uninstalling.

- Allowing the user to override this check, but making that a little bit hard

    - I think that might be good.

- Making the check a little more forgiving, e.g. check for /opt/sandstorm/latest instead of /opt/sandstorm

    - Tentatively I like this, but I'm worried that it'd be _too_ forgiving and then help users find themselves in the same situation as before, where they should have uninstalled before doing a new install but we didn't guide them appropriately.

- Checking for /opt/sandstorm being empty and allowing the check to pass

    - I'm slightly worried that the user might have `/opt/sandstorm/.placeholder` or some other singular file in the directory, or e.g. `/opt/sandstorm/README.txt` which is a note from one sysadmin to another, and we'd then fail this check, but really the user should be allowed to proceed.

paulproteus:
Close #2626

Add tests for this case, too.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1094,12 +1094,16 @@ choose_install_dir() {
     DIR=$(prompt "Where would you like to put Sandstorm?" "$DEFAULT_DIR")
   fi
 
-  if [ -e "$DIR" ]; then
+  # Check for the existence of any partial Sandstorm installation. Note that by default, the
+  # Sandstorm uninstall process will retain $DIR/sandstorm.conf and $DIR/var. Since the install
+  # script can't reliably use those to preseed a new Sandstorm install, we still bail out in that
+  # situation.
+  if [ -e "$DIR/sandstorm.conf" ] || [ -e "$DIR/var" ] || [ -e "$DIR/sandstorm" ] ; then
     # Clear the previous line, since in many cases, it's a "echo -n".
     error ""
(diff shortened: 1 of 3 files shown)


paulproteus:
Self-merging because 100% (install.sh + installer-tests).

paulproteus:
I ran a subset of installer-tests (the specific tests modified in this commit) and they passed. I'll also ask Jenkins to check my work now.

PR #2634, opened 2016-10-07
sandstorm-http-bridge: don't double-report errors upon restoring identities

dwrensha:
`Sandstorm.getSavedIdentity()` returns a pipelined identity capability. If the `restore()` call fails, the error is reported when the app tries to use the returned pipelined identity capability. Currently, the same error is also unconditionally reported by sandstorm-http-bridge; an app that calls `whenResolved()` on the returned cap and attempts to gracefully deal with the error still ends up with the error logged by sandstorm-http-bridge. That could be confusing for someone trying to debug an app.

### DIFF
--- a/src/sandstorm/sandstorm-http-bridge.c++
+++ b/src/sandstorm/sandstorm-http-bridge.c++
@@ -1106,6 +1106,8 @@ public:
         kj::StringPtr textIdRef = textId;
         KJ_ASSERT(liveIdentities.insert(std::make_pair(
           textIdRef, IdentityRecord { kj::mv(textId), kj::mv(identity) })).second);
+      }, [] (auto e) {
+        // Ignore the error here because the returned capability will report it upon use.
       }));
 
       return kj::mv(identity);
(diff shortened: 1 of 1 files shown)


kentonv:
I'm not super-excited about muting errors like this. If the app never uses the identity, then the error would never be reported at all.

Do we expect that https://github.com/sandstorm-io/capnproto/pull/378 will fix the main source of errors here?

kentonv:
OK, on further reading I see that it makes enough sense in this use case. The app wouldn't be loading the identity if they weren't going to use it.

dwrensha:
Another source of errors is when a user gets revoked from a grain. A grain might restore an identity capability and wait for it to resolve, intending to call `SessionContext.activity()` with it upon success. Upon failure to resolve, the grain might sensibly assume that the identity has been revoked. It's annoying if sandstorm-http-bridge unconditionally dumps such failures to the grain log; the grain might want to report them elsewhere, or not at all.

PR #2644, opened 2016-10-11
Support RHEL 7, CentOS 7, and Arch by not using userns when unavailable

kentonv:
This turned out significantly less horrible than imagined.

Until now our assumption had been that supporting non-userns systems would require turning the supervisor into a setuid binary. Setuid binaries are scary. Also, the supervisor isn't actually a separate binary from other Sandstorm components these days, which would be a pain to undo.

But I came up with a trick that avoids the need for setuid binaries: Instead, root privileges are handed down from the bundle runner through the "backend" process to the supervisor. The backend process runs with _effective_ UID non-root, but still has a _real_ UID of root, which allows it to run the supervisor as root. No setuid binaries.

This does mean that an RCE vulnerability in the backend process would be able to do more damage, since it could potentially setuid(0) itself. However, the backend is a relatively simple process that is not directly exposed to the internet nor to apps, so this risk seems small -- and in many cases less-bad than the risk involved in dropping a setuid binary onto the system.

For systems that support userns, this change should have no effect. Support for userns is discovered dynamically at startup. Note that I've deleted the userns-related code from the install script, which means the installer will no longer go out of its way to enable userns, which may mean that a lot of people even with newer kernels end up using the non-userns path.

With this, I successfully tested this on CentOS 7.2 (kernel 3.10).

@amluto Curious for your thoughts.

@paulproteus Please review installer changes.

@jparyani Please review C++ changes.

### DIFF
--- a/src/sandstorm/run-bundle.c++
+++ b/src/sandstorm/run-bundle.c++
@@ -180,6 +180,37 @@ bool isKernelNewEnough() {
   return true;
 }
 
+bool isUserNsAvailable() {
+  Subprocess child([]() {
+    if (getuid() == 0) {
+      if (setuid(1000) < 0) {
+        // setuid() failed?
+        return 2;
+      }
+    }
+
+    if (unshare(CLONE_NEWUSER) < 0) {
+      return 1;
+    }
+
+    return 0;
+  });
+
+  int status = child.waitForExit();
+  switch (status) {
+    case 0:
+      return true;
+    case 1:
+      return false;
+    case 2:
+      KJ_LOG(ERROR, "setuid() failed when trying to test if unprivileged userns works");
+      return true;
+    default:
+      KJ_LOG(ERROR, "userns test process exited with unexpected status code", status);
+      return true;
+  }
+}
+
 // =======================================================================================
 // id(1) handling
 //
@@ -1417,16 +1448,28 @@ private:
     restoreResolvConfIfNeeded();
   }
 
-  void dropPrivs(const UserIds& uids) {
+  void dropPrivs(const UserIds& uids, bool keepRealUid = false) {
     // Defense in depth: Don't give my children any new caps for any reason.
     KJ_SYSCALL(prctl(PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0));
 
     if (runningAsRoot) {
       KJ_SYSCALL(setresgid(uids.gid, uids.gid, uids.gid));
       KJ_SYSCALL(setgroups(uids.groups.size(), uids.groups.begin()));
-      KJ_SYSCALL(setresuid(uids.uid, uids.uid, uids.uid));
+
+      if (keepRealUid) {
+        // We're prepping to run the backend and user namespaces are not available, therefore the
+        // backend needs to keep its superuser powers stashed in order to hand them off to the
+        // grain supervisors so that they can set up sandboxes. Instead of creating a suid binary
+        // (with all the danger that entails), we merely drop the effective UID, but raise it again
+        // when invoking the supervisor.
+        KJ_SYSCALL(seteuid(uids.uid));
+      } else {
+        KJ_SYSCALL(setresuid(uids.uid, uids.uid, uids.uid));
+      }
     } else {
       // We're using UID namespaces.
+
+      KJ_ASSERT(!keepRealUid);
 
       // Defense in depth: Drop all capabilities from the set of caps which my children are allowed
       //   to ever have.
@@ -2191,13 +2234,20 @@ private:
         }
       }
 
-      dropPrivs(config.uids);
+      // If we're not running as root, we have to use user namespaces. Otherwise, dynamically
+      // check if they're available. If not, we'll need to pass superuser privileges on to the
+      // backend.
+      bool avoidUserns = runningAsRoot && !isUserNsAvailable();
+      kj::Maybe<uid_t> sandboxUid;
+      if (avoidUserns) sandboxUid = config.uids.uid;
+
+      dropPrivs(config.uids, avoidUserns);
       clearSignalMask();
 
       auto paf = kj::newPromiseAndFulfiller<Backend::Client>();
       TwoPartyServerWithClientBootstrap server(kj::mv(paf.promise));
       paf.fulfiller->fulfill(kj::heap<BackendImpl>(*io.lowLevelProvider, network,
-        server.getBootstrap().castAs<SandstormCoreFactory>()));
+        server.getBootstrap().castAs<SandstormCoreFactory>(), sandboxUid));
 
       // Signal readiness.
       write(outPipe, "ready", 5);
--- a/src/sandstorm/run-bundle.c++
+++ b/src/sandstorm/run-bundle.c++
@@ -1430,15 +1430,15 @@ private:
   }
 
   void dropPrivs(const UserIds& uids) {
+    // Defense in depth: Don't give my children any new caps for any reason.
+    KJ_SYSCALL(prctl(PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0));
+
     if (runningAsRoot) {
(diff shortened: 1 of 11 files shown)


kpreid:
> Note that I've deleted the userns-related code from the install script, which means the installer will no longer go out of its way to enable userns, which may mean that a lot of people even with newer kernels end up using the non-userns path.

Why not keep it in the install script but make it optional (or at least _check and warn_ about userns)? I-as-of-back-when-I-installed-Sandstorm want the best security I can get easily but have no idea what userns is or whether it's enabled.

paulproteus:
Out of curiosity, can we use seccomp to block the `setuid(0)` that the backend could hypothetically do?

Regardless of the above, I'm very excited about these changes, and will review the installer component shortly. I've filed https://github.com/sandstorm-io/sandstorm/issues/2645 with a list of docs changes we'll seemingly need.

Also, I wonder if this will affect `vagrant-spk enter-grain`. If so, we should probably do a coordinated release or something.

install.sh
```diff
@@ -457,151 +456,10 @@ Mac system, you can use Vagrant or another virtualization tool. See our install
...
-  # If the test binary failed, but the sysctl is set to 1, then allow the install to
-  # continue. Perhaps the user has /tmp mounted noexec or something like that.
-  if [ "$(</proc/sys/kernel/unprivileged_userns_clone)" = "1" ] ; then
-    return
-  fi
-
-  # If /proc/sys is mounted read-only, then they won't be able to run sysctl with our
-  # help. Similarly if there's no /etc/sysctl.conf we won't be able to help.
-  #
-  # The /proc/sys issue applies to Docker containers and probably other sorts of containers, too.
```
paulproteus: Given this change, I should re-test in Docker and update our docs. I added a note to the docs issue #2645.

src/sandstorm/run-bundle.c++
```diff
@@ -172,21 +172,45 @@ KernelVersion getKernelVersion() {
 
 bool isKernelNewEnough() {
   auto version = getKernelVersion();
-  if (version.major < 3 || (version.major == 3 && version.minor < 13)) {
+  if (version.major < 3 || (version.major == 3 && version.minor < 10)) {
```
paulproteus: Super duper exciting! :tada:

paulproteus: This will also help on Synology NAS boxes.

src/sandstorm/supervisor.c++
```diff
@@ -644,14 +663,21 @@ kj::MainBuilder::Validity SupervisorMain::addCommandArg(kj::StringPtr arg) {
...
   setupSupervisor();
 
   // Exits if another supervisor is still running in this sandbox.
   systemConnector->checkIfAlreadyRunning();
 
-  SANDSTORM_LOG("Starting up grain.");
+  if (setuidMode == nullptr) {
+    SANDSTORM_LOG("Starting up grain. Sandbox type: userns");
+  } else {
+    SANDSTORM_LOG("Starting up grain. Sandbox type: setuid");
```
paulproteus: As I recall, it's not literally a setuid binary that is setting up the sandbox, right? If so, the string "setuid" here is a little misleading. Maybe

> Starting up grain. Sandbox type: root

or

> Starting up grain. Sandbox type: uid-switching

Although I guess "setuid" doesn't mean "setuid root", so I _guess_ this is OK. But anyway when people read this I think they're going to think "setuid root", so I personally advise against calling it the "setuid sandbox".

src/sandstorm/supervisor.h
```diff
@@ -97,6 +98,7 @@ class SupervisorMain: public AbstractMain {
   bool devmode = false;
   bool seccompDumpPfc = false;
   bool isIpTablesAvailable = false;
+  kj::Maybe<uid_t> setuidMode;
```
paulproteus: I guess we call it the "setuid sandbox" here as an implementation detail, not just in strings that admins will see.

Still, I prefer if we call it something else to minimize admin confusion that we're using setuid root binaries, e.g the user-switching or uid-switch sandbox.

src/sandstorm/util.c++
```diff
@@ -879,6 +879,14 @@ Subprocess::Subprocess(Options&& options)
         KJ_SYSCALL(dup2(options.moreFds[i], STDERR_FILENO + 1 + i));
       }
 
+      // Drop priveleges if requested.
```
paulproteus: Typo: "priveleges" != "privileges"

kentonv:
@kpreid Unfortunately it's not totally cut-and-dry. Although enabling userns makes Sandstorm more secure, it possibly makes other parts of your system less secure, since there have been many kernel CVEs in the userns system. This is why Arch has refused to enable userns, for example: they feel it's too insecure.

These trade-offs are subtle and the typical user is not likely to be able to evaluate them. Presenting them with a choice would therefore probably cause fewer users to complete installation, and presenting them with a warning would probably upset users unnecessarily.

@paulproteus 

> Out of curiosity, can we use seccomp to block the setuid(0) that the backend could hypothetically do?

No, because it actually needs to setuid(0) when running the supervisor.

> Also, I wonder if this will affect vagrant-spk enter-grain. If so, we should probably do a coordinated release or something.

No, as long as vagrant-spk is still using the userns approach.

zarvox:
> But I came up with a trick that avoids the need for setuid binaries: Instead, root privileges are handed down from the bundle runner through the "backend" process to the supervisor. The backend process runs with effective UID non-root, but still has a real UID of root, which allows it to run the supervisor as root. No setuid binaries.

Does this add the requirement that the backend run as root, rather than as an unprivileged user?  I seem to remember that "all processes run as non-root" was previously a supported configuration.

kentonv:
> Does this add the requirement that the backend run as root, rather than as an unprivileged user? I seem to remember that "all processes run as non-root" was previously a supported configuration.

(As discussed in person:) No. When unprivileged userns is available, nothing has changed, so you can still run with no root privileges at all. When unprivileged userns is not available, then root privileges are required, but that's strictly better than before, where you couldn't run Sandstorm at all.

paulproteus:
As discussed in person, I'm concerned that if there are hidden bugs in this, we might not know until new users start to use this.

Therefore my request to Kenton is that he delay the installer change for about a week to find out if there's any surprise breakages.

My request to myself is that I go find 1-10 people willing to run their Sandstorm server _without_ userns for a week, and ask them if they run into any problems that way.

kentonv:
We observed that this change creates a potential problem in that getuid() inside the sandbox will now return a potentially different value on every machine, whereas with userns it always returns 1000. Some apps may inadvertently rely on being run as UID 1000. For example, apps that insist on the existence of a valid entry it /etc/passwd for the current UID may fail.

I tried installing and opening every app currently on the app market. The only app that showed a problem was RoundCube. The problem here seems to be that Dovecot's config file needs to contain its UID, and if the UID in the config is different from getuid(), it tries to setuid(), which doesn't work.

We can probably fix the RoundCube bug by having it generate a config dynamically at startup, or something.

In order to prevent future problems with hard-coded UIDs in apps, we could make the userns sandbox choose a random UID on every startup, or perhaps on every grain creation. (This would randomize only the perceived UID as returned by getuid(); the actual physical UID would be unaffected.) Developers would then notice and fix any problems before shipping their app.

paulproteus:
> I tried installing and opening every app currently on the app market. The only app that showed a problem was RoundCube. The problem here seems to be that Dovecot's config file needs to contain its UID, and if the UID in the config is different from getuid(), it tries to setuid(), which doesn't work.

This is some impressive manual QA. :tada:

kentonv:
@paulproteus - Well, I only opened them. I didn't actually do anything with most of them, besides look at the first page and verify it had no obvious errors.

FWIW mass-installing apps is easy through the pre-installed apps panel. :)

kentonv:
Changes:
* The new sandboxing mode is now called the "privileged sandbox".
* I've moved the installer changes to #2656, so that we don't release it immediately.

Issue #2643, opened 2016-10-11
install.sh: Stop assuming nc is present, then using port 25 without actually checking properly


kentonv:
If we're seeing a lot of `E_NEVER_LISTENED` in the logs, it's because the new installer script defaults to listening for SMTP on port 25, but port 25 is often already taken by the local MTA.

We should fix this ASAP.

paulproteus:
Massive forehead slap. I'll fix and write a test. Thanks.

paulproteus:
@kentonv , I can't seem to reproduce this in my "test lab" of `installer-tests`. I've pushed a new test that already passes without any changes to `install.sh`. I've also manually examined the resulting Sandstorm install, and don't currently see how we'd end up in the situation you describe.

However, we did get at least one report on IRC from someone who had this problem. Perhaps Sandstorm itself should handle the problem differently on startup, rather than by crashing.

It's also very likely I simply haven't figured out how to reproduce the install.sh bug and it is a real bug, so let's figure that out when you have a chance.

paulproteus:
The cause of this seems to be the following:

- install.sh only insists that `nc` is present if you are doing a sandcats install

- nc isn't present on our DigitalOcean CentOS test box

- install.sh assumes that any failure of nc to connect to localhost:25 is due to the port not being bound; however, it forgets that the absence of nc can also cause this failure!

I'll fix the above.

kentonv:
OK, but I *did* do a Sandcats install and observed this problem, so apparently it actually doesn't insist that `nc` is present even for sandcats?

paulproteus:
> OK, but I did do a Sandcats install and observed this problem, so apparently it actually doesn't insist that nc is present even for sandcats?

It turns out that the check happens at the wrong time. I'll fix that in the same PR. Thanks greatly for catching that.

PR #2646, opened 2016-10-11
installer-tests: Add test for installing with an MTA

paulproteus:
See also: #2643


paulproteus:

zarvox:
Fascinating failure:

screenshot

Was DNS out briefly or something?

PR #2649, opened 2016-10-11
installer-tests: Start Postfix separate from install

paulproteus:
This allows slower machines (like Jenkins) to be able to complete this test.


paulproteus:
Self-merging because 100% installer-tests

PR #2651, opened 2016-10-12
Move info/error messages in the login dialog to the top of the form

zarvox:
Placing errors below the thing the user is interacting with is bad because:

* the errors can wind up completely off-screen if they appear below the fold
* users are less likely to associate the error message with the action they
  just took if it is in a section that they haven't read/interacted with on the
  page.

It is better to place errors at the top of the context to which they apply
than to place them below the context to which they apply.

Ideally, we'd place errors associated with a particular login mechanism just
above *that login mechanism's login section* but I don't think we currently
track those separately and this seems like a strict improvement.

screenshot1

### DIFF
--- a/shell/client/accounts/login-buttons.html
+++ b/shell/client/accounts/login-buttons.html
@@ -150,6 +150,8 @@ licenses, included in the LICENSES directory.
 
 <template name="loginButtonsList">
   <div class="login-buttons-list">
+  {{> _loginButtonsMessages "" }}
+
   {{#each services}}
     {{#with data=loginTemplate.data linkingNewIdentity=linkingNewIdentity}}
       {{> Template.dynamic template=../loginTemplate.name }}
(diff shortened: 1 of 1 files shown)


neynah:
Nice! This is already an improvement but I would prefer if we used the same error styling (redish bg, dark text) seen in other places. Then we can also avoid the error message looking buttony like the other buttons on this login dialog.

zarvox:
> I would prefer if we used the same error styling (redish bg, dark text) seen in other places

+1.  That said, at that point we should also be putting it atop the specific identity provider button you attempted to use, which is a much larger change than "move this one line up a few lines", which is what this opportunistic patch was and probably requires a little more thought/planning.

kentonv:
In testing I'm finding this is a little bit jarring in that it causes controls you are interacting with to move around. E.g. when you click on "send login email" the button jumps downwards out from under your mouse cursor.

kentonv:
Also frustrating is that because of this jumping around, it's hard for the user to track the state changes to the e-mail login component in particular. When you type in an address, the component resizes to become a form asking for the token. But when all the elements in the login list are jumping up and down at the same time, it's hard to tell what happened.

Sorry but I think as-is this change is more harm than good. I understand and agree with the intent but I think we need to think more carefully about how two do it right. I'm reverting this for now.

kentonv:
I reverted this, but then I failed to pull the revert before doing a release, so it was released anyway.

/facepalm

PR #2654, opened 2016-10-13
docs: Airgap/offline: Remark on switching network on/off

paulproteus:


paulproteus:
Self-merging because 100% pure docs

PR #2656, opened 2016-10-14
installer: Don't require userns. Accept kernel 3.10.

kentonv:
Depends on #2644.

We've decided to delay releasing the installer changes until we've been able to test the privileged (non-userns) sandbox mode more thoroughly, because once we update the installer, users of Debian and Ubuntu will probably start using the privileged sandbox by default.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -493,12 +493,10 @@ maybe_enable_userns_sysctl() {
   fi
 
   # Enable it.
-  echo "NOTE: Enabling unprivileged user namespaces because you passed -d."
-  if ! sysctl -wq kernel.unprivileged_userns_clone=1; then
-    echo "WARNING: 'sysctl -w kernel.unprivileged_userns_clone=1' failed. If you" >&2
-    echo "  are using Docker, you will need to run this command manually outside" >&2
-    echo "  of Docker and update /etc/sysctl.conf. For more info, see:" >&2
-    echo "    https://docs.sandstorm.io/en/latest/install/#option-6-using-sandstorm-within-docker" >&2
+  if sysctl -wq kernel.unprivileged_userns_clone=1 2>/dev/null; then
+    echo "NOTE: Enabled unprivileged user namespaces because you passed -d."
+  else
+    # Apparently we can't. Maybe we're in a Docker container. Give up and use privileged sandbox.
     return
   fi
 
@@ -518,6 +516,7 @@ __EOF__
     # We couldn't make the change permanent, so undo the change. Probably everything will work
     # fine with the privileged sandbox. But if it doesn't, it's better that things fail now rather
     # than wait for a reboot.
+    echo "NOTE: Never mind, not enabling userns because can't write /etc/sysctl.d."
     sysctl -wq "kernel.unprivileged_userns_clone=$OLD_VALUE" || true
     return
   fi
--- a/install.sh
+++ b/install.sh
@@ -463,6 +463,51 @@ assert_usable_kernel() {
   fi
 }
 
+maybe_enable_userns_sysctl() {
+  # This function enables the Debian/Ubuntu-specific unprivileged
+  # userns sysctl, if the system has it and we want it.
+
+  if [ "$USE_DEFAULTS" != "yes" ] ; then
+    # Only do this when -d is passed. -d means "use defaults suitable for app development", and
+    # we want userns enabled for app development if possible since it enables UID randomization
+    # which helps catch app bugs. For the rest of the world, we're fine using the privileged
+    # sandbox instead.
+    return
+  fi
+
+  if [ "no" = "$CURRENTLY_UID_ZERO" ] ; then
+    # Not root. Can't do anything about it.
+    return
+  fi
+
+  if [ ! -e /proc/sys/kernel/unprivileged_userns_clone ]; then
+    # No such sysctl on this system.
+    return
+  fi
+
+  if [ "$(sysctl -n kernel.unprivileged_userns_clone)" = "1" ]; then
+    # Already enabled.
+    return
+  fi
+
+  # Enable it.
+  echo "NOTE: Enabling unprivileged user namespaces because you passed -d."
+  sysctl -wq kernel.unprivileged_userns_clone=1
+
+  # Also make sure it is re-enabled on boot. If sysctl.d exists, we drop our own config in there.
+  # Otherwise we edit sysctl.conf, but that's less polite.
+  local SYSCTL_FILENAME="/etc/sysctl.conf"
+  if [ -d /etc/sysctl.d ] ; then
+    SYSCTL_FILENAME="/etc/sysctl.d/50-sandstorm.conf"
+  fi
+
+  cat >> "$SYSCTL_FILENAME"  << __EOF__
+
+# Enable non-root users to create sandboxes (needed by Sandstorm).
+kernel.unprivileged_userns_clone = 1
+__EOF__
+}
+
 assert_dependencies() {
   if [ -z "${BUNDLE_FILE:-}" ]; then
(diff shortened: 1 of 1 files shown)


kentonv:
Looks like the mail test is currently failing due to the Roundcube bug where it wants the UID to be 1000.

paulproteus:
So far, my feedback is:

Per our discussion today, I believe this needs to try to use the `userns` sandbox if:

- You are installing in `-d` mode ("defaults" for "development")
- You have `/proc/sys` mounted read-write
- The system (via looking in `/proc/sys`) shows that there exists a sysctl for enabling userns

The one thing it should do, in that case, is to create the `sysctl.d` file (...or modify `sysctl.conf` if you don't have a `sysctl.d`? but you will have one in the one case that we care about, which is vagrant-spk using this, so we could just go for the cleaner `sysctl.d` approach).

@kentonv is that something you'll do as a patch to this, or is that my job? :)

It does mean we still can remove the x86-64 binary in install.sh. :)

Also I should presumably add an installer-test for this case, where `-d` prefers the userns sandbox.

kentonv:
@paulproteus OK, I added code to enable the userns sysctl specifically when `-d` is used.

install.sh
```diff
@@ -457,152 +456,56 @@ Mac system, you can use Vagrant or another virtualization tool. See our install
...
 
-  if [ "${PRINT_USERNS_PROMPT}" = "yes" ] ; then
-    echo "Sandstorm requires sysctl kernel.unprivileged_userns_clone to be enabled."
-    echo "Currently, it is not enabled on your system."
-    if prompt-yesno "Shall I enable it for you ${ENABLE_SYSCTL_STRING}?" yes; then
-      ACCEPTED_SYSCTL_SWITCH="yes"
-    fi
+  # Also make sure it is re-enabled on boot. If sysctl.d exists, we drop our own config in there.
+  # Otherwise we edit sysctl.conf, but that's less polite.
+  local SYSCTL_FILENAME="/etc/sysctl.conf"
```
paulproteus: Arguably feel free to bail on enabling the sysctl if `/etc/sysctl.d` doesn't exist. Not a show-stopper, just an opportunity for slightly fewer lines of code. vagrant-spk will have `/etc/sysctl.d`.

install.sh
```diff
@@ -463,6 +463,51 @@ assert_usable_kernel() {
...
+    # No such sysctl on this system.
+    return
+  fi
+
+  if [ "$(sysctl -n kernel.unprivileged_userns_clone)" = "1" ]; then
+    # Already enabled.
+    return
+  fi
+
+  # Enable it.
```
paulproteus: If `/proc/sys` is read-only, we should probably avoid setting the flag. It's read-only within Docker.

We have the following code currently within install.sh:

```
  # The /proc/sys issue applies to Docker containers and probably other sorts of containers, too.
  if egrep -q '/proc/sys\s+proc\s+ro' /proc/mounts || [ ! -e /etc/sysctl.conf ] ; then
```

Presumably we should use the same `egrep` test.

install.sh
```diff
@@ -463,6 +463,51 @@ assert_usable_kernel() {
...
+  echo "NOTE: Enabling unprivileged user namespaces because you passed -d."
+  sysctl -wq kernel.unprivileged_userns_clone=1
```
paulproteus: I'm concerned that if this fails, we have no code to handle the failure. The current install.sh does things this way:

```
    sysctl -w kernel.unprivileged_userns_clone=1 >/dev/null \
      || fail "E_SYSCTL_WRITING_FAILED" "'sysctl -w" \
              "kernel.unprivileged_userns_clone=1' failed. If you are inside docker, please run the" \
              "command manually inside your host and update /etc/sysctl.conf."
```

If the `sysctl` command fails, arguably we want to abort this function and continue with the install.

Same with the writing to `$SYSCTL_FILENAME`.

Curious if you agree.

install.sh
kentonv: Would `test -w` work instead of grepping mounts?

kentonv:
I've added a commit to handle when sysctl -w fails, and also when writing to sysctl.d fails.

install.sh
```diff
@@ -485,14 +485,22 @@ maybe_enable_userns_sysctl() {
...
+
+  if [ "$OLD_VALUE" = "1" ]; then
     # Already enabled.
     return
   fi
 
   # Enable it.
   echo "NOTE: Enabling unprivileged user namespaces because you passed -d."
-  sysctl -wq kernel.unprivileged_userns_clone=1
+  if ! sysctl -wq kernel.unprivileged_userns_clone=1; then
```
kentonv: For some reason I assumed Docker installs required userns, but in retrospect that doesn't make sense. OK, removed the warning.

paulproteus:
Status update: Looks good to me; In 24 hours, I'll merge then ask for a install.sh release.

Let's wait the 24 hours so the new Roundcube update can flow to people's servers via auto-updates, then release this at/around 5 PM Pacific tomorrow.

I've tested everything I want to test by then and updated the docs with a PR <https://github.com/sandstorm-io/sandstorm/pull/2694> and updated the website with a PR <https://github.com/sandstorm-io/sandstorm-website/pull/253>, which I can merge+deploy after we do the install.sh release.

install.sh
```diff
@@ -457,151 +456,69 @@ Mac system, you can use Vagrant or another virtualization tool. See our install
...
+  # userns sysctl, if the system has it and we want it.
 
-  # Let's see if there's a Debian-style sysctl. If not, bail out because we have no idea how to help
-  # the user.
-  if [ ! -e /proc/sys/kernel/unprivileged_userns_clone ]; then
-    fail "E_NO_USERNS" "Your kernel does not appear to be compiled with" \
-         "support for unprivileged user namespaces (CONFIG_USER_NS=y), or something else is" \
-         "preventing creation of user namespaces. This feature is critical for sandboxing." \
-         "Read more at: " \
-         "https://docs.sandstorm.io/en/latest/administering/faq/#enabling-user-namespaces"
```
paulproteus: Notably: In https://github.com/sandstorm-io/sandstorm/pull/2694 I plan to remove this section from the docs. Happily, install.sh is also about to stop referring to it.

This does mean that old copies of install.sh will link here. I keep wondering if we should avoid breaking that link. But I guess people will be OK.

paulproteus:
This change does mean that there is now no way to do a headless install on a Debian system that doesn't enable this sysctl; furthermore, we now don't ask if that's OK up-front. In https://docs.sandstorm.io/en/latest/administering/install-script/ you'll see that using `-d` is part of our suggestions for headless installs.

@kentonv if that's OK with you, it's OK with me. It's a little suboptimal but there's so complexity in the requirements that I can live with it.

If it's not OK with you, one thing we could do is add a `OVERRIDE_TRY_TO_ENABLE_USERNS=no` environment variable which would become one of the documented overrides at https://docs.sandstorm.io/en/latest/administering/install-script/ . 

A different option is we could "just" warn people as part of the "Are you sure?" printout, like we used to.

Now it's my turn to be apologetic about not catching this in earlier review!

I'll adjust docs to accommodate the current behavior, and you can see what I mean.

kentonv:
@paulproteus I'm not too worried about that, but I think the solution if we feel it's needed is to introduce a new flag for unattended non-dev installs. I don't feel the need to do it in this PR, in any case.

paulproteus:
Works for me for now.

paulproteus:
@kentonv once this is live, I'll click merge on:

- https://github.com/sandstorm-io/sandstorm-website/issues/248 (and then do a website deploy)

- https://github.com/sandstorm-io/sandstorm/pull/2694 (to update docs to say it works on RHEL etc.)

PR #2670, opened 2016-10-17
install.sh: Use /dev/net in bash where available

paulproteus:


paulproteus:
/dev/tcp rather.

Self-merging due to 100% install.sh change. Will run installer-tests in Jenkins immediately.

PR #2671, opened 2016-10-18
"Randomize" sandbox UID and GID in dev mode.

kentonv:
This helps detect bugs in apps where they have hard-coded a particular UID or GID.

### DIFF
--- a/src/sandstorm/supervisor.c++
+++ b/src/sandstorm/supervisor.c++
@@ -905,9 +905,22 @@ void SupervisorMain::unshareOuter() {
     KJ_SYSCALL(unshare(CLONE_NEWUSER | CLONE_NEWNS | CLONE_NEWIPC | CLONE_NEWUTS | CLONE_NEWPID));
 
     // Map ourselves as 1000:1000, since it costs nothing to mask the uid and gid.
+    uid_t fakeUid = 1000;
+    gid_t fakeGid = 1000;
+
+    if (devmode) {
+      // "Randomize" the UID and GID in dev mode. This catches app bugs where the app expects the
+      // UID or GID to be always 1000, which is not true of servers that use the privileged sandbox
+      // rather than the userns sandbox. (The "randomization" algorithm here is only meant to
+      // appear random to a human.)
+      time_t now = time(nullptr);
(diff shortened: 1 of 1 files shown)


src/sandstorm/supervisor.c++
```diff
@@ -905,9 +905,22 @@ void SupervisorMain::unshareOuter() {
...
+    uid_t fakeUid = 1000;
+    gid_t fakeGid = 1000;
+
+    if (devmode) {
+      // "Randomize" the UID and GID in dev mode. This catches app bugs where the app expects the
+      // UID or GID to be always 1000, which is not true of servers that use the privileged sandbox
+      // rather than the userns sandbox. (The "randomization" algorithm here is only meant to
+      // appear random to a human.)
+      time_t now = time(nullptr);
+      fakeUid = now * 4721 % 2000 + 1;
```
paulproteus: These are some weird constants, and I'd prefer you pull data from `/dev/urandom` or something that was more defensible, even though I know it's supposed to be "only random to humans". But out of curiosity, what's with the constants?

zarvox: `<drew-having-fun>`

These constants can be reduced by the following equivalences:

```
now * 4721 % 2000 + 1 =
(now * 4000 + now * 721) % 2000 + 1 =
((now * 4000 % 2000) + (now * 721 % 2000)) % 2000 + 1 =
now * 721 % 2000 + 1
```

A similar proof applies to the `fakeGid` case.

```
fakeUid = now * 721 % 2000 + 1;
fakeGid = now * 791 % 2000 + 1;
```

Both 791 and 721 are coprime with 2000, so each is a generator over the group of integers between 0 and 2000.  This means they'll cover each possible UID over the course of 2000 seconds.

791 and 721 share a factor of 7, so you will see repeated (uid:gid) pairs every (7 * 113 * 103) * 2000 seconds, which could theoretically be improved if they didn't share a factor to simply their product.

`</drew-having-fun>`

No action required IMO. :)

kentonv: These "constants" are two prime numbers arbitrarily chosen from https://primes.utm.edu/lists/small/1000.txt

PR #2674, opened 2016-10-18
docs: Fix broken Markdown for a link

paulproteus:
Thanks @mokomull


paulproteus:

PR #2675, opened 2016-10-18
docs: Add Azure Active Directory docs

paulproteus:
Also:

- Make this page very wide, so that the images can be seen clearly.

- Add commented-out docs for Windows Server Active Directory, whose
  text & screenshots I haven't edited yet.

### DIFF
--- a/docs/administering/active-directory.md
+++ b/docs/administering/active-directory.md
@@ -82,7 +82,7 @@ follow these instructions.
 
 ![Screenshot of this step](https://alpha-evgl4wnivwih0k6mzxt3.sandstorm.io/azure-ad-config/11.png)
 
-- Now, visit your Sandstorm server and begin configuring SAML login. In the Sandstorm SAML config, make sure your "Entity ID" is the full URL to your Sandstorm server, including the leading "https://", which you will also use as the Azure AD "Identifier". Azure AD requires that the entity ID be a full URL, not a hostname. You will also need to copy the Sandstorm server's Service URL to your clipboard.
+- Now, visit your Sandstorm server and begin configuring SAML login. In the Sandstorm SAML config, adjust the "Entity ID" to be the full URL to your Sandstorm server, including the leading "https://", which you will also use as the Azure AD "Identifier". Azure AD requires that the entity ID be a full URL, not a hostname. You will also need to copy the Sandstorm server's Service URL to your clipboard.
 
 ![Screenshot of this step](https://alpha-evgl4wnivwih0k6mzxt3.sandstorm.io/azure-ad-config/13-blanked.png)
 
--- a/docs/administering/active-directory.md
+++ b/docs/administering/active-directory.md
@@ -168,7 +168,7 @@ Here is a screenshot tour, using example.sandcats.io as an example Sandstorm ser
   the Sandstorm service provider metadata URL. This is typically of the form:
   https://example.sandcats.io/_saml/config/default
 
-![Screenshot of this step](https://alpha-evgl4wnivwih0k6mzxt3.sandstorm.io/windows-server-ad-config/13-blanked.png)
+![Screenshot of this step](https://alpha-evgl4wnivwih0k6mzxt3.sandstorm.io/windows-server-ad-config/26-blanked.png)
 
 - Back in the "Add Relying Party Trust Wizard," choose the radio button for: "Import data about the
   relying party published online or on a local network".  Under "Federation metadata address (host
@@ -296,7 +296,9 @@ Here is a screenshot tour, using example.sandcats.io as an example Sandstorm ser
     - Under "SAML cert for above provider", copy the contents of the certificate you extracted in part
       2, above.
 
-    - Use the full server URL as the entity ID. For example, https://example.sandcats.io
+    - For the entity ID, keep the Sandstorm default, which is your server hostname. For example,
+      example.sandcats.io. If you change this value, logins will not work until you resynchronize AD
+      FS with Sandstorm's SAML metadata.
 
     - Finally, click "Save."
 
--- a/docs/administering/active-directory.md
+++ b/docs/administering/active-directory.md
@@ -141,9 +141,12 @@ Sandstorm.  Once you have done this, you can use Active Directory to choose whic
 domain are allowed to use Sandstorm.
 
 To connect to Active Directory via SAML, you must enable Active Directory Federation Services (AD
-FS). We will assume you have already done this. We also assume that the Sandstorm server can
-reach the Active Directory server, e.g. they are on the same VPN, or firewall rules permit them
-to communicate.
+FS). We will assume you have already done this. We also assume that your users' web browsers can
+reach the the Sandstorm server as well as the AD server. With SAML, all communication between the
+Service Provider (Sandstorm) and the Identity Provider (AD) is through the browser. The only
+server-to-server communication happens during setup, where AD FS fetches a configuration XML file
+from Sandstorm. If desired, the admin can avoid this server-to-server communication be by
+downloading the XML from Sandstorm and then uploading it to AD separately.
 
 Note that AD FS requires that your Sandstorm server supports HTTPS with a valid certificate. You can
 use a private CA for this, but setting up a private CA is beyond the scope of this document.
--- a/docs/administering/active-directory.md
+++ b/docs/administering/active-directory.md
@@ -6,11 +6,14 @@ integration requires [Sandstorm for Work.](for-work.md)
 
 Sandstorm can rely on AD to authenticate users.
 
-Active Directory exists as a cloud product called Windows Azure as well as a software product called
-Windows Server. Sandstorm supports both products.
+Active Directory exists as part of a cloud product called Windows Azure as well as a software
+product called Windows Server. Sandstorm supports both products.
 
-Sandstorm can be configured to use single sign-on in a configuration where Sandstorm never sees the
-user's passwords. This relies on Sandstorm's support for SAML 2.0.
+Sandstorm can be configured to use single sign-on in a configuration where AD has full control over
+the login form, allowing you to enjoy AD's support for multi-factor authentication, Windows
+credentials login, and a consistent sign-in experience for all apps within your organization. As a
+security benefit, the Sandstorm software never sees the user's passwords. This relies on Sandstorm's
+support for SAML 2.0.
 
 This document provides textual advice as well as a great deal of screenshots to allow you to proceed
 with confidence. If you have questions, please email support@sandstorm.io. We want to help you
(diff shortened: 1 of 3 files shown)


paulproteus:
Requesting a +1/-1 from @kentonv.

I know you suggested editing the screenshots, but I personally find them very readable as-is without e.g. arrows.

I also made one text change that seemed important to me to make, and yet I wasn't sure I understood the original text, so I wanted to know if it is an acceptable text change! Specifically:


In the original text, you wrote this:

> - Click the checkbox to complete configuration.
> - Click: "Assign accounts"
> - Choose "All Users" from the drop-down.

But I didn't see any screenshot with "Assign accounts" or "All users" in them. So I removed that  text in the instructions. But I might be missing something. Can you clarify?

I'll wait on the above before I merge this.

paulproteus:
FWIW:

https://github.com/sandstorm-io/sandstorm/blob/44b59aa7fd1b9f0df3990400748f0b0c9d1352e0/docs/administering/active-directory.md

is a good way to preview these changes!

paulproteus:
Question for @kentonv:

- In the Windows Server SAML docs, I notice that the Entity ID in Sandstorm is `example.sandcats.io` not `https://example.sandcats.io`. Is that intentional? Does it matter?

- In the context of the "Add Transform Claim Rule Wizard", you wrote this bullet point:

> - Under "Attribute store", choose "Active Directory".

But I didn't see any information about that in the screenshots.

Also @kentonv , for Azure, do we use the "old portal"? Presumably that means we'll need to re-do this layout & screenshot work in the future? Should we switch to the "new portal"?

ocdtrekkie:
I'm going to spend some time reviewing this, probably tonight, from the Windows system admin perspective.

ocdtrekkie:
"ActiveDirectory" needs a space in the Overview section.

paulproteus:
Yippee! Thanks @ocdtrekkie !

FWIW, make sure to review the current version, rather than the version that I link to above that's specific to a single GitHub commit ID.

https://github.com/sandstorm-io/sandstorm/blob/docs-for-azure-activedirectory/docs/administering/active-directory.md is a good way to get that.

Note also that the screenshots will be a little less wide when they go live in the docs. I can put up a test version of the generated docs if you like, but hopefully it doesn't matter much.

ocdtrekkie:
Yeah, I'm looking at the latest commit. I should use the Review feature, so I don't spam you with each one.

docs/administering/active-directory.md
```diff
@@ -0,0 +1,304 @@
...
+![Screenshot of this step](https://alpha-evgl4wnivwih0k6mzxt3.sandstorm.io/azure-ad-config/13.png)
+
+![Screenshot of this step](https://alpha-evgl4wnivwih0k6mzxt3.sandstorm.io/azure-ad-config/16.png)
+
+- Back in the Azure AD configuration, copy the line titled "Single Sign-On Service URL". In the Sandstorm config, paste this under "SAML provider entry point URL:".
+
+![Screenshot of this step](https://alpha-evgl4wnivwih0k6mzxt3.sandstorm.io/azure-ad-config/17.png)
+
+- In the Azure AD config, check the box: "Confirm that you have configured single sign-on as
+  described above." Clik the "next" arrow and enter your email address for service alerts from
```
ocdtrekkie: "Click"

kentonv:
> In the Windows Server SAML docs, I notice that the Entity ID in Sandstorm is example.sandcats.io not https://example.sandcats.io. Is that intentional? Does it matter?

Sandstorm defaults to the bare hostname as the Entity ID. For Windows Server, it doesn't matter because it picks up the correct entity ID via the XML configuration blob -- the most important thing is that people don't change it after they've already configured the AD FS side. So I intentionally made sure the form field had its default value here, to avoid making people think they had to change it.

Azure, on the other hand, seems to require that the entity ID is a full URL.

> In the context of the "Add Transform Claim Rule Wizard", you wrote this bullet point:
>
> > Under "Attribute store", choose "Active Directory".
>
> But I didn't see any information about that in the screenshots.

It's the drop-down right in the middle of the dialog:



> Also @kentonv, for Azure, do we use the "old portal"? Presumably that means we'll need to re-do this layout & screenshot work in the future? Should we switch to the "new portal"?

AFAICT the new portal does not yet support configuring an arbitrary SAML endpoint. I noted this in the bullet points: "If this option is not displayed: Make sure you are using the classic portal. This option does not yet appear in the new portal. This option may require upgrading to Azure AD Premium."

docs/administering/active-directory.md
```diff
@@ -0,0 +1,304 @@
+This document explains how to integrate with Microsoft Active Directory (also known as AD).  This
+allows all users, or specific users within your Windows domain, to log into a Sandstorm server. The
+integration requires [Sandstorm for Work.](for-work.md)
+
+## AD & Sandstorm: Overview
+
+Sandstorm can rely on AD to authenticate users.
+
+Active Directory exists as a cloud product called Windows Azure as well as a software product called
```
kentonv: "exists as a" -> "exists as part of a"

Otherwise I feel like this is saying that Windows Azure and Windows Server are versions of Active Directory, rather than bundles that include versions of Active Directory...

docs/administering/active-directory.md
```diff
@@ -0,0 +1,304 @@
...
+Sandstorm can be configured to use single sign-on in a configuration where Sandstorm never sees the
+user's passwords. This relies on Sandstorm's support for SAML 2.0.
+
+This document provides textual advice as well as a great deal of screenshots to allow you to proceed
+with confidence. If you have questions, please email support@sandstorm.io. We want to help you
+successfully set up Sandstorm!
+
+Note that you can also set up Sandstorm to integrate with Active Directory using LDAP bind to
+authenticate users. We recommend Active Directory Federation Services or Microsoft Azure AD Single
+Sign-On instead, which use SAML. SAML has the advantage that users never type passwords into
```
kentonv: I think "use" is correct -- he mentioned two different products earlier in the sentence.

docs/administering/active-directory.md
```diff
@@ -0,0 +1,304 @@
...
+Sandstorm. If you must use LDAP instead, see the general [Sandstorm for Work documentation about
```
kentonv: I think the main reason people will want to use SAML is not because Sandstorm sees their passwords with LDAP but rather because SAML allows AD full control over the login form, which enables things like multi-factor auth, Windows credentials login, etc. So I think that should be mentioned here...

docs/administering/active-directory.md
```diff
@@ -0,0 +1,304 @@
...
+- Open Sandstorm's SAML login configuration.
+
+    - Under "SAML provider entry point URL", enter your AD FS login entry point. This is usually
+      "/adfs/ls" at the host associated with your Active Directory server. For example, if your Active
+      Directory server is ad.example.com, the URL would be: https://ad.example.com/adfs/ls
+
+    - Under "SAML cert for above provider", copy the contents of the certificate you extracted in part
+      2, above.
+
+    - Use the full server URL as the entity ID. For example, https://example.sandcats.io
```
kentonv: As mentioned elsewhere, it's important that people *don't* change the Entity ID at this point, otherwise they'll break the config that has already been synced into AD FS. The default entity ID does not include "https://" and works just fine here.

docs/administering/active-directory.md
```diff
@@ -0,0 +1,304 @@
...
+add a new Relying Party, namely the Sandstorm web application. Each Sandstorm server publishes
+federation metadata information at a URL available within the SAML configuration page.  You must
+configure a rule that sends the user email address to Sandstorm, as well as a rule that declares
+email addresses are the persistent identifier. You must also export the custom SAML certificate to
+Sandstorm.  Once you have done this, you can use Active Directory to choose which users in your
+domain are allowed to use Sandstorm.
+
+To connect to Active Directory via SAML, you must enable Active Directory Federation Services (AD
+FS). We will assume you have already done this. We also assume that the Sandstorm server can
+reach the Active Directory server, e.g. they are on the same VPN, or firewall rules permit them
```
kentonv: Actually, Sandstorm does not need to reach the AD server. Only the client's browser needs to reach it. With SAML all communication between the SP and IdP is through the browser.

The only server<->server communication happens during setup, where AD FS fetches the XML config blob from Sandstorm. So it's actually Sandstorm that needs to be accessible from AD. (Though, this could be worked around by the admin downloading the XML from Sandstorm and then uploading it to AD separately.)

docs/administering/active-directory.md
ocdtrekkie: You are in fact correct here. :/ This is just hard to read and sounds wrong because of Microsoft's tendency of naming schemes for things that are four words or longer.

docs/administering/active-directory.md
paulproteus: Thanks, will fix

docs/administering/active-directory.md
paulproteus: Thanks for noticing that this is confusing. I'll see if I can do a minor rephrasing.

docs/administering/active-directory.md
paulproteus: Fascinating. Thanks! I'll adjust the text accordingly.

docs/administering/active-directory.md
paulproteus: +1, see soon-to-come revised text

docs/administering/active-directory.md
paulproteus: +1, see soon-to-come revised text

paulproteus:
@ocdtrekkie and @kentonv - this is now ready for re-review. I believe I have addressed all feedback.

I'm still not sure about this aspect. I wrote in a comment at the top:

> In the original text, you [@kentonv] wrote this:
>
> > - Click the checkbox to complete configuration.
> > - Click: "Assign accounts"
> > - Choose "All Users" from the drop-down.
>
> But I didn't see any screenshot with "Assign accounts" or "All users" in them. So I removed that text in the instructions. But I might be missing something. Can you clarify?

Other than that I'm happy to ship this.

kentonv:
> Click the checkbox to complete configuration.

This should perhaps have been "click the checkmark button" -- it's a circular button in the lower-right of the window which would normally be labeled "OK" or "Finish" but some designer decided giving it a checkmark would be clearer.

> Click: "Assign accounts"

Oh, sorry. I should have mentioned this. So after you click the check to complete that stage of configuration, you get sent back to this screen:



I didn't take a new screenshot since it's the same screen as before. This is where you click "Assign accounts".

> Choose "All Users" from the drop-down.

On this screen:



The drop-down that says "Groups" needs to be changed to "All Users" (as it is in the following shot). Unfortunately I couldn't get a screenshot of the drop-down being open since the drop-down is rendered as a separate window thus isn't included in the shot. (You might have noticed the weird shot where the drop-box seems to be activated but there's no menu below it... that was my attempt to take a screenshot but it apparently didn't work.)

ocdtrekkie:
I am happy with all of the changes since my last review.

paulproteus:
OK! I think I've addressed all relevant feedback and will now merge. Post-merge feedback is OK too.

PR #2676, opened 2016-10-19
Rework sign-in overlay postMessage API

zarvox:
The previous sign-in overlay did not behave well in the presence of scrolling.
Now, we show a modal overlay rather than painting over the app at the same
time.

This change also makes the sign-in dialog's label optionally explicit, which means
we can say "Create account" in the setup wizard for the first user, or "Sign in" if you're
trying to promote an existing user to admin.


neynah:
Exciting! Could we see a screenshot?

zarvox:
before-click

create-account

sign-in

PR #2678, opened 2016-10-19
Marginally re-organize Active Directory docs

paulproteus:


paulproteus:
This moves SAML to the top so that people looking for AD will find it faster.

PR #2680, opened 2016-10-20
installer-tests: Add CentOS test

paulproteus:


paulproteus:

PR #2693, opened 2016-10-24
docs: Offer advice on unattended-upgrades

paulproteus:
Thanks to Mark King for contacting us with a question that inspired this note.


paulproteus:
Self-merging b/c 100% docs.

PR #2694, opened 2016-10-24
docs: Explain that userns is now optional

### LINKED ISSUE #2645: docs: Explain that userns is now optional
Once https://github.com/sandstorm-io/sandstorm/pull/2644 lands, we'll need to make some changes to docs. This issue tracks finding those required docs changes, and then making them.

## List of docs that will need to be changed

- [x] "Enabling user namespaces" in https://docs.sandstorm.io/en/latest/administering/faq/#enabling-user-namespaces

- [x] "Sandstorm uses user namespaces to set up its own sandbox, but then disallows the sandboxed app from creating its own nested namespaces, rendering these vulnerabilities unexploitable" https://docs.sandstorm.io/en/latest/using/security-non-events/#linux-kernel (@kentonv perhaps you should be the one to change this one?)

- [x] Re-test in Docker, and update this doc: https://docs.sandstorm.io/en/latest/install/#option-6-using-sandstorm-within-docker (also this might be a good time to add an installer-tests for that)

- [x] Add docs on the two kinds of sandboxing we do (userns and no-userns) somewhere. I presume the security-practices page is the right place for this. Having a technical note about this will also help me when I field questions like "Doesn't Sandstorm fail on RHEL 7?" - then I can say "It works now on RHEL 7 - see this part of the docs." I can take point on writing this.

## List of docs that seemingly don't need changes

- https://docs.sandstorm.io/en/latest/using/security-practices/#server-sandboxing - this page doesn't seem to mention userns.

paulproteus:
Close #2645


kentonv:
I feel like we should remove the whole FAQ entry and maybe replace it with a different entry entitled "Installing and running without root privileges".

I don't think any user should bother themselves with the details of the userns sandbox vs. the privileged sandbox unless they are seeking to avoid running as root. Giving the user too much information about things that don't really matter to them can actually be harmful, since it potentially wastes their time or even scares them away from the product.

paulproteus:
@kentonv +1. That's something that was vaguely on my mind, but I had figured that if you wanted less info, you would have said as much on the earlier GitHub issue. I prefer your strategy and I'll revise with that in mind.

kentonv:
Sorry for failing to suggest this earlier!

paulproteus:
No problem. I'm happy enough as-is; we have a day for me to revise this before the change goes live, and I accept some amount of churn in pull requests as just part of life!

paulproteus:
Thanks @kentonv . Requesting re-review.

paulproteus:
Added another commit that is perhaps controversial ("Explain that -d is development defaults") since it means there's no way to get a non-systemwide-userns-enabled unattended install; see https://github.com/sandstorm-io/sandstorm/pull/2656 for discussion.

kentonv:
lgtm

paulproteus:
Self-merging because Kenton says it's OK, plus 100% docs :)

PR #2697, opened 2016-10-25
docs: Explain strange DNS lookups in admin faq

paulproteus:
Thanks to @zarvox for the text.

I'm going to self-merge now, to avoid latency, but post-merge feedback is very welcome.

### DIFF
--- a/docs/administering/faq.md
+++ b/docs/administering/faq.md
@@ -658,3 +658,44 @@ tutorial.](https://davejamesmiller.wordpress.com/2013/03/14/installing-dnsmasq-w
 **Email configuration within Sandstorm.** Some Sandstorm features expect email delivery to be
 configured. If you cannot configure email on your server, those features may be disabled or not work
 properly. However, this is not specific to an offline/airgap network.
+
+## Why do I see strange DNS lookups in my server log?
+
+When viewing your server's admin log, you might see DNS-related messages like this. They are
+harmless and can be ignored. They are probably the result of bots scanning the Internet for security
+issues that do not affect Sandstorm.
+
+```
+Error: Host "sandstorm-www.notyourdomain.example.com" must have exactly one TXT record.
+    at server/pre-meteor.js:728:16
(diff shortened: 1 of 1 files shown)


paulproteus:
self-merge because 100% docs

docs/administering/faq.md
```diff
@@ -658,3 +658,44 @@ tutorial.](https://davejamesmiller.wordpress.com/2013/03/14/installing-dnsmasq-w
...
+Sandstorm supports publishing static sites at domains that are not part of its "wildcard domain."
+This allows you to publish a website at a domain like example.com from your Sandstorm server,
+preserving the domain. To achieve this, you would make a DNS A record for example.com pointing at
+your Sandstorm server. This ensures that when users navigate to example.com in their browser, the
+browser sends the request to Sandstorm.
+
+Once Sandstorm receives this request, it needs to know which grain it should dispatch this request
+to. This is supported by having the user create a DNS TXT record for sandstorm-www.example.com
+containing the grain's "public ID". If no such DNS record is found, Sandstorm takes that as an
+indication that the domain owner has not indicated their desire to publish a particular Sandstorm
```
zarvox: the root "indicate" is used twice in this line which reads oddly

docs/administering/faq.md
```diff
@@ -658,3 +658,44 @@ tutorial.](https://davejamesmiller.wordpress.com/2013/03/14/installing-dnsmasq-w
...
+grain at the domain from the HTTP request.
+
+What's happening in your log traces is that Sandstorm is receiving HTTP requests for those domains,
+performing the DNS lookup to see what grain's static publishing folder should be used to answer
+those requests, and the DNS lookup is failing, which is distinct from returning no record.
```
zarvox: Actually, based on the fact that you also get a log message for 0 TXT records, this statement is not quite right - it might be better to say something like:

"...should be used to answer those requests, and isn't receiving a DNS response in the expected format."

paulproteus:
Thanks, both are maybe fixed in https://github.com/sandstorm-io/sandstorm/pull/2699 !

PR #2700, opened 2016-10-25
Pass user/identity ID to sessions for standalone domains

jparyani:
This fixes a bug in powerbox offer + standalone domains

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -428,6 +428,8 @@ Meteor.methods({
       let maybeUserId = null;
       let maybeIdentityId = null;
       if (neverRedeem) {
+        // Powerbox offers expect that sessions will have a userId attached to them, and will fail
+        // without it if the user is actually logged in.
         maybeUserId = this.userId;
         maybeIdentityId = identityId;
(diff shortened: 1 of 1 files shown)


shell/server/proxy.js
```diff
@@ -425,8 +425,15 @@ Meteor.methods({
                                "User is not authorized to open this grain.");
       }
 
-      const opened = globalBackend.openSessionInternal(apiToken.grainId, null, null,
-                                                       title, apiToken, cachedSalt);
+      let maybeUserId = null;
+      let maybeIdentityId = null;
+      if (neverRedeem) {
```
zarvox: Please add comments explaining why we do this.

kentonv:
I'm actually pretty confused why we aren't _always_ passing in these values to openSession... Does this mean that powerbox requests similarly fail in an incognito view?

paulproteus:
Is this required for moving the feature key vendor AKA Sandstorm for Work dashboard forward? If so, I'd like to request a code release on Oasis that contains this fix in order to minimize delay.

PR #2703, opened 2016-10-26
Fix some issues / simplify code in handling of standalone domains and incognito.

kentonv:


shell/server/proxy.js
```diff
@@ -425,8 +425,13 @@ Meteor.methods({
                                "User is not authorized to open this grain.");
       }
 
-      // Pass in userId/identityId because in the case of a user being logged in, powerbox offers
-      // require it.
+      // Even in incognito mode, we want to record the user ID on the session. The user ID is not
+      // leaked to the app in any way -- it is used to decide what options to show in a powerbox
+      // request. Even if the user is incognito, they should be able to offer their full range of
```
dwrensha: Jason's original comment talked about offers. This new comment talks about requests. With this change, do both requests and offers work for incognito logged-in users?

dwrensha:
Looks good to me, once you clarify whether this pertains to powerbox requests, offers, or both.

shell/server/proxy.js
kentonv: It's definitely requests that were broken. I think Jason mixed them up in the comment.

PR #2707, opened 2016-10-27
Fix powerbox requests for standalone domains.

kentonv:
Claiming the request was failing if the identity didn't have direct access to the grain, because a MembraneRequirement was being added requiring that the identity have such access. Instead, we should add a token requirement.


kentonv:
(Would have @dwrensha review but he's out today.)

dwrensha:
👍

PR #2713, opened 2016-10-28
Fix updates for non-root installs.

### LINKED ISSUE #2692: Server crashed after update to 192
Had the server go down after the automatic update to 192.

```
** Time: Sat Oct 22 18:50:51 2016
Checking for updates on channel dev...
Downloading: https://dl.sandstorm.io/sandstorm-192.tar.xz
Checking signature...
Signature is valid.
** Successfully updated; restarting.
```

```
** Restarting to apply update
** Shutting down due to signal
Not killing Dev daemon because it is not running.
*** Uncaught exception ***
sandstorm/run-bundle.c++:1345: failed: unshare(CLONE_NEWNS): Operation not permitted
stack: 0x4e62d4 0x4e1bdf 0x507a96 0x50796d 0x49219d 0x492ca0
** Server monitor died. Aborting.
```

After manually starting the server it's working fine.

kentonv:
Currenly, for servers running with no root privileges, each update crashes the server. Starting the server back up again works fine.

The problem was introduced in the change introducing the privileged sandbox. On non-root installs, Sandstorm's outer container needs to use a user namespace. Within that user namespace, the UID was being mapped to zero. This caused a newly-added assert in supervisor.c++ to fail. Rather than roll back the assert, I decided to change the UID mapping to non-zero, *completely ignoring* the comment *right above* the code that explained why it really does need to be zero. :(

This change goes back to mapping as UID zero and removes the asserts instead.

Note that UID zero inside the user namespace is not actually root on the system. Also, we explicitly drop crapabilities so that the front-end process does not even get within-namespace root rights.

Unfortunately, because of the nature of the bug, the update which fixes the bug will again fail -- but updates after that will succeed.

Fixes #2692.

### DIFF
--- a/src/sandstorm/backup.c++
+++ b/src/sandstorm/backup.c++
@@ -106,9 +106,6 @@ void BackupMain::bind(kj::StringPtr src, kj::StringPtr dst, unsigned long flags)
 }
 
 bool BackupMain::run(kj::StringPtr grainDir) {
-  KJ_ASSERT(geteuid() != 0, "can't run as user ID zero (root)");
-  KJ_ASSERT(getegid() != 0, "can't run as group ID zero (root)");
-
   // Enable no_new_privs so that once we drop privileges we can never regain them through e.g.
   // execing a suid-root binary, as a backup measure. This is a backup measure in case someone
   // finds an arbitrary code execution exploit in zip/unzip; it's not needed otherwise.
--- a/src/sandstorm/supervisor.c++
+++ b/src/sandstorm/supervisor.c++
@@ -663,9 +663,6 @@ kj::MainBuilder::Validity SupervisorMain::addCommandArg(kj::StringPtr arg) {
 // =====================================================================================
 
 kj::MainBuilder::Validity SupervisorMain::run() {
-  KJ_ASSERT(geteuid() != 0, "can't run as user ID zero (root)");
-  KJ_ASSERT(getegid() != 0, "can't run as group ID zero (root)");
-
   isIpTablesAvailable = checkIfIpTablesLoaded();
 
   setupSupervisor();
--- a/src/sandstorm/run-bundle.c++
+++ b/src/sandstorm/run-bundle.c++
@@ -1318,8 +1318,8 @@ private:
       // But if our UID is zero, then the file's attributes are ignored and all capabilities are
       // inherited.
       writeSetgroupsIfPresent("deny\n");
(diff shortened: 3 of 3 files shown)


src/sandstorm/supervisor.c++
```diff
@@ -663,9 +663,6 @@ kj::MainBuilder::Validity SupervisorMain::addCommandArg(kj::StringPtr arg) {
 // =====================================================================================
 
 kj::MainBuilder::Validity SupervisorMain::run() {
-  KJ_ASSERT(geteuid() != 0, "can't run as user ID zero (root)");
```
paulproteus: Does this mean that if you set `SANDSTORM_USER=root` it'll now run the supervisor + grains as root?

paulproteus: Let me change this to an actual "GitHub review..." one sec :)

src/sandstorm/backup.c++
```diff
@@ -106,9 +106,6 @@ void BackupMain::bind(kj::StringPtr src, kj::StringPtr dst, unsigned long flags)
 }
 
 bool BackupMain::run(kj::StringPtr grainDir) {
-  KJ_ASSERT(geteuid() != 0, "can't run as user ID zero (root)");
```
paulproteus: I don't see how these assert() changes in backup.c++ are relevant. Does that mean on the privileged sandbox, downloading grain backups theoretically doesn't work right now? Let me go check that.

kentonv: We need to remove the asserts because, for non-root installs, the perceived euid inside the outer container is zero, because that's what the user namespace maps the current user to. Without this change, backups would break for non-root installs.

kentonv:
> Should someone add a test somewhere that this doesn't break in the future?

Maybe the installer tests could test doing a `sandstorm update` and verify that the server comes back up and still works?

We could also add an update test to the regular test suite, but seemingly it would _only_ test the non-root case, and it seems like we should test cases where installing as root, too.

kentonv:
>  haven't manually tested it. Should I, or do you trust your existing manual test procedure for this @kentonv ?

I have tested it manually, yes.

kentonv:
Merging for release.

I'm confident in this change because:
1) Aside from removing asserts (which can't possibly break anything), the only changes are to code which is only executed for non-root installs.
2) I manually tested that this fixes updates of non-root installs.
3) The automated tests operate on a non-root install.

I do agree we should add upgrade tests, but I'm not very familiar with the installer tests so I might need help understanding where to put those.

PR #2714, opened 2016-10-28
Shutdown trashed grains immediately

### LINKED ISSUE #2648: Trashing a grain doesn't shut it down
It appears that trashing a running grain does not shut down the app -- it continues to run normally until it times out. It should be proactively killed, since the grain is no longer accessible anyway.

jparyani:
Fixes #2648

### DIFF
--- a/shell/shared/grain-shared.js
+++ b/shell/shared/grain-shared.js
@@ -14,6 +14,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import { waitPromise } from "/imports/server/async-helpers.js";
+
 Meteor.methods({
   // Methods defined in this file have meaningful latency compensation (client-side prediction)
   // potential.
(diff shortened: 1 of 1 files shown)


shell/shared/grain-shared.js
```diff
@@ -14,6 +14,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import { waitPromise } from "/imports/server/async-helpers.js";
```
kentonv: Doesn't this fail on the client? This file runs on both server and client (for client-side prediction purposes). I think you need to guard your code accordingly.

BTW, meteor-promise adds a `.await()` method to promises which you can use instead of importing `waitPromise()`.

shell/shared/grain-shared.js
```diff
@@ -57,6 +57,21 @@ Meteor.methods({
...
                        { $set: { "trashed": new Date() } },
                        { multi: true });
+
+      const grainsOwned = Grains.find({
+        userId: { $eq: this.userId },
+        _id: { $in: grainIds },
+      }, { fields: { _id: 1, }, });
+      grainsOwned.forEach((grain) => {
+        try {
+          Sessions.remove({ grainId: grain._id, });
```
kentonv: Seems like this shouldn't be in the try/catch? Then the entire try/catch can be in if (!this.isSimulation).

jparyani: Theoretically mongo operations could throw exceptions, although I will say we rarely ever catch/deal with them...

kentonv: That's the thing: I don't think we want to catch and squelch such exceptions.

PR #2716, opened 2016-10-30
nginx configuration tweaks for geolocation

mrflos:
1. Passes the remote address in the header so that sandstorm can use the remote address in apps like Piwik (fixes https://github.com/zarvox/piwik/issues/10 )
2. Changes server_name so that sandstorm api access url will be recognized.

### DIFF
--- a/docs/administering/sample-config/nginx-example.conf
+++ b/docs/administering/sample-config/nginx-example.conf
@@ -24,7 +24,7 @@ server {
 # Configuration for Sandstorm shell and apps, over HTTPS.
 server {
   listen 443;
-  server_name example.com *.example.com;
+  server_name example.com *.example.com ~^sandstorm-[^.]*\.example\.com$;
 
   ssl on;
   ssl_certificate /etc/nginx/ssl/sandstorm.crt;
@@ -55,6 +55,7 @@ server {
     proxy_http_version 1.1;
     proxy_set_header Upgrade $http_upgrade;
     proxy_set_header Connection $connection_upgrade;
+    proxy_set_header X-Real-IP $remote_addr;
   }
 
   # Allow large spk uploads from the /install form and allow grains to receive large uploads.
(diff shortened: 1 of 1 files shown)


garply:

docs/administering/sample-config/nginx-example.conf
```diff
@@ -24,7 +24,7 @@ server {
 # Configuration for Sandstorm shell and apps, over HTTPS.
 server {
   listen 443;
-  server_name example.com *.example.com;
+  server_name example.com *.example.com ~^sandstorm-[^.]*\.example\.com$;
```
zarvox: My reading of this is that `~^sandstorm-[^.]*\.example\.com$` should already be covered by `*.example.com`.  Am I missing something?

mrflos: In fact my configuration was with sandstorm installed on a subdomain.  
`server_name sandstorm.example.com ~^sandstorm-[^.]*\.example\.com$ *.sandstorm.example.com;
`  
In this case, it was needed for ghost blog static publishing:  
`sandstorm-www.host.example.com IN TXT xql6hui0fqaguiazpnk3`  

This was a while ago, maybe it is not necessary anymore?

kentonv: Adding this regex isn't correct for our example. Arguably we should have another example showing how to wedge Sandstorm into a wildcard that also has non-Sandstorm things on it, by using the `sandstorm-*.example.com` trick. But we can't really merge that into this example without it being inconsistent.

If you revert the server_name changes I'm happy to merge the X-Real-IP changes!

mrflos: @kentonv @zarvox : done.

paulproteus:
Thanks all - this is great, and I'm going to merge this straight-away now.

Issue #2462, opened 2016-08-28
[supervisor] Keep alive grains during network activity (Davros sync, Groovebasin playback)


digitalcircuit:
## In short
* Don't shut down grains with active network connections
 * Monitor grain network activity and ongoing connections, tracking upload/download rates
 * When activity falls below a threshold for a period of time, shut down grain as before
 * Criteria should account for high loss, low speed networks (*e.g. 2G modems*)

Criteria | Rank | Reason
---------|---------|-------------
Impact | ★★☆ *2/3* | Allow larger grain uploads/downloads, better handle slow networks
Risk | ★★☆ *2/3* | Rogue/broken grains could stay alive when not intended or desired
Intrusiveness | ★☆☆ *1/3* | No expected backwards-incompatible changes

## Rationale
Sandstorm grains may make use of long-running connections without a need to remain constantly running.  For greater efficiency, Sandstorm should detect network activity and keep the grain alive, perhaps with some sort of developer or user opt-in/opt-out.

Two examples so far:
* [Davros](https://github.com/mnutt/davros ) exposes a WebDav endpoint for the ownCloud sync clients in order to automatically upload and download files from the grain to one's devices.  While ownCloud initiates new connections by breaking apart large files (*1 GB+*) into smaller chunks, on slow networks (*e.g. 350 KB/s*) individual chunks don't upload in time for Sandstorm to count the grain as still alive, shutting down the grain and interrupting the ongoing sync.
* [Groovebasin](https://github.com/dwrensha/groovebasin ) serves a continuous stream for music playback, which the Sandstorm interface keeps alive via JavaScript pings.  On desktop browsers, this works fine.  On mobile devices, browsers may stop sending the JavaScript pings when switching tabs, apps, or turning off the screen.  Playback continues to function right until Sandstorm kills the grain.
 * *Some browsers will also stop the playback stream after a while, but not nearly as quick as Sandstorm shuts down the grain.  This is likely unrelated to Sandstorm itself; Groovebasin may need a mobile app that acquires a wake-lock.*

### Davros
1.  [Set up a Davros grain](https://oasis.sandstorm.io/appdemo/8aspz4sfjnp8u89000mh2v1xrdyx97ytn8hq71mdzv4p4d8n0n3h )
2.  [Connect the ownCloud sync client](https://owncloud.org/install/#install-clients ) to Davros (*the Clients tab*)
3.  Open the Sandstorm grain log
[...]

Issue #2218, opened 2016-07-13
"remote exception: grain supervisor is dead"


elimisteve:
```
HTTP 500 from https://api-168ddaae....oasis.sandstorm.io/rows; response: `Error: Error: remote exception: grain supervisor is dead
C++ location:(javascript):??
type: disconnected
```

Any idea what causes this?

Context: I'm hammering the CrypTag Sandstorm app pretty hard from my client app trying to recreate the nginx 504 Gateway issue that I experienced the other day that @kentonv said was probably related to the June 12th update and the fact that the CrypTag Sandstorm app was built at about the same time and probably is messed up. I upgraded the CrypTag Sandstorm web app and packaged it with the latest Sandstorm version, but I still got the error. So I created a custom version that turns DEBUG on, but I haven't run into the 504 again... but I _did_ run into the except above. Thanks.

kentonv:
@elimisteve So, this is embarrassing, but we have a bug this week where accepting an app update notification does not actually update the app. So it's possible that you saw the problem continue to occur because the app hadn't actually been updated -- unless you manually uploaded your update or manually went back to the app market and clicked "install".

We've pushing an update this weekend which will correct the problem and re-notify people of missed updates.

kentonv:
As for the new error, my first guess is that Sandstorm isn't treating HTTP API requests as keep-alives for the grain, so the grain shuts itself down after a couple minutes even if it is receiving API requests. The grain will immediately start back up on the next request, but if you catch it at just the right time, maybe it produces this error? This is just a guess, but if correct then this is in fact a Sandstorm bug.

elimisteve:
@kentonv Re: your first comment -- okay that makes perfect sense, because I manually uploaded a new `.spk` file to my local Sandstorm Vagrant VM for testing, and I couldn't replicate the error that I was still getting from Oasis, despite having v0.8.2 of CrypTag accepted into the Sandstorm App Store.

Though I guess what I'm saying is only true if the bug you mention didn't apply to manually uploaded apps/SPKs. Is that the case?

zarvox:
Yes, the bug only affected automatic updates being applied via the button in the notification.  Manually-uploaded SPKs or even manually installing from the App Market would have successfully installed the new package.

kentonv:
@elimisteve Right, manually-uploaded SPKs -- or even going to the app store and clicking "install" -- worked correctly. It's only the auto-update notification that didn't work.

[...]

PR #2728, opened 2016-11-03
Keep grains alive if they have recent API session activity.

dwrensha:
We already have logic that periodically culls the proxy objects being used by API sessions. This patch adds a call to `proxy.supervisor.keepAlive()` for each proxy that we decide not to clean up. It also accounts for requests and responses that are streaming.

This handles the most urgent pieces of #2462 and #2218.

### DIFF
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -1751,7 +1751,7 @@ class Proxy {
           content.body.stream.close();
           response.rejectResponseStream(new Error("client disconnected"));
         } else {
-          const idx = this.streamingResponseCounter.toString();
+          const idx = this.streamingResponseCounter;
           this.streamingResponseCounter += 1;
           const streamHandle = content.body.stream;
           const rawResponseStream = new ResponseStream(
@@ -2066,7 +2066,7 @@ class Proxy {
           }
         });
 
-        const idx = this.streamingRequestCounter.toString();
+        const idx = this.streamingRequestCounter;
         this.streamingRequestCounter += 1;
         this.streamingRequests[idx] = request;
         destructor = () => { delete this.streamingRequests[idx]; };
@@ -2150,7 +2150,7 @@ class Proxy {
           promise.serverStream.sendBytes(head);
         }
 
-        const socketIdx = this.websocketCounter.toString();
+        const socketIdx = this.websocketCounter;
         this.websockets[socketIdx] = socket;
         this.websocketCounter += 1;
         pumpWebSocket(socket, promise.serverStream, () => { delete this.websockets[socketIdx]; });
--- a/shell/server/proxy.js
+++ b/shell/server/proxy.js
@@ -683,8 +683,10 @@ class ApiSessionProxies {
                 this.newBucket[oldHashedToken][oldHashedParams];
           const oldProxy = this.oldBucket[oldHashedToken][oldHashedParams];
           if (oldProxy && !newProxy) {
-            if (Object.keys(oldProxy.websockets).length > 0) {
-              // A client has an open websocket. Keep the proxy around.
+            if (Object.keys(oldProxy.websockets).length > 0 ||
+                Object.keys(oldProxy.streamingRequests).length > 0 ||
+                Object.keys(oldProxy.streamingResponses).length > 0) {
+              // A client has an open stream. Keep the proxy around.
               this.put(oldHashedToken, oldHashedParams, oldProxy);
             } else {
               // We can close this proxy and forget about it.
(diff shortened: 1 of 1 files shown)


paulproteus:
I read the patch, and I'm not sure what parts of #2462 and #2218 are _not_ fixed by this. Curious if you can say!

dwrensha:
If you open a grain frame in its own browser tab, as suggested in #2462, then the `keepSessionAlive` method never gets called, and the grain can shut down.

The "grain supervisor is dead" error reported in #2218 should *never* happen, and seems to indicate some bug having to do with grain restarts. I expect this patch to make that error much more unlikely, but not to fix the root cause.

paulproteus:
Gotcha, thanks!

shell/server/proxy.js
```diff
@@ -1711,11 +1743,22 @@ class Proxy {
...
           response.rejectResponseStream(new Error("HEAD request; content doesn't matter."));
+        } else if (response.socket.destroyed) {
+          // The client has already closed the stream. No additional "close" event will be
+          // emitted. Annoyingly, we cannot rely on `response.write()` throwing an exception
+          // in this case, so if we don't handle it specially we could end up with a
+          // `ResponseStream` that gets stuck waiting for a "drain" event that never arrives.
+          content.body.stream.close();
+          response.rejectResponseStream(new Error("client disconnected"));
         } else {
+          const idx = this.streamingResponseCounter.toString();
```
kentonv: Stringifying idx here probably just means that V8 parses it again when you use it as an index below. Integer indexes (including stringified integers) are handled specially by V8.

kentonv:
The code looks good but this is a somewhat-scary change, in that a bug could mean that we leak grains. Any thoughts on how to test this / how to roll it out safely?

dwrensha:
I did a fair amount of manual testing, one outcome of which was the discovery of the need for [this special case](https://github.com/dwrensha/sandstorm/blob/8f7c5f47a74af64c6556c9c20a4d2b2d48f2c1d6/shell/server/proxy.js#L1746-L1752). I instrumented to code to monitor the number of active streaming requests/responses and then tested uploads with gitweb and downloads with both gitweb and davros (using curl as the client for davros), attempting to induce errors by pressing control-c at random times.

digitalcircuit:
@dwrensha Thanks for this!  In the next few days I'll try to confirm it's fixed with Davros/OwnCloud, and if so I'll update the bug report.

Agreed that the Groovebasin case is not yet fixed, and I'm not sure about the best way of handling that.  One could argue it's not Sandstorm's fault that mobile browsers seem to block background JavaScript requests when the screen's off.

digitalcircuit:
Unfortunately, the Davros bug seems more involved.  Trying to upload a 2.3 GB file at 350 KB/s succeeded up until the last chunk (245 of 246), after which the connection was closed.

On the plus side, **the Sandstorm supervisor seems to work properly** for case 1 of 2, no longer terminating the grain!

I'll poke @mnutt over on the [Davros project](https://github.com/mnutt/davros) when I have a bit more time.  Case 2 of 2 (Groovebasin) likely remains broken, too, but this pull request is complete.  Thanks for the hard work!

Meanwhile, for the curious...
```
** SANDSTORM SUPERVISOR: Grain still in use; staying up for now.
PUT /remote.php/webdav/2016-09-16%2012-14-12.flv-chunking-2593448364-246-243 201 66926.988 ms - 0
PUT /remote.php/webdav/2016-09-16%2012-14-12.flv-chunking-2593448364-246-244 201 65024.802 ms - 0
[crashes sometime between here...]
PUT /remote.php/webdav/2016-09-16%2012-14-12.flv-chunking-2593448364-246-245 201 3018.785 ms - 0
PROPFIND /remote.php/webdav/ 207 3.728 ms - -
PROPFIND /remote.php/webdav/ 207 25.119 ms - -
PROPFIND /remote.php/webdav/ 207 3.457 ms - -
PROPFIND /remote.php/webdav/ 207 21.342 ms - -
PROPFIND /remote.php/webdav/ 207 3.802 ms - -
GET /ocs/v2.php/apps/notifications/api/v1/notifications?format=json 200 3.139 ms - -
PROPFIND /remote.php/webdav/ 207 26.050 ms - -
[...and here, corrupting the below file]
GET /remote.php/webdav/2016-09-16%2012-14-12.flv 200 20.189 ms - 2451230160
```

kentonv:
@digitalcircuit Hmm, that sounds to me like a new, different bug. The app didn't shut down according to the log, so something else corrupted the file, presumably?

PR #2738, opened 2016-11-04
Support spaces in paths in static web publishing.

### LINKED ISSUE #2725: Web Publishing doesn't serve static files with spaces
Steps I took to reproduce:

* create a Davros grain, add a file with spaces in it.
* turn on web publishing (this currently just symlinks davros's `data` directory to `www`)
* go to the file, attempt to access its public URL.

It appears that it doesn't urldecode the URL paths before serving the files.

(zarvox comment from IRC: _just thinking out loud: it should probably be the supervisor that handles decoding the path before trying to open the file it serves_)

kentonv:
Fixes #2725.

### DIFF
--- a/shell/server/pre-meteor.js
+++ b/shell/server/pre-meteor.js
@@ -76,6 +76,9 @@ function wwwHandlerForGrain(grainId) {
     // Strip leading "/".
     if (path[0] === "/") path = path.slice(1);
 
+    // URI-decode the rest. Note that this allows filenames to contain spaces and question marks.
+    path = decodeURIComponent(path);
+
     let type = mime.lookup(path);
     const charset = mime.charsets.lookup(type);
     if (charset) {
--- a/shell/server/pre-meteor.js
+++ b/shell/server/pre-meteor.js
@@ -77,7 +77,8 @@ function wwwHandlerForGrain(grainId) {
     if (path[0] === "/") path = path.slice(1);
 
     // URI-decode the rest. Note that this allows filenames to contain spaces and question marks.
-    path = decodeURIComponent(path);
+    const encoded = path;
+    path = decodeURIComponent(encoded);
 
     let type = mime.lookup(path);
     const charset = mime.charsets.lookup(type);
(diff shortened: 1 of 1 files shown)


dwrensha:
This would be a breaking change with at least Hacker CMS, which automatically escapes spaces in filenames.

kentonv:
I guess that's true. I guess the alternative right now is to tell apps to name files using escaped names.

@mnutt Thoughts?

mnutt:
@dwrensha how do those files look on disk and how are they referred to in URLs? They have `%20`s in the filenames? I have files on disk with spaces, and am not able to refer to them at all with URLs.

dwrensha:
On disk, the file names have spaces escaped as "%20" and they can be refered to in URLs with a path like "/my%20file". After the change in this PR, such URLs would need to be double-escaped, like "/my%2520file".

I agree that #2725 is a bug, but I'm worried that some people might be depending on the old behavior.

Maybe we should add a parameter to `HackSessionContext.getPublicId()` to indicate which escaping style is expected by the app.

mnutt:
Hmm, agreed that we don't want to break existing behavior, and adding a new parameter probably makes sense. Another option could be to first look for the file with spaces, and if it doesn't exist then fall back to `%20`-escaped filenames.

kentonv:
OK, I added a commit that falls back to the URL-encoded path if the decoded path doesn't work. Thoughts?

shell/server/pre-meteor.js
```diff
@@ -127,6 +131,13 @@ function wwwHandlerForGrain(grainId) {
           });
           response.end("redirect: /" + path + "/");
         } else if (status === "notFound") {
+          if (path !== encoded) {
+            // Compatibility hack: Try geting the URL-encoded version of the path. We used to not
```
dwrensha: s/geting/setting/

kentonv: Actually "getting" is what I intended, as in we're "getting" the file from the grain.

Though I meant it with two t's...

mnutt:
Thanks, this is great!

PR #2742, opened 2016-11-06
don't serve web publishing requests from trashed grains

### LINKED ISSUE #2739: grains in trash can still serve public static content from /var/www
When a grain is in the trash, any content it has published should no longer be publicly available.

dwrensha:
Fixes #2739.

### DIFF
--- a/shell/server/pre-meteor.js
+++ b/shell/server/pre-meteor.js
@@ -43,6 +43,10 @@ const webPublishingHandlers = {};
 //   - Web publishing hosts
 //   - Standalone grain hosts
 
+deleteWebPublishingHandler = (publicId) => {
+  delete webPublishingHandlers[publicId];
+};
+
 const dnsCache = {};
 // Unfortunately, node's DNS library doesn't cache results, so we do our own caching.
 // Unfortunately, node's DNS library also dosen't give us TTLs. So, we'll cache for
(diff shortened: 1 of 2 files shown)


kentonv:
Not sure how I missed this when doing code reviews last week. Sorry!

But looking at the code, it seems it won't work correctly on Blackrock where there are multiple frontend replicas. Is it worth a Mongo subscription to watch for changes? Not sure.

zenhack:
Is this fix still applicable? There are some conflicts in pre-meteor.js, but I don't know if those are serious or just incidental.

Maybe @kentonv's comment about blackrock is no longer relevant, since Oasis is gone?

zenhack:
...or is alpha running blackrock?

ocdtrekkie:
Nope, Alpha is ordinary Sandstorm. Predates Blackrock entirely.

ocdtrekkie:
@zenhack If we want to implement this, we should also would need to fix it after #3231.

kentonv:
I suspect this PR would need to be completely rewritten due to the gateway having been moved to C++.

zenhack:
Ok, yeah looking a bit more careful much of the code it's touching doesn't really exist any more (even though it's not touching the gateway directly); we should probably close this.

PR #2751, opened 2016-11-11
Add quota display to the grainlist

### LINKED ISSUE #2744: Notify on 90% quota usage and show more quota information
Currently we only inform the user that they've reached their quota by blocking them from creating new grains after they hit it.

A customer requests that we improve this:

> There is no visual cue for the user that they are over quota (apart from the link when they try to open a new grain).  It would be great if there were an additional bit of information in the grain menu, e.g.:
>
> Grains I own: 177
> My grains' total size: 1.68GB
> My quota: 2GB
> My available space: 320MB
>
> and it would also be helpful if a notification was displayed in the notifications menu at around 90% of quota.

jparyani:
Fixes #2744 
image

### DIFF
--- a/shell/client/grain/grainlist-client.js
+++ b/shell/client/grain/grainlist-client.js
@@ -322,8 +322,12 @@ Template.sandstormGrainListPage.helpers({
     return prettySize(this._db.getUserQuota(Meteor.user()).storage);
   },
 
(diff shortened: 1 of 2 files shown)


jparyani:
Updated image:
image

PR #2752, opened 2016-11-12
Setup wizard base url changes

### LINKED ISSUE #2030: Sandstorm install: When base URL has changed, "Sign in to become admin" fails.
1. Asheesh changes base URL 
2. Begin setup using admin token URL
3. Click on "Sign in to become admin"
4. Click on "Configure login"
5. See "You are not logged in as an admin. To access the admin panel..."

Proposal: if at least 1 identity provider is not enabled the user should not see a "Sign in to become admin" button and instead be taken through the install flow.

@paulproteus may have more comments.

image
image
image

zarvox:
Changing BASE_URL can deconfigure Google and GitHub login, which can leave
the server in a situation where you can't log in, but you're directed to
sign in to create an admin account.  We should prompt you to reconfigure login.

Also, in the transition to the new admin UI, we stopped help text when the BASE_URL was changed.  This changeset adds a note and additional instructions when you go to reconfigure Google or GitHub login after a BASE_URL change.

Fixes #2030.

setup-wizard-start

setup-wizard-idp

google-instructions

github-instructions

### DIFF
--- a/shell/client/admin/identity-providers.html
+++ b/shell/client/admin/identity-providers.html
@@ -75,6 +75,16 @@
 {{#modalDialogWithBackdrop onDismiss=onDismiss}}
   <h2>Google login configuration</h2>
 
+  {{#if formerBaseUrl}}
+    <p class="flash-message warning-message" tabindex="-1">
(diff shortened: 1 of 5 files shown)


kentonv:
That warning message is awesome.

PR #2753, opened 2016-11-13
fixed the short text for the update console command

alandrees:
Simple text fix for the command line uninstall option.


garply:

dwrensha:
Thanks! This looks not quite right, though. Shouldn't you instead replace the *second* instance of "Generate admin token" ?

dwrensha:
For consistency with the other commands, the replacement text should probably be in the imperative mood, like "Uninstall Sandstorm."

alandrees:
@dwrensha ahhh yeah, good point.  I'll fix and create another PR.

Issue #1759, opened 2016-04-04
Grains fail to execute on fresh install on Debian 8.3, specifically with Scaleway default kernel


patrickod:
Setup: Debian 8.3 on Scaleways

```
root@scw-ca9206:~# uname -a
Linux scw-ca9206 4.4.4-std-3 #1 SMP Tue Mar 8 17:31:34 UTC 2016 x86_64 GNU/Linux
```

Symptoms: Sandstorm installs normally, and runs as expected allowing apps to be installed. However each time a grain is executed it fails, reporting an RPC error in the sandstorm interface like so:
```
Error: remote exception: expected result >= minBytes; Premature EOF
C++ location:(remote):??
type: failed
```

Inspecting the logs of the individual grains we see the following:
```
** SANDSTORM SUPERVISOR: Starting up grain.
*** Uncaught exception ***
sandstorm/supervisor.c++:1131: failed: seccomp_load(ctx): Invalid argument
stack: 0x5c12ee 0x5c071e 0x5e8fc9 0x5e8fba 0x4a84cf
```

Appears the issue is lack of support for seccomp on the default Scaleways installations.

paulproteus:
Hi @patrickod ,

Thanks for filing this bug!

I'm hopeful that we can use this bug to get the attention of Scaleway.com support staff to add seccomp support to their default kernel. So for now, let's keep the issue open here, on the Sandstorm issue tracker, until we can provide a decent experience for users of Scaleway.

As context for Scaleway people: Sandstorm is open source Linux server software that lets users make collaborative or personal documents using a Google Docs-like interface. The Sandstorm community, rather than core team, maintains the apps that users use to create those documents, so the Sandstorm core team focuses on building a consistent user interface to any app, and making sure it's safe for users to run those apps. (There are many other accurate summaries of Sandstorm, but perhaps that's the most relevant one for now.)

Sandstorm uses seccomp to decrease the Linux kernel's attack surface, so that one app can't attack another app. Right now, since Scaleway servers disable seccomp, Sandstorm crashes before it can launch an app.

Over the past two years of me watching people set up Sandstorm, Scaleway is the only service provider I have seen that actively disables seccomp by default.

At this point, the Sandstorm team has a few options:

- Accept that Scaleway will keep seccomp disabled, and print a big sad failure message early-on during the install process to tell people Scaleway's kernels don't support Sandstorm, but if they do the complicated step of switching to a different scaleway.com kernel, then maybe it will work.

[...]

PR #2756, opened 2016-11-15
Added assert for seccomp in install.sh

RedRudeBoy:
I've made the changes directly in github, please test it before merge.

Detect if seccomp is available:
https://github.com/sandstorm-io/sandstorm/issues/1759
Code from:
http://outflux.net/teach-seccomp/autodetect.html

Is "Seccomp filtering" also necessary?

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -481,6 +481,20 @@ assert_usable_kernel() {
   fi
 }
 
+assert_kernel_has_seccomp() {
+  ret = prctl(PR_GET_SECCOMP, 0, 0, 0, 0);
+  if (ret < 0) {
+    switch (errno) {
+      case ENOSYS:
+        fail "seccomp not available: pre-2.6.23"
+      case EINVAL:
+        fail "seccomp not available: not built in"
+      default:
+        fail "seccomp not available: unknown code"
+    }
+  }
+}
+
 maybe_enable_userns_sysctl() {
   # This function enables the Debian/Ubuntu-specific unprivileged
   # userns sysctl, if the system has it and we want it.
@@ -2150,6 +2164,7 @@ set_umask
 assert_on_terminal
 assert_linux_x86_64
 assert_usable_kernel
+assert_kernel_has_seccomp
 detect_current_uid
 assert_dependencies
 assert_valid_bundle_file


garply:

install.sh
```diff
@@ -481,6 +481,20 @@ assert_usable_kernel() {
   fi
 }
 
+assert_kernel_has_seccomp() {
+  ret = prctl(PR_GET_SECCOMP, 0, 0, 0, 0);
```
paulproteus: Hi @RedRudeBoy ,

Thanks for submitting this change!

I agree that it can be useful to test for seccomp existing.

However, I believe this change that you are suggested would introduce invalid bash syntax. It looks like you're writing C, but since `bash` is executing this script, I believe it would error-out with a syntax error.

Let me know what you think. Thank you again for submitting it!

I'll wait for your reply before testing the change.

paulproteus:
Hi again @RedRudeBoy ,

I just saw that you mentioned this pull request in a support ticket. Thank you for being willing to file it.

I think that we'd need to compile it and embed a binary in the install script. This might be worth doing - we've done it before when testing for user namespaces.

I think that a bigger question here is - is Sandstorm OK running on systems that don't have seccomp available? If so, one fix might be to allow Sandstorm to downgrade to not using seccomp for sandboxing. I'll let @kentonv comment on that.

Either way, it is probably smart for the Sandstorm codebase to check if seccomp is present. It could log an error and fail to start up in that case, or it could proceed with some sandboxing removed. Can I request you to file an issue asking for that, @RedRudeBoy ? I think that putting the change into Sandstorm itself is going to be more maintainable than adding something to the install script and it'll have a similar positive effect.

Once that issue is filed, it probably makes sense to close this PR since I think the code you proposed isn't valid bash.

Let me know what you think. Thanks again for bringing up the issue! Having Sandstorm work reliably in a variety of Linux systems is a goal of ours, and you've opened our eyes to the possibility that seccomp might not be available.

kentonv:
> is Sandstorm OK running on systems that don't have seccomp available?

I think the answer to this should be "no".

In theory, today, the effect of skipping the seccomp filter would be that it would be relatively easy for an application to break out of its sandbox (any time a new kernel vulnerability occurs, which is often). This may be OK if you believe that all users of the Sandstorm servers are trustworthy and know better than to install potentially-malicious apps.

However, in the future, we expect to implement sandbox features using seccomp. Specifically we would tell seccomp to raise SIGSYS, and then via LD_PRELOAD we should load a library into the app which handles the signal by emulating the syscall. This emulation may intentionally do things like remap paths or invoke RPCs, which could mean that without a seccomp filter installed, the app would behave differently (and incorrectly).

It's true that we don't have any immediate plans to do this sort of thing, but it would be sad if we bind our hands such that we can't make such chanages in the future.

Also, I worry that allowing users to turn off seccomp will take the pressure off of hosting providers to fix this. There is no reason I know of why a hosting provider should disable seccomp -- this seems to be an oversight on their part and the right answer is to get them to fix it.

=======================

We could probe for seccomp using a new tiny rust binary embedded in the install script, like we used to do for userns. However, another option might be to wait until Sandstorm has started up, and then present an error message in the setup wizard UI. Seccomp is only strictly required before starting an app, so we don't necessarily have to block install, unless we think there's a significant advantage to knowing about the problem a couple minutes earlier.

FWIW, we actually need not just seccomp but SECCOMP_MODE_FILTER.

RedRudeBoy:
Ups, you're right, the code must be compiled in binary. I put the code in install.sh for avoid the installation of Sandstorm if the kernel is not supported.

Closing the pull request and asking (again) to my service provider to fix this.

Thank you!

paulproteus:
Thanks @RedRudeBoy ; I've taken the liberty of filing #2757 in which we will presumably follow Kenton's suggested path of showing an error message in the web-based setup wizard.

Issue #2633, opened 2016-10-07
Notifications from shared-with-me grains with missing icons don't display correctly


kentonv:
When testing @dwrensha Wekan changes, I observed that the bell menu was showing up empty even when I definitely had Notifications waiting. The console showed the stack trace below. I suspect this has to do with the fact that this Wekan build is missing icons.

Note that the problem only occurs for users who do not own the grain. The grain owner sees notifications rendered correctly with an identicon.

```
Exception in template helper: TypeError: Cannot read property 'assetId2xDpi' of null
at e (https://oasis.sandstorm.io/ad7921a02c5a97967428a0c554f60da892aad1d5.js?meteor_js_resource=true:146:2999)
at Function.g.iconSrcForDenormalizedGrainMetadata (https://oasis.sandstorm.io/ad7921a02c5a97967428a0c554f60da892aad1d5.js?meteor_js_resource=true:146:3416)
at https://oasis.sandstorm.io/ad7921a02c5a97967428a0c554f60da892aad1d5.js?meteor_js_resource=true:196:13560
at https://oasis.sandstorm.io/ad7921a02c5a97967428a0c554f60da892aad1d5.js?meteor_js_resource=true:42:1836
at https://oasis.sandstorm.io/ad7921a02c5a97967428a0c554f60da892aad1d5.js?meteor_js_resource=true:42:1659
at Array.forEach (native)
at Function.A.each.A.forEach (https://oasis.sandstorm.io/ad7921a02c5a97967428a0c554f60da892aad1d5.js?meteor_js_resource=true:1:865)
at h.Cursor.forEach (https://oasis.sandstorm.io/ad7921a02c5a97967428a0c554f60da892aad1d5.js?meteor_js_resource=true:42:1581)
at h.Cursor.map (https://oasis.sandstorm.io/ad7921a02c5a97967428a0c554f60da892aad1d5.js?meteor_js_resource=true:42:1805)
at e (https://oasis.sandstorm.io/ad7921a02c5a97967428a0c554f60da892aad1d5.js?meteor_js_resource=true:196:12601)
```

zarvox:
Looks like `iconObj` in [iconSrcFor](https://github.com/sandstorm-io/sandstorm/blob/8d5092309068eb892bc0233f5a1b22ed70bcc07b/shell/packages/sandstorm-identicons/helpers.js#L75) was `null` rather than `undefined`.

[...]

PR #2758, opened 2016-11-16
Fix exception when opening notifications menu which causes notifications not to show.

kentonv:


dwrensha:
This fixes #2633, as far as I can tell.

shell/packages/sandstorm-identicons/helpers.js
```diff
@@ -65,7 +65,7 @@ const bytesToBase64 = function (bytes) {
 };
 
 const iconSrcFor = function (appId, iconObj, staticPrefix, usage) {
-  if (iconObj === undefined) {
+  if (iconObj === undefined || iconObj === null) {
```
dwrensha: My stylistic choice would be `if (!iconObj)`, but this seems fine too.

dwrensha:
👍

PR #2764, opened 2016-11-18
don't bother setting grainInfo.icon if there is no icon

dwrensha:
Currently, the object returned here can have `grainInfo.icon = undefined`. When this gets written to Mongo, that `undefined` (somewhat surprisingly) morphs into a `null`, which can lead to bugs like #2633.

It's cleaner to not the `icon` field at all if there is nothing to put there.


kentonv:
IIRC Mongo's null/undefined behavior is covered in this talk: https://www.youtube.com/watch?v=amaR4Aqe0s0

(It's a pretty fun talk. Recommended.)

PR #2767, opened 2016-11-19
Add experiment that gives a grain limit of 100 to 10% of new free, including demo users.

kentonv:
We want to see if this makes users more likely to "get hooked" and become paying users or refer their friends.

### DIFF
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -133,6 +133,8 @@ const collectionOptions = { defineMutationMethods: Meteor.isClient };
 //               (possibly "free") before they can create grains (but not when opening someone
 //               else's shared grain). The goal of the experiment is to determine whether this
 //               prompt scares users away -- and also whether it increases paid signups.
+//       freeGrainLimit: Value is "control" or or a number indicating the grain limit that the
+//               user should receive when on the "free" plan, e.g. "Infinity".
 //   stashedOldUser: A complete copy of this user from before the accounts/identities migration.
 //                   TODO(cleanup): Delete this field once we're sure it's safe to do so.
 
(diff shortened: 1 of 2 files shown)


kentonv:
Changed it to 100 instead. Making it "unlimited" could lead to confusion around the referral program also offering "unlimited" grains.

paulproteus:
Thanks for noticing this, btw! I'm very +1 to this experiment.

zarvox:
LGTM

oresmus:
I caution you about interpreting this experiment as if it will give insight to what would happen if you adopted a global change to the free grain limit from 5 to 100. A user with the limit of 100 (assuming you make it clear to such users that most new free users will still have a limit of 5) will still likely use the limit of 5 in deciding whether to recommend the service to a friend or coworker. If they get used to having 100 grains, it might even make them *less* likely to recommend it than if they get used to it themselves with the max of 5 grains, since they might think making use of it with only 5 grains is almost impossible (having not learned to adapt to this themselves).

I think a more valuable experiment (requiring no code change unless your current logs aren't good enough) would just be to try to track user referral behavior vs. grain usage, and determine how many grains were typically in use by users who decided to try to get friends or coworkers to use the service. You can assume that they found the service "usable" with that number of grains in use.

oresmus:
Another important point is to make it clear to new users who receive this higher limit whether it's permanent for them -- will they get this free higher service level forever? What they guess about that may affect their behavior which you're studying, and it may also be morally and/or legally important to make this clear to them.

And another one is that, if they are aware of this being an experiment and of why you're doing it, this awareness may also affect their behavior in some way.

kentonv:
@oresmus Most users who end up with 100 grains will probably assume that this is what everyone else gets too.

Keep in mind that anyone can get unlimited grains by referring someone else, per: https://oasis.sandstorm.io/referrals

kentonv:
Results:

https://oasis.sandstorm.io/shared/adcB3wqewVam8TuOWwf5m3dlAvlTQAudIO4WRE6nsV1

Summary:

* Not enough data to say if this reduced paid signups.
* Free users with an 100-grain limit use 27% more storage than those with a 5-grain limit.
* Users with a 5-grain limit are 50% more likely to refer / share with others than are those with a 100-grain limit (although only a couple percent of users use the referral program either way).

Conclusion:

Instead of increasing the grain limit, perhaps we should emphasize the referral program. Once you've referred someone, the grain limit goes away. Currently, though, the referral program is somewhat obscure and is not directly presented as an option when the grain limit is hit. Nevertheless, the experiment results seem to clearly show that the grain limit drives referrals for some users. If it were more obvious, we might get a more viral effect.

PR #2774, opened 2016-11-26
Add `tls` parameter to `IpRemoteHost.getTcpPort()`

dwrensha:
The implementation is mostly a copy/paste of the `TcpPort.connect()` method.

Using this, I have successfully sent an SMS message through https://api.twilio.com from a Sandstorm grain.

This should allow an outgoing HTTP/HTTPS driver to be implemented entirely as an app, without the need to include a certificate bundle in the SPK.

### DIFF
--- a/shell/server/drivers/ip.js
+++ b/shell/server/drivers/ip.js
@@ -18,6 +18,7 @@ import Bignum from "bignum";
 import { PersistentImpl } from "/imports/server/persistent.js";
 const Future = Npm.require("fibers/future");
 const Net = Npm.require("net");
+const Tls = Npm.require("tls");
 const Dgram = Npm.require("dgram");
 const Capnp = Npm.require("capnp");
 
@@ -341,6 +342,35 @@ TcpPortImpl = class TcpPortImpl {
       });
     });
   }
+
+  connectTls(downstream) {
+    const _this = this;
+    let resolved = false;
+    return new Promise((resolve, reject) => {
+      const client = Tls.connect({ host: _this.address, port: _this.port }, () => {
+        resolved = true;
+        resolve({ upstream: new ByteStreamConnection(client) });
+      });
+
+      client.on("data", (data) => {
+        downstream.write(data);
+      });
+
+      client.on("close", (hadError) => {
+        downstream.done();
+      });
+
+      client.on("error", (err) => {
+        if (resolved) {
+          client.write = errorWrite;
+        } else {
+          // upstream hasn't been resolved yet, so it's safe to reject
+          reject(err);
+        }
+      });
+    });
+  }
+
 };
 
 const errorWrite = (data) => {
--- a/shell/server/drivers/ip.js
+++ b/shell/server/drivers/ip.js
@@ -300,8 +300,8 @@ IpRemoteHostImpl = class IpRemoteHostImpl {
     }
   }
 
-  getTcpPort(portNum) {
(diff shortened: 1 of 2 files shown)


kentonv:
I'd argue for pushing this up even further: Decide when creating the TcpPort -- or, indeed, the IpNetwork -- whether it is for TLS or cleartext communications. No interface changes needed; only a change in powerbox request parameters.

This approach separates semantics from logistics. The code which initiates the connection and speaks the application-layer protocol does not need to know whether the underlying transport is TLS-encrypted. Higher-level dependency-injecting code (or in some cases the user / admin) can more easily make this decision.

The HTTP/HTTPS driver would actually request two IpNetwork implementations upfront: The cleartext one and the TLS-backed one.

FWIW this also matches the KJ TLS interface proposal, which provides a new implementation of kj::Network that forms strictly TLS connections rather than plain ones. https://github.com/sandstorm-io/capnproto/pull/335/files#diff-204753c7aa052f9f092166751e25d89bR113

dwrensha:
Updated so that the decision about TLS is made when the `TcpPort` is created.

I'm not convinced that it would be good to hoist this the whole way up to the `IpNetwork` level, as @kentonv suggested. I think Sandstorm admins would be annoyed that they need to set up capabilities which seem to be mostly redundant with each other.

@kentonv: can you provide a more concrete example of the kind of dependency-injection that you'd like to enable?

kentonv:
If requesting two caps would be annoying, we can solve that with a "bundle" interface like:

    interface IpNetworkPlusTls {
      getPlain @0 () -> (network :IpNetwork);
      getTls @1 () -> (network :IpNetwork);
    }

Again, this helps separate semantics from logistics. Imagine a distributed protocol involving many nodes talking to each other. The implementation of this protocol would probably request an IpNetwork capability to work with rather than an IpRemoteHost, much less a TcpPort.

Now imagine two different contexts in which you might want to use this protocol: On a private network, where encryption is irrelevant, and on the public internet, where it is essential.

With your approach, to support both of these cases, the protocol implementation component would need to accept a boolean to its constructor telling it whether to request TLS on each connection it opens. This is at best ugly. At worst, the implementation doesn't bother providing the boolean and instead hard-codes one value or the other. Now to get it to do what you want, you have to write a custom IpNetwork implementation that intercepts the getTcpPort() method and toggles the `tls` parameter.

If TLS is instead normally provided by an alternative IpNetwork capability, then these potential problems go away.

Other considerations:
* TLS fundamentally requires additional configuration. 95% of use cases may be covered by shipping a standard set of trusted certificates, but in some use cases the user will want to provide their own certificate to trust -- possibly in the context of a specific requesting grain rather than globally. So we should provide that configurability, but on the other hand, it's weird to present these options if the app isn't actually going to use TLS. So it's nice to know at request time.
* It's potentially comforting to the user to know at request time that the app will strictly used TLS, because it is requesting only the TLS capability.
* What if someone wants to use an encrypted transport *other than* TLS? This seems likely easier to support if TLS is not given special status.

I would more generally argue that all of the standard arguments against singletons apply to a lesser extent to this sort of situation. In some sense, having a "bool tls" parameter when creating a port makes TLS support into a singleton "relative to" the IpNetwork. http://www.object-oriented-security.org/lets-argue/singletons

dwrensha:
Thanks for the elaboration! Closing in favor of #2776.

PR #2776, opened 2016-11-27
Add option for IpNetwork to be layered on top of TLS.

dwrensha:
This is an implementation of @kentonv's [suggestion](https://github.com/sandstorm-io/sandstorm/pull/2774#issuecomment-263086336) from #2774.

This changes the database representation of `IpNetwork` sturdyrefs and includes a migration to convert existing sturdyrefs of the form `ipNetwork: true` into `ipNetwork: {encryption: {none: null}}`.

Adds a "TLS-encrypted outbound network access" table to the "Network capabilities" section of the admin panel. Also in the "Network capabilities" section, fixes some bugs, most of which were introduced when we switched to using the `claimRequest()` flow. `cap.owner.grain.introducerIdentity` is now obsolete!

### DIFF
--- a/shell/server/drivers/ip.js
+++ b/shell/server/drivers/ip.js
@@ -246,7 +246,7 @@ class IpNetworkImpl extends PersistentImpl {
     // However, we keep that code path active because there is in principle no reason why it
     // should always fail, and we wish to allow for a possible future in which we let users
     // specify custom certificate authorities, in which case it might be more likely for TLS
-    // to be be expected to authenticate raw IP addresses.
+    // to be expected to authenticate raw IP addresses.
 
     return { host: new IpRemoteHostImpl(addressToString(address), this.tls) };
   }
--- a/shell/server/drivers/ip.js
+++ b/shell/server/drivers/ip.js
@@ -18,6 +18,7 @@ import Bignum from "bignum";
 import { PersistentImpl } from "/imports/server/persistent.js";
 const Future = Npm.require("fibers/future");
 const Net = Npm.require("net");
+const Tls = Npm.require("tls");
 const Dgram = Npm.require("dgram");
 const Capnp = Npm.require("capnp");
 
@@ -234,16 +235,17 @@ const addressType = (address) => {
 };
 
 class IpNetworkImpl extends PersistentImpl {
-  constructor(db, saveTemplate) {
+  constructor(db, saveTemplate, tls) {
     super(db, saveTemplate);
+    this.tls = tls;
   }
 
   getRemoteHost(address) {
-    return { host: new IpRemoteHostImpl(address) };
+    return { host: new IpRemoteHostImpl(address, this.tls) };
   }
 
   getRemoteHostByName(address) {
-    return { host: new IpRemoteHostImpl(address) };
+    return { host: new IpRemoteHostImpl(address, this.tls) };
   }
 };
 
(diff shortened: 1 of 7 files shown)


src/sandstorm/ip.capnp
```diff
@@ -50,6 +50,21 @@ interface IpNetwork @0xa982576b7a2a2040 {
   # access". The IpNetwork capability is a dangerous capability that should only be granted to
   # trusted drivers. Only the Sandstorm server administrator is likely to possess this capability.
 
+  struct PowerboxTag {
+    # Tag to be used in a `PowerboxDescriptor` to describe an `IpNetwork`.
+
+    encryption :union {
```
dwrensha: Good point!

I've now boxed the union into an `Encyrption` struct, but I've kept it as a field of `IpNetwork.PowerboxTag` rather than adding it as an independent tag, mostly because doing so fits better with our current `FrontendRefRegistry` code. I've added a note explaining that other interfaces may use `Encryption` as an independent tag. Hopefully, most such capabilities can be implemented in userland and hence will not need to have a `frontendRef` representation.

shell/server/drivers/ip.js
```diff
@@ -234,16 +235,17 @@ const addressType = (address) => {
...
 class IpNetworkImpl extends PersistentImpl {
-  constructor(db, saveTemplate) {
+  constructor(db, saveTemplate, tls) {
     super(db, saveTemplate);
+    this.tls = tls;
   }
 
   getRemoteHost(address) {
-    return { host: new IpRemoteHostImpl(address) };
+    return { host: new IpRemoteHostImpl(address, this.tls) };
```
kentonv: At first I thought "TLS authenticates hostnames, therefore the by-address method should probably throw an exception if using TLS", but apparently TLS can in theory authenticate IP addresses, it's just that none of the trusted CAs will ever sign such a certificate.

Since at present we don't provide a way to specify a custom CA, this path will never work for TLS, but it makes sense to keep here for later when we do provide custom CAs.

Maybe worth adding a comment?

shell/server/drivers/ip.js
```diff
@@ -290,35 +307,49 @@ Meteor.startup(() => {
...
   }
 
   getTcpPort(portNum) {
-    return { port: new TcpPortImpl(this.address, portNum) };
+    return { port: new TcpPortImpl(this.address, portNum, this.tls) };
   }
 
   getUdpPort(portNum) {
+    if (this.tls) {
+      const error = new Error("UDP over TLS is not yet supported");
```
kentonv: Technically this would be "TLS over UDP", or perhaps more precisely "DTLS over UDP", rather than "UDP over TLS".

shell/server/drivers/ip.js
dwrensha: added

kentonv:
I can't remember if I had a standing objection to this, or if it just dropped off the radar. Do you remember?

dwrensha:
As I remember it, you convinced me that outgoing HTTPS should use a separate pseudo-driver, and that made this change seem less urgent.

I think the TLS functionality added here is still probably a good idea, but it might be good to verify that this design works for various protocols that may need it, e.g. IRC.

This pull request also fixes some glaring bugs in the admin UI. I suppose in retrospect that perhaps those fixes ought not to have been bundled with the more controversial changes.

kentonv:
Eh I'm going to merge it.

PR #2780, opened 2016-11-28
Fix up the logic for garbage collecting autoupdate apps

### LINKED ISSUE #2772: Exception while invoking method 'updateApps' Error: No such package on server
Today on my Sandstorm I got offered an update to Rocket.Chat version 0.46.0. However, clicking "apply updates" generated a JS alert that said "500 Internal Server Error."

Checking my `sandstorm.log`, I see the following:

```
Exception while invoking method 'updateApps' Error: No such package on server: 6fc85797bb55a4f8f140a18047b46759
    at packages/sandstorm-autoupdate-apps/autoupdate-apps.js:47:15
    at Array.forEach (native)
    at [object Object].updateApps (packages/sandstorm-autoupdate-apps/autoupdate-apps.js:44:14)
    at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1711:12)
    at packages/ddp-server/livedata_server.js:711:19
    at packages/ddp-server/livedata_server.js:709:40
    at packages/ddp-server/livedata_server.js:707:46
    at Session.method (packages/ddp-server/livedata_server.js:681:23)
```

Maybe the update got pulled or obsoleted, and my local Sandstorm isn't in sync? In any case, seems like there are two bugs to solve:

1. Improving usability when errors like this _do_ occur - the JS alert could be more specific about what actually happened
2. Making sure that this particular error (no such package on server) doesn't happen again

jparyani:
Before this change, if a user on the server manually installed the package before the autoupdate downloaded the package, and then the user deleted all their references to the package, the package would be deleted.

I don't think this actually fixes #2772, but it could be what @zarvox ran into.

### DIFF
--- a/shell/server/installer.js
+++ b/shell/server/installer.js
@@ -57,8 +57,8 @@ const deletePackageInternal = (pkg) => {
     const action = UserActions.findOne({ packageId: packageId });
     const grain = Grains.findOne({ packageId: packageId });
     const notificationQuery = {};
-    notificationQuery["appUpdates." + pkg.appId] = { $exists: true };
-    if (!grain && !action && !(pkg.isAutoUpdated && Notifications.findOne(notificationQuery))
+    notificationQuery["appUpdates." + pkg.appId + ".packageId"] = packageId;
+    if (!grain && !action && !Notifications.findOne(notificationQuery)
         && !globalDb.getAppIdForPreinstalledPackage(packageId)) {
       Packages.update({
         _id: packageId,


zarvox:
Agreed that we should always consider references to packages in app update notifications live, regardless of whether the package is marked `isAutoUpdated` or not.

Patch LGTM.

kentonv:
Is there any way we can index this so that this query doesn't do a full table scan over notifications?

jparyani:
I don't think we can, at least with how the `appUpdates` currently works. Its keys are the appIds, and so we'd want to create an index of the form `appUpdates.$.packageId`, but I don't think that's possible to do.

In retrospect, the design of the appUpdates object was a mistake, and it should probably be an array of documents, which would be easy to index/query.

PR #2781, opened 2016-11-28
prepend '/' to paths in ExternalWebSessionImpl

dwrensha:
The `path` parameter to `WebSession`'s methods should never include the leading `'/'`. (See [this](https://github.com/sandstorm-io/sandstorm/blob/v0.196/src/sandstorm/web-session.capnp#L98-L99) and [this](https://github.com/sandstorm-io/sandstorm/blob/v0.196/shell/server/proxy.js#L1863).)

In contrast, the `path` argument in a call to node's [`http.request()`](https://nodejs.org/api/http.html#http_http_request_options_callback) should have a leading `'/'`. So  this pull request adds it.

I noticed this problem when I tried making a hacky HTTP pseudodriver by copying this code.

### DIFF
--- a/shell/server/drivers/external-ui-view.js
+++ b/shell/server/drivers/external-ui-view.js
@@ -125,7 +125,15 @@ ExternalWebSession = class ExternalWebSession {
     return new Promise((resolve, reject) => {
       const options = _.clone(session.options);
       options.headers = options.headers || {};
-      options.path = "/" + path;
+
+      // According to the specification of `WebSession`, `path` should not contain a
+      // leading slash, and therefore we need to prepend "/". However, for a long time
+      // this implementation did not in fact prepend a "/". Since some apps might rely on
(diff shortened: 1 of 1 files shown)


jparyani:
Unfortunately, there are now apps that rely on the formerly buggy behavior. You'll have to check and handle both cases I think.

dwrensha:
what apps use this?

jparyani:
Specifically, our internal dashboard does. Beyond that, I'm not sure, but it is a public API that anyone could have used about a year+.

That being said, I did fix a major breaking issue in it a couple months back, and no one had complained about it not working, so maybe we are the only consumer.

kentonv:
Should we temporarily add the '/' only if it doesn't exist, and then update the dashboard app, and then change the logic to always add it?

dwrensha:
OK, I've added the temporary compatibility hack.

PR #2788, opened 2016-11-29
Fix potential issue with login email under IE

jparyani:
We'll ignore the port when checking rootUrl because it's mostly meaningless in this context, and because IE will sometimes include it, as opposed to most browsers omitting it for 80/443.


paulproteus:
>  IE will sometimes include it, as opposed to most browsers omitting it for 80/443

TIL

kentonv:
Hmm, could this also be solved by changing the client side to use .hostname rather than .host?

PR #2793, opened 2016-12-03
Redesign demo/appdemo to show up as a modal

### LINKED ISSUE #2657: Demo: Landing page redesign
Changes proposed:
1. Provide a visual hint/preview of the demo so that users know what they're getting into. Some users voiced that they were hesitant to click on "Try a quick demo" since they had no idea what they would be doing next.
2. Added string "with all the apps you need, all in one place." to be consistent with our website. It also hints that Sandstorm is a work space where users can use many apps
3. Change string from "Try a quick demo" -> "Start" and also using arrows instead of the play icon. Play icon gives the impression there is some sort of video so it should be removed. "Try a quick demo" hints at non-permanence; users can actually create things and carry over their data to an actual account so instead I chose to use the string "Start". Also, the band is now styled it as a button to make it more obvious that it should be clicked.
4. Styling improvements

Proposed design (draft 1):

oasis-demo1 1

Mobile (draft 1):

oasis-demo-landing-mobile1

Current design:

oasis-demo-old1

jparyani:
Fixes #2657

A slight change worth noting, we now no longer show the guided tour for appdemo users (see setting `userNeedsShareAccessHint`). This is because the double stacked modals felt very weird, and after discussing with @paulproteus, we decided to push this to happen after account creation (which is actually broken at the moment and has been for a while).

image

image


kentonv:
Seems this broke some demo-related tests?

kentonv:
It looks like everything we changed this week other than this affects only self-hosters, so I'll simply not push an Oasis updated today and will instead push a hotfix whenever we get these tests updated.

shell/client/demo-client.js
```diff
@@ -103,21 +88,17 @@ Router.map(function () {
...
 
       // TODO(cleanup): Is there a better way to do this? Before multi-identity, we could use the
       //   `userCallback` option of `Accounts.callLoginMethod()`, but now that doesn't work because
       //   it fires too early.
-      //
-      // Note that we don't use Template.instance().autorun() because the template gets destroyed
-      // and recreated during account creation.
       let done = false;
-      const handle = Tracker.autorun(function () {
+      handle = Tracker.autorun(function () {
```
kentonv: What happens if `handle` is already assigned? It seems like you leak the old handle...?

jparyani: It turns out that `handle` being non-local wasn't actually necessary after refactoring this route to use `onRun`.

dwrensha:
The introjs overlay shows up for me now when I visit /demo in an incognito window in Chrome on Fedora:

demo-overlay

This seems wrong.

jparyani:
It turns out you only see that in the case where you have no preinstalled apps (and my local server does, so I never noticed).

Fixed in #2803

kentonv:
Following a second appdemo link after starting a demo seems to log you out of the old demo, which means you lose all your data. Looking at how to fix...

kentonv:
Really weird timing issues he. By doing `console.log(Meteor.loggingIn())` at the top of the `onRun()`, I somehow cause the whole thing not to load. No error.

kentonv:
Fixed a bunch of bugs: #2805

PR #2805, opened 2016-12-17
Fix multiple bugs in new demo splash

kentonv:


kentonv:
Merging for release but review would still be appreciated.

PR #2806, opened 2016-12-23
Implement app-to-app powerbox flow

kentonv:
Currently only raw-API apps can be responders. Future work will extend sandstorm-http-bridge to allow HTTP apps to be responders.

~~To do still in this PR: Add tests. I have a test program which needs some cleaning up, then I'll need to write the test case.~~ TESTS HAVE BEEN ADDED


shell/server/proxy.js
```diff
@@ -230,6 +231,12 @@ Meteor.setInterval(() => {
 
 const proxiesByHostId = {};
 
+function parsePowerboxDescirptorList(list) {
```
zarvox: typo: `parsePowerboxDescriptorList`

ditto below where you have autocompleted the same spelling

shell/server/proxy.js
```diff
@@ -321,17 +336,30 @@ Meteor.methods({
       throw new Meteor.Error(403, "Unauthorized", "User is not authorized to open this grain.");
     }
 
+    const sessionFields = {};
+    if (options.powerboxRequest) {
+      sessionFields.powerboxRequest = {
+        descriptors: parsePowerboxDescirptorList(options.powerboxRequest.descriptors),
```
zarvox: `parsePowerboxDescriptorList` here

shell/server/proxy.js
```diff
@@ -425,6 +460,14 @@ Meteor.methods({
                                "User is not authorized to open this grain.");
       }
 
+      const sessionFields = {};
+      if (options.powerboxRequest) {
+        sessionFields.powerboxRequest = {
+          descriptors: parsePowerboxDescirptorList(options.powerboxRequest.descriptors),
```
zarvox: `parsePowerboxDescriptorList` here

kentonv:
I went ahead and did what I've always wanted: Add an official test app directly to the Sandstorm repo, which is built along with everything else and passed directly into the test suite. No need to host the spk elsewhere.

For now it is only used in the new powerbox tests. Someday, perhaps, we'll merge all the other test apps into this one.

kentonv:
Garply, retest this please

kentonv:
(FYI: I just upgraded the build server to node 4.)

kentonv:
OK, I don't understand the latest test failure. It works locally. It looks like on the CI, the powerbox popup doesn't appear when it should.

kentonv:
Is there some way to show the browser console log on test failure?

zarvox:
I'm not aware of any convenient way to capture the browser's console log when CI observes a test failure at present.

I suppose it's theoretically possible for us to use [`.execute()`](http://nightwatchjs.org/api#execute) to monkey-patch `window.console` at the beginning of the testcase to append messages to a global array that you can extract with a second `.execute()` block later so you can dump it to the server log.  Or maybe those could live in a [beforeEach](http://nightwatchjs.org/guide#using-before-each-and-after-each-hooks) and afterEach block?

kentonv:
Oh, I think I know the problem here. I bet that the Iceweasel on the CI doesn't support ES6, and the test app has some ES6 syntax in it. I guess I can change it to use strictly ES5 but I wonder how hard it would be to update the browser used in CI... @jparyani ?

jparyani:
Ya, we should definitely update iceweasel. The current version is a year+ out of date.

kentonv:
YAY TESTS PASS!

kentonv:
@dwrensha @zarvox Might either of you be able to look at this in the next day or two? If not I will merge it now, but if you're available and can review (in any amount of depth) that would be awesome.

dwrensha:
I'll take a look tomorrow morning.

src/sandstorm/test-app/test-app.c++
```diff
@@ -0,0 +1,219 @@
...
+    context.getResults().setSession(
+        kj::heap<WebSessionImpl>(params.getUserInfo(), params.getContext(),
+                                 params.getSessionParams().getAs<sandstorm::WebSession::Params>()));
+
+    return kj::READY_NOW;
+  }
+
+  kj::Promise<void> newRequestSession(NewRequestSessionContext context) override {
+    auto params = context.getParams();
+
```
dwrensha: trailing whitespace

kentonv:
All comments addressed.

src/sandstorm/test-app/test-app.capnp
```diff
@@ -0,0 +1,82 @@
+@0xdb849475182f3bb4;
+
+#$import "/capnp/c++.capnp".namespace("sandstorm::testapp");
+
+using Powerbox = import "/sandstorm/powerbox.capnp";
+using Grain = import "/sandstorm/grain.capnp";
+using Spk = import "/sandstorm/package.capnp";
+
+const testAppHtml :Data = embed "test-app.html";
```
kentonv: I love embedding resources directly into the binary! It's so much easier to deal with and there's really no down side.

PR #2807, opened 2016-12-28
Rotate logs -- both sandstorm.log and grain logs.

### LINKED ISSUE #2636: Consider splitting grain logs from backups
@zenhack on IRC noted that his irc-idler grain is currently running with debug logging, and as a result, he has a grain with 300MB of grain log and maybe 12k of actual sandbox content.  Downloading a grain backup, sadly, results in fetching all the logs to have a chance to look at the grain storage.

There is also no user-facing way to truncate the grain log other than deleting the whole grain.

A proposal:
- Make it possible to download the grain log, perhaps from the window that tails the debug log
- Then, don't include the grain log in the backup.  For the purposes of backing up and restoring grains, the log is not generally useful.  For the purposes of debugging, being able to view or download the whole debug log inline is probably better for users. 
- Possibly allow grain owners to truncate the log.

Another approach would be to make the logging code smarter and retain logs for a shorter window of time, rather than indefinitely.  This is a larger policy decision, though, and since we don't currently split the logs nor annotate them with any time information, this would probably be a bit more engineering work.

kentonv:
Have been wanting to do this a long time but continuously procrastinated...

As written this will save 512k-1MB of grain logs at a time and 1MB-2MB of main log. These numbers could well be adjusted if desired.

Fixes #2636.

### DIFF
--- a/shell/server/admin-server.js
+++ b/shell/server/admin-server.js
@@ -576,16 +576,42 @@ Meteor.publish("adminLog", function (token) {
   const fd = Fs.openSync(logfile, "r");
   const startSize = Fs.fstatSync(fd).size;
 
+  // Difference between the current file offset and the subscription offset. Can be non-zero when
+  // logs have rotated.
+  let extraOffset = 0;
+
+  if (startSize < 8192) {
+    // Log size is less than window size. Check for rotated log and grab its tail.
+    const logfile1 = SANDSTORM_LOGDIR + "/sandstorm.log.1";
+    if (Fs.existsSync(logfile1)) {
+      const fd1 = Fs.openSync(logfile1, "r");
+      const startSize1 = Fs.fstatSync(fd1).size;
+      const amountFromLog1 = Math.min(startSize1, 8192 - startSize);
+      const offset1 = startSize1 - amountFromLog1;
+      const buf = new Buffer(amountFromLog1);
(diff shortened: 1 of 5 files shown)


ocdtrekkie:
Is it safe to assume after this releases, we'll be able to truncate our grain logs by simply opening the grain?

kentonv:
@ocdtrekkie Not quite. As written, if your current grain log is larger than 512k, then on next open it will be moved to "log.1" (in its entirety) and a new log would be started. "log.1" would only be deleted when the new log reaches 512k (and then is itself moved to replace "log.1").

Maybe I should change the logic so that it only moves the last 512k into "log.1"? That way really gigantic existing logs would be immediately cut down to size on next open...

ocdtrekkie:
In fairness, if you have issues like me with 90 MB grains because of log files, it'll probably manage to hit that 512k mark pretty darn quick. And obviously for any grains newer than this change, there is no difference, effectively, with that change.

kentonv:
@ocdtrekkie I went ahead and added a commit which cuts the log down when rotating it, so now you should see space reclaimed the first time you open a grain after the update.

Note that this update probably won't go live on Oasis until Jan 8, because it will require a full update (with 20-minute downtime) rather than a quick update, due to modifying C++ code. Normally such updates are pushed at midnight Saturday night / Sunday morning (for minimal disruption), but this weekend we're all going to be occupied at that time, obviously. :)

PR #2821, opened 2017-01-14
[http-bridge] Asynchronously wait for app to accept connections

### LINKED ISSUE #2779: sandstorm-http-bridge: don't block capnp connections on HTTP server.
According to `git blame`, this `TODO(soon)` is ~2.5 years old:

https://github.com/sandstorm-io/sandstorm/blob/master/src/sandstorm/sandstorm-http-bridge.c%2B%2B#L2227

I just banged into it again yesterday.

Basically, sandstorm-http-bridge won't listen for capnp rpc connections until it's found an HTTP server.

dwrensha:
Fixes #2779.

The only places where we need to `connect()` to the app are in methods of `WebSessionImpl`, so it suffices to return promise capabilities from `UiViewImpl::newSession()` that resolve once we've successfully made the initial connection.

cc @zenhack

### DIFF
--- a/src/sandstorm/sandstorm-http-bridge.c++
+++ b/src/sandstorm/sandstorm-http-bridge.c++
@@ -2196,8 +2196,10 @@ public:
                                 kj::Timer& timer,
                                 bool loggedSlowStartupMessage,
                                 int numTriesSoFar) {
-    return address->connect().then([this](auto x) -> void {
-      KJ_LOG(WARNING, "App successfully started listening for TCP connections!");
+    return address->connect().then([this, loggedSlowStartupMessage](auto x) -> void {
+      if (loggedSlowStartupMessage) {
+        KJ_LOG(WARNING, "App successfully started listening for TCP connections!");
+      }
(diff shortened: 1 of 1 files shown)


src/sandstorm/sandstorm-http-bridge.c++
```diff
@@ -2173,6 +2192,35 @@ public:
     });
   }
 
+  kj::Promise<void> connectLoop(kj::Own<kj::NetworkAddress>&& address,
+                                kj::Timer& timer,
+                                bool loggedSlowStartupMessage,
+                                int numTriesSoFar) {
+    return address->connect().then([this](auto x) -> void {
+      KJ_LOG(WARNING, "App successfully started listening for TCP connections!");
```
kentonv: This log message seems overly chatty -- won't it be printed at every startup? IMO better to log only in the problem case.

dwrensha: Done.

PR #2833, opened 2017-01-24
SAML login: Don't require password authentication.

kentonv:
Apparently, this clause actually prevents Sandstorm from working with ADFS installations that use non-password mechanisms like Kerberos.

I do not know why SAML lets the client specify the authentication mechanism, nor do I know why our code specified password authentication here (seems to have been inherited from the old meteor-accounts-saml code).

I've tested this change against: SimpleSamlPhp, Windows Server ADFS, Azure AD.

@abliss This is the change you wanted, right?


kentonv:
@jparyani Any idea why this clause was in there?

jparyani:
It's from the upstream meteor-accounts-saml as you guessed. It seems that RequestedAuthnContext is documented as optional, so I'm not sure why it was in there.

Issue #2834, opened 2017-01-26
Tried to build with clang even though it isn't available


strugee:
When I clone Sandstorm on my Arch Linux system and invoke `make`, I get a bunch of normal output and then a bunch of errors ending with the following:

```
[snipped]
compile: kj/test.c++
  src/ekam-rules/compile.ekam-rule: line 119: clang++: command not found
compile: sodium/crypto_onetimeauth_poly1305/donna/poly1305_donna.c
compile: sodium/crypto_onetimeauth_poly1305/onetimeauth_poly1305.c
  src/ekam-rules/compile.ekam-rule: line 119: clang: command not found
compile: joyent-http/http_parser.c++
  src/ekam-rules/compile.ekam-rule: line 119: clang++: command not found
==== build failed. You might need to: make update-deps ====
make: *** [Makefile:257: tmp/.ekam-run] Error 1
```

The result of `make | sponge sandstorm-make-log.txt` can be found [here](https://github.com/sandstorm-io/sandstorm/files/733085/sandstorm-make-log.txt). I'm happy to provide more debugging, but I'm not sure where to look.

Here's some version info:

```
% gcc --version
gcc (GCC) 6.2.1 20160830
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

% make --version
GNU Make 4.2.1
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

kentonv:
There's documentation here on prerequisites for building Sandstorm:

https://docs.sandstorm.io/en/latest/install/#option-4-installing-from-source

It looks like you are missing Clang.

I suppose the bug here is that this section of the docs wasn't easy to find. Any suggestions on how we could make it more visible?

strugee:
@kentonv ah yeah. I got confused since (IIRC) the `ekam` docs said it worked on GCC.

[...]

PR #2836, opened 2017-01-26
Handle Sandstorm being bound on secondary addresses in Sandcats

strugee:
Fixes sandstorm-io/sandcats#153.

I'm not quite sure if this actually works, as I can't actually get a local dev environment set up (and even if I could I'm not particularly sure how to test Sandcats-related stuff since it calls out to the network service). However, it does pass JSHint, so at least it's not totally busted.

Question: does Sandcats determine the IP to set the DNS record to from the IP used to send the POST request? Or is it embedded in the HTTP body? AFAICT it's the former, but I could be wrong as I don't fully understand everything this code does yet.

Also, I fixed a whitespace issue as a bonus.

### DIFF
--- a/shell/server/sandcats.js
+++ b/shell/server/sandcats.js
@@ -26,7 +26,7 @@ const Url = Npm.require("url");
 import { SANDSTORM_ALTHOME } from "/imports/server/constants.js";
 
 const SANDCATS_HOSTNAME = (Meteor.settings && Meteor.settings.public &&
-                         Meteor.settings.public.sandcatsHostname);
+                           Meteor.settings.public.sandcatsHostname);
 const SANDCATS_VARDIR = (SANDSTORM_ALTHOME || "") + "/var/sandcats";
 
 const ROOT_URL = Url.parse(process.env.ROOT_URL);
--- a/shell/server/sandcats.js
+++ b/shell/server/sandcats.js
@@ -63,6 +63,10 @@ const pingUdp = () => {
     }
   });
 
+  socket.on("error", (err) => {
+    throw err;
+  });
+
   socket.bind({ address: process.env.BIND_IP }, () => {
     socket.send(message, 0, message.length, 8080, SANDCATS_HOSTNAME, (err) => {
       if (err) {
--- a/shell/server/sandcats.js
+++ b/shell/server/sandcats.js
@@ -70,7 +70,7 @@ const pingUdp = () => {
       }
     });
 
-    Meteor.setTimeout(() => {
+    setTimeout(() => {
       socket.close();
     }, 10 * 1000);
   });
--- a/shell/server/sandcats.js
+++ b/shell/server/sandcats.js
@@ -80,6 +80,7 @@ const performSandcatsRequest = (path, hostname, postData, errorCallback, respons
   const options = {
     hostname: hostname,
     path: path,
+    localAddress: process.env.BIND_IP,
     method: "POST",
     agent: false,
     key: fs.readFileSync(SANDCATS_VARDIR + "/id_rsa"),
--- a/shell/server/sandcats.js
+++ b/shell/server/sandcats.js
@@ -63,15 +63,17 @@ const pingUdp = () => {
     }
   });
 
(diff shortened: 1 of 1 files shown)


garply:

shell/server/sandcats.js
```diff
@@ -63,21 +63,24 @@ const pingUdp = () => {
     }
   });
 
-  socket.send(message, 0, message.length, 8080, SANDCATS_HOSTNAME, (err) => {
-    if (err) {
-      console.error("Couldn't send UDP sandcats ping", err);
-    }
-  });
+  socket.bind({ address: process.env.BIND_IP }, () => {
```
kentonv: Hmm, how are errors reported here? E.g. if the bind IP isn't a local address. Do they go to the callback, or does this throw? It's weird to me that bind() even takes a callback since the underlying syscall does not block anyway.

kentonv:
Garply, ok to test

shell/server/sandcats.js
strugee: @kentonv I'm not sure. The docs don't make it clear: https://nodejs.org/api/dgram.html#dgram_socket_bind_options_callback

Given the preceding section in that link, it seems like we should be listening to the `error` event? Does that seem right to you?

strugee:
@kentonv the machine I'm on currently is OS X, so that's a no-go. Logging into a Linux machine produced some build errors - see #2834. I think there was another build failure after I installed clang; I can file a separate issue about that. Also, docs were scarce. I can file a separate issue about that too if you want.

I think I'll try out this fix in production, since it doesn't require any Sandcats server-side changes. I'll report back.

strugee:
Also, CI is failing but I'm not sure how to interpret the results. Apps are failing to install, which doesn't make sense to me since AFAICT this doesn't affect anything having to do with that...

kentonv:
Oh, since this is a server-side (as opposed to browser-side) file, I think you could actually test this by simply finding the `sandcats.js` in your production Sandstorm server and swapping this file in over it. It should be the same file.

The CI failures definitely aren't your fault. I'm looking into it.

strugee:
Oh, neat! I just assumed all the JS was glommed together regardless of whether it was server or browser code. I'll try that in production.

strugee:
@kentonv nope, it all gets concatenated. That being said it's not minified so I can just manually copy the changes into production.

strugee:
So! The `localAddress: process.env.BIND_IP` definitely works. 🎉

However when I add in the `socket.bind()` change I get this in the logs:

```
/programs/server/packages/meteor.js:1060
    throw new Error("Meteor code must always run within a Fiber. " +
    ^

Error: Meteor code must always run within a Fiber. Try wrapping callbacks that you pass to non-Meteor libraries with Meteor.bindEnvironment.
    at Object.Meteor._nodeCodeMustBeInFiber (packages/meteor/dynamics_nodejs.js:9:1)
    at [object Object]._.extend.get (packages/meteor/dynamics_nodejs.js:21:1)
    at withoutInvocation (packages/meteor/timers.js:4:1)
    at bindAndCatch (packages/meteor/timers.js:13:1)
    at Object._.extend.setTimeout (packages/meteor/timers.js:29:1)
    at Socket.<anonymous> (server/sandcats.js:73:5)
    at Socket.g (events.js:260:16)
    at emitNone (events.js:67:13)
    at Socket.emit (events.js:166:7)
    at startListening (dgram.js:121:10)
```

(Note that that stack trace isn't quite accurate because my making manual changes means that the sourcemap is out of sync with the actual JS file, so the `sandcats.js` line number is probably incorrect.)

strugee:
@kentonv just pushed a fix for the crashes - see the commit message for the rationale. I bet that's what was causing CI to fail, too.

And with that, this is ready for merge! Everything works in production for me :)

kentonv:
Oh, the tests are failing because the build machine is out of disk space. Heh.

kentonv:
Thanks for the fix!

strugee:
@kentonv sure! Thanks for the merge (and for Sandstorm generally)! \o/

PR #2838, opened 2017-01-27
Extend http-bridge to support app-to-app powerbox HTTP APIs

kentonv:
Commits 3acb128 through c970f29 implement the ability to *export* HTTP APIs from an http-bridge app. In the bridge config, the user may define a list of APIs the grain supports. (Currently these are singleton declarations -- the grain cannot implement multiple sub-objects. For most apps, this is probably fine, because grains themselves are usually the appropriate unit of export.)

Commits cd92c08 through 73bd11d implement the ability to *import* HTTP APIs in an http-bridge app. http-bridge is extended to export an HTTP proxy on localhost:15239. The `HTTP_PROXY` and `http_proxy` environment variables are set appropriately. A special route, `http://http-bridge/session/<session-id>/claim`, wraps the `SessionContext.claimRequest()` method, converting a request token obtained via a (client-side) powerbox request into an access token. The access token can be used in subsequent requests in the `Authorization: Bearer` header to make requests directly to the requested capability, which is assumed to be a `WebSession`.

Hence, it is now possible for apps to utilize the powerbox without interacting with Cap'n Proto at all.

The import half of this depends on the new KJ HTTP library: https://github.com/sandstorm-io/capnproto/pull/404

### DIFF
--- a/src/sandstorm/api-session.capnp
+++ b/src/sandstorm/api-session.capnp
@@ -70,9 +70,8 @@ interface ApiSession @0xc879e379c625cdc7 extends(WebSession.WebSession) {
     # Google Calendar API version 2 as "https://apidata.googleusercontent.com/caldav/v2". However,
     # you would not normally request "https://api.github.com/users" because "GitHub users" is not
     # considered a separate API but rather one part of the overall GitHub API. Note that
-    # `canonicalUrl` should never end with a '/', because request paths to `ApiSession` are
-    # required to start with a '/', so this would result in two consecutive '/'s which is usually
-    # wrong.
+    # `canonicalUrl` should never end with a '/', because a '/' is added implicitly to separate
+    # the API URL from the individual request's path.
     #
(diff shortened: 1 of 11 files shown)


kentonv:
I am sort of shocked that this works. After several weeks of work it just all came together in the last 24 hours and... did what it was supposed to do.

I'm also excited that the bridge-proxy code could eventually replace proxy.js -- we could make the front-line HTTP server be C++! Would be a lot more efficient.

kentonv:
Garply, retest this please

kentonv:
Tests pass locally on my machine. It seems like something is up with the build server -- it's suddenly failing in the same way for several different PRs.

kentonv:
Garply, retest this please

kentonv:
The test failures -- which I can't reproduce locally -- don't appear to be caused by this change (the same failures are appearing on other PRs).

This change is relatively low-risk for release in that it only affects newly-built apps, hence if there's a critical problem it will only block new app releases rather than break things already in production.

So I'm gonna merge and release!

PR #2848, opened 2017-02-03
Fix bug where Sandcats wouldn't update or renew certs if BIND_IP was 127.0.0.1

kentonv:
This is common for people who use sniproxy.


kentonv:
Merging for immediate release since some sandcats servers are currently suffering from expired certs.

PR #2868, opened 2017-02-11
Initial draft of search API.

kentonv:
DO NOT MERGE

There are some `TODO(now)`s in there regarding parts I'm still thinking about, but wanted to get some initial feedback on this.

### DIFF
--- a/src/sandstorm/indexer.capnp
+++ b/src/sandstorm/indexer.capnp
@@ -1,0 +1,139 @@
+# Sandstorm - Personal Cloud Sandbox
+# Copyright (c) 2017 Sandstorm Development Group, Inc. and contributors
+# All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+@0xcae98639575b2b35;
+
+$import "/capnp/c++.capnp".namespace("sandstorm");
+
+using Grain = import "grain.capnp";
+
+interface IndexingSession(Metadata) {
+  # This is a UiView session type, created by calling UiView.newSession().
+  #
+  # Sandstorm requests a session of this type when it wants to index a grain for search purposes.
+  #
+  # Indexing sessions, like any other sessions, *must* pay attention to the UserInfo passed to
+  # `newSession()`; only content which is visible to that user can be indexed.
+  #
+  # Note that, as an optimization, Sandstorm may start out assuming that all users see exactly
+  # the same content, and may do all indexing as an anonymous user, perhaps with no permissions.
+  # However, if the app logs an ActivityEvent that specifies that it requires specific permissions
+  # or is visible only to certain users, Sandstorm uses that as a hint to index the path associated
+  # with that event separately specifically for people who can see said event. Hence, if your app
+  # has content that is visible to some users but not others, and that needs to be indexed, it
+  # should be sure to log appropriate activity events.
+
+  indexAll @0 (indexer :GrainIndexer(Metadata));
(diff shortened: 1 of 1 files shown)


src/sandstorm/indexer.capnp
```diff
@@ -0,0 +1,139 @@
...
+  # metadata format.
+  #
+  # TODO(someday): Spec out how this works. Not needed for MVP of search.
+}
+
+# ========================================================================================
+# Indexer implementation
+
+interface IndexerSession {
+  # A session type specifically implemented by the indexer app. The app's UiView's newSession()
```
dwrensha: How many indexer app grains are there? Just one? One per user? Maybe the are created on demand when the user manually sets up some grains as indexable?

src/sandstorm/indexer.capnp
```diff
@@ -0,0 +1,139 @@
...
+
+  indexPaths @1 (paths :List(Text), indexer :GrainIndexer(Metadata));
+  # Index a specific list of paths. If a path in the list doesn't exist, calls `indexer.index()`
+  # with `content` = null for that path.
+}
+
+interface GrainIndexer(Metadata) {
+  # Capability used to index the content of a grain.
+  #
+  # This is a one-way capability. Although GrainIndexer is implemented by the indexer and is called
```
dwrensha: I only vaguely understand what you're intending to protect against by making this a "one-way" capability. Should we be worried that capabilities could be smuggled through the `metadata` field of `IndexableContent`?

src/sandstorm/indexer.capnp
```diff
@@ -0,0 +1,139 @@
...
+  #
+  # Sandstorm requests a session of this type when it wants to index a grain for search purposes.
+  #
+  # Indexing sessions, like any other sessions, *must* pay attention to the UserInfo passed to
+  # `newSession()`; only content which is visible to that user can be indexed.
+  #
+  # Note that, as an optimization, Sandstorm may start out assuming that all users see exactly
+  # the same content, and may do all indexing as an anonymous user, perhaps with no permissions.
+  # However, if the app logs an ActivityEvent that specifies that it requires specific permissions
+  # or is visible only to certain users, Sandstorm uses that as a hint to index the path associated
```
dwrensha: Does this mean that Sandstorm tries to index any grain that calls `SessionContext.activity()`?

src/sandstorm/indexer.capnp
```diff
@@ -0,0 +1,139 @@
...
+# limitations under the License.
+
+@0xcae98639575b2b35;
+
+$import "/capnp/c++.capnp".namespace("sandstorm");
+
+using Grain = import "grain.capnp";
+
+interface IndexingSession(Metadata) {
+  # This is a UiView session type, created by calling UiView.newSession().
```
dwrensha: So I guess this should have `extends(UiSession)`?

kentonv:
I added comments answering all of @dwrensha's questions.

ocdtrekkie:
How well do you expect this to scale? Global full-text search of say... documents... seems reasonable to expect Sandstorm servers to do, but what happens if I have an eBook library with a few hundred books of text in my Sandstorm?

Will it be possible to define what type of content is being indexed? To extend the above example, if I have an eBook library, and a lot of personal documents, it may be hard to find personal documents (data I generated) I am looking for if it also returns a lot of results from books (data I have).

kentonv:
@ocdtrekkie I would guess that the tech can handle indexing your ebooks. Lucene is designed to be able to index millions or even billions of documents, and I assume Lucy is similar. As a rule of thumb, the index ought to be similar in size to the text it is indexing.

As for usability issues, that's a good point. I think we would handle this by adding search operators and such. This initial design doesn't really get into how those will work, but it can be extended in that direction in the future.

src/sandstorm/indexer.capnp
```diff
@@ -0,0 +1,168 @@
...
+
+  threadPath @2 :Text;
+  # Optional path of thread of which this item is a part, if any. See `ActivityEvent.thread`.
+
+  metadata @3 :Metadata;
+  # Metadata for search operators, e.g. "subject:", "author:", etc. Each app can define its own
+  # metadata format.
+  #
+  # TODO(someday): Spec out how this works. Not needed for MVP of search.
+}
```
zenhack: It's not obvious to me how an indexer would be able to use "generic" operators like this while treating the metadata as completely opaque.

If the plan is to not implement this right away anyway, is there a good reason to even include this method? We can always add more later...

PR #2869, opened 2017-02-16
Support comma-delimited emailDomain suffixes, optionally with leading "*."

audreyt:
For example, in the gov.tw instance we need to allow all .gov.tw users using `/\.gov\.tw$/` in this field.

If this makes sense, we can change the description of `Users with an email address at this domain will be members of this server's organization.` to mention the regex syntax.

### DIFF
--- a/shell/packages/sandstorm-db/db.js
+++ b/shell/packages/sandstorm-db/db.js
@@ -1109,7 +1109,16 @@ _.extend(SandstormDb.prototype, {
     const ldapEnabled = orgMembership && orgMembership.ldap && orgMembership.ldap.enabled;
     const samlEnabled = orgMembership && orgMembership.saml && orgMembership.saml.enabled;
     if (emailEnabled && emailDomain && identity.services.email) {
-      if (identity.services.email.email.toLowerCase().split("@").pop() === emailDomain) {
+      if (/^\/(.*)\/$/.test(emailDomain)) {
+        try {
+          const emailDomainRegex = new RegExp(RegExp.$1);
+          if (emailDomainRegex.test(identity.services.email.email.toLowerCase().split("@").pop())) {
+            return true;
+          }
+        } catch (e) {
+          console.error("Invalid emain romain regex:", emailDomain);
+        }      
+      } else if (identity.services.email.email.toLowerCase().split("@").pop() === emailDomain) {
         return true;
       }
     } else if (ldapEnabled && identity.services.ldap) {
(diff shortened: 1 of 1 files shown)


garply:

paulproteus:
FWIW, I am generally +1 on this. I appreciate that it keeps the existing functionality working properly.

kentonv:
Hmm, using regexes here makes me nervous. I'd like to propose instead that we support a leading wildcard, e.g. `*.gov.tw`, which works like DNS and TLS certificate wildcards. We could also support specifying multiple domains, comma-delimited.

Reasons:

First, I worry that regexes could complicate other features or refactorings we might do in the future. For example, what happens when we need to search the database for all users in the organization, and we don't want to do a table scan? Indexing for regex search is hard, but indexing a suffix search is straightforward. What happens if, say, we implement hosted organizations, and we need to verify that the org admin really owns the domain? Automatically verifying a regex is probably impossible, but verifying wildcards is a solved problem.

Second, regexes are not user-friendly and I can imagine less-knowledgeable users getting it wrong in ways that compromise their security. For example, I might accidentally enter `/.example.com$/` and then someone could register `evilexample.com` and own me. Or I might enter `/.*\.example\.com/` and then someone could register `example.com.evil.com` and own me.

audreyt:
OK! Re-implemented as per @kentonv .

shell/packages/sandstorm-db/db.js
```diff
@@ -1109,8 +1109,17 @@ _.extend(SandstormDb.prototype, {
...
     const samlEnabled = orgMembership && orgMembership.saml && orgMembership.saml.enabled;
     if (emailEnabled && emailDomain && identity.services.email) {
-      if (identity.services.email.email.toLowerCase().split("@").pop() === emailDomain) {
-        return true;
+      const domainSuffixes = emailDomain.split(/\s*,\s*/);
+      for (let i = 0; i < domainSuffixes.length; i++) {
+        const suffix = domainSuffixes[i];
+        const domain = identity.services.email.email.toLowerCase().split("@").pop();
+        if (suffix.startsWith("*.")) {
+          if (domain.endsWith(suffix.substr(2))) {
```
kentonv: I think this should be:

    if (domain.endsWith(suffix.substr(1))) {

Otherwise `*.gov.tw` matches `evilgov.tw`.

Otherwise looks good.

audreyt: Sharp eyes! Thanks, fixed.

kentonv:
Thanks!

kentonv:
(Expected release date: March 4.)

sraimund:
Could you please add the comma as a separator and the wildcard in the note below the text field "Users authenticated via email address. Domain: " (organization settings in the admin panel) and/or in the docs (https://docs.sandstorm.io/en/latest/administering/for-work/)?

I first tried a semicolon and locked myself out...

Thank you!

PR #2870, opened 2017-02-16
Support connecting to external HTTP APIs via the Powerbox

kentonv:
All requests for an `ApiSession` will offer the option to specify an HTTP URL manually. If the user specifies a URL containing a username/password or a webkey, the authorization mechanism is understood and used (without passing details to the app).

Additionally, requests with an HTTP `canonicalUrl` in their tag will offer a one-click option to use that URL. If the `canonicalUrl` refers to a well-known OAuth API service (currently recognizes Github and some Google API hosts), an OAuth handshake will occur.


shell/packages/sandstorm-db/db.js
```diff
@@ -492,6 +500,15 @@ ApiTokens = new Mongo.Collection("apiTokens", collectionOptions);
 //         address :Text;
 //       }
 //       identity :Text;
+//       http :group {
+//         url :Text;
+//         auth :union {
+//           none :Void;
+//           bearer :Text;
+//           basic :group { username :Text; password :Text; }
```
dwrensha: I wish we could encrypt these fields somehow. I don't see a good way to do that, though.

shell/server/drivers/external-ui-view.js
```diff
@@ -41,10 +45,207 @@ ExternalUiView = class ExternalUiView {
...
+      "Content-Type": "application/x-www-form-urlencoded",
+      "Accept": "application/json",
+    },
+    content: "client_id=" + encodeURIComponent(config.clientId)
+          + "&client_secret=" + encodeURIComponent(config.secret)
+          + "&refresh_token=" + encodeURIComponent(refreshToken)
+          + "&grant_type=refresh_token",
+  });
+
+  console.log(response.data);
```
dwrensha: Stray debug print?

shell/packages/sandstorm-db/db.js
```diff
@@ -426,6 +426,14 @@ ApiTokens = new Mongo.Collection("apiTokens", collectionOptions);
 //       verifiedEmail: An VerifiedEmail capability that is implemented by the frontend.
 //                      An object containing `verifierId`, `tabId`, and `address`.
 //       identity: An Identity capability. The field is the identity ID.
+//       http: An ApiSession capability pointing to an external HTTP service. Object containing:
```
dwrensha: Where in the UI will the user be able to audit such frontendRefs?

shell/packages/sandstorm-db/db.js
kentonv: We could encrypt them using the ApiToken as a key, since we only store the hash of the API token in the database. Child ApiTokens would store an encrypted copy of the parent token (in addition to the parent token ID as they do today). But a complete solution would require that app storage be encrypted as well, since that's where the tokens are ultimately stored. That is, of course, a big project...

shell/packages/sandstorm-db/db.js
kentonv: Good question. I don't plan to implement that in this PR, but it's important.

kentonv:
OK, I added a bunch of commits to implement encryption of these HTTP credentials!

Caveat: The encryption key is the token given to the grain. So, if the grain stores that token unencrypted, and if an attacker gets a copy of the grain's storage _and_ a copy of Sandstorm's Mongo database, then the attacker can decrypt the credentials.

However, an attacker who gets _only_ the Mongo database cannot decrypt any credentials within it. I suspect that getting the database and grain storage is in many cases significantly harder than getting just the database. For example, it's easy to imagine a Mongo injection bug in the Sandstorm shell which allows an attacker to get access to all ApiTokens, but it's harder to imagine a bug in which Sandstorm delivers raw grain storage data to attackers.

Some day, we will implement transparent encrypted grain storage in Blackrock, which will take things a step further.

shell/packages/sandstorm-db/db.js
```diff
@@ -508,6 +537,37 @@ ApiTokens = new Mongo.Collection("apiTokens", collectionOptions);
...
+//
+// Additionally, some ApiToken entries contain tokens to third-party services, e.g. OAuth tokens
+// or even passwords. Such tokens are encrypted, using the ApiToken entry's own full token (which,
+// again, is not stored in the database) as the encryption key.
+//
+// When such encryption is applied, the cipher used is ChaCha20. All API tokens are 256-bit base64
+// strings, hence can be used directly as the key. No MAC is applied, because this scheme is not
+// intended to protect against attackers who have write access to the database -- such an attacker
+// could almost certainly do more damage by modifying the non-encrypted fields anyway. (Put another
+// way, if we wanted to MAC something, we'd needto MAC the entire ApiToken structure, not just
```
dwrensha: s/needto/need to/

shell/packages/sandstorm-db/db.js
```diff
@@ -426,13 +426,33 @@ ApiTokens = new Mongo.Collection("apiTokens", collectionOptions);
 //       verifiedEmail: An VerifiedEmail capability that is implemented by the frontend.
 //                      An object containing `verifierId`, `tabId`, and `address`.
 //       identity: An Identity capability. The field is the identity ID.
+//       http: An ApiSession capability pointing to an external HTTP service. Object containing:
+//           url: Base URL of the external service.
+//           auth: Authentitation mechanism. Object containing one of:
```
ocdtrekkie: "Authentication"

shell/server/drivers/external-ui-view.js
```diff
@@ -41,10 +45,207 @@ ExternalUiView = class ExternalUiView {
...
+
+        parsedUrl.auth = null;
+        request.url = Url.format(parsedUrl);
+      }
+
+      if (parsedUrl.hash) {
+        if ("none" in request.auth) {
+          request.auth = { bearer: parsedUrl.hash.slice(1) };
+        } else {
+          throw new Meteor.Error(400, "Can't supprot multiple authentication mechanisms at once");
```
ocdtrekkie: "support"

kentonv:
Garply, retest this please.

Issue #2517, opened 2016-09-09
docs: Add a way to contact us


paulproteus:
Context:
- As a docs user,
- I am looking for help.
- I might want want a way to contact the Sandstorm core team, but I don't particularly have one.
- Therefore, you could imagine a "Get help from us!" button somewhere in the docs UI.

PR #2874, opened 2017-02-24
Add e-mail support button to docs footer

neynah:
"E-mail support" button added to the footer. Fixes: #2517 

image

### DIFF
--- a/docs/sandstorm_readthedocs_modified_theme/base.html
+++ b/docs/sandstorm_readthedocs_modified_theme/base.html
@@ -104,6 +104,8 @@
           </div>
 	  {%- block footer %}
           {% include "footer.html" %}
+          <div id="get-help">Need help?<a href="https://groups.google.com/forum/#!forum/sandstorm-dev" class="btn btn-neutral">Sandstorm-dev Group</a> <a href="https://kiwiirc.com/client/irc.freenode.net/?channel=#sandstorm" class="btn btn-neutral">IRC</a>
++         </div>
 	  {% endblock %}
         </div>
       </div>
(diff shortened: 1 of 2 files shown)


kentonv:
Hmm. Given that we're not full-time anymore, I think we should direct people towards sandstorm-dev / IRC / other community options and deprecating support@ (which is pretty much now an alias for me personally). Maybe we could link to `sandstorm.io/community` here or maybe we need a better landing page specifically for "where do I ask questions"?

neynah:
I think sandstorm-dev & IRC make the most sense for the time being until we have a better "where do I ask questions?" landing page as you mention. Perhaps this can live somewhere like sandstorm.io/help.

neynah:
Updated the commit to display "Sandstorm-dev Group" and "IRC"

image

docs/sandstorm_readthedocs_modified_theme/base.html
```diff
@@ -104,6 +104,8 @@
           </div>
 	  {%- block footer %}
           {% include "footer.html" %}
+          <div id="get-help">Need help?<a href="https://groups.google.com/forum/#!forum/sandstorm-dev" class="btn btn-neutral">Sandstorm-dev Group</a> <a href="https://kiwiirc.com/client/irc.freenode.net/?channel=#sandstorm" class="btn btn-neutral">IRC</a>
++         </div>
```
kentonv: Oops, there's still a stray `+` here which shows up on the final site as a "+" to the right of the buttons.

kentonv: I'll fix it.

neynah: Oops! Thanks.

PR #2877, opened 2017-02-28
Upload/download backups on DDP host rather than main host.

kentonv:
When a separate DDP host is specified, it should be used for all "dynamic" interaction, where the main host is used only for static content (e.g. static HTML, Javascript, CSS, images).

See also: https://sandstorm.io/news/2017-02-28-cloudbleed


dwrensha:
What about spk uploads? It looks like it would be easy to make this change for them as well.

I suppose nothing needs to be done about profile image uploads, because they already go through the "static" host.

kentonv:
Good point, done.

(I wasn't really thinking about spk uploads since they generally do not contain secrets, but for performance reasons it's still good to upload them directly.)

PR #2879, opened 2017-03-06
Fix SAML login for Azure active directory accounts

miyazakijunichi:
When I  trying to logged in with Azure AD,  an error occurred with message `SAML profile did not contain an email address`  and I could not log in.
As far as I can see, this issue is caused by the field name of SAML response about email is different between Azure AD accounts and Microsoft accounts. (ref., https://onlinehelp.tableau.com/current/online/en-us/saml_config_azure_ad.htm)
This patch is fix of this issue.

### DIFF
--- a/shell/imports/server/accounts/saml-utils.js
+++ b/shell/imports/server/accounts/saml-utils.js
@@ -354,6 +354,11 @@ SAML.prototype.validateResponse = function (samlResponse, callback) {
           profile.email = microsoftEmail;
         }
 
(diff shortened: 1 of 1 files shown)


garply:

kentonv:
Guh, why is AD so weird? Looking at your link, it indeed documents that there are two different possible names for the e-mail field...

kentonv:
Thanks for the fix! Probable next release date is March 18th.

PR #2885, opened 2017-03-11
Draft of cron API.

kentonv:
@mrdomino 

I feel like I'm forgetting some things here, but this is my first take.

### DIFF
--- a/src/sandstorm/grain.capnp
+++ b/src/sandstorm/grain.capnp
@@ -126,6 +126,63 @@ interface SandstormApi(AppObjectId) {
   # This method is here on `SandstormApi` rather than on `Identity` in order to ease a possible
   # future transition to a model where identity IDs are not globally stable and each grain has a
   # separate (possibly incompatible) map from identity ID to account ID.
+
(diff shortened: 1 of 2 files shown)


mrdomino:
Wow, somehow capnp is outputting a definition of `SandstormApi<AppObjectId>::MINIMUM_SCHEDULING_SLACK` without a template decl.

kpreid:
It looks like in this interface, if an app calls `schedulePeriodic`, then `save`, then forgets the token, it cannot ever cancel the scheduling internally. __How about giving the callback some means of communicating "please stop"?__

While I have no specific scenario other than something contrived like "the app schedules its periodic thing and then immediately crashes", this in general allows the possibility of resynchronizing grain-versus-environment state after any kind of failure (change existant callback(s) to return "stop", then schedule with a fresh callback).

(A similarly robust alternative would be @mrdomino 's scheduling proposal with the ability to list everything currently scheduled, but I consider that design to make things less composable within a grain since it creates a free-for-all namespace and invites "cancel everything on the list".)

ocdtrekkie:
This was included in #3172 and can be closed.

PR #2887, opened 2017-03-12
WIP: Extend WebSession to support range requests.

kentonv:
This is just the protocol change for now. Someone additionally needs to update `proxy.js` and `sandstorm-http-bridge.c++` and perhaps `bridge-proxy.c++`.

### DIFF
--- a/src/sandstorm/web-session.capnp
+++ b/src/sandstorm/web-session.capnp
@@ -138,6 +138,12 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
       matchesNoneOf @7 :List(ETag);  # If-None-Match
     }
 
+    rangeHint @10 :Range;
+    # Requests that the server returns a range of bytes from the document rather than the entire
+    # document. If the server accepts the range hint, it will return a `content` response with
+    # `statusCode` set to `partialContent`. However, the server is also free to ignore the
+    # `rangeHint`.
+
     additionalHeaders @3 :List(Header);
     # Additional headers present in the request. Only whitelisted headers are
     # permitted.
@@ -173,6 +179,7 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
     mimeType @0 :Text;
     content @1 :Data;
     encoding @2 :Text;  # Content-Encoding header (optional).
+    range @3 :Range;    # Content-Range header (optional).
   }
 
   struct PutContent {
@@ -181,6 +188,7 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
     mimeType @0 :Text;
     content @1 :Data;
     encoding @2 :Text;  # Content-Encoding header (optional).
+    range @3 :Range;    # Content-Range header (optional).
   }
 
   struct ETag {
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
@zenhack How do you feel about this issue and WIP? It was specifically mentioned as a pain point recently.

src/sandstorm/web-session.capnp
```diff
@@ -181,6 +188,7 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
     mimeType @0 :Text;
```
zenhack: A bit of a tangent, but: would it actually break anything to delete PutContent per the comment? the layout is exactly the same so it should be wire compatible. The type IDs are different, but it's hard for me to imagine that anything uses that.

src/sandstorm/web-session.capnp
```diff
@@ -190,6 +198,28 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
...
+    # Start point of the range, in bytes.
+
+    end :union {
+      # Endpoint of the range, in bytes.
+
+      endOfContent @1 :Void;
+      position @2 :UInt64;
+    }
+
+    fullSize :union {
```
zenhack: As far as I can tell, this is only valid in responses. Maybe we should remove it, since we only currently use the Range struct in requests?

src/sandstorm/web-session.capnp
```diff
@@ -243,11 +273,22 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
...
+      # via `Context.rangeHint`. This status is only allowed when a `rangeHint` was present.
+      #
+      # (In traditional HTTP, the server could return an arbitrary range that has nothing to do
+      # with the range the client requested. Since this is useless in practice and annoying to
+      # check for, WebSession only allows the server to return exactly the range the client
+      # wanted. If an http-bridge app returns a range that doesn't match what the client asked
+      # for, sandstorm-http-bridge will throw an exception instead.)
 
-      # This seems to fit better here than in the 3xx range
+      # TODO(apibump): Only the three status codes above should actually be used. The status codes
```
zenhack: s/three/four/

src/sandstorm/web-session.capnp
kentonv: It'd be wire-compatible, but not source-compatible. Some apps could have their builds broken. Maybe we don't care that much...

src/sandstorm/web-session.capnp
kentonv: Huh. It's been a long time since I wrote this, but I wonder if I had intended to include a `Range` under `Response.content` but forgot... otherwise I'm not sure how the `Content-Range` header is supposed to be constructed for the response.

zenhack: See the comment you wrote below -- looks like the idea was to just insist that the app returns the range that was actually asked for, so the supervisor only needs to check the status code.

kentonv: But how is the overall size communicated? I guess the final `Content-Range` header would always end up ending in `/*`?

BTW, what makes you say that `fullSize` is not valid for requests? It doesn't appear in a GET request (`Range` header), but a PUT that is uploading a range (`Content-Range` header) could include it, right? (However, I'm not sure if it's actually worth supporting range PUTs...)

zenhack: Yes, I was only thinking about GET requests. I guess I don't feel that strongly about this.

src/sandstorm/web-session.capnp
zenhack: I doubt very many apps actually use this directly, and for the ones that do it's a trivial fix. But probably doesn't make sense to change it as part of this patch.

PR #2901, opened 2017-03-27
add 'connect-src wss: ws' so that Safari can use a websocket on the main host

### LINKED ISSUE #2578: This server seems to be unable to create a WebSocket
I'm suddenly getting the WebSocket message on all sandstorm servers I run, all configured with nginx reverse proxy, all according to the mentioned [nginx-example.conf](https://github.com/sandstorm-io/sandstorm/blob/master/docs/administering/sample-config/nginx-example.conf).

Is this a false alert?

```
WARNING: This server seems to be unable to create a WebSocket. Sandstorm may appear to run fine, but many apps will fail to run properly. Learn more. Usually this is caused by a reverse proxy like nginx or Apache that needs to be reconfigured. If you think you are seeing this message in error, please reload this page or file an issue.
```

dwrensha:
Fixes #2578.

Our `connect-src` already includes `*`. According to [this document](https://webkit.org/blog/6830/a-refined-content-security-policy/), last year WebKit changed the meaning of the `*` in a backwards-incompatible way, such that now it only matches the scheme of the current page. So we still need to add a `ws` and/or `wss` scheme.

See also: https://github.com/meteor/meteor/issues/7772#issuecomment-265600327

### DIFF
--- a/shell/server/shell-server.js
+++ b/shell/server/shell-server.js
@@ -60,6 +60,8 @@ BrowserPolicy.content.allowImageOrigin(staticAssetHost);
 BrowserPolicy.content.allowScriptOrigin(staticAssetHost);
 BrowserPolicy.content.allowFontOrigin(staticAssetHost);
 BrowserPolicy.content.allowConnectOrigin(staticAssetHost);
+BrowserPolicy.content.allowConnectOrigin("wss:");
+BrowserPolicy.content.allowConnectOrigin("ws:");
 
 Meteor.publish("grainsMenu", function () {
   if (this.userId) {


shell/server/shell-server.js
```diff
@@ -60,6 +60,8 @@ BrowserPolicy.content.allowImageOrigin(staticAssetHost);
 BrowserPolicy.content.allowScriptOrigin(staticAssetHost);
 BrowserPolicy.content.allowFontOrigin(staticAssetHost);
 BrowserPolicy.content.allowConnectOrigin(staticAssetHost);
+BrowserPolicy.content.allowConnectOrigin("wss:");
```
kpreid: Doesn't this allow WebSocket connections to _any_ origin, and if so, isn't that unwanted?

dwrensha: As part of [meteor's default policy](https://github.com/meteor/meteor/blob/cf4133c7632137d870136f02100622aa7e8fcd95/packages/browser-policy-content/browser-policy-content.js#L139), we already have `connect-src: *`, which allows websocket connects to any origin in browsers other than Safari, if I understand correctly.

PR #2919, opened 2017-04-21
Add account settings tests.

wdk122:


garply:

kentonv:
Hey, just wanted to mention, I saw this, and this is really cool! Nice work figuring out the test framework without even asking any questions. :) I plan to look closer this weekend.

For the account settings test, I think we should also test that the profile information is correctly passed on to apps. This will require using a test app, though, which raises a bunch of new questions. There's already a C++ test app in the repository that uses the raw API, but it may also make sense to create a new Node.js-based test app that uses sandstorm-http-bridge, to test both. We also have a variety of out-of-tree test apps which aren't very well organized currently, but we could also consider adopting one of those. Anyway, let's chat about this over the weekend.

kentonv:
Garply, OK to test.

wdk122:
I've created a new pull request at [#2939](https://github.com/sandstorm-io/sandstorm/pull/2939) to follow up.

PR #2920, opened 2017-04-25
fix errors reported by clang++ 4.0

dwrensha:
`realpath()` and `signal()` return a pointer, but `KJ_SYSCALL()` expects a function that returns an integer.


kentonv:
I guess these calls seem unlikely to need EINTR handling...

PR #2926, opened 2017-04-30
Add man page references re: mount propagation type

zenhack:
A while back I'd been just playing around with these things, and I derived the same set of arguments by reading the docs -- so when I was poking through this code and saw the AFAICT comment, I knew that couldn't be right. I've changed the comment to point to the relevant docs, and supply a quick summary.


garply:

kentonv:
It appears the man page has been updated! These definitely weren't documented at the time I wrote the comment. :)

PR #2927, opened 2017-04-30
Correctly pass through error bodies for Powerbox HTTP requests

kentonv:
Previously, error responses were replacing the response body with the error status name, e.g. "Not Found". For HTTP APIs, error bodies are often important.

Since HTTP APIs often want to return non-HTML error bodies, I extended web-session.capnp to be able to represent these.

However, I have not updated sandstorm-http-bridge or proxy.js to understand this protocol for now. I would like to rewrite sandstorm-http-bridge in terms of kj-http at some point, which would be a good time to update it.

This also fixes a bug: The content-type header on 200-OK responses wasn't being set correctly (it was clobbering the Content-Language header instead).


kentonv:
@urakagi Hoping this helps you out! I'll release it tonight.

PR #2932, opened 2017-05-04
Remove dead code

zenhack:
I was futzing around with typescript, decided to run it on sandstorm to see what it caught without any annotations -- this is what I got.

Note that I'm not really clear on the purpose of that code, so for all I know this may represent a real bug, but it also may.


garply:

kentonv:
I guess it's debatable whether the line should be removed or moved above the `throw`. This is a code path that should only be possible to hit if the calling code is buggy, so all bets are off by that point. So I'll go ahead and merge this. Thanks!

PR #2935, opened 2017-05-08
Send a notification warning affected users that identities are changing.

kentonv:
See explanation here: https://github.com/sandstorm-io/sandstorm-website/pull/289

We send a bell notification to all users who have multiple identities where the display names of those identities are not all the same. We also show a warning on the accounts page for these users.

We specifically look for users with varying names on the assumption that if all the names are the same, then the user really only has one identity anyway, and the coming changes will be a strict improvement for them. I didn't feel it was necessary to compare profile pictures, handles, or preferred pronouns because if the display name is the same, it seems unlikely that the profiles vary significantly in these other factors.

### DIFF
--- a/shell/imports/server/migrations.js
+++ b/shell/imports/server/migrations.js
@@ -787,6 +787,48 @@ function setIpBlacklist(db, backend) {
   }
 }
 
+function notifyIdentityChanges(db, backend) {
+  // Notify users who might be affected by the identity model changes.
+  //
+  // Two types of users are affected:
+  // - Users who have multiple identities with differing names.
+  // - Users who have identities that are shared with other users.
+  //
+  // However, the second group seems like it must be a subset of the first group. So we only check
+  // for the first.
+
+  const names = {};
+  Meteor.users.find({ "profile.name": { $exists: true } }, { fields: { "profile.name": 1 } })
+      .forEach(user => {
+    names[user._id] = user.profile.name;
+  });
+
+  Meteor.users.find({ loginIdentities: { $exists: true } },
+                    { fields: { loginIdentities: 1, nonloginIdentities: 1 } }).forEach(user => {
+    let previousName = null;
+    let needsNotification = false;
+    SandstormDb.getUserIdentityIds( user).forEach(identityId => {
+      const name = names[identityId];
+      if (!name || (previousName && previousName !== name)) {
+        needsNotification = true;
+      }
+      previousName = name;
+    });
+
+    if (needsNotification) {
+      db.collections.notifications.upsert({
+        userId: user._id,
+        identityChanges: true,
+      }, {
+        userId: user._id,
+        identityChanges: true,
+        timestamp: new Date(),
+        isUnread: true,
+      });
+    }
+  });
+}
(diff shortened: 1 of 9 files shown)


shell/imports/server/migrations.js
```diff
@@ -787,6 +787,49 @@ function setIpBlacklist(db, backend) {
...
+  // Two types of users are affected:
+  // - Users who have multiple identities with differing names.
+  // - Users who have identities that are shared with other users.
+  //
+  // However, the second group seems like it must be a subset of the first group. So we only check
+  // for the first.
+
+  const names = {};
+  Meteor.users.find({ "profile.name": { $exists: true } }, { fields: { "profile.name": 1 } })
+      .forEach(user => {
```
dwrensha: What about profiles for which the `name` field is not filled in yet? (See https://github.com/sandstorm-io/sandstorm/blob/v0.207/shell/packages/sandstorm-db/profile.js#L189-L222.)

dwrensha: Ah, looks like account with such identities will *always* get an `identityChanges` notification.

shell/imports/server/migrations.js
```diff
@@ -787,6 +787,49 @@ function setIpBlacklist(db, backend) {
...
 
+function notifyIdentityChanges(db, backend) {
+  // Notify users who might be affected by the identity model changes.
+  //
+  // Two types of users are affected:
+  // - Users who have multiple identities with differing names.
+  // - Users who have identities that are shared with other users.
+  //
+  // However, the second group seems like it must be a subset of the first group. So we only check
+  // for the first.
```
dwrensha: Why must the second group be a subset of the first? Suppose I have an account where github:dwrensha is a login identity and email:david@sandstorm.io is a non-login identity, and I have another account where google:dwrenshaw is a login identity and email:david@sandstorm.io is again a non-login identity. There's nothing stopping all of the associated profiles from having the same name "David Renshaw".

kentonv: Hmm. I wasn't really thinking about shared identities that are shared between accounts belonging to the same human, and hence have the same name as the user's other identities. I was thinking about brand accounts, which would presumably have a different name from the user's primary identity.

It seems to me, though, that the use case you described is largely unaffected by the change, or at least it's not terribly important to notify them in advance. They aren't at risk of accidentally revealing a secret identity when the change occurs.

So I'm not going to worry about detecting this case. (I know it wouldn't be super-hard but my time to work on sandstorm stuff is very limited...)

kentonv: Also there are probably like two people in the world that actually have the setup you describe and they are both in this thread. :)

shell/imports/server/migrations.js
kentonv: I think in practice almost all identities will have the profile filled in since you're forced to submit your profile immediately after first login.

I guess adding calls to `SandstormDb.fillInProfileDefaults()` would not be very hard but, again, not sure it's worth spending time to re-test, so, meh. Maybe some extra people will get notifications.

Issue #293, opened 2015-03-15
spk pack: no file found to satisfy requirement


JamborJan:
I'm currently finishing the sandstorm port of paperwork (https://github.com/twostairs/paperwork).

While ```spk dev``` runs as expected, generates a proper ```sandstorm-files.list``` and let's me use the app in dev mode, ```spk pack``` bothers me with a ton of files it claims it doesn't find. For example:

```
No file found to satisfy requirement: etc/php5/cli/conf.d/20-pdo_sqlite.ini
```

I have tested with some files, copying would help. I could now copy the files from my OS to a folder in the app but that shouldn't be the solution of the problem, or am I wrong?

I used @jparyani way of porting apps via Docker.

kentonv:
It's possible that this is a file that existed a previous time that you ran `spk dev`, but no longer exists now. If that's the case -- meaning presumably you don't actually need this file -- then you can delete it from `sandstorm-files.list` and try again. If you find that there are a lot of such files, you may want to delete `sandstorm-files.list` altogether and then re-run `spk dev` to regenerate it (make sure to test all the features of your app so that all needed files are detected).

paulproteus:
Hi JJ!

Some notes here:

* Try, if possible, to provide the full list of files that aren't being
found. That list of files can provide a pattern. I'm pretty familiar with
Debian/Ubuntu packaging tools so can usually quickly help debug given that
information.

* I agree with Kenton that the files probably existed last time you ran
"spk dev" but don't exist now, and can probably be removed from the files
list.

Thanks for working on this! I hope it's been informational, and I also hope
that you can give feedback on this (and future) packaging tools so that we
can make them truly easy to use.

kentonv:
> Try, if possible, to provide the full list of files that aren't being found.

Sadly, `spk pack` currently doesn't provide this, as it throws an exception on the first file it can't find. But that's probably a bug we should fix...

JamborJan:
Okay there are two things I didn't take that much care about:
- the ```sandstorm-files.list```grows and grows with every attempt you have trying to build your app. So I had a 3.4 MB big file instead of 4KB which it has now after creating it new
- the "Docker process" to build an app uses the exported Docker environment. You should delete this folder before running a new Docker build process otherwise you have kind of recursively /opt/app/dockerenv/opt/app/dockerenv/... in your app.

Building the app worked now. I'm not quite sure but maybe it would be good that ```spk dev``` would always create the file list from scratch. 

Thanks
[...]

PR #2939, opened 2017-05-13
Tests for account settings passing to a Meteor app

wdk122:
This is a followup to [#2919](https://github.com/sandstorm-io/sandstorm/pull/2919). I've added a small Meteor test app that echoes the user's profile data, tests that verify this behavior, and make rules to install meteor-spk, pack the spk, and repack the spk only if there are changes between the current spk and the test app source files.

We might want to discuss refactoring my changes to the Makefile to make it better. These changes might break the "Run Tests -- Manual" portion of the sandstorm/tests README.


garply:

kentonv:
Garply, OK to test.

Makefile
```diff
@@ -150,9 +150,29 @@ update: sandstorm-$(BUILD)-fast.tar.xz
...
 
-test: sandstorm-$(BUILD)-fast.tar.xz test-app.spk
+test: sandstorm-$(BUILD)-fast.tar.xz test-app.spk pack-meteor-testapp
 	tests/run-local.sh sandstorm-$(BUILD)-fast.tar.xz test-app.spk
 
+meteor-testapp-clean:
+	rm -rf tests/assets/meteor-testapp.spk meteor-spk-0.3.2 meteor-testapp/.meteor-spk
+
+pack-meteor-testapp:
+	@if [ ! -f ./meteor-spk-0.3.2/meteor-spk ]; then \
```
kentonv: I think it would make sense for this to be a separate make rule for `meteor-spk-0.3.2/meteor-spk`.

Makefile
```diff
@@ -150,9 +150,29 @@ update: sandstorm-$(BUILD)-fast.tar.xz
...
+		echo "Downloading meteor-spk into" $(PWD)"..." && \
```
kentonv: Recommend using `$(call color,downloading meteor-spk)` rather than echo.

Makefile
```diff
@@ -150,9 +150,29 @@ update: sandstorm-$(BUILD)-fast.tar.xz
...
+		curl https://dl.sandstorm.io/meteor-spk-0.3.2.tar.xz | tar Jxf -; \
+	fi; \
+	if [ ! -f ./tests/assets/meteor-testapp.spk ]; then \
+		cd meteor-testapp && \
+		echo "No meteor-testapp.spk found in sandstorm/tests/assets, packing one now..." && \
+		../meteor-spk-0.3.2/meteor-spk pack ../tests/assets/meteor-testapp.spk; \
+	elif test "$$(find meteor-testapp -name '*' -newer ./tests/assets/meteor-testapp.spk -not -path "*node_modules*" -not -path "*.meteor-spk*" -not -path "*.meteor*" | wc -l)" != "0" ; then \
```
kentonv: It looks like you're sort of re-implementing what `make` does here. Instead, you could have the make rule itself be called `tests/assets/meteor-testapp.spk`, and declare all the source files (and `meteor-spk-0.3.2/meteor-spk`) as dependencies. Make will then automatically rebuild the file whenever the dependencies change.

tests/tests/account-settings.js
```diff
@@ -89,3 +90,42 @@ module.exports["Test profile changes"] = function (browser) {
...
+    .click('#confirmInstall')
+    .waitForElementVisible('button.action', medium_wait)
+    .click('button.action')
+    .pause(15000)
+    .grainFrame()
+    .executeAsync(function (done) {
+      done(Meteor.sandstormUser());
+    }, [], function (result) {
+      this
+        .assert.containsText('#id', result.value.id)
```
kentonv: Hmm, it looks to me like these assertions here are checking that the content displayed on the screen matches the meteor user data present in the iframe.

What we'd really like to do, though, is check that the info matches what we've set on the account settings page -- that is, we're testing that the profile is being passed correctly from Sandstorm.

So what we want to do is go to the account settings page, set a particular name, handle, etc., and then open the meteor test app and verify that the profile data there matches what we set.

meteor-testapp/.meteor/packages
```diff
@@ -0,0 +1,23 @@
...
+
+standard-minifier-css@1.3.4   # CSS minifier run for production mode
+standard-minifier-js@2.0.0    # JS minifier run for production mode
+es5-shim@4.6.15                # ECMAScript 5 compatibility for older browsers.
+ecmascript@0.7.2              # Enable ECMAScript2015+ syntax in app code
+shell-server@0.2.3            # Server-side component of the `meteor shell` command
+
+autopublish@1.0.7             # Publish all data to the clients (for prototyping)
+insecure@1.0.7                # Allow all DB writes from clients (for prototyping)
+jacksingleton:accounts-sandstorm-dev
```
kentonv: FWIW I think `jacksingcleton:accounts-sandstorm-dev` is obsolete, as I effectively merged its functionality into `accounts-sandstorm` at some point...

kentonv:
Oops, looks like the build fails on the CI because `spk` is not pre-installed there. You'll need to change the Makefile to use `./bin/spk` as built by the ekam run. Since `spk` is being called inside the `meteor-spk` script, you'll probably need to do something like `PATH="$PWD/bin:$PATH" meteor-spk`.

wdk122:
In this pull request, account-settings.js has two test cases: one for verifying profile changes after submitting the form and seeing a flash success message, and another for checking that profile information passes to the app. Should I merge these into one test case?

wdk122:
The new build error is:

>No rule to make target 'meteor-testapp/.meteor-spk', needed by 'tests/assets/meteor-testapp.spk'.  

I used the expression `$(shell find meteor-testapp -type f -o -type d)` to list all files and directories in the meteor-testapp directory as dependencies for building meteor-testapp.spk, which works on my machine, but not in Jenkins. Should I only list a few specific files as dependencies?

Makefile
```diff
@@ -415,3 +415,16 @@ test-app-dev: tmp/.ekam-run
...
+
+# ====================================================================
+# meteor-testapp.spk
+
+meteor-spk-0.3.2/meteor-spk:
+	@$(call color,downloading meteor-spk)
+	@curl https://dl.sandstorm.io/meteor-spk-0.3.2.tar.xz | tar Jxf -
+
+tests/assets/meteor-testapp.spk: meteor-testapp meteor-spk-0.3.2/meteor-spk $(shell find meteor-testapp -type f -o -type d)
+	@cd meteor-testapp && PATH="$$PATH:$$PWD/bin" && ../meteor-spk-0.3.2/meteor-spk pack ../tests/assets/meteor-testapp.spk
```
kentonv: You probably need to set the PATH before changing directories, otherwise PWD will have the wrong value (since it will point at the meteor-testapp directory). You should also put `$PWD/bin` at the front of the list rather than the back so that it gets searched first.

    @PATH="$$PWD/bin:$$PATH" && cd meteor-testapp && ../meteor-spk-0.3.2/meteor-spk pack ../tests/assets/meteor-testapp.spk

Makefile
```diff
@@ -415,3 +415,16 @@ test-app-dev: tmp/.ekam-run
...
 	spk dev -Isrc -Itmp -ptmp/sandstorm/test-app/test-app.capnp:pkgdef
+
+# ====================================================================
+# meteor-testapp.spk
+
+meteor-spk-0.3.2/meteor-spk:
+	@$(call color,downloading meteor-spk)
+	@curl https://dl.sandstorm.io/meteor-spk-0.3.2.tar.xz | tar Jxf -
+
+tests/assets/meteor-testapp.spk: meteor-testapp meteor-spk-0.3.2/meteor-spk $(shell find meteor-testapp -type f -o -type d)
```
kentonv: I think the CI error is because the CI actually runs Make as:

    make deps clobber-deps ci-clean fast test

The `$(shell ...)` is expanded at the start, when parsing the makefile. Later on, `ci-clean` deletes some files from under `meteor-testapp`, including files which the `$(shell)` found earlier, hence those files are still listed as dependencies. So then, when Make gets around to building `meteor-testapp.spk`, it thinks these dependencies are missing.

Unfortunately I think you are going to need to list source files more explicitly here. You don't have to list everything, and you can use wildcards. Maybe something like:

    meteor-testapp/client/* meteor-testapp/server/* meteor-testapp/*.js meteor-testapp/.meteor/*

Makefile
wdk122: When I prepend sandstorm/bin to PATH as written above, I get an error when trying to build meteor-testapp.spk:

>*** Uncaught exception ***
>sandstorm-pkgdef.capnp:2: failed: Import failed: /sandstorm/package.capnp
>Makefile:430: recipe for target 'tests/assets/meteor-testapp.spk' failed
>make: *** [tests/assets/meteor-testapp.spk] Error 1

Is this because there is a sandstorm binary inside sandstorm/bin that is the first sandstorm encountered in PATH?

kentonv: No, I think this is because `spk` doesn't know where to find the capnp files. The installed spk finds the files by looking for the correct directory relative to itself, but now we're running the one that isn't installed, so it doesn't know where to look. To fix this, add `-I../src` to the `meteor-spk` command line -- this tells it to find the capnp files in `../src` (which contains `sandstorm/package.capnp`).

wdk122: From `spk help` it looks like the command should be:
> ../meteor-spk-0.3.2/meteor-spk -I../src pack ../tests/assets/meteor-testapp.spk

... but the above change yields this error:
> No file found to satisfy requirement: 
Makefile:430: recipe for target 'tests/assets/meteor-testapp.spk' failed
make: *** [tests/assets/meteor-testapp.spk] Error 1

Based on the discussion in [293](https://github.com/sandstorm-io/sandstorm/issues/293), I ran `meteor-spk dev` to try to get rid of that error, and got a new error:
> *** Uncaught exception ***
sandstorm/spk.c++:1115: failed: lstat(target.cStr(), &stats): No such file or directory; target = /home/bill/sandstorm/bin/bin/sandstorm-http-bridge
stack: 0x5503a8 0x54e593 0x54d6d1 0x54d60d 0x4f9041 0x4f9cc5
Makefile:430: recipe for target 'tests/assets/meteor-testapp.spk' failed
make: *** [tests/assets/meteor-testapp.spk] Error 1

kentonv: You need to put the `-I../src` after `pack`. `spk` allows you to put it on either side, but the `meteor-spk` wrapper will get confused if you put it before.

Specifically what's happening here is the `meteor-spk` script looks at the first argument to decide what extra setup it needs to do before running `spk`. In the case of `pack` it arranges the dependencies where `spk` can find them. But in your case it sees `-I../src` which it doesn't understand, so it just passes along to `spk` verbatim, but then `spk` can't find the files it needs since `meteor-spk` didn't set them up.

kentonv:
Garply, retest this please.

Makefile
```diff
@@ -415,3 +415,16 @@ test-app-dev: tmp/.ekam-run
...
+
+meteor-spk-0.3.2/meteor-spk:
+	@$(call color,downloading meteor-spk)
+	@curl https://dl.sandstorm.io/meteor-spk-0.3.2.tar.xz | tar Jxf -
+
+tests/assets/meteor-testapp.spk: meteor-testapp meteor-spk-0.3.2/meteor-spk meteor-testapp/client/* meteor-testapp/server/* meteor-testapp/.meteor/*
+	@PATH="$$PWD/bin:$$PATH" && cd meteor-testapp && ../meteor-spk-0.3.2/meteor-spk pack -I../src ../tests/assets/meteor-testapp.spk
+
+meteor-testapp-clean:
+	rm -rf tests/assets/meteor-testapp.spk meteor-spk-0.3.2 meteor-testapp/.meteor-spk
```
kentonv: It occurs to me that we probably don't want to delete meteor-spk as part of `ci-clean` (which the CI runs before every build), since that's just going to force it to re-download over and over the same file which never changes. But rebuilding the .spk seems like a good idea.

Usually we don't create a separate "clean" rule for each component of the build but instead add files to the list of the existing clean rules. So maybe add meteor-spk to the main `clean` rule and the other things to the `ci-clean` rule.

kentonv:
Oops, now the problem is the missing keyring. You'll need to extract the signing key for the meteor test app from your keyring (see `spk help`), commit it to the repo, and then add a flag like the `-ksrc/sandstorm/test-app/test-app.key` that is used for the C++ test app (see the make rules above your new ones).

kentonv:
(It's OK to submit the signing key for the test app since it's not a real app and we'll never publish it.)

kentonv:
Jenkins is down for a while, but you could verify this locally by running the tests in a VM where Sandstorm isn't installed.

zenhack:
I think I'd like to adopt this and try to get it merged. In particular having a test app that uses the bridge will be helpful for testing the stuff I speced out in #3176.

PR #2947, opened 2017-06-10
Honor header whitelist for Powerbox HTTP APIs.

### LINKED ISSUE #2944: X-Sandstorm-* headers don't pass through the proxy of powerbox http request
When I make a http request through powerbox http proxy, almost all headers are blocked including those are prefixed as "X-Sandstorm" or "X-Sandstorm-App". Changing case does not change anything.
We need a way to let customized headers pass through the proxy and reach the target server.

kentonv:
Fixes #2944


kentonv:
Garply, retest this please

PR #2948, opened 2017-06-10
meteor update (to Meteor 1.5)

kentonv:


kentonv:
Will retry supposedly-broken test locally.

PR #2952, opened 2017-06-17
Support ETag/If-None-Match/304 Not Modified through powerbox HTTP APIs.

### LINKED ISSUE #2950: Add support for "If-*" headers to powerbox
Just talked with @kentonv about this.
We'd like to use "If-None-Match" to enable requests that return a 304 response code.
It looks like ETag in the response headers might necessary as well, if it's not there already.

kentonv:
Fixes #2950


kentonv:
Jenkins is down but I ran the tests locally and they passed.

PR #2978, opened 2017-09-03
Fixed forgotten rename in the setup wizard

xarinatan:
Which caused the 'Begin Sandstorm Setup' button to not work after installation. 

Note, there might be more instances of stuff that needs renaming from Identity => Login, although I did a grep -ri on "adminProviderIdentityTable" and this was the only instance I could find.


kentonv:
Thanks for noticing this and alerting me! It turns out there's a bunch of bugs in the setup wizard. I've pushed release 0.214 with commit 42fa4ce to fix them.

PR #2998, opened 2017-10-28
Small ui fixes

m-burg:
While translating I noticed some strings were showing up wrong or twice, I have removed those from the templates.


kentonv:
Thanks!

PR #2999, opened 2017-10-28
Dutch translation

m-burg:
I have translated everything except for the style guide.

When saving your preferences the confirmation still shows up as 'Saved changes', but I don not know enough web development to fix that.

Also some text boxes are a little too small for the words, this is noticeable for instance in the user settings.

### DIFF
--- a/shell/i18n/nl.i18n.json
+++ b/shell/i18n/nl.i18n.json
@@ -1,0 +1,1787 @@
+{
+  "error": "Fout: %s",
+  "shell": {
+    "layout": {
+      "demoExpired": {
+        "title": "Demo verlopen",
+	"canUpgrade": "Bedankt voor het uitproberen van de Sandstorm Demo! Uw demo accaunt is verlopen. Als u Sandstorm wilt blijven gebruiken en uw data permanent op wilt slaan, log dan alstublieft in.",
+        "cannotUpgrade": "Bedankt voor het uitproberen van de Sandstorm Demo! Uw demo accaunt is verlopen.",
+        "signOut": "Uitloggen"
+      },
+      "adminAlert": {
+        "icon": "!",
+        "title": "Bericht van de administrator"
+      },
+      "demoModal": {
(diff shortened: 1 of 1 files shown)


shell/i18n/nl.i18n.json
```diff
@@ -0,0 +1,1787 @@
...
+        "explanation": "Betaalde abonnees kunnen tot 30 GB gratis opslagruimte krijgen, 2 GB per persoon die ze verwijzen, naast de hoeveelheid opslag van hun account niveau. Eigenaren van gratis accounts kunnen een onbeperkte hoeveelheid korrels en 50MB opslagruimte per persoon die ze verwijzen verdienen, tot in totaal 2GB."
+      }
+    },
+    "about": {
+      "title": "Over Sandstorm",
+      "intro": {
+        "explanation": "<a href=\"https://sandstorm.io\">Sandstorm.io</a> is een open source hosting platform voor persoonlijke/privé instanties van web apps. Gebruikers kunnen software uploaden en installeren via een simpele app-store achtige interface. Naast betere privacy <a href=\"https://blog.sandstorm.io/news/2014-07-21-open-source-web-apps-require-federated-hosting.html\">is dit de enige manier om Open Source web apps haalbaar te maken.</a>",
+        "version": "Versie: %s",
+        "unofficialWarning": "<b>WAARSCHUWING:</b> deze server draait een onofficiële build van Sandstorm die aanpassingen kan bevatten die niet zijn gecontroleerd door de ontwikkelaars."
```
zenhack: It's in the existing localizations too, so orthoganal to this pr, but we should probably not be using the `<b>` tag. @kentonv, maybe we should change this to `<strong>`?

kentonv:
Awesome, thanks!

shell/i18n/nl.i18n.json
kentonv: I don't have a strong opinion (... so to speak), but would accept a pull request.

m-burg: I'll get on it

PR #3000, opened 2017-10-28
Add vim swap files to .gitignore

zenhack:


kentonv:
Hmm, but doesn't the presence of a swap file hint that maybe you forgot to save some of your work, which is something you'd want to know when you type `git status`?

kentonv:
BTW congrats on getting PR #3000.

zenhack:
Yay round numbers.

The way I tend to work, it's not uncommon to leave an editor window open even if the file is in a stable state; I'll often have a separate terminal open to run things like make, etc. The problem you're describing seems plausible a-priori, but at least personally I don't think I've ever actually had that happen in practice. I'm also in the habit of saving very frequently though, as it triggers plugins that run pretty-printers, various linters, etc.

zenhack:
...I *have* however accidentally commited swap files in the past.

zenhack:
I think I'll go ahead and close this; it's cluttering up my list of open PRs, and it seems like we disagree on the issue (and it isn't terribly important to me).

PR #3002, opened 2017-10-29
Update zh-TW.i18n.json

audreyt:
Change &mdash; into unicode character; otherwise it's displayed doubly-escaped


kentonv:
Thanks!

PR #3003, opened 2017-10-29
Ask for help on improving translations

m-burg:
People might find things that are wrongly translated or could be worded better. I think it is good to be more explicit on what to do in that scenario.


ocdtrekkie:
FWIW, I am not sure how responsive an issue/discussion on a translation would be. It's likely one person who did the prior translation which may or may not be currently responsive. I think a better piece of advice would be to just file a PR, and tag the GitHub user who did the original translation so they are notified, and can comment if they'd like to.

PR #3007, opened 2017-11-13
Handle HTTP requests in C++

kentonv:
This adds a new C++ component called the "gateway". All HTTP requests first hit the gateway, where they are then forwarded to other components. By default, requests are forwarded to the shell (the Meteor / Node.js process), but requests to grain hosts can be sent directly to the corresponding grain, skipping the shell.

The hope is that this will solve the memory leaks we've observed on Oasis, by offloading a lot of work from the Node.js process. Maybe it will improve performance, too.

As a side note, this change eliminates `run-dev.sh` in favor of a new `sandstorm dev-shell` command which actually allows the dev version of the shell to use the listen socket opened by the main server, rather than force the main server to close its socket so that the dev shell can open its own.

This change is far from done. TODO list:
- [x] Purge old session ID -> WebSession mappings from gateway.
- [x] Revoke WebSession capabilities when user loses access.
- [ ] Optimize HTTP requests for more than just UI sessions. Currently these all just get forwarded to the shell.
    - [ ] API sessions
    - [ ] Static asset hosting (e.g. avatars; `static.oasis.sandstorm.io`)
    - [ ] Static web hosting (i.e. what the Wordpress app uses)
        - [x] Public ID hostnames (randomly-generated)
        - [ ] Foreign hostnames (via CNAME and TXT)
- [x] Support Sandcats HTTPS.
- [x] Support SMTP including STARTTLS.
- [x] Bug fixes.
    - [x] Fix powerbox test failures.
    - [x] Run whole test suite with gateway enabled.
    - [x] Don't let gateway connect until migrations are done.
    - [x] Redirect HTTP ports to HTTPS where appropriate.
    - [x] Build BoringSSL in release mode as appropriate.
    - [x] Fix problem where connections sometimes fail after upgrade.
- [ ] In Blackrock, make sure we can actually replace nginx entirely.
    - [x] Support configurable HTTPS keys / certs.
    - [x] Support serving /terms and /privacy by forwarding to static content from a grain.
    - [x] Provision gateway machine with static IP.
    - [ ] Use [Cloudflare Workers](https://blog.cloudflare.com/introducing-cloudflare-workers/) for maintenance notice, since the gateway will go down during upgrades.
- [ ] Push a release where this is available but off by default, then let people try turning it on.
- [ ] Push a release where it's on by default.

Blackrock bits in: https://github.com/sandstorm-io/blackrock/pull/47


kentonv:
LEEEEEEROYYYYYYYYYYYYYYYYYYYYYYYYYYYYY

PR #3013, opened 2017-12-02
fix github login setup url in zh-CN and zh-TW

peterlee0127:
fix github login setup url.
https://github.com/settings/applications/newi -> https://github.com/settings/applications/new


kentonv:
Thanks!

PR #3014, opened 2017-12-22
Fix install script for less frequent releases

ocdtrekkie:
Reported via IRC. Sandstorm builds used to reliably occur weekly, so a month old installer was not surprising. As Sandstorm builds now happen infrequently, this causes a failure. As the tolerance here was four times the frequency of the average build, and we still regularly have new builds roughly monthly, I propose changing this to 120 days, or four times the current release cadence.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1372,9 +1372,9 @@ __EOF__
     fi
 
     if [ ! -e "$BUILD_DIR/buildstamp" ] || \
-       [ $(stat -c %Y "$BUILD_DIR/buildstamp") -lt $(( $(date +%s) - 30*24*60*60 )) ]; then
+       [ $(stat -c %Y "$BUILD_DIR/buildstamp") -lt $(( $(date +%s) - 120*24*60*60 )) ]; then
       rm -rf "$BUILD_DIR"
-      fail "E_PKG_STALE" "The downloaded package seems to be more than a month old. Please verify that your" \
+      fail "E_PKG_STALE" "The downloaded package seems to be more than 120 days old. Please verify that your" \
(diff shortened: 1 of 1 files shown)


kentonv:
So, the 30-day limit is actually a thing that pushes me to do releases. When you see the releases in the changelog that are like "nothing changed except dependencies", that's a release I did because I knew the limit was coming up.

Forcing me to do releases every month is probably a good thing for security hygiene, so I actually think we should keep it for now. I'll do a release today.

ocdtrekkie:
Will this be a problem for people who may go back one build for troubleshooting as well though? Not sure the workflow people use for that.



I did see one dependencies-only release and was guessing this might have been why.

kentonv:
I think the code in question only applies when auto-downloading the latest package. If you provide an already-downloaded tarball, it doesn't do the check. And if you do `sudo sandstorm update <tarball>` that definitely doesn't do the check.

dckc:
I just got the "more than a month old" error; I'm trying to set up a server to transfer from oasis. I see the latest release is Nov 23.

ocdtrekkie:
Being Christmas Day, I suspect a release won't happen today, @dckc 

If you can modify the install.sh file and run it as demonstrated here, you should be good to go for now.

dckc:
Right; I tweaked the install script using the clue here. I just figured I would nudge things along in hopes of saving the next person some trouble.

Merry Christmas!

kentonv:
Sorry about that! I thought I had another week, oops. Just pushed an update, should be fixed now.

Annoyingly, updating meteor broke tests, so I had to revert. Will have to look into it later, too busy moving...

PR #2985, opened 2017-09-27
i18n implementation and Chinese (Traditional/Simplified) translation

urakagi:
This pull request includes `tap-i18n` framework implementation and Chinese translation.
Please note that opposite to the guideline, language is auto-selected by browser's `navigator.language` in this PR, so users cannot set their language manually, and `Accept-Language` header is ignored. Also, `capnp` related texts are not touched. This will be left for further work.

PR #3015, opened 2017-12-27
Add Finnish translation to Sandstorm

xet7:
Added Finnish translation to Sandstorm.

### DIFF
--- a/shell/i18n/fi.i18n.json
+++ b/shell/i18n/fi.i18n.json
@@ -1,0 +1,1786 @@
+{
+  "error": "Virhe: %s",
+  "shell": {
+    "layout": {
+      "demoExpired": {
+        "title": "Demo vanhentunut",
+        "canUpgrade": "Kiitos kun kokelit Sandstorm Demoa! Demo tilisi on vanhentunut. Jos haluat jatkaa Sandstormin käyttämästä ja pitää tietosi tallessa, ole hyvä ja kirjaudu sisään.",
+        "cannotUpgrade": "Kiitos kun kokelit Sandstorm Demoa! Demo tilisi on vanhentunut.",
+        "signOut": "Kirjaudu ulos"
+      },
+      "adminAlert": {
+        "icon": "!",
+        "title": "Viesti adminilta"
+      },
+      "demoModal": {
+        "header": "Sandstorm on web-tuottavuusohjelmisto jonka voi asentaa omalle palvelimelle.",
+        "withAppDemo": "ohjelmien kuten %s kanssa.",
+        "withoutAppDemo": "kaikki tarvitsemasi ohjelmat samassa paikassa.",
+        "startButton": {
+          "withAppDemo": "Kokeile %s »",
+          "withoutAppDemo": "Käynnistä »"
+        },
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
i,i it's awesome to see another localization for Sandstorm.

kentonv:
Nice!

My grandparents on my mother's side immigrated from Finland. Unfortunately I don't know the language myself, but I'm still extra-happy to see this.

xet7:
@kentonv 

How can I test this? Will it be included to Sandstorm release sometime?

kentonv:
@xet7 To test now you'll need to build from source:

https://docs.sandstorm.io/en/latest/install/#option-4-installing-from-source

It would be great if you could do that to verify your translation before it's released.

I will probably do the next release around Jan 6, or maybe Jan 13.

xet7:
@kentonv 

Ok I will install from source and test it.

xet7:
@kentonv 

I did compile Sandstorm on Pop!_OS 64bit that is based on Ubuntu 17.10.

I will later send fixes to this Finnish translation, to use better words.

Not everything is translated, for example when installing Wekan some popups/tooltips were in English. Are there translations also in other places? Is App Market content also translatable, where new apps are installed from? Is it possible to translate Sandstorm website?

ocdtrekkie:
The Sandstorm App Market should be pretty trivial to translate. Ideally for simplicity we'd want to use the same i18n package on this Meteor app: https://github.com/sandstorm-io/sandstorm-app-market

The App Market has not a whole ton of text of it's own, so I shouldn't think this would be a huge project. Most of the issue is that the App Market's apps all have text, which currently doesn't localize, even though there exists a design to do so.

If you look at Wekan's pkgdef, Maxime had put in some of the French translations, including a very outdated changelog in French: https://github.com/wekan/wekan/blob/devel/sandstorm-pkgdef.capnp and https://github.com/wekan/wekan/blob/devel/meta/t9n-changelog/fr.md

ocdtrekkie:
This is the PR that added tap:i18n to Sandstorm itself: https://github.com/sandstorm-io/sandstorm/pull/2985 and the first commit is really, most of the change from a technical standpoint. The rest of that giant PR is string replacement.

ocdtrekkie:
Note that translating the main sandstorm.io website would be very much not even remotely trivial. The sandstorm.io website is mostly static html, and many of the images contain text.

kentonv:
@xet7 Sandstorm's i18n is definitely incomplete. The i18n hooks were added in #2985, but not everything was hooked. I'm happy to accept PRs to add more hooks.

I'd also be happy for someone to take over the app market and add i18n hooks there, though as Jake points out the app descriptions come out of the app packages and so won't be translated unless the package maintainer adds translations. There is indeed support in the metadata structures for specifying translations, although I suspect the way it's done currently isn't very maintainable -- it's organized by message rather than by language, which probably means it'll be a mess for anyone supporting more than a couple languages. We should probably develop a better way for apps to provide separate translation tables, but that's a big project.

Regarding the Sandstorm web site, I'd rather not try to translate it currently, as doing so would make it pretty difficult for me to make changes to the site going forward -- and the site is currently in need of an overhaul to make it less commercial.

ocdtrekkie:
I'm gonna tag @caasi and @urakagi in here since they implemented localization of Sandstorm itself. Either or both of them might be interested in implementing the same for the app market, perhaps?

urakagi:
I'm not sure about the structure of the app market. Is it as same as Sandstorm shell so add tap-18n and modifying templates will do the work?

ocdtrekkie:
Should be more or less the same thing, yeah. AFAIK, there's also massively less text in the app market. Most of the text in it comes from the app packages themselves.

PR #3033, opened 2018-02-06
Added support for OpenRC in the installer script. (3030)

lramage94:
This change has been tested on the following systems,
- alpine:latest
- gentoo:stage3-amd64
- ubuntu:14.04

Signed-off-by: Lucas Ramage <ramage.lucas@openmailbox.org>


lramage94:
I don't know what's wrong with me today. I'm having issues attaching the correct issue.

Let's try here.

#3033

kentonv:
Hi @lramage94,

Thanks for working on this! Sorry for the delay in responding; I only work on Sandstorm on weekends and sometimes I forget to check e-mail. :/

If I understand correctly, the main difference between openrc vs. sysvinit is that instead of:

    update-rc.d sandstorm defaults

we have to do:

    rc-update add sandstorm

Perhaps, then, a much simpler change to the install script would be to replace this line:

    update-rc.d sandstorm defaults

With:

    if [ "$(which update-rc.d)" != "" ] then
      update-rc.d sandstorm defaults
    elif [ "$(which rc-update)" != "" ] then
      rc-update add sandstorm
    else
      echo "WARNING: I couldn't figure out how to make the Sandstorm init script active on" >&2
      echo "  your system; neither update-rc.d nor rc-update commands seem to exist. Sandstorm" >&2
      echo "  will not start automatically at boot until you mark its init script active." >&2
    fi

The nice thing about doing the change this way is that we can be pretty confident that no new bugs have been introduced and do not need to do much testing.

What do you think?

lramage94:
Sure I can do that.

lramage94:
Ping @kentonv

kentonv:
Sorry this fell off the radar... I haven't been keeping up on Sandstorm enough lately. :/

Thanks, this change looks good.

PR #3038, opened 2018-02-19
Delete proxy.js and most of pre-meteor.js in favor of Gateway

kentonv:
The magic is in commit 4edb2ae.

The rest of this PR fills out the remaining missing pieces needed for proxy.js to really be able to go away.

proxy.js is pretty much the oldest code in Sandstorm, and some of the ugliest. I'm so pleased to have it go away.


zookozcash:
Good work. :-)

Issue #3044, opened 2018-02-22
No way to change the interface language


jidanni:
User cannot find a way to make the particular account he is using become English.

jidanni:
Signed out it is English, signed in, it is Chinese. Maybe because my name is half Chinese characters. Anyway my top browser language is English.

xet7:
@jidanni 

Sandstorm uses same language that user's webbrowser has as top browser language in browser settings.

jidanni:
@xet7 It has dug down to my second language!!
![0](https://user-images.githubusercontent.com/1230959/36542598-39fffd0a-181c-11e8-81e0-b58f673eac62.jpg)

jidanni:
https://storm.debian.net/grain/uNF6W7gzmxGQHshgb5yCBS

xet7:
@jidanni 

Current translations are here:
https://github.com/sandstorm-io/sandstorm/tree/master/shell/i18n

Not everything is translated yet, and not everything is added to be translatable at all.

Please remove all other languages from your browser, only leave Chinese.

xet7:
@jidanni

If you get some of text in Chinese after logging in, translations are already working correctly. Not everything is in Chinese yet.

jidanni:
Here we see the left menu is English, and the right smaller menu is Chinese.
I want to make both of them English.
![1](https://user-images.githubusercontent.com/1230959/36542872-0fe08ae8-181d-11e8-90f9-f3ed1cb21532.jpg)

ocdtrekkie:
Definitely strange, Sandstorm doesn't do a lot to pick languages, so usually if anything, people get stuck with English. I think adding a manual override setting to set a Sandstorm user's language has been talked about before. I wonder how its picking when it sees multiple in your browser, and if other software using the same localization package has the same problem.

jidanni:
All I know is all the big website always allow the user to override whatever the browser says anyway.
One can never guess all the types of users, and one doesn't want them to have to dig into their browser settings due to being a multicultural person.

xet7:
@jidanni 

In that screenshot at top Sandstorm is translated to Chinese.

But below it you seem to be running EtherPad, that is installed to Sandstorm from Sandstorm App Market https://apps.sandstorm.io/ .

For that EtherPad app, direct link is:
https://apps.sandstorm.io/app/h37dm17aa89yrd8zuqpdn36p6zntumtv08fjpu8a8zrte7q1cn60

[...]

PR #3046, opened 2018-02-27
Make en-US uses en too

urakagi:
Fix issue #3044.


kentonv:
Thanks!

urakagi:
It seems like I broke the feature, my computer does not show Chinese now. I'll find what's wrong.

Issue #3047, opened 2018-02-27
Outdated docs for hacking on sandstorm shell


dmitry-sher:
## UPDATE
Easiest way to hack on sandstorm shell is as follows:

 - install latest VirtualBox (5.2.6 is good)
 - import VM, made by @xet7 . read #3036 for details. [VM link](https://releases.wekan.team/Wekan-Sandstorm-0.77.ova)
 - set up VM's networking as appropriate (i prefer NAT with some port like host's 4222 as guest 22 and host's 127.0.0.1:6080 -> guest's 0.0.0.0:6080). ssh to VM with wekan:wekan `ssh -p 4222 wekan@localhost`. You can add your keys `ssh-copy-id -p 4222 wekan@localhost`
 - update `/opt/sandstorm/sandstorm.conf` in VM. set `BIND_IP=0.0.0.0`. comment out `#UPDATE_CHANNEL=dev`. restart sandstorm `sudo service sandstorm restart`. Now you can access sandstorm from host OS on port 6080
 - add wekan user to sandstorm group. `sudo usermod -a -G sandstorm wekan`. logout and login back for changes to take place.
 - go to `~/repos/sandstorm` and `make update`
 - run `sandstorm dev-shell` from the repo root. 

## Below are my previous attempts to do it on Ubuntu 16.04.3. You can safely skip it.

1. The commands

```
cd shell
sudo service sandstorm stop-fe
./run-dev.sh
```
Are not working anymore. Current command is `sandstorm dev-shell` in project root.

If you run it, you would see a multitude of problems.

1a. Preparations to run hacking.

```
cd shell
meteor npm install
```

2. When trying to run `sandstorm dev-shell`. From the latest master commit 760faa2452dcd3e9f76544ab527a630aa0210ef6

```
=> Errors prevented startup:                  
   
   While processing files with fourseven:scss (for target web.browser):
   /client/styles/shell.scss: Scss compiler error: File to import:
   {}/client/styles/_icons.scss not found in file:
   /home/sherd/sandstorm/shell/{}/client/styles/shell.scss
```

As you can see, the file is missing in the sources:
https://github.com/sandstorm-io/sandstorm/tree/master/shell/client/styles.
It's generated. See how to generate it below.

3. You would miss capnp.node and capnp.js. You can find them in `/opt/sandstorm/latest/node_modules`. Copy everything from that folder to `shell/node_modules`.

[...]

Issue #2464, opened 2016-08-28
Support range requests


mrdomino:
This would be nice for a stupid project of mine. I bet it'd also make Groovebasin song previews seekable. See also #1499

rchrd2:
👍 I second this request. Also for an mp3-based app.

Unlike #1499, my app isn't using the special static hosting (though it could if necessary (just have to learn how to use it)).

I wonder if it's really just a matter of passing through the headers to the app, and then it's up to the app to support the range request. If so, then seems like low hanging fruit.

rchrd2:
@kentonv Is this as simple as passing the headers through and letting the app's http server handle those headers?

Michael-S:
@rchrd2   the Groove Basin app in Sandstorm for music hosting is really good, and supports mp3s.  I've been using it off my personal Sandstorm installation for over a year.

kentonv:
@rchrd2 Almost. I only want to use the header whitelist for things that are very application-specific. Range requests are a standard part of HTTP, so should be handled by extending the HTTP-over-Cap'n-Proto structures to represent them.

As it happens I actually have a PR open proposing such protocol changes: https://github.com/sandstorm-io/sandstorm/pull/2887

But the code hasn't been updated to implement the protocol.

PR #3048, opened 2018-03-02
Clang 6 is now required, minor docs update.

Michael-S:
I got that from the change in Makefile,
commit 01d7cc9fca808db97c70cb5f2d9c54b18bccc8c2

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -142,7 +142,7 @@ Please install the following:
 * `golang-go`
 * `cmake`
 * discount (markdown parser)
-* [Clang compiler](http://clang.llvm.org/) version 3.4 or better
+* [Clang compiler](http://clang.llvm.org/) version 6 or better
 * [Meteor](http://meteor.com)
 
 On Debian or Ubuntu, you should be able to get all these with:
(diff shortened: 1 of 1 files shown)


docs/install.md
```diff
@@ -152,6 +152,9 @@ On Debian or Ubuntu, you should be able to get all these with:
         golang-go
     curl https://install.meteor.com/ | sh
 
+You may need to do some extra work to get Clang 6, Debian and
+Ubuntu derivatives can use https://apt.llvm.org/
```
zenhack: It might be better to (a) put this before the commands above; users are less likely to install the old version and only then find out that they need a different one, and (b) mention which versions of ubuntu/debian have a new enough version, and which need to use the llvm project's repos.

xet7:
I did list Sandstorm deps at https://github.com/sandstorm-io/sandstorm/issues/3047#issuecomment-369002784 and there is VirtualBox appliance at https://github.com/wekan/wekan/releases . It would be nice to have all deps included in commands.

docs/install.md
Michael-S: Good points.  I'll do a little research and resubmit.

Michael-S:
Hello @xet7  - I did see that you listed it in that issue.  That's what inspired me to update the docs.  I'm trying to make an attempt at #2464 but I need my build working first.

Michael-S:
Per my comment above, I'm going to try again.  Thanks for the feedback, @zenhack  and @xet7

PR #3049, opened 2018-03-03
Clang v6 doc tweak

Michael-S:
I'm not sure quite how much detail we want to put about adding a new software package repository to a Debian or Ubuntu installation.  I gave a reference to the page for obtaining Clang 6 builds for Debian and Ubuntu, but not the full details for adding them.

And of course this covers the Clang dependency bump to version 6.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -145,15 +145,20 @@ Please install the following:
 * [Clang compiler](http://clang.llvm.org/) version 6 or better
 * [Meteor](http://meteor.com)
 
-On Debian or Ubuntu, you should be able to get all these with:
+Debian 10 and Ubuntu 18.04 are planned to have Clang 6 available
+in the default software package repositories. Until then,
+you can configure your Debian or Ubuntu installation to
+obtain the the software packages from the information on
+https://apt.llvm.org/
+
+On Debian or Ubuntu, you should be able to get all the remaining
+dependencies other than Clang 6 with:
 
     sudo apt-get install build-essential libcap-dev xz-utils zip \
-        unzip strace curl clang discount git python zlib1g-dev \
+        unzip strace curl discount git python zlib1g-dev \
         golang-go
     curl https://install.meteor.com/ | sh
 
-You may need to do some extra work to get Clang 6, Debian and
-Ubuntu derivatives can use https://apt.llvm.org/
 
 ### Get the source code
 
--- a/docs/install.md
+++ b/docs/install.md
@@ -142,7 +142,7 @@ Please install the following:
 * `golang-go`
 * `cmake`
 * discount (markdown parser)
-* [Clang compiler](http://clang.llvm.org/) version 3.4 or better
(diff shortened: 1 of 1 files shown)


zenhack:
Yeah that's pretty much what I had in mind.

FYI, you don't need to open a new PR every time you change things; you can just push more commits to the same branch and it will update the old one.

Michael-S:
Thanks.  I didn't know.

kentonv:
Hmm, is clang 5 readily available on Debian / Ubuntu? It might be good enough. I went straight from 4 to 6 without checking 5 because I was trying to solve an issue and I saw that I already had clang 6 installed.

Michael-S:
I'm just curious, do you remember how you installed clang-6?  Using the site I linked?  Built from source?  Something else?

I'm spoiled by a lot of other software I've installed on Linux, usually the instructions are "at a prompt, type or paste: ....   " and everything just works.  You know, like "curl https://install.sandstorm.io | bash"  

So I'm surprised that for newer versions of LLVM/Clang the user is expected to know how to modify their sources files before running the install. 

I'll try building with Clang 5 and see what happens.  Is the Makefile the only thing I'll need to change?

xet7:
Using newer clang on Ubuntu requires using PPA, and that was broken last time it was tried at https://github.com/sandstorm-io/sandstorm/issues/3047#issuecomment-368884767

I also did try to compile Sandstorm on Ubuntu 14.04 64bit, but failed.

I only got Sandstorm compiled on Debian 10 Buster where all clang versions 4, 5 and 6 are directly available at .deb package repos.

xet7:
When compiling at Debian 10, I did not make any changes to Makefile.

kentonv:
@Michael-S It looks like Debian Sid (aka Unstable) has clang-5.0 and clang-6.0 packages in the main apt repository, and that's what I'm using. I notice that clang 6 is a release candidate version so I should probably back up to clang 5. Running a test build right now, seems to work.

kentonv:
BTW @Michael-S, you'd be surprised how many people insist, _angrily_, that Sandstorm's install process should be something like LLVM's, not `curl|bash`. :)

kentonv:
OK I changed the Makefile to clang 5 instead of clang 6. It would still be very helpful if people could figure out and document what's necessary to get clang 5 installed on release versions of Debian and Ubuntu.

Michael-S:
If a package is in the _default_ software repositories, then of course sudo apt install is better than curl | bash because (hopefully) the project engineers have vetted the software in the default repositories.

But otherwise I agree, curl | bash is equally as risky as using an untrusted repo and a lot less effort, so why not?   And to be clear, I read the discussion on why Sandstorm iterates too rapidly for it to be practical to include it in default repositories of any Linux distribution with a fixed release cycle, and I agree with the reasoning.

Michael-S:
This has been resolved in the interim.

PR #3053, opened 2018-03-06
Added a DNS note.

Michael-S:
I set up Sandstorm on a VM for testing, and the local.sandstorm.io
address would not work for me.  It took me a few hours to track
down the problem, and it would have been longer if not for
TimMc on IRC.  My DD-WRT router was running dnsmasq and
blocking the resolution to 127.0.0.1.  So I've added a note
in the hope that nobody else stumbles on the same problem.

### DIFF
--- a/docs/administering/wildcard.md
+++ b/docs/administering/wildcard.md
@@ -87,6 +87,12 @@ company behind Sandstorm) maintains `local.sandstorm.io` as a wildcard domain wh
 same as `localhost`. This allows you to run Sandstorm for development without needing to own a
 domain name or configure wildcard DNS for a subdomain.
 
+Note that sometimes DNS in routers and firewalls such as dnsmasq in DD-WRT can block the
+local.sandstorm.io name resolution to `127.0.0.1`.  In that case, for the machine developing
+Sandstorm please switch your DNS from using your router or firewall as a DNS server to using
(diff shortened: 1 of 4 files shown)


Michael-S:
But if this kind of headache is too obscure to worry about, I'm not offended if the request is closed. It only took ten minutes to throw together.

kentonv:
Hmm, maybe we should actually start recommending using `*.localhost`, which Chrome supports and I think other browsers do too?

publicarray:
This could be a security feature to protect against [DNS Rebinding](https://en.wikipedia.org/wiki/DNS_rebinding). Ideally these DNS requests do not go outside the local network. (could be added to a hostfile)

Yes localhost. is correct, see: https://tools.ietf.org/html/rfc6761

This is interesting (by Tavis Ormandy): https://www.securityfocus.com/archive/1/486606/30/0/threaded

ocdtrekkie:
My only real question here is if this note should be duplicated in four places. I think it likely is an edge case, and while it should be documented, I am not sure it needs to be repeatedly so. Especially with the updated docs making search functional again.

ocdtrekkie:
I'm going to close this one. I'm not sure we've seen this issue about, I don't really like recommending people use Google DNS, and I'm not fond of placing this particular advice in four places, as noted. I think figuring out how to handle your own networking is probably a bit out of scope for our docs anyways.

PR #3056, opened 2018-03-10
Fix language detect fail bug

urakagi:
I don't know the problem is return or indexOf, but anyway I change the code more similar to the original one. I still can't build up latest sandstorm :/

### DIFF
--- a/shell/client/shell-client.js
+++ b/shell/client/shell-client.js
@@ -40,18 +40,14 @@ if (Meteor.isClient) {
 
     if (navigator.languages) {
       navigator.languages.forEach(lang => {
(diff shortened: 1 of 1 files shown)


kentonv:
The problem was that `langMap` doesn't have an `indexOf()` method, since `langMap` is an object, not an array.

I'll merge this and make sure it's actually fixed. In the future, please be sure to let me know if a PR is untested -- when I merged the previous PR I assumed you had tested it so I didn't try it myself.

urakagi:
@kentonv Sorry for not mentioning untested. Confirmed this PR fixes the problem. Thank you!

PR #3057, opened 2018-03-15
Docs: Update PHP version in example

ocdtrekkie:
As per sandstorm-io/vagrant-spk#211. Also MkDocs isn't making these repo URLs into links, so I'm trying to force that.


kentonv:
Thanks!

PR #3060, opened 2018-03-17
Updated install docs for Fedora 27 packages

Michael-S:
These are all of the packages I needed to install for
a successful build of Sandstorm as of early March, 2018.


kentonv:
Thanks!

PR #3077, opened 2018-06-04
Support GitLab and Bitbucket with Bug Report link

ocdtrekkie:
I haven't tested this, but it's fairly simple code. I believe all my parenthesis are correct.

I am baffled why this doesn't use `startsWith` like the app market in https://github.com/sandstorm-io/sandstorm-app-market/pull/134


kentonv:
I think `startsWith()` was introduced in ES6. The style here was idiomatic before ES6. So probably either this was written before we started using ES6, or the developer who wrote it (quite possibly me) didn't know / forgot about ES6 `startsWith()`.

PR #3091, opened 2018-09-28
Add headers module enable comment

burkemw3:


troyjfarrell:
Thanks for this.  I think that @burkemw3 is trying to make [apache-virtualhost.conf](https://github.com/sandstorm-io/sandstorm/blob/master/docs/administering/sample-config/apache-virtualhost.conf) more internally consistent.  Other sections contain text like this change:
```
#    # If this request did not want a websocket, then give it http:// access to Sandstorm. To make this
#    # work, you might need to run:
#    #
#    # sudo a2enmod proxy_http
```

ocdtrekkie:
@troyjfarrell Indeed. Though I think if we are going to merge this, we'd want it to explain itself, as the other sections do, specifying why it is there and what it does. Otherwise, there is just a comment that maybe looks like an executable command. Also, whitespace from the section above would be nice.

ocdtrekkie:
This file was already updated by #3164 which had substantial technical changes, this PR was never really explained and had no practical impact anyways, as far as I can tell. Closing it.

PR #3136, opened 2019-05-04
Propagate error on LDAPS SSL cert verification failure

### LINKED ISSUE #3012: LDAP login no longer working
Sandstorm Version: 0.221
Operating System: CentOS 7

Trying to login with LDAP and after putting in a username and password and clicking LOG IN, it simply does nothing. /opt/sandstorm/var/log/sandstorm.log doesn't show an error. Effects every user on different PCs. It stopped working within the last couple weeks.

precisionconage:
Fixes #3012.

The problem was that failing to verify the cert emits an `error` rather than calling the error callback, so `errorFunc` would catch the error and `ldapBindFut` would never return/throw.

### DIFF
--- a/shell/imports/server/accounts/ldap.js
+++ b/shell/imports/server/accounts/ldap.js
@@ -89,8 +89,6 @@ LDAP.prototype.ldapCheck = function (db, options) {
       }, errorFunc);
     }
 
-    client.on("error", errorFunc);
-
     let username = options.username;
     let domain = _this.options.defaultDomain;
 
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
Thanks for this! Silent failures are always a little unfortunate for troubleshooting.

shell/imports/server/accounts/ldap.js
```diff
@@ -111,22 +109,25 @@ LDAP.prototype.ldapCheck = function (db, options) {
...
+      client.bind(_this.options.searchBindDn, _this.options.searchBindPassword, searchBindErrorFunc);
 
       try {
         ldapBindFut.wait();
       } catch (err) {
         return { error: err };
       }
     }
+
+    client.on("error", errorFunc);
```
kentonv: Is it intended that we may end up registering multiple on-error handlers? What is the effect of running both? You moved this line down here, so I guess ordering is important? Might be worth a comment explaining that this is expected, otherwise it looks like it could be a mistake.

PR #3152, opened 2019-09-02
Implement Mass Transfers.

kentonv:
This makes it easy to mass-transfer grains to another Sandstorm server. A UI lets you choose which grains to transfer, watch the transfer happen, and check and retry any errors.


ocdtrekkie:
This is very exciting functionality to have! ...Though I'm a bit nervous about it being implemented.

PR #3157, opened 2019-09-12
docs: mention OS environment, Debian Stretch

curbengh:
https://github.com/sandstorm-io/vagrant-spk/blob/4d16fb17583c946753e49b1877d41e5e2d18b2dd/vagrant-spk#L73


ocdtrekkie:
I was trying to see if there was a better place in the Docs to mention this, but there really isn't. This seems like a fine place to note it to me.

https://docs.sandstorm.io/en/latest/vagrant-spk/design/ might be a good spot, but it's kinda a historic text, rather than a current document.

https://docs.sandstorm.io/en/latest/vagrant-spk/customizing/ might be a good spot, although as noted in our discussion, it's not something easy to "customize" or change, but from the title "Understanding" and customizing vagrant-spk, that explains how vagrant-spk works, it might be an alright place.

curbengh:
What about https://docs.sandstorm.io/en/latest/vagrant-spk/installation/?

> In this tutorial, we use VirtualBox to emulate a computer on which we run Debian 9 (Stretch), controlled via Vagrant.

Perhaps replace "In this tutorial," with "By default,"?

ocdtrekkie:
"By default" would suggest you can change it. And in the case of vagrant-spk, you really can't. Though your Node stack changes are release-independent, so if the other stacks were updated similarly, it might be a bit easier in the future to change platforms.

curbengh:
> "By default" would suggest you can change it. And in the case of vagrant-spk, you really can't.
yep, that's what I thought too; what about removing "In this tutorial,"? It sort of imply that Debian is only used in the tutorial, despite being the only option (unless user fork the entire vagrant-spk).

What do you think of mentioning Debian in the Installation tutorial instead, I'm assuming most users read that before "Platform Stacks" page.

ocdtrekkie:
I am kinda thinking in the Overview of https://docs.sandstorm.io/en/latest/vagrant-spk/customizing/ is the best place, as it goes into detail about what vagrant-spk does and how it works, particularly in the section about what different files in it are for.

I don't think the intention was for users to need to know or care what vagrant-spk was running in the background (as in many cases, vagrant-spk users may be Mac or Windows users who are not super opinionated about Linux distros).

curbengh:
> I don't think the intention was for users to need to know or care what vagrant-spk was running in the background (as in many cases, vagrant-spk users may be Mac or Windows users who are not super opinionated about Linux distros).

If users are fine with the default stacks, indeed there's no need to know about the distro. But it's essential for `diy` stack, i.e. adding 3rd-party apt repo, so they need to know which distro and its version is used.

PR #3158, opened 2019-09-12
docs(nginx-example): update ssl config

curbengh:
Prioritize newer recommendations and mention older ones if the platform uses older nginx+openssl.

Changes:
- Security:
  - Disable `ssl_session_tickets`
  - [`ssl_dhparam`](https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_dhparam) to support EDH ciphers.
  - `ssl_ciphers` always use GCM mode for AES
  - Increase `Strict-Transport-Security` expiry to 2 years (from 1 year).
  - Retain `ssl_session_timeout 5m`; Mozilla [recommends](https://github.com/mozilla/server-side-tls/issues/198) 1d for performance and comes with security tradeoff.

- Performance:
  - Enable http/2
  - Enable `ssl_session_cache`

- Misc:
  - Remove [redundant](https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl) `ssl on`
  - Disable `ssl_prefer_server_ciphers`. Not exactly security related, but it's more of respecting browser's choice, e.g. Chrome prefers `CHACHA20-POLY1305`.
  - [OCSP stapling](https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_stapling) is mentioned but not enabled by default as it requires intermediate cert, which is not always present.

Ref: https://nginx.org/en/docs/http/ngx_http_ssl_module.html

### DIFF
--- a/docs/administering/sample-config/nginx-example.conf
+++ b/docs/administering/sample-config/nginx-example.conf
@@ -47,6 +47,13 @@ server {
   # future then you should probably delete this line.
   add_header Strict-Transport-Security "max-age=63072000" always;
 
+  # OCSP stapling, requires Intermediate cert
+  #ssl_stapling on;
+  #ssl_stapling_verify on;
+
+  # verify chain of trust of OCSP response using Root CA and Intermediate certs
+  #ssl_trusted_certificate /path/to/root_CA_cert_plus_intermediates;
+
   location / {
     proxy_pass http://127.0.0.1:6080;
 
--- a/docs/administering/sample-config/nginx-example.conf
+++ b/docs/administering/sample-config/nginx-example.conf
@@ -27,7 +27,6 @@ server {
   listen 443 ssl http2;
   server_name example.com *.example.com;
 
-  ssl on;
   ssl_certificate /etc/nginx/ssl/sandstorm.crt;
   ssl_certificate_key /etc/nginx/ssl/sandstorm.key;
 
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
Note that while the vagrant-spk tool uses Debian Stretch, Sandstorm is not targeted explicitly towards it, and Sandstorm *should* be runnable on current, well-supported stable versions of most flavors of Linux.

The Nginx config example in the Docs should be able to safely include newer recommendations, though if it's so new a user might be likely to run into problems depending on their choice of distro, you may wish to note it in a comment.

curbengh:
Noted, I'll prioritize newer recommendations and mention older ones if the platform uses older nginx+openssl.

garrison:
Have you looked at the results of running https://www.ssllabs.com/ssltest/ against this example configuration, by chance?  Same question for #3164.

ocdtrekkie:
My assumption is that replacing `  ssl_protocols TLSv1 TLSv1.1 TLSv1.2;` with `  ssl_protocols TLSv1.2 TLSv1.3;` nearly guarantees an at minimum *better* Qualys test. But I don't have the time or spare hardware to spin up tests for this. It's plausible @curbengh has a Sandstorm server they could test against for at least one of these, but they haven't responded to your comment in a couple weeks so they may not be listening. :(

PR #3162, opened 2019-09-16
Adapt sync-langs.py for Python 3

troyjfarrell:
With these changes, the script runs under Python >=2.7. With the sunsetting of Python 2, this seems like a good idea.


ocdtrekkie:
Seems like a good idea, especially considering it's a pretty small change.

Issue #3027, opened 2018-01-22
Provide a mechanism for apps to export capabilities while using sandstorm-http-bridge


zenhack:
Right now afaik there is no way to export (persistent) objects via powerbox requests/offers from an app which is using sandstorm-http-bridge.

In terms of api, I *think* the only piece that's missing is a way to restore/load the objects. One approach is to add a method to [SandstormHttpBridge][1] to supply an object that handles [MainView.restore/drop][2]

[1]: https://github.com/sandstorm-io/sandstorm/blob/master/src/sandstorm/sandstorm-http-bridge.capnp#L25
[2]: https://github.com/sandstorm-io/sandstorm/blob/master/src/sandstorm/grain.capnp#L671

PR #3172, opened 2019-12-09
Cron API

zenhack:
This is a continuation of #2922; it superseeds that PR and also #2885. So far, I have:

- Merged back in changes from master, and fixed some bitrot.
- Drafted some changes to the API, to address the two-phase commit issue. See also my comments [here][1], and the commit message for e8631ed.

Note also that:

- I've gone back on the idea of attaching this API to the powerbox.
- Per discussion with @ocdtrekkie, I think it would be a good idea to let the user know that jobs are scheduled. Therefore, the API now expects a name for the job.

[1]: https://github.com/sandstorm-io/sandstorm/pull/2922#issuecomment-561365864


zenhack:
@kentonv, would appreciate your thoughts when you have the time.

zenhack:
Note that the `ScheduledJob` struct is completely unrelated to the (now deleted) interface of the same name in the previous API.

src/sandstorm/grain.capnp
```diff
@@ -127,66 +127,88 @@ interface SandstormApi(AppObjectId) {
...
+  # A job to be scheduled by `SandstormApi.schedule()`.
+
+  name @0 :Util.LocalizedText;
+  # A name for the job. This will be used to describe the job to the user.
+
+  callback @1 :Callback;
+  # A callback to actually execute when it is time to run the job.
   #
-  # Returns a persistable ScheduledJob handle that allows cancellation of the job.
+  # In addition to implmenting the `Callback` interface below, this must also
```
ocdtrekkie: s/implmenting/implementing

src/sandstorm/grain.capnp
```diff
@@ -127,66 +127,88 @@ interface SandstormApi(AppObjectId) {
...
   #
-  # In order to improve resource utilization, Sandstorm prefers imprecise scheduling -- this allows
-  # it to choose a time of low activity to invoke the callback. Therefore, Sandstorm guarantees
-  # that the callback will be called once per scheduling period, but does not make any guarantees
-  # about exactly when in that period the call will occur. If you need more precise scheduling,
-  # you will need to use `scheduleAt()`.
+  # - Where possible, save all needed information in the `objectId` returned
+  #   by `AppPersistent.save()`; this avoids the race condition entirely.
+  # - In your implementation of `MainView.restore()`, be sure to correctly
+  #   handle attempts to restore unrecognized capabilities, by rasing an
```
ocdtrekkie: s/rasing/raising

src/sandstorm/grain.capnp
```diff
@@ -127,66 +127,88 @@ interface SandstormApi(AppObjectId) {
...
+  interface Callback {
+    run @0 () -> (cancelFutureRuns :Bool = false);
+    # Run the job. If the return value has `cancelFutureRuns = true` and this is a periodic job,
+    # the job will be cancelled and will not be run again.
+  }
 
-  confirm @0 ();
-  # Activates the job. This method only has an effect the first time it is called.
+  schedule :union {
+  # The actual scheudule of the job.
```
ocdtrekkie: s/scheudule/schedule

src/sandstorm/grain.capnp
```diff
@@ -127,66 +127,88 @@ interface SandstormApi(AppObjectId) {
...
+
+    oneShot :group {
+    # Run the job eactly once. Currently Unimplemented.
```
ocdtrekkie: s/eactly/exactly

shell/packages/sandstorm-db/scheduled-jobs-db.js
```diff
@@ -0,0 +1,145 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const Crypto = Npm.require("crypto");
+
+const PERIOD_MILLIS = {
+  yearly:  1000 * 60 * 60 * 24 * 365,
+  monthly: 1000 * 60 * 60 * 24 * 30,
+  weekly:  1000 * 60 * 60 * 24 * 7,
```
zenhack: Observation: Weekly is present in some relevant parts of the code, and absent in others.

shell/packages/sandstorm-db/scheduled-jobs-db.js
```diff
@@ -0,0 +1,145 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const Crypto = Npm.require("crypto");
+
+const PERIOD_MILLIS = {
+  yearly:  1000 * 60 * 60 * 24 * 365,
+  monthly: 1000 * 60 * 60 * 24 * 30,
```
zenhack: If we're going to go to the trouble of providing symbolic periods instead of just doing a numeric value, we should be smart enough to know that a month isn't always 30 days, leap-years are a thing, etc.

zenhack:
(Note that this should not be merged in its current state, as it's half-implemented and broken because of it).

zenhack:
I noticed that `AppPersistent.save` actually includes a similar `name` value in its result, so maybe we can skip the name field in `ScheduledJob`, since we're going to use that anyway?

src/sandstorm/grain.capnp
```diff
@@ -126,6 +126,99 @@ interface SandstormApi(AppObjectId) {
...
+  # locally to recognize the task. There is a possibility that between scheduling
+  # the job and actually saving the information, The app may be killed, the server
+  # may crash, or some other failure may occur that prevents the app from saving
+  # the information. We recommend dealing with this in one of the following ways:
+  #
+  # - Where possible, save all needed information in the `objectId` returned
+  #   by `AppPersistent.save()`; this avoids the race condition entirely.
+  # - In your implementation of `MainView.restore()`, be sure to correctly
+  #   handle attempts to restore unrecognized capabilities, by raising an
+  #   exception with `type = failed`. This will cause Sandstorm to cancel
```
kentonv: I think cancelling the schedule if it throws an exception would lead to scheduling being unreliable.

Exceptions may be thrown for transient reasons. A "failed" exception in KJ means "there is a bug in the code", but the nature of the bug could be just about anything. It could be a failure to allocate memory, or a race condition. It could be that the grain is temporarily in a transitional state due to poorly-written storage code. It could be that the grain was updated to a new version of the app which has a bug, but an urgent bugfix update is on its way.

Generally, "reliable" message delivery systems have to retry on error, with some sort of back-off to avoid spinning and burning resources. If retries keep failing, a human should be alerted to the problem, but the system cannot give up without a human telling it to.

For example, consider Stripe webhooks. Stripe sends an HTTP request to your server to notify you of some event, like a failed payment. If your server returns a 4xx or 5xx error, Stripe retries later. If requests keep failing, Stripe will actually send the developer an e-mail to say "hey go fix your server so we can deliver these events". Stripe will never let an event go undelivered. This has saved my ass many times! If not for this behavior, every time we deployed a buggy webhook handler I would have been stuck going back through Stripe events manually to figure out what was missed.

zenhack: Yeah, I see your point. Spitballing: what if, instead of using exceptions, we modified the return type of `restore` to allow a "not-recognized" response? This would have the same advantages to the existing proposal, but without the possibility of confusing it with a different type of error.

src/sandstorm/grain.capnp
```diff
@@ -126,6 +126,99 @@ interface SandstormApi(AppObjectId) {
   # This method is here on `SandstormApi` rather than on `Identity` in order to ease a possible
   # future transition to a model where identity IDs are not globally stable and each grain has a
   # separate (possibly incompatible) map from identity ID to account ID.
+
+  schedule @11 ScheduledJob;
```
kentonv: Let's say the app calls `schedule()` and it throws a DISCONNECTED exception (or, really, any exception). The app doesn't really know for sure if the exception occurred before or after the scheduled job was persisted. What should it do? If it calls `schedule()` again, then it seems like it might end up scheduling two copies of the job, which seems problematic.

This is why we end up stuck with two-phase commit. :/

Alternatively, the other option is to make `schedule()` idempotent, so that retrying after error is always OK. That could mean that any time you schedule a job, any previously-scheduled job is removed. Or, if we feel that's too limiting and that a grain really needs to be able to schedule more than one job, we could make the grain assign names to jobs. Scheduling a new job with the same name as an existing job deletes the old job. We could also provide a way to list known jobs, which could allow an app to examine and clean up its own state.

src/sandstorm/grain.capnp
kentonv: Well, if the app explicitly knows that this is an event which should be un-scheduled, it could always construct a callback that returns false.

My worry about having a "not found" error code is that I've seen too many real-world examples where buggy or misconfigured applications incorrectly returned "not found".

src/sandstorm/grain.capnp
zenhack: I think I prefer the idea of having the app provide an id for the job to avoid duplication vs. two-phase commit. Adding a method to list existing jobs seems like a good possible extension, though that could be a separate pr.

src/sandstorm/grain.capnp
zenhack: Having the app set `cancelFutureRuns = true` sounds reasonable to me.

src/sandstorm/grain.capnp
kentonv: It's kind of sad that this is sort of abandoning capability-based design by introducing a namespace, but I guess it is a lot easier to understand in this case.

I wonder if we should take it an extra step and not make the app provide a persistent callback, but instead have a method on `MainView` to invoke a scheduled job, passing the job ID...

src/sandstorm/grain.capnp
zenhack: Hm, but the trouble with that is when calling `restore` you don't know how the returned capability is going to be used, so choosing to return something that responds to `Callback.run` is a bit awkward in general.

More spitballing: would unscheduling if run raises `unimplemented` be better than `restore` raising `failed`? I still worry about something bubbling up from a nested call, my gut is that's not a great idea.

src/sandstorm/grain.capnp
zenhack: That *would* simplify the API substantially.

Conceptually its a bit sad, but it's not really operationally any different I think. With the current design you have:

1. App calls schedule(), passing a capability.
2. Sandstorm immediately asks the app for a token (bits) to represent that capability.
3. Sandstorm stores those bits.
4. Later, Sandstorm asks the app to turn those bits back into a capability, and calls run on it

Vs:

1. App calls schedule, passing a token (bits) representing the job.
2. Sandstorm stores those bits
3. Later, Sandstorm asks the app to interpret those bits and run a job.

I don't think the indirection of having a liveref really gains you anything here; the two are basically isomorphic, just with more steps in the current design.

src/sandstorm/grain.capnp
kentonv: If the app knows that `restore()` is trying to restore a cron callback, then it can return a dummy callback that returns false.

If the app does not know what kind of object is being restored, then it seems like fundamentally the app cannot signal explicitly that the schedule should be canceled.

FWIW since apps are allowed to define their own SturdyRef type, the app can encode in the ref what type of ref it is, so can potentially know this way when to return a dummy callback without having to store anything... but then again we're back to the problem where the app is pretty fragile, and a bad database query that happens to return no results can mean the job gets unscheduled when it shouldn't have.

I really think the right answer is to alert the user that this background job has stopped working. If the user wants to cancel the job, they can do so, but they can also try to debug the problem.

src/sandstorm/grain.capnp
kentonv: Good point that we don't actually need a new method on `MainView`... the app can just pass the appropriate SturdyRef at job creation.

I guess the fundamental problem with the "practical" approach here is separation of concerns. One can imagine a future where Sandstorm apps are commonly built on top of frameworks where deals of things like SturdyRefs are managed at the framework level while the application business logic code only concerns itself with LiveRefs. Such a framework will potentially need to special-case the scheduling API, if the scheduling API uses SturdyRefs explicitly. It could be kind of ugly.

On the other hand, with the purist approach, in the framework-happy future, the added tedium of implementing a persistent LifeRef will be handled by the framework so won't be a problem for apps.

But I dunno, I can definitely see the argument that we should design for the clear near term rather than a future than may never come. Your call I guess.

zenhack:
Ok, thought about this a bit, the path I like the best right now is: the API will stay as is currently specified, but per your suggestion a `restore()` failure will not cancel the job; if the app wants to clean up unknown jobs it will have to return a callback that noops and reports the job should be cancelled.

The latest commit updates the docs accordingly; I'll update the implementation soonish.

shell/packages/sandstorm-db/scheduled-jobs-db.js
```diff
@@ -0,0 +1,127 @@
...
+  weekly:  1000 * 60 * 60 * 24 * 7,
+  daily:   1000 * 60 * 60 * 24,
+  hourly:  1000 * 60 * 60,
+};
+
+SandstormDb.prototype.addPeriodicScheduledJob = function (grainId, name, callback, period) {
+  check(grainId, String);
+  // FIXME(zenhack): check that name is a LocalizedText (how?)
```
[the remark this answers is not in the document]
zenhack: My current understanding is that it actually shouldn't be necessary here, since name is always something we pass in from a deserialized capnproto message, and so it can't be malformed. @kentonv, is my understanding here correct?

zenhack:
oneShot jobs work now too.

kentonv:
FYI I do plan to review this, but things are quite busy for me right now getting all my stuff moved across the country. Sorry for the delay! Worst-case I'll be able to look at this on the weekend of Jan 11th but hoping to find time long before that.

zenhack:
Re: testing, I have a utility app I used to validate this here:

https://github.com/zenhack/sandstorm-sched-test

It would be nice to automate the testing. However, as written it's not currently possible to check for "yes the job actually ran" without waiting for up to a minute in the case of oneShot jobs, and an hour in the case of periodic ones, which seems like a tall ask for something to run as part of the regular test suite. I did sanity check it manually.

I'd hacked the above together while actually developing the feature. I went to test it I looked into building on top of the existing test app, but found that that would be slightly awkward, since until #3027 is solved (or we add some other bridge-specific workaround) you can't actually use this API with sandstorm-http-bridge in the way (since you can't implement your own restore()). For test purposes you could probably skip the web API entirely, but since I'd mostly written the above app I figured I'd finish it up and wait to get others to weigh in. I'm not in love with the idea of pulling in Go just for one test (though I guess we do already have it as a transitive build dep due to boringssl).

Interested in others' thoughts about how to approach the testing.

shell/packages/sandstorm-db/scheduled-jobs-db.js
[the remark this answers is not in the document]
kentonv: Right. Type checking is necessary on Meteor RPC methods but not Cap'n Proto RPC methods.

LocalizedText is kind of a sad story, we never actually implemented any sort of localization with it and I'm not sure if it is even designed appropriately.

src/sandstorm/supervisor.c++
```diff
@@ -1992,6 +1992,32 @@ public:
...
+      case ScheduledJob::Schedule::PERIODIC:
+        req.getSchedule().setPeriodic(sched.getPeriodic());
+        break;
+      default:
+        KJ_UNIMPLEMENTED("Unknown schedule type.");
+    }
+    // There aren't any actually results to copy over, but we do want
+    // to wait for the SandstormCore to return before we do, so the
+    // app doesn't prematurely think the scheduling is complete.
+    return req.send().then([](auto) -> void {});
```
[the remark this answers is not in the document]
kentonv: There is an `.ignoreResult()` method on Promises, which is equivalent to what you wrote.

kentonv:
So, we have a selenium-based test suite under `/tests`. It runs Sandstorm with a special "testing mode" config option that causes it to expose some methods that can be used to control certain things.

Maybe what you could do is implement a new testing-mode method that fast-forwards time, thus testing that the task gets executed?

For the app side, we actually have a raw-API test app that gets built and packed as part of the C++ build:

https://github.com/sandstorm-io/sandstorm/tree/master/src/sandstorm/test-app

You should extend this and then use it in the tests. The idea had been that we'd gradually get rid of all our out-of-tree test apps and use this instead. Downloading SPKs from the internet as part of the test suite is crazy -- especially downloading so many different versions of the same app.

(See also #2939, which introduces a second test app based on Meteor, which is particularly useful for testing http-bridge features and integration. IIRC that PR was almost ready to merge but the author didn't see it through to completion for whatever reason. Maybe someone can finish it off? That said, as you point out, this won't help you at present -- unless you add cron functionality to http-bridge.)

shell/packages/sandstorm-db/db.js
```diff
@@ -2394,8 +2424,8 @@ if (Meteor.isServer) {
     check(intervalMs, Number);
     check(callback, Function);
 
-    if (intervalMs < 120000) {
-      throw new Error("less than 2-minute cleanup interval seems too fast; " +
+    if (intervalMs < 45000) {
+      throw new Error("less than 45-second cleanup interval seems too fast; " +
```
kentonv: Do we really want to be polling for jobs every minute, or do we maybe want to set `minimumSchedulingSlack` to something higher?

zenhack: I'm fine with just making the slack larger. Maybe 5 minutes?

kentonv: SGTM. If it turns out to be a problem we can always reduce it later without breaking anything.

zenhack:
Status update: I'm struggling with getting the test suite to run; when I run make test I get this when meteor tries to start:

```
W20200106-16:20:26.784(-5)? (STDERR) /home/isd/.meteor/packages/meteor-tool/.1.8.1.1e7bax4.0pqe++os.linux.x86_64+web.browser+web.browser.legacy+web.cordova/mt-os.linux.x86_64/dev_bundle/server-lib/node_modules/fibers/future.js:280
W20200106-16:20:26.784(-5)? (STDERR) 						throw(ex);
W20200106-16:20:26.784(-5)? (STDERR) 						^
W20200106-16:20:26.784(-5)? (STDERR) 
W20200106-16:20:26.784(-5)? (STDERR) Error: Can't find capnp.node.
W20200106-16:20:26.784(-5)? (STDERR)     at packages/sandstorm-capnp.js:53:11
W20200106-16:20:26.785(-5)? (STDERR)     at packages/sandstorm-capnp.js:61:4
W20200106-16:20:26.785(-5)? (STDERR)     at packages/sandstorm-capnp.js:69:3
W20200106-16:20:26.785(-5)? (STDERR)     at /tmp/meteor-test-run2akxu9.ikh0c/.meteor/local/build/programs/server/boot.js:419:36
W20200106-16:20:26.785(-5)? (STDERR)     at Array.forEach (<anonymous>)
W20200106-16:20:26.785(-5)? (STDERR)     at /tmp/meteor-test-run2akxu9.ikh0c/.meteor/local/build/programs/server/boot.js:228:19
W20200106-16:20:26.785(-5)? (STDERR)     at /tmp/meteor-test-run2akxu9.ikh0c/.meteor/local/build/programs/server/boot.js:479:5
W20200106-16:20:26.785(-5)? (STDERR)     at Function.run (/tmp/meteor-test-run2akxu9.ikh0c/.meteor/local/build/programs/server/profile.js:510:12)
W20200106-16:20:26.785(-5)? (STDERR)     at /tmp/meteor-test-run2akxu9.ikh0c/.meteor/local/build/programs/server/boot.js:478:11
=> Exited with code: 1
=> Your application is crashing. Waiting for file change.
```

Looks like more path issues re: node-capnp. It starts up fine when doing both `sandstorm dev-shell` and using the version already unpacked in `/opt/sandstorm`.

I'm not sure what's different. I've verified that the bundle that's being passed in to the test suite has a `capnp.node` in a place where the logic from the `sandstorm-capnp` package should find it, and it does seem to end up in the tmp-sandstorm directory created by the test runner.

I think its just tests that still need to happen before this can land.

zenhack:
I managed to work around the issue I'm having with the tests by just running them against my local dev instance, rather than the temporary one the test scripts set up.

I updated the test app with the functionality we'll need for the tests, and fixed a bug I caught while doing that.

Currently looking into what the least invasive way to mock the time is. It's fiddlier than I expected, since right now `periodicCleanup()` is using `setTimeout`, and I don't think there's any way to hook that after it's been called, so we can't just monkeypatch `Date` from a new Meteor method; It wouldn't have any effect after `periodicCleanup` has been called. We could change periodicCleanup to compute the next time to run by looking at the current time, rather than just bumping the increment, but that still wouldn't cause things to happen any sooner than the next check. Right now the best I've got feels way too complicated:

* Instead of directly using `setTimeout`, `periodicCleanup` should put jobs to run in a centralized priority queue, sorted by earliest time.
* On wakeup, `periodicCleanup` runs all jobs from the priority queue whose run time is not in the future.
* We add a meteor method that monkey-patches `Date`, and also runs ready jobs from the priority queue. This avoids needing to wait for the next call scheduled by `periodicCleanup`.

That feels way to invasive just for a simple test hook. I'll keep thinking about, suggestions welcome.

kentonv:
I think you could add a test-only meteor method that directly runs the function that checks for cron jobs, bypassing `periodicCleanup` altogether. It could be specific to cron jobs.

src/sandstorm/test-app/test-app.c++
```diff
@@ -204,8 +241,8 @@ public:
 
     // The `CLIENT` side of a `capnp::TwoPartyVatNetwork` does not serve its bootstrap capability
     // until it has initiated a request for the bootstrap capability of the `SERVER` side.
-    // Therefore, we need to restore the supervisor's `SandstormApi` capability, even if we are not
-    // going to use it.
+  // Therefore, we need to restore the supervisor's `SandstormApi` capability, even if we are not
```
kentonv: Seems like an inadvertent indentation change here.

zenhack: Yes, but this whole comment went away later, since we actually use it now.

zenhack:
> I think you could add a test-only meteor method that directly runs the function that checks for cron jobs, bypassing periodicCleanup altogether. It could be specific to cron jobs.

This sounds reasonable, I'll implement.

zenhack:
There are tests now, and they pass. @kentonv, whenever you have some time I think this is just waiting on another review pass.

ocdtrekkie:
OMG 🎉

kentonv:
It looks like the "can't find capnp" problem was actually introduced by this PR. I think it might be specifically because you added sandstorm-capnp as a dependency of sandstorm-db. I think the error is showing up when running unit tests, which are separate from the selenium tests. The unit tests don't run inside the usual sandstorm container, they run in some /tmp directory where capnp.node isn't present... trying to come up with a fix now.

zenhack:
Ack, sorry about that.

As a longer term thing: does it still make sense to be building (node-)capnp in tree? Historically capnp development was being driven by sandstorm development, and so it made sense to check them out together and hack on them at the same time. But capnp isn't missing much we need anymore, afaik sandstorm should work against the release anyway.

It seems like pulling in node-capnp from npm would simplify things a bit since we wouldn't need special-case logic for it. Obviously not the best stop gap though, since master is broken _now_ and the current version of node-capnp isn't even on npm right now.

kentonv:
Hmm, the tests don't pass for me.

I did:

```
TESTCASE="apps/schedule.js" make test
```

I saw:

```
[Schedule] Test Suite
=========================

Running:  Test periodic tasks
18:51:35.101 INFO [ActiveSessionFactory.apply] - Capabilities are: {
  "acceptSslCerts": true,
  "browserName": "chrome",
  "javascriptEnabled": true,
  "name": "Schedule"
}
18:51:35.103 INFO [ActiveSessionFactory.lambda$apply$11] - Matched factory org.openqa.selenium.grid.session.remote.ServicedSession$Factory (provider: org.openqa.selenium.chrome.ChromeDriverService)
Starting ChromeDriver 2.43.600233 (523efee95e3d68b8719b3a1c83051aa63aa6b10d) on port 22774
Only local connections are allowed.
18:51:36.074 INFO [ProtocolHandshake.createSession] - Detected dialect: OSS
18:51:36.346 INFO [RemoteSession$Factory.lambda$performHandshake$0] - Started new session d857936e8b87855c59775814939ce96c (org.openqa.selenium.chrome.ChromeDriverService)
 ✔ Passed [ok]: login completed successfully
 ✔ Element <.app-list> was visible after 600 milliseconds.
 ✔ Element <#upload-app> was visible after 590 milliseconds.
 ✔ Element <#step-confirm> was visible after 1105 milliseconds.
 ✔ Element <.app-list> was visible after 567 milliseconds.
 ✔ Element <.grain-list-table tr.action button.action> was visible after 73 milliseconds.
 ✔ Element <#grainTitle> was visible after 273 milliseconds.
 ✔ Testing if element <#grainTitle> contains text: "Untitled Sandstorm Test App instance".
 ✔ Element <iframe.grain-frame> was present after 44 milliseconds.
 ✔ Element <#grain-frame-NNFq8h3SxDbhGi5sFRT3ha> was visible after 67 milliseconds.
 ✔ Element <#do-schedule-hourly> was present after 32 milliseconds.
 ✔ Element <.grainlog-contents > pre> was visible after 72 milliseconds.
 ✔ Testing if element <.grainlog-contents > pre> contains text: "Running job hourly".
 ✔ Expected element <.grainlog-contents > pre> text to contain: "Running job hourly
Running job hourly"

OK. 14 assertions passed. (18.183s)

Running:  Test canceling tasks
 ✔ Passed [ok]: login completed successfully
 ✔ Element <.app-list> was visible after 579 milliseconds.
 ✔ Element <#upload-app> was visible after 580 milliseconds.
 ✖ Timed out while waiting for element <#step-confirm> to be present for 60000 milliseconds.  - expected "visible" but got: "not found"
    at Object.exports.command (/home/kenton/src/sandstorm/tests/commands/uploadTestApp.js:34:6)


FAILED:  1 assertions failed and 3 passed (1m 2s / 62842ms)
18:52:56.052 INFO [ActiveSessions$1.onStop] - Removing session d857936e8b87855c59775814939ce96c (org.openqa.selenium.chrome.ChromeDriverService)

 _________________________________________________

 TEST FAILURE:  1 assertions failed, 17 passed. (1m 21s)

 ✖ schedule

   - Test canceling tasks (1m 2s)
   Timed out while waiting for element <#step-confirm> to be present for 60000 milliseconds.  - expected "visible" but got: "not found"
       at Object.exports.command (/home/kenton/src/sandstorm/tests/commands/uploadTestApp.js:34:6)
   SKIPPED:
   - Test one-shot tasks
```

It looks like it's failing on app upload, which is odd, since other tests that depend on the test app seem to work? I haven't dug in.

I'm going to go ahead and release this anyway, since this is probably a test bug, and it's a new feature anyhow.

zenhack:
I hit the 'hanging on app upload' problem a few times while developing, and I remember it being something stupid with my setup, but now I'm drawing a blank... I'll see if I can dig around a bit and jog my memory.

zenhack:
Re: not finding node-capnp, it seems like for that case it should work to also search relative to `__dirname`, but I tried adding that and it looks like `__dirname` isn't actually defined for some reason. There are closed and allegedly fixed issues for meteor about that, not sure what's going on there.

kentonv:
Oh I fixed the node-capnp issue with a9d79c78bebc8c05cffe2953dfe320e77f591194. It's a hack but... fine.

I think it would be trickier than it looks to use node-capnp from npm. To build that module, you need to have capnproto installed in the standard way. So, to build sandstorm, either capnp would have to be installed on the build system or we'd have to build capnproto twice as part of the build. Either seems problematic... I think it's probably better to leave it be for now.

zenhack:
Some more details on what's going on with the test failure: it looks like the app isn't getting uninstalled between tests, so the confirmation dialog doesn't appear when the second test tries to install the app. 

I'm not sure why this doesn't affect the tests in `apps/powerbox.js`, or where the cleanup is supposed to occur.

zenhack:
Re: node-capnp, I'm fine leaving things as is if we've got a solution. That said, expecting a system wide install doesn't seem that onerous to me; most distros have packages for it these days.

Issue #3174, opened 2019-12-17
Capnproto APIs: make mapping to legacy protocols more direct.


zenhack:
This is sortof a meta issue regarding design API philosophy. I'll pick on `WebSession` to make the problem concrete, but similar arguments apply to other APIs that wrap existing protocols, e.g. the stuff in `email.capnp`.

Right now, implementing a sandstorm app that doesn't use `sandstorm-http-bridge` is uncompelling, even in a language with good capnproto support, because:

* Most languages folks would want to use to build web apps have good support for HTTP, standard interfaces for working with it within the language, and lots of libraries and frameworks built atop those interfaces that make building apps easier.
* Using the capnproto APIs directly, without any kind of compatibility shim, means forgoing the use of much of that language's web ecosystem.
* A possible solution would be to implement the language's standard HTTP interface on top of the `WebSession` API. However, the complexity of the `WebSession` API and the fact that its mapping to HTTP is not terribly straightforward means that this is rather a lot of work. I did a good chunk of this for Go, though the implementation is incomplete, and I can't in good conscience recommend that anyone else take this approach.

It seems like the current APIs are designed to capture a higher-level view of the protocols' semantics, but ironically the end result is that developers wishing to use these APIs would have to work at a lower level of abstraction than would normally be the case.

I'd propose replacing the web session interface with something simpler, which maps more directly to HTTP.

Advantages of this approach:

* This would make wrapping the capnproto API with language-standard interfaces easier, and we then could easily have libraries that take e.g. a standard python wsgi app, a Go http.Handler, Haskell's WAI interface, etc. and serve it as a Sandstorm app.
* It would mean not having to also write shims around everything else Sandstorm exposes via capnproto. It would be nice not to have to deal with issues like #3027.

Spitballing a possible alternative:

```capnp
interface HttpSession {
  request @0 (method :Text, headers :List(Util.KeyValue), responder :Responder)
    -> (requestBody :Util.ByteStream)
}

interface Responder {
  respond @0 (status :UInt16, headers :List(Util.KeyValue))
    -> (responseBody :Utill.ByteStream);
}
```

In terms of upgrade path, we could have Sandstorm try the existing WebSession methods, and if they throw Unimplemented, try calling `HttpSession.request` instead.

Thoughts?

kentonv:
Check this out, which we use in Cloudflare Workers:

https://github.com/capnproto/capnproto/blob/master/c++/src/capnp/compat/http-over-capnp.capnp

[...]

PR #3176, opened 2019-12-21
Spec out a solution to #3027

zenhack:
This introduces additions to capnproto interfaces to allow
sandstorm-http-bridge to delegate MainView.restore/drop and
UiView.getViewInfo to the app, while still handling HTTP
translation.

Would like opinions on the design.

//cc @kentonv, @ocdtrekkie

### DIFF
--- a/src/sandstorm/sandstorm-http-bridge-internal.capnp
+++ b/src/sandstorm/sandstorm-http-bridge-internal.capnp
@@ -30,9 +30,6 @@ struct BridgeObjectId {
   # Recall that Sandstorm obfuscates object IDs automatically, such that clients cannot see the
   # contents and the app can trust that the ID passed from Sandstorm is authentic. Hence, we can
   # put all the metadata we need directly in this structure and let Sandstorm store it.
-  #
-  # Note the AppHooks.BridgeObjectId defined in `sandstorm-http-bridge.capnp` is wire
-  # compatible with this, but exposes less information to apps.
 
   union {
     application @0 :AnyPointer;
--- a/src/sandstorm/sandstorm-http-bridge-internal.capnp
+++ b/src/sandstorm/sandstorm-http-bridge-internal.capnp
@@ -30,6 +30,9 @@ struct BridgeObjectId {
   # Recall that Sandstorm obfuscates object IDs automatically, such that clients cannot see the
   # contents and the app can trust that the ID passed from Sandstorm is authentic. Hence, we can
   # put all the metadata we need directly in this structure and let Sandstorm store it.
+  #
+  # Note the BridgeObjectId defined in `sandstorm-http-bridge.capnp` is wire compatible
+  # with this, but exposes less information to apps.
 
   union {
     application @0 :AnyPointer;
--- a/src/sandstorm/sandstorm-http-bridge-internal.capnp
+++ b/src/sandstorm/sandstorm-http-bridge-internal.capnp
@@ -31,8 +31,8 @@ struct BridgeObjectId {
   # contents and the app can trust that the ID passed from Sandstorm is authentic. Hence, we can
   # put all the metadata we need directly in this structure and let Sandstorm store it.
   #
-  # Note the BridgeObjectId defined in `sandstorm-http-bridge.capnp` is wire compatible
-  # with this, but exposes less information to apps.
+  # Note the AppHooks.BridgeObjectId defined in `sandstorm-http-bridge.capnp` is wire
+  # compatible with this, but exposes less information to apps.
 
   union {
     application @0 :AnyPointer;
--- a/src/sandstorm/sandstorm-http-bridge.capnp
+++ b/src/sandstorm/sandstorm-http-bridge.capnp
@@ -42,44 +42,46 @@ interface SandstormHttpBridge {
   # grain's database, allowing it to be fetched later with `getSavedIdentity()`.
 }
 
-interface AppHooks {
+interface AppHooks (AppObjectId) {
   # When connecting to the bridge's domain socket at /tmp/sandstorm-api, the
   # application may supply this as a bootstrap interface. If the app chooses
   # to do so, it should also set `bridgeConfig.expectAppHooks = true` in the
   # package's PackageDefinition.
 
-  getViewInfo @0 () -> :Grain.UiView.ViewInfo;
+  getViewInfo @0 () -> Grain.UiView.ViewInfo;
   # Like Grain.UiView.GetViewInfo. If AppHooks is supplied, the bridge will
   # delegate UiView.GetViewInfo to this method. If it raises unimplemented,
   # the bridge will fall back to reading the viewInfo from the bridgeConfig.
 
-  restore @0 (objectId :BridgeObjectId) -> (cap :Capability);
-  # Like Grain.MainView.restore, but the objectId is wrapped to work
-  # with the bridge; see the comments for `BridgeObjectId`.
+  restore @1 (objectId :AppObjectId) -> (cap :Capability);
+  # Like Grain.MainView.restore. If an app implements AppHooks.restore, any
+  # capabilities it provides which implement AppPersistent should return
+  # a value of type BridgeObjectId (below), with the `application` variant
+  # set to the value that should be passed to AppHooks.restore/drop.
 
-  drop @1 (objectId :BridgeObjectId);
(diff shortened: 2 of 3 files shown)


ocdtrekkie:
Honestly, I believe I can wrap my head around the problem you are trying to solve, but any design implications here are beyond my level of ability to grasp.

zenhack:
@ocdtrekkie, fair. I specifically tagged you because you'll need this to actually use the functionality implemented in #3172 (in this design you'd need to implement `AppHooks.restore()` in order for your jobs to actually run). I don't expect you to anticipate every design implication, but seeing as this is something you might end up having to use, an impression of the API would be helpful. If this and #3172 landed soon, do you think you could actually make progress on getting ttrss to auto-update?

ocdtrekkie:
FWIW, I am unsure how equipped I will be to implement it. One of the reasons I put a bounty out on updating the app as well. ;)

I mean, as someone who fundamentally doesn't understand this PR, I guess what I'd be looking for is a Sandstorm docs page at some point that says "here's an example in some language of the most stupid-simple way to have an app launch itself every few hours", which folds into the whole problem of people having a hard time figuring out how to package Sandstorm apps.

I mean, part of me wants to say "why not have a bridgeConfig setting that launches this app roughly x often" for apps that more or less do what they need to by default, but just need to be opened regularly. But I know that wouldn't be helpful for any apps that need to do any custom scheduling. And I believe Kenton is iffy on adding a lot of extraneous things to the package definition.

zenhack:
I could be talked into adding something like this to bridgeConfig:

```capnp
struct PeriodicTask {
  command @0 :Command;
  period @1 :SchedulingPeriod;
}

periodicTasks @5 :List(PeriodicTasks);
```

And having a shortcut to do this particular task is probably a good idea; if folks have something in vaguely crontab format, they shouldn't have to do a ton of work to get it to run on sandstorm. And even with an understanding of these APIs, I'm of the opnion that the work involved to get a basic "run this command once a day/hour/etc" is greater than it ought to be.

But more broadly I'm a little wary of getting into a situation where we end up just wrapping every new feature we add in some quick-and-dirty thing added to sandstorm-http-bridge. At some point we need to tackle the problem of developer comfort with the capnp apis head-on.

ocdtrekkie:
Indeed, but that is likely heavily dependent on developers building for Sandstorm, as opposed to developers packaging for Sandstorm. (Side note though, I would hardly call myself a "developer".)

Arguably sandstorm-http-bridge is a way of giving conventional apps/developers quick and dirty access to Sandstorm APIs for "legacy" apps, so having a pile of switches might be absolutely in-spec for it.

Ultimately the best case for getting people used to using the actual API is going to be example code someone can grab and modify for their JS app or the like.

zenhack:
Quoting Jacob Weisz (2019-12-23 15:28:55)

> Indeed, but that is likely heavily dependent on developers building
> for Sandstorm, as opposed to developers packaging for Sandstorm. (Side
> note though, I would hardly call myself a "developer".)

Granted.

> Arguably sandstorm-http-bridge is a way of giving conventional
> apps/developers quick and dirty access to Sandstorm APIs for "legacy"
> apps,

I think this is only true for a definition of "legacy" that's highly
impractical: anything that actually wants to speak HTTP. See also #3174.
I don't think removing it from the stack is a realistic option even for
greenfield projects built specifically for Sandstorm (I've tried).

But you're right that the bridge is a good place for compatibility
stuff, and given that the cron pr adds a new method to SandstormApi
(which I am also generally inclined to be conservative about), adding a
new field to BridgeConfig seems reasonable.

> Ultimately the best case for getting people used to using the actual
> API is going to be example code someone can grab and modify for their
> JS app or the like.

Point taken. I agree docs and examples in particular are a big hole in
the developer experience right now.

Anyway, this is getting a bit off topic for this pr (which addresses an
issue more general than just the scheduling API).

zenhack:
Fixed a few problems I found, and made the API slightly more type safe.

src/sandstorm/sandstorm-http-bridge-internal.capnp
```diff
@@ -30,6 +30,9 @@ struct BridgeObjectId {
   # Recall that Sandstorm obfuscates object IDs automatically, such that clients cannot see the
   # contents and the app can trust that the ID passed from Sandstorm is authentic. Hence, we can
   # put all the metadata we need directly in this structure and let Sandstorm store it.
+  #
+  # Note the AppHooks.BridgeObjectId defined in `sandstorm-http-bridge.capnp` is wire
```
kentonv: Instead of defining parallel, compatible messages, can you instead pull out the `application` part of this message and pass only that on to the app? I think that's what I had in mind here.

src/sandstorm/sandstorm-http-bridge.capnp
```diff
@@ -41,3 +41,47 @@ interface SandstormHttpBridge {
...
+  restore @1 (objectId :AppObjectId) -> (cap :Capability);
+  # Like Grain.MainView.restore. If an app implements AppHooks.restore, any
+  # capabilities it provides which implement AppPersistent should return
+  # a value of type BridgeObjectId (below), with the `application` variant
+  # set to the value that should be passed to AppHooks.restore/drop.
+
+  drop @2 (objectId :AppObjectId);
+  # Like Grain.MainView.drop. See the comments for restore.
+
+  struct BridgeObjectId {
```
kentonv: I think this struct doesn't need to be public at all if sandstorm-http-bridge automatically wraps/unwraps `AppObjectId` into/out of the private version of `BridgeObjectId`.

src/sandstorm/sandstorm-http-bridge-internal.capnp
zenhack: I agree that would be a better API, but it seems much more complex to implement -- unless I'm missing something, the bridge would have to intercept all incoming calls from the platform to AppPersistent.save(), so that it could wrap the sturdyref. I'm not sure how to approach that in way that seems worth the trouble. Open to ideas.

src/sandstorm/sandstorm-http-bridge.capnp
zenhack: See my other comment, but I'm unsure how to do the wrapping part, since that happens in AppPersistent.save()

src/sandstorm/sandstorm-http-bridge.capnp
```diff
@@ -41,3 +41,47 @@ interface SandstormHttpBridge {
...
+      # The object ID is in a format understood by the application, not by http-bridge.
+      # When objects supplied by the app (as opposed to the bridge itself) implement
+      # Grain.AppPersistent, save() should always use this variant, and store their
+      # own data in the pointer.
+
+      bridge @1 :AnyPointer;
+      # The object ID is in a format understood by sandstorm-http-bridge. The bridge
+      # will use this to manage persistent objects it implements itself (for example,
+      # HTTP apis), but applications should not set this variant on objects they
+      # create, and the bridge will never pass this to `AppHooks.restore` or `drop`.
```
zenhack: Just noticed I could probably drop this last clause, since the type makes it clear the bridge does the unwrapping anyway (in my first pass it did not).

src/sandstorm/sandstorm-http-bridge-internal.capnp
kentonv: This is what membranes are for. :) Take a look at how the supervisor process uses membranes to intercept `save()` calls destined for the app and then wraps them into [SupervisorObjectId](https://github.com/sandstorm-io/sandstorm/blob/master/src/sandstorm/supervisor.capnp#L381).

https://github.com/sandstorm-io/sandstorm/blob/master/src/sandstorm/supervisor.c++#L1716

zenhack: Aha, I didn't realize the capnproto supported this in a generic way -- that makes this much easier. I'll update the schema.

zenhack: Updated.

kentonv:
LGTM. Should I merge this or did you intend this PR only for feedback purposes?

zenhack:
If you're happy with the API go ahead and merge, and I'll do the implementation as a second pr.

zenhack:
If you're happy with the API go ahead and merge, and I'll do the implementation as a second pr.

Issue #3171, opened 2019-11-29
Build failure on current Archlinux


zenhack:
I was trying to build from source just, and hit the error below. This is on current Archlinux. This bit looks relevant:

```
    In file included from /ekam-provider/canonical/sandstorm/web-session-bridge.c++:17:
    In file included from /ekam-provider/canonical/sandstorm/web-session-bridge.h:22:
    In file included from /ekam-provider/canonical/sandstorm/util.h:35:
    In file included from /usr/lib64/gcc/x86_64-pc-linux-gnu/9.2.0/../../../../include/c++/9.2.0/set:61:
    /usr/lib64/gcc/x86_64-pc-linux-gnu/9.2.0/../../../../include/c++/9.2.0/bits/stl_set.h:193:9: error: no matching member function for call to '_M_insert_range_unique'
            { _M_t._M_insert_range_unique(__first, __last); }
              ~~~~~^~~~~~~~~~~~~~~~~~~~~~
```

But I haven't been able to sort out what's going on. Here's the full set of errors:

```
   /ekam-provider/canonical/sandstorm/supervisor.c++:791:19: warning: 'readdir_r' is deprecated [-Wdeprecated-declarations]
          int error = readdir_r(dir, &entry, &eptr);
                      ^
    /usr/include/dirent.h:186:28: note: 'readdir_r' has been explicitly marked deprecated here
         __nonnull ((1, 2, 3)) __attribute_deprecated__;
                               ^
    /usr/include/sys/cdefs.h:251:51: note: expanded from macro '__attribute_deprecated__'
    # define __attribute_deprecated__ __attribute__ ((__deprecated__))
                                                      ^
    /ekam-provider/canonical/sandstorm/supervisor.c++:1903:29: warning: lambda capture 'this' is not used [-Wunused-lambda-capture]
        return req.send().then([this, context](auto args) mutable -> void {
                                ^~~~~
    In file included from /ekam-provider/canonical/sandstorm/supervisor.c++:17:
    In file included from /ekam-provider/canonical/sandstorm/supervisor.h:20:
    In file included from /ekam-provider/canonical/sandstorm/abstract-main.h:20:
    In file included from /ekam-provider/c++header/kj/main.h:28:
    In file included from /ekam-provider/c++header/kj/array.h:28:
[...]

PR #3178, opened 2020-01-08
Fix for part of issue #3171.

zenhack:
See the comment. For reference, the relevant detection logic is here:

https://github.com/lovell/detect-libc/blob/5173db0bc47608f89986d9125c2d868c08d5f429/lib/detect-libc.js#L42

I'm not in love with this solution, but I'm not sure what else to do.

### DIFF
--- a/make-bundle.sh
+++ b/make-bundle.sh
@@ -133,6 +133,13 @@ tar xf $OLD_BUNDLE_PATH --transform=s/^${OLD_BUNDLE_BASE}/bundle/ $OLD_MONGO_FIL
 
 cp $(which zip unzip xz gpg) bundle/bin
 
+# 'node-fibers' depends on a package (detect-libc) that uses various heuristics
+# to work out what libc implementation & version it was linked against. The more
+# reliable ones use these commands, so we include them to increase the chances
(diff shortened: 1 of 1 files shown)


make-bundle.sh
```diff
@@ -133,6 +133,13 @@ tar xf $OLD_BUNDLE_PATH --transform=s/^${OLD_BUNDLE_BASE}/bundle/ $OLD_MONGO_FIL
 
 cp $(which zip unzip xz gpg) bundle/bin
 
+# 'node-fibers' depends on a package (detect-libc) that uses various heuristics
+# to work out what libc implementation & version it was linked against. The more
+# reliable ones use these commands, so we include them to increase the chances
+# of success. Notably, without these detecting the correct libc fails if the
+# bundle was built on current Archlinux (as of Jan. 2020).
+cp $(which ldd getconf) bundle/bin
```
kentonv: Wait, it shells out to these at runtime!?!??! Are you sure it doesn't only need this at build time?

Oh my god you're right, node-fibers uses it to decide what the name of the fibers node module should be, in order to open it... uggghhhh.

kentonv:
Whelp, OK, I guess. Thanks for the fix!

PR #3180, opened 2020-01-13
external-ui-view: Don't restrict GitHub to /users

zenhack:
I don't understand why this was previously restricted to the /users
path on the GitHub api; this ought to match for the entire API.

### DIFF
--- a/shell/server/drivers/external-ui-view.js
+++ b/shell/server/drivers/external-ui-view.js
@@ -64,7 +64,7 @@ function getOAuthServiceInfo(url) {
       service: "google",
       endpoint: "https://www.googleapis.com/oauth2/v4/token",
     };
-  } else if (url.startsWith("https://api.github.com/users")) {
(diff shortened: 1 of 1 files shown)


kentonv:
I feel like there was a story behind this, but I don't remember what it was. But I can't see any reason why this change would break anything, so... ok, let's do it.

shell/server/drivers/external-ui-view.js
```diff
@@ -64,7 +64,7 @@ function getOAuthServiceInfo(url) {
       service: "google",
       endpoint: "https://www.googleapis.com/oauth2/v4/token",
     };
-  } else if (url.startsWith("https://api.github.com/users")) {
+  } else if (url.startsWith("https://api.github.com/")) {
```
kentonv: After merging and then reading other PRs I realize that you probably need to change this so that it also accepts `https://api.github.com` (with no trailing `/`). But it shouldn't accept `https://api.github.com.evil.com`, of course.

Issue #3179, opened 2020-01-13
ApiSession doesn't work with GitHub's OAuth.


zenhack:
I've been messing around with trying to get a basic demo of api-session working, using the GitHub API. However, I'm getting authentication failures (returned as 404s, but they don't distinguish in order to not leak info). I tracked this down to how the oauth token is passed in the `Authorization` header; sandstorm prefixes the token with `Bearer`:

https://github.com/sandstorm-io/sandstorm/blob/80ef65f8207cbc91ebbabaa83819a7ca33a60837/shell/server/drivers/external-ui-view.js#L45

However, the GitHub api expects a prefix of `token`:

https://developer.github.com/v3/#oauth2-token-sent-in-a-header

We probably need to add a bit of per-service logic here.

zenhack:
Actually nevermind; this seems to work with curl using `Bearer` as well as `token`; I'll keep troubleshooting I guess. Closing.

PR #3181, opened 2020-01-13
Fix bug where some api-session requests had an extra / in the path.

zenhack:
See the comment; if this.path is '/' then joining it and the rest of the
path with another '/' gives us a duplicate slash, which is not what we
want.

Note that there's a normative change to the comment here as well: I don't think it makes sense to plan remove the workaround for hack session down the line. I don't think being rigid about this really buys us anything, though it might cause trip up developers if they make the mistake of adding the slash themselves. I am willing to change the wording on the comment if you feel strongly though.

This turned out to be the real source of the problems I was seeing when I opened #3179; I have a demo using the GitHub api that will work once this and #3180 are merged.

### DIFF
--- a/shell/server/drivers/external-ui-view.js
+++ b/shell/server/drivers/external-ui-view.js
@@ -399,14 +399,16 @@ ExternalWebSession = class ExternalWebSession extends PersistentImpl {
         options.headers["user-agent"] = "sandstorm app";
       }
 
-      if (this.fromHackSession) {
-        // According to the specification of `WebSession`, `path` should not contain a
-        // leading slash, and therefore we need to prepend "/". However, for a long time
-        // this implementation did not in fact prepend a "/". Since some apps might rely on
-        // that behavior, we only prepend "/" if the path does not start with "/".
-        //
-        // TODO(soon): Once apps have updated, prepend "/" unconditionally.
-        options.path = path.startsWith("/") ? path : "/" + path;
+      // In theory, we should always prepend a slash to the path ourselves because,
+      // according to the specification of `WebSession`, `path` should not contain
+      // one, but in practice we need to be a little more flexible, for two reasons:
+      //
+      // 1. For a long time the `HackSession` implementation did not in fact
+      //    prepend a "/", so some apps may rely on that behavior.
+      // 2. If the session's path is at the root of a domain, it will have the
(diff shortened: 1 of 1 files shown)


kentonv:
So, auto-coalescing of slashes in general bugs me because it leads to security vulnerabilities around empty strings. Say there's an API whose URLs are like `https://api.example.com/<username>/foo`, and you want to construct a capability to give to me that only allows access to my own username, which I choose. So I choose my username as `""` -- the empty string. You then construct a capability to `https://api.example.com/` -- you concatenated the username on the end there but it's an empty string! If I then try to access the path `zenhack/secrets` under this capability, I should get `https://api.example.com//zenhack/secrets` which is an invalid URL and 404, but instead due to coalescing of `/`s I get `https://api.example.com/zenhack/secrets` and I steal your secrets.

If you actually wanted to give me a capability to the root of the API, you would have expressed it as `https://api.example.com` -- note the lack of a trailing `/`. See line 352 in this file where it handles this case.

But maybe this is a lost cause. Path joining was invented wrong in 1970 or whatever and trying to introduce sanity now just confuses people. :/

zenhack:
Aha, the trailing slash in the requested url is probably part of why this didn't work.

...however, right now if you *do* omit the trailing slash, the logic in `getOAuthServiceInfo` doesn't recognize it as matching one of those, per your comment here: https://github.com/sandstorm-io/sandstorm/pull/3180/files#r366673359

I guess we could change the check to something like `x == "https://api.github.com" || x.startsWith("https://api.github.com/")`

Sanity would probably require not using path strings :/

I'll see about the solution above.

kpreid:
The question that comes to mind on hearing this is: what do the HTTP and URI RFCs have to say? I searched RFC 3986, RFC 2616, and RFC 7320, and as far as I can tell there is nothing that grants license to treat an empty path component like an absent one or vice versa (except of course to the server determining the final interpretation of paths) — except in the case where `https://example.com` is equivalent to `https://example.com/`.

If I understand the comment and the code change correctly, it is about handling only and exactly that last case — but the principle should _not_ be generalized to middle slashes. (But I don't understand what exactly this is doing in Sandstorm; I'm just presenting some context that seemed important for someone to have checked.)

zenhack:
Closing in favor of #3182

PR #3185, opened 2020-01-17
Fix Docs

abliss:
See commit comments.


ocdtrekkie:
We probably can close the previous PR, it's included in this one.

But search is broken again on your test grain. :(

abliss:
Good catch, thanks. I've fixed the search and the landing page. Redeployed to test grain at https://7125z831x29zsn814csg.rapidraven.sandcats.io:6443/ .

ocdtrekkie:
I'm on my phone right now, but I don't see anything I'd be concerned about right now. Assuming I'm not missing some new regression, I think we're probably there.

Issue #3075, opened 2018-06-01
Docs: vagrant-spk design document is obsolete/old info


ocdtrekkie:
There's some interesting things in it like the goals/direction of vagrant-spk at the top of the about section which should probably be spun out into a separate article, but much of it, concerning how it "should work" is not accurate, as it was all development prior to implementation. I am not sure there's significant value in retaining the original design document in docs other than introducing potential confusion.

This is probably an issue I should solve, but I figured I would document it here and maybe ask @paulproteus or @zarvox if they had any opinions on that.

ocdtrekkie:
@zenhack found this file and reminded me I should take care of this. Reviewing the document again, my personal opinion is we should just drop it entirely. I think the goals/direction info could maybe be stashed in vagrant-spk's readme or a wiki page, but it's not valuable as docs and we need our docs to be informative and organized.

I will open a PR to delete this page and remove it from the table of contents.

PR #3194, opened 2020-01-27
Docs: fix some bitrot re: vagrant-spk up/vm up.

zenhack:
Stuff was moved into the `vm` namespace way back, apparently this never got updated.


ocdtrekkie:
I am not sure we should fix this because the design document is largely historical and very inaccurate today. People should not use it to learn how to use vagrant-spk, and the correct outcome may be to strip it of any remaining unique information/value and then purge this page.

--
 Jacob Weisz
 inbox@jacobweisz.com

ocdtrekkie:
See #3075

zenhack:
I'll let you make the call. I didn't really look at the document very closely tbh.

ocdtrekkie:
I am going to close this one then, but highly prioritize getting rid of the entire page (now that docs changes are a thing again). While vagrant-spk's design doc may be interesting, it's probably not appropriate for it to be in Docs at all, and likely a big distraction from better curated documentation on how to use and work with vagrant-spk.

PR #3200, opened 2020-02-03
Update vagrant-spk stack list to reflect master

ocdtrekkie:
We do tell people to use GitHub master by default, so we should update this in Docs immediately.


ocdtrekkie:
I should note that even though I have permission to merge docs, I don't intend to merge any of my own changes to master without someone else approving it. Maybe @abliss or @zenhack would be so kind.

PR #3201, opened 2020-02-03
Remove vagrant-spk design doc

### LINKED ISSUE #3075: Docs: vagrant-spk design document is obsolete/old info

ocdtrekkie:
Fixes #3075. It's outdated and can lead to confusion. It may have interesting historical value, but then it's place is in git history, not git master. Updating it isn't really worthwhile because it would duplicate the efforts of the tutorial and other documents. Furthermore, even the design goals are not necessarily super important to retain, as years have passed and the goals of the project may shift over time, especially as different people take the lead.

If there's a strong desire to retain this, I would perhaps stash this in a wiki page, but I would not keep it in Docs, which should be relatively good at getting people to the information they need to develop apps and operate Sandstorm.

I'd love @paulproteus or @zarvox opinion on just dropping this document from Docs.


ocdtrekkie:
Retaining a copy on the wiki would be super easy, we would just revert the last change at https://github.com/sandstorm-io/sandstorm/wiki/Porting-an-app-with-vagrant-spk

PR #3206, opened 2020-02-07
Fix Vagrantfile's Windows RAM calculation

ocdtrekkie:
See https://github.com/sandstorm-io/vagrant-spk/pull/241 for the same fix on vagrant-spk, see https://github.com/sandstorm-io/vagrant-spk/pull/240 for details on how the existing command responds (listing half the RAM twice instead of providing a single valid amount of RAM).

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -56,12 +56,8 @@ EOF
     cpus = `nproc`.to_i
     total_kB_ram = `grep MemTotal /proc/meminfo | awk '{print $2}'`.to_i
   elsif host =~ /mingw/
-    # powershell may not be available on Windows XP and Vista, so wrap this in a rescue block
-    begin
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
As an aside (since I'd want to do it in vagrant-spk too/first)... I am really tempted to remove the rescue block. What on earth are we doing with code designed to prevent failure on XP and Vista?

zenhack:
I'm fine with junking the rescue, given that both of those OSes are EOL.

ocdtrekkie:
I'll test removing it and then add that change here and to vagrant-spk. There are still people adamantly using Windows 7, but I know of nobody militantly attached to Windows Vista. Furthermore, as I noted in the vagrant-spk repo, I believe this bug should error out because the PowerShell is wrong, but the rescue block likely was suppressing the fact that this code was broken.

ocdtrekkie:
Tested vagrant-spk with the original broken version which gave me a 512 MB RAM VM (the default), and with the corrected command and the rescue block removed, and got a 4 GB RAM VM (a quarter of my system RAM).

kentonv:
I have no idea what this does but I trust you. :)

Issue #3188, opened 2020-01-19
spacejam tests broken as of Meteor 1.8.2


kentonv:
When I try to update to Meteor 1.8.2, the tests no longer run. It looks like they fail in the spacejam stage. The tests hang after printing this:

```
=> App running at: http://localhost:4096/
spacejam: meteor is ready
spacejam: spawning phantomjs
phantomjs: Running tests at http://localhost:4096/local using test-in-console
TYPE_MISMATCH_ERR: DOM Exception 17: The type of an object was incompatible with the expected type of the parameter associated to the object.

  http://localhost:4096/packages/random.js?hash=12299754ae4ff2f712be4beac9e83f1a2767afc3:186
  http://localhost:4096/packages/random.js?hash=12299754ae4ff2f712be4beac9e83f1a2767afc3:278
  http://localhost:4096/packages/random.js?hash=12299754ae4ff2f712be4beac9e83f1a2767afc3:229
  http://localhost:4096/packages/random.js?hash=12299754ae4ff2f712be4beac9e83f1a2767afc3:249
  http://localhost:4096/packages/ddp-client.js?hash=eb08b9f44ae64f2796293bd3e55b9d578fa4893d:774 in subscribe
  http://localhost:4096/packages/ecmascript-runtime-client.js?hash=5b30537e50eb15e04414a25f8255a85e32b4a76c:7682 in bound
  http://localhost:4096/packages/autoupdate.js?hash=5aadf4beac05bd9d8656e96a2c5f41ef8d237f29:200
  http://localhost:4096/packages/autoupdate.js?hash=5aadf4beac05bd9d8656e96a2c5f41ef8d237f29:204
  http://localhost:4096/packages/modules-runtime.js?hash=e45bec2ec3e2453a582f7d943d64a6c995a6f6e3:352 in fileEvaluate
  http://localhost:4096/packages/modules-runtime.js?hash=e45bec2ec3e2453a582f7d943d64a6c995a6f6e3:248 in require
  http://localhost:4096/packages/modules-runtime.js?hash=e45bec2ec3e2453a582f7d943d64a6c995a6f6e3:268 in require
  http://localhost:4096/packages/autoupdate.js?hash=5aadf4beac05bd9d8656e96a2c5f41ef8d237f29:393
  http://localhost:4096/packages/autoupdate.js?hash=5aadf4beac05bd9d8656e96a2c5f41ef8d237f29:400
[...]

PR #3208, opened 2020-02-09
Add a Github action to automatically build and test

abliss:
Currently the test fails, so we don't fail the job for that (to reduce noisy
emails). Instead we upload the test log for inspection.

My next plan is to try to disable the failing tests, then switch the job to fail for any failing test, then fix and re-enable the tests one by one. It'd be nice if someone could browse through the test log and let me know if there are any tests that shouldn't be disabled (e.g. because the failure appears to be an issue with the runner, not with the test).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,77 @@
+name: build_and_test
+
+on: [push, pull_request]
+
+env:
+  BUILD: 9999
+  CCACHE_DIR: ${{ github.workspace }}/.ccache"
+  # TODO: remove this. See https://github.com/sandstorm-io/sandstorm/issues/3188
+  SKIP_UNITTESTS: 1
+jobs:
+  build_and_test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: install dependencies
+        run: |
+            export DEBIAN_FRONTEND=noninteractive
+            echo "The build of boringssl requires golang."
+            echo "The github runner seems to have some version of golang already; installing golang-go breaks."
+            which go || sudo apt-get install golang-go
+
+            sudo apt-get install -y build-essential libcap-dev xz-utils zip unzip strace curl discount git python zlib1g-dev cmake ccache
+      - name: install meteor
+        run: |
+            curl https://install.meteor.com/ | sh
+      - name: cache ccache files
+        uses: actions/cache@v1.1.0
+        with:
+          path: ${{ env.CCACHE_DIR }}
+          key: ccache-${{ github.run_number }}
+          restore-keys: ccache-
+      - name: print and zero ccache stats
+        run: |
+            mkdir -p "$CCACHE_DIR"
+            ccache -s
+            ccache -z
+      - name: make fast
+        run: |
+            set -x
+            export CCACHE_COMPRESS="true"
+            export CCACHE_COMPRESSLEVEL="6"
+            export CCACHE_MAXSIZE="1G"
+            #     export CCACHE_DEBUG=1
+            export CCACHE_SLOPPINESS="time_macros"
+            export PATH=/usr/lib/ccache:$CLANGDIR:$PATH
+
+            make BUILD=${{ env.BUILD }} \
+              CC="$(which clang)" \
+              CXX="$(which clang++)" \
+              fast
+      - name: print ccache stats
+        run: |
+            ccache -s
+      #      # for debugging ccache misses
+      #      find . -iname *ccache* | tar -czf ccache-debug.tar.gz -T /dev/stdin
+      #- name: upload ccache debug tarball
+      #  uses: actions/upload-artifact@v1
+      #  with:
+      #    name: ccache-debug.tar.gz
+      #    path: ccache-debug.tar.gz
+      - name: upload sandstorm tarball
+        uses: actions/upload-artifact@v1
+        with:
+          name: sandstorm-${{ env.BUILD }}-fast.tar.xz
+          path: sandstorm-${{ env.BUILD }}-fast.tar.xz
+      - name: test
+        run: |
+          set -o pipefail
+          # Currently these tests never pass.
+          # TODO: When they are fixed, remove the || true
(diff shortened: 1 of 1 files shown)


abliss:
Example run, for inspection: https://github.com/sandstorm-io/sandstorm/runs/434692231?check_suite_focus=true

abliss:
Example run on a warm ccache is here: https://github.com/abliss/sandstorm/runs/434691487?check_suite_focus=true
`make fast` goes down from 17m40s to 12m32s

ocdtrekkie:
So if I read the test log right, the four tests failed are:

tests/account-settings (#3225)
tests/account (<kentonv> Also broken, but my fault: tests/account because I removed the ability to upgrade demo accounts,)
apps/schedule (#3226)
apps/web-publishing (<kentonv> and apps/web-publishing which has been broken ever since I rewrote the HTTP handling into C++)

That suggests you have a working tool here, because the test failures we're seeing are all ones Kenton sees on his build machine. I am going to mark this ready-for-review.

kentonv:
Answering my previous question about how ccache gets shared between builds: https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows

Neat!

Is there any way that people building locally can download the cache from CI builds?

.github/workflows/build.yml
```diff
@@ -0,0 +1,76 @@
...
+        uses: actions/upload-artifact@v1
+        with:
+          name: sandstorm-${{ env.BUILD }}-fast.tar.xz
+          path: sandstorm-${{ env.BUILD }}-fast.tar.xz
+      - name: test
+        run: |
+          set -o pipefail
+          # Currently these tests never pass.
+          # TODO: When they are fixed, remove the || true
+          (make BUILD=${{ env.BUILD }} test |& tee testlog.txt) || true
```
kentonv: We will need to add `SKIP_UNITTESTS=1` (before `make`) due to #3188. Otherwise the test run hangs before it even runs any interesting tests.

abliss:
I added `SKIP_UNITTESTS=1` to the env, but the `make test` still fails.

ocdtrekkie:
On line 1952 of the test log seems to be the results of the tests, defining which ones failed. Pasted below. Those are the tests that, as noted, we currently expect to fail. The SKIP_UNITTESTS=1 is a new requirement since you originally opened this PR because of the Meteor update, I believe. Which was required to still get these four tests to fail, as they should.

` TEST FAILURE:  4 assertions failed, 1015 passed. (9m 8s)

 ✖ tests/account-settings

   - Test profile changes passing to testapp (1m 3s)
   Timed out while waiting for element <#name> to be present for 30000 milliseconds.  - expected "visible" but got: "not found"
       at Object.module.exports.Test profile changes passing to testapp (/home/runner/work/sandstorm/sandstorm/tests/tests/account-settings.js:116:6)
       at _combinedTickCallback (internal/process/next_tick.js:132:7)

 ✖ tests/account

   - Test link credentials (7.456s)
   Timed out while waiting for element <form.account-profile-editor> to be present for 5000 milliseconds.  - expected "visible" but got: "not found"
       at Object.module.exports.Test link credentials (/home/runner/work/sandstorm/sandstorm/tests/tests/account.js:53:6)
       at _combinedTickCallback (internal/process/next_tick.js:132:7)
   SKIPPED:
   - Test try login with non-login credential
   - Test link credential from unused account

 ✖ apps/schedule

   - Test canceling tasks (1m 2s)
   Timed out while waiting for element <#step-confirm> to be present for 60000 milliseconds.  - expected "visible" but got: "not found"
       at Object.exports.command (/home/runner/work/sandstorm/sandstorm/tests/commands/uploadTestApp.js:34:6)
   SKIPPED:
   - Test one-shot tasks

 ✖ apps/web-publishing

   - Web publishing with grain shutdown (9.84s)
   Testing if element <#result> contains text: "Shutdown success".  - expected "Shutdown success" but got: "Success"
       at Object.<anonymous> (/home/runner/work/sandstorm/sandstorm/tests/apps/web-publishing.js:69:17)`

abliss:
> Is there any way that people building locally can download the cache from CI builds?
I don't see a way. But if someone wants to use  ccache locally, they can build up their own cache pretty quickly.

.github/workflows/build.yml
```diff
@@ -0,0 +1,77 @@
+name: build_and_test
+
+on: [push, pull_request]
+
+env:
+  BUILD: 9999
```
kentonv: Why set build to 9999 instead of leaving it default? By default the build will be set to 0, which Sandstorm's updater logic understands to be a dev build. With build number 9999, the updater logic will believe this is legitimately a far-future build and this may lead to confusion if someone actually installs it.

.github/workflows/build.yml
```diff
@@ -0,0 +1,77 @@
...
+        run: |
+            ccache -s
+      #      # for debugging ccache misses
+      #      find . -iname *ccache* | tar -czf ccache-debug.tar.gz -T /dev/stdin
+      #- name: upload ccache debug tarball
+      #  uses: actions/upload-artifact@v1
+      #  with:
+      #    name: ccache-debug.tar.gz
+      #    path: ccache-debug.tar.gz
+      - name: upload sandstorm tarball
```
kentonv: Jake points out that uploading the tarball for every build is going to eat up our storage space -- we get 2GB free, and then GitHub will start charging me. Maybe we shouldn't bother?

ocdtrekkie: On additional research, I believe for a public repo it still doesn't cost us anything. It just... feels wasteful to upload every build. Are we likely to download this when troubleshooting tests?

abliss:
I set it to 9999 because  I was having trouble with 0, but I think I solved
that trouble some other way, so I can set it back.

I thought uploading the bundle might be useful to e.g. check the
functionality introduced in a PR, or to have a handy way to run sandstorm
off the tip of a feature branch. But I'm happy to remove the upload if
others don't think it'll be useful.

ocdtrekkie:
I think #3241 fixes these questions successfully? And was fun to play with.

PR #3214, opened 2020-02-11
Docs: Change vagrant-spk port to 6090

ocdtrekkie:
This changes all the vagrant-spk tutorials to specify using port 6090 instead of port 6080, which should be true for new packages made after the merge of https://github.com/sandstorm-io/vagrant-spk/pull/260

I also drive-by-deleted a line about the minimum version of vagrant-spk required to use enter-grain, because that was four years ago and the Docs reflect the current state of vagrant-spk anyways, which is very much different than four years ago.

### DIFF
--- a/docs/vagrant-spk/packaging-tutorial-meteor.md
+++ b/docs/vagrant-spk/packaging-tutorial-meteor.md
@@ -155,26 +155,6 @@ noise.
 Eventually, you will get your shell back. At this point, you can
 continue to the next step.
 
-**Troubleshooting note**: If the `vagrant-spk vm up` command fails, it
-could be because you already have Sandstorm installed on your
-laptop. You can recognize this error via the following red text:
-
-```bash
-Vagrant cannot forward the specified ports on this VM, since they
-would collide with some other application that is already listening
-on these ports. The forwarded port to 6080 is already in use
-on the host machine.
-```
-
-If you see that, run:
-
-```bash
-sudo service sandstorm stop
-```
-
-and halt any other `vagrant-spk` virtual machines you might be using
-to develop other apps.
-
 ## Connect your app to your local Sandstorm server
 
 Apps run differently in Sandstorm, compared to `meteor deploy`, so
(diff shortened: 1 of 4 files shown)


ocdtrekkie:
I felt the troubleshooting notes could be deleted wholesale because they primarily related to collision with Sandstorm installs, which will no longer happen, and details about vagrant-spk collisions are detailed lower in the tutorial. (Nobody should have a vagrant-spk collision until they have gone through the tutorial at least once, I hope. And the troubleshooting note didn't explain how to fix this case anyhow.)

docs/vagrant-spk/packaging-tutorial-meteor.md
```diff
@@ -224,8 +204,6 @@ but it does mean that the domain name running Sandstorm needs
 DNS. You can rest assured that your interactions with
 `local.sandstorm.io` stay entirely on your computer.
 
-<!--(**Editor's note**: We should make localhost:6080 work, so that people don't have to learn about `local.sandstorm.io`.)-->
```
zenhack: I think this is still relevant, just the port number needs updating too.

docs/vagrant-spk/packaging-tutorial.md
```diff
@@ -196,9 +177,6 @@ the experience of using your app as other users.
 Over the next few steps, we will prepare the app so that it is visible in that
 Sandstorm instance.
 
-<!--(**Editor's note**: We should make localhost:6080 work, so that people don't have to learn about `local.sandstorm.io`.)-->
```
zenhack: Duplicate of the other comment, but same thing, this is still relevant, just the port is out of date.

docs/vagrant-spk/packaging-tutorial-meteor.md
ocdtrekkie: We have this issue tracked as https://github.com/sandstorm-io/vagrant-spk/issues/73 and I marked it as a goal for vagrant-spk 1.1. I figured we should delete these because invisible comments in the docs page is not a good place to track the issue.

zenhack: Works for me; as long as we're still tracking it.

ocdtrekkie:
Fun, I actually had to re-run the Action because the Sandstorm grain gave me a 500 error, which I assume was a fluke.

Issue #2520, opened 2016-09-09
Redundancy of "was: Untitled Wekan board" and identical titles


neynah:
When the previous grain title is "Untitled Wekan board" or [the same grain title as it currently is] I think displaying "was:" text here is redundant and adds no value. I think we should not display previous titles in cases where the title is "Untitled..." or the same title as it currently is. Curious as to what other's think.
![image](https://cloud.githubusercontent.com/assets/855341/18398148/474cf394-7680-11e6-93b2-06fac139c072.png)

kentonv:
I agree that if the titles match exactly the "was" shouldn't be displayed -- but I thought it did that already, so that seems to be a bug.

Regarding "Untitled...", what we should really do is prevent people from sharing a grain that starts with "Untitled...". Google Docs does this: it prompts you to choose a title before you share.

neynah:
>Regarding "Untitled...", what we should really do is prevent people from sharing a grain that starts with "Untitled...". Google Docs does this: it prompts you to choose a title before you share.

Yes, we should do this!

dwrensha:
> I agree that if the titles match exactly the "was" shouldn't be displayed -- but I thought it did that already, so that seems to be a bug.

One way something like this can happen is:

  1. Alice shares grain A with Title "first title" to Bob
  2. Bob changes the title to "second title"
  3. Alice changes the title to "second title"
  4. Bob now sees the grain as "second title (renamed from: second title)"

PR #3220, opened 2020-02-13
postMessage API for getting the grain's title.

zenhack:
The subscription part mentioned in the docs is not implemented yet. I plan on digging into this a bit more later, but wouldn't mind a pointer as to where a good place to add the subscription would be.

### DIFF
--- a/docs/developing/path.md
+++ b/docs/developing/path.md
@@ -80,6 +80,8 @@ sample](https://github.com/Azeirah/brainstorm/blob/ca01c7d2b0ae7f0480b93d7e37e19
 and [per-page-load code
 sample](https://github.com/paulproteus/semantic-mediawiki-sandstorm/blob/445151c033a85da5e586d1a401abea8179b599b2/resources/src/startup.js#L64).
 
+## Using the grain's title in your app
+
 If you wish to make use of the grain's title inside your app, you
 can obtain this information by making a postMessage request:
 
--- a/docs/developing/path.md
+++ b/docs/developing/path.md
@@ -80,6 +80,35 @@ sample](https://github.com/Azeirah/brainstorm/blob/ca01c7d2b0ae7f0480b93d7e37e19
 and [per-page-load code
 sample](https://github.com/paulproteus/semantic-mediawiki-sandstorm/blob/445151c033a85da5e586d1a401abea8179b599b2/resources/src/startup.js#L64).
 
+If you wish to incorporate the grain's title into the page's title, you
+can obtain this information by making a postMessage request.
+
+```js
+var getGrainTitleRpcId = 0;
+
+// Sandstorm will reply via postMessage, so we need to set up a handler:
+window.addEventListener('message', function(event) {
+  if(event.source !== window.parent) {
+    // SECURITY: ignore messages not from the parent.
+    return;
+  }
+  if(event.data.rpcId === getGrainTtileRpcId) {
+    // Set the page title here.
+  }
+})
+
+// Now make the request:
+window.postMessage({
+  getGrainTitle: {},
+  rpcId: getGrainTitleRpcId,
+  // If subscribe is true, sandstorm will push future updates to the
+  // grain's title. If it is false or absent, the app will not be
+  // notified of updates.
+  subscribe: true,
+}, '*')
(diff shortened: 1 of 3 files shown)


ocdtrekkie:
I love the combined PR of code + docs, as that was a big reason for the initial decision to host Docs inside the main repo!

I would give this new documentation it's own subheading within the page though, as I feel it is a different activity than updating the URL and page title from your app.

zenhack:
@ocdtrekkie, done.

kentonv:
I can't quite remember if there was some reason why we didn't support this. I can't think of a good one at the moment but it feels like something I've thought about before, and apparently we never did it...

Did you happen to find or remember any old conversations on this topic?

zarvox:
My vague recollections are that the grain's title is a property unique to the user viewing the grain (petname), rather than a globally-unique name, and I forget offhand if the implementation is taking into account the viewer's conceptualization of the grain title or the original (owner's) grain title, or what the denormalization story was.

ocdtrekkie:
That sounds familiar.

zenhack:
Right, so this should only be used for doing transient UI things, but
apps might abuse it by either storing the title persistently and using
it somehow, or doing things that have global effects. I can see the
argument for not allowing this from fear of misuse. At a minimum, if we
do this, the docs should make clear what the appropriate uses are.

Quoting Jacob Weisz (2020-02-15 14:57:12)

kentonv:
Hmm I guess if Jake shared a grain with me and I renamed it (for myself) "Stupid Jake's dumb document that I hate", it'd be unfortunate if that name was communicated with the grain. :)

However, I have trouble imagining that this is a realistic danger, and refusing to tell the app about the grain title probably harms overall UX.

But I suppose it would be best if we could specifically tell the app about the grain's title _as set by the owner_, and not about any renames made by a user with whom it was shared...

ocdtrekkie:
With #2520 and all, I kinda wonder how valuable people having their own grain titles really is, considering the complexity it adds. It feels like a "multiple identities" type feature to me. Are there any other platforms where people share documents with each other where each user can set their own name on the document?

Would it be plausible to say in this case that the grain has a canonical official title, that's used by this API, but leave the existing petname functionality for people who want to leave themselves notes on what a grain is, knowing that the official grain title need not be hidden or treated as information that shouldn't be shared?

Perhaps the "was:" functionality could be transitioned to a more simple system where the (was:) always shows the owner-official grain title, if it differs from how you renamed it?

kentonv:
I think the realistic use cases for petnames are:

* The grain owner has chosen the title poorly, or left it "untitled", and the recipient wants a way to clarify for themselves what the heck this thing is. (Note that we could also -- like Google Docs does -- block users from sharing "untitled" grains, to solve the most common case of this problem.)
* The grain owner shares a grain, and then subsequently renames it, leaving the recipient confused when they go to look for the grain again later or see a grain with a title they don't recognize. By remembering the title at the time of sharing, we are able to add the "was:" note to help with this. (I'm not sure how often this really happens in practice, though...)

Use cases that are NOT realistic include:

* The recipient wants to add private notes to the grain title (as in my example above). I don't think anyone does this or would reasonably expect it to be hidden from the grain owner.
* Defense against a grain owner maliciously renaming a grain to match some other grain for a phishing attack. I don't think this is really prevented, since the grain owner could also name the grain maliciously _before_ sharing. I think to avoid this problem we need to more clearly indicate the provenance of a grain in your grain list. Awkwardly, we currently have a policy of not revealing the owner's identity, nor the identity of the person who shared with you (if different)... I wonder if that was a mistake.

ocdtrekkie:
Would it be fair to suggest a feature such as this API should always return the grain owner's title?

zenhack:
I'm fine with always returning the grain owner's title. I'm not sure whether the PR does that as-is.

There are conflicts, but I'm going to wait until master is fixed before I try to resolve them.

kentonv:
Yes, I think the API should always return the grain owner's title.

zenhack:
Ok, finally got around to finishing this up. It now always uses the owner's title, per discussion, and the `subscribe` option now works. I also rebased on top of master.

PR #3221, opened 2020-02-13
Implement support for AppHooks

### LINKED ISSUE #3027: Provide a mechanism for apps to export capabilities while using sandstorm-http-bridge

zenhack:
Resolves #3027.

I'm still debugging an issue or two, but otherwise this is representative.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,7 @@ EKAM=ekam
 #   in libstdc++. See also issue #3171.
 METEOR_DEV_BUNDLE=$(shell ./find-meteor-dev-bundle.sh)
 METEOR_SPK_VERSION=0.3.2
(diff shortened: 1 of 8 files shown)


zenhack:
Ok, tests are passing, I'm removing the WIP tag.

PR #3222, opened 2020-02-14
Allow apps using the bridge to respond to powerbox requests & offers

zenhack:
Starting with just spec'ing out the solution I described in #3197 precisely; I've done the capnproto parts of this, and will push subsequent commits that update other relevant parts of the docs. Once that's in place and folks are happy with the API, I'll start implementing. I'll probably wait until #3221 is merged and rebase on top of that to avoid potential conflicts.

PR #3228, opened 2020-02-17
sandstorm-http-bridge: BridgeContext: make the sessions map private.

zenhack:
Per the TODO.

I found myself wanting this when working on #3222, figured it made sense to submit it as its own pr. I'll probably send a couple other cleanup patches; there are some bits of the relevant code that have some duplication and/or are a bit hard to follow.


zenhack:
On second thought I'm going to close this and just include the patch in that pr; it's looking like the merge conflicts are going to be a bit annoying if we apply this separately.

PR #3231, opened 2020-02-18
Frontend refactor: move everything to `imports/`

zarvox:
tl;dr move everything to the imports/ folder and take explicit responsibility for load order.  Eventually, move toward a model where each file imports everything it depends on, and there are very few imports made strictly for their side-effects to the namespace.

Modern JS project structure and tooling has changed a lot in the past few years, and the layout of the sandstorm shell limits what tools we can readily make full use of.  For instance, it would be very nice if we could rely on ESLint to tell us if we're making use of a global that we haven't imported or to eventually use TypeScript to tell us we're failing to handle a possible shape of data.

Meteor has moved to a model where they generally encourage folks to make use of `npm` over Meteor packages, and Sandstorm already relies on many more npm packages than third-party Meteor packages.  I believe the only Meteor packages that we use that are *not* maintained by either us or by Meteor themselves are the `iron:router` package and the `fourseven:scss` package, the latter of which does not even impact JS.  In the interest of simplifying the answer to the questions "where does code come from?", we can move our own packages into `imports` and prefer to use a single npm `package.json` to pull in dependencies.

This will allow easier refactoring as all load order questions within Sandstorm become answerable in one place -- an explicit series of imports in a single server or client main file.  We can over time reduce the number of imports-for-side-effects by restructuring code to only make use of things that it explicitly imports (and thus, that something else explicitly exports).  Then, we can burn down the length of the import list in the client/server main files.

This pull request moves all JS code and HTML templates from `client/`, `server/`, `shared/`, `lib/`, and `packages/` to live in reasonable places under `imports/`.  It sets up a single client `main.js` and server `main.js` entry point for the client and server, which each explicitly import all the files that Meteor would previously have done in the same load order.


ocdtrekkie:
To my knowledge, nobody is currently working on significant frontend work right now. I am not super versed in the JS world, but I feel the shell both should be the easiest place to iterate, and is the part of Sandstorm that I feel has been kept up with the least for a while. Ensuring working on it is more accessible to modern JS developers is a really worthy goal.

Also, hi!

zenhack:
It's exciting to see an old face jump back in.

Generally positive about this change; I am in favor of explicit imports & exports and minimal import-time side effects. As @ocdtrekkie mentions I don't think anyone has any WIP stuff on the shell right now. The stuff I'm working on is on the bridge, @abliss has been working on CI stuff.

Re: typescript, now that we've updated to meteor 1.8.2 I think we can just `meteor add typescript` and start writing code with `.ts` extensions.

I'd like to keep the window of "don't touch anything" small if possible though. How long do you think this would take to get past that period?

zarvox:
This branch now covers "move all the code into `imports/` but still relies on all the side-effect ordering.  That said, I wouldn't mind landing what I have here so far if it passes tests; this makes it pretty clear where code is supposed to live going forward. :)

The next steps will be approximately:

* go through each of the files imported in each `main.js` for their side effects from top to bottom (and be careful about the ones that are imported on both client and server).
* for each file:
  * explicitly import everything that file depends on (including all the Meteor packages and whatnot)
  * extract the things they modify in the global namespace into some object or objects that they explicitly export
* remove the file from the imports in main.js and add imports of the newly-exported symbols in each file in the list that cares about observing those side-effects.

zarvox:
I'm hitting some test failures when I run the testsuite locally, so probably don't merge this quite yet unless you have reason to believe I haven't broken anything.  Trying to figure out if they're just flakiness or true issues.

Also, once upon a time we had tests that ran automatically against PRs?  Do those still exist?

ocdtrekkie:
Four tests currently fail in master. Bit of detail on #3208 comments.

ocdtrekkie:
If #3208 gets merged, which it should, since it only fails tests that Kenton also fails when he builds on his machine, then we will have tests run automatically against PRs. The Jenkins server is long gone, but that PR gets tests running under GitHub Actions (which is free! woo!)

zarvox:
This PR is failing the same four tests locally, plus one more:

``` ✖ tests/basic

   - Test title (1.494s)
   Testing if the page title equals "Sandstorm".  - expected "Sandstorm" but got: ""
       at Object.Test title (/home/zarvox/git/sandstorm/tests/tests/basic.js:27:15)
       at _combinedTickCallback (internal/process/next_tick.js:132:7)
   SKIPPED:
   - Test login command```

Seems possible that importing a template with a `<head><title>Sandstorm</title></head>` section doesn't actually apply it to the document, and I need to figure out how to do that with imports.

**Comment by zarvox** [comment]
https://www.meteor.com/tutorials/blaze/templates is not particularly helpful here.  It appears to be carefully sidestepping what appears to be the fact that importing a template that contains a `<head>` section does not add it to the <head> on initial pageload by leaving that bit in the template that they leave in `client/main.html`.  After doing likewise, this PR passes the same suite of tests that are passing on master (so the integration tests minus the four known failures detailed in #3208).

**Comment by kentonv** [comment]
Awesome work!

This change will probably require more thorough manual testing than usual, to catch obscure breakages not covered by the tests. But, I think it makes sense to merge now to minimize conflicts with any other work, and then we'll just have to spend some time testing things before the next release.

Also, it sounds like subsequent work will be more incremental and less likely to conflict with any concurrent changes, which is great.

**Comment by zenhack** [comment]
Master now looks like this for me:

Screenshot_2020-02-22 Sandstorm

Did you confirm this was working before merging?

**Comment by zenhack** [comment]
From sandstorm.log:

```
kj/compat/http.c++:5109: error: HttpService threw exception after generating a partial response; too late to report error to client; exception = kj/compat/http.c++:1851: failed: previous HTTP message body incomplete; can't write more messages
stack: 5747e0 573ba0
kj/compat/http.c++:5048: error: unhandled exception in HTTP server; exception = kj/compat/http.c++:1851: failed: previous HTTP message body incomplete; can't write more messages
stack: 5747e0 57c290
kj/compat/http.c++:5109: error: HttpService threw exception after generating a partial response; too late to report error to client; exception = kj/compat/http.c++:1851: failed: previous HTTP message body incomplete; can't write more messages
stack: 5747e0 573ba0
kj/compat/http.c++:5048: error: unhandled exception in HTTP server; exception = kj/compat/http.c++:1851: failed: previous HTTP message body incomplete; can't write more messages
stack: 5747e0 57c290

```

zenhack:
False alarm; when fixing the merge conflicts on #3155, somehow I ended up with a `changelog.html` not part of the repo. Not sure how, but deleting it fixes this.

kentonv:
I think this is because `changelog.html` is generated during `make` (based on `CHANGELOG.md`), and this PR changes the location of that file, so the old one was left sitting around in our work trees, and Meteor happily incorporated that into the build, leading to a duplicate-defined template, leading startup to throw an exception before registering any routes, hence the Iron Router error page.

zenhack:
Aha, that makes sense.

PR #3242, opened 2020-02-23
`meteor add typescript`

zenhack:
First step towards starting to use typescript. This adds the meteor package, which causes the meteor build system to compile `.ts` and `.tsx` extensions to javascript. I also renamed a couple trivial files to make sure doing so doesn't break anything.

From here we can convert old code at our leisure, and write new code in typescript where convenient.


zenhack:
Huh, build is failing. I'll investigate further.

ocdtrekkie:
Is this working then at this point?

zenhack:
Yeah, the last commit I pushed fixed the build. This is good to go.

kentonv:
yay!

PR #3244, opened 2020-02-23
Docs: Remove Google Fonts

ocdtrekkie:
This replaces Lato and Roboto Slab from Google Fonts with sandstorm.io's hosted Source Sans Pro. Right now, we don't have a monospace font in sandstorm-website, so it'll use a system font for code quotes, which we should improve. (I am considering Source Code Pro, since it's designed to complement Source Sans Pro.)

It looks like the color on the sidebar was modified for Sandstorm, but it was a much blue-r purple than common to Sandstorm. I took the purple from the leftmost side of the app store header, which I felt immediately made the Docs feel more unified with the Sandstorm brand.

I went for minimum work to succeed when using `mkdocs serve` here. Arguably the font styles should be extracted from `theme.css` and inserted into `extra.css`, and I may reorganize this later.

I'd love some others to test how they feel this looks (the `mkdocs serve` instructions in the README are pretty straightforward) before it's merged.


ocdtrekkie:
As a note to my theme.css modifications, all I did was Find/Replace `"Lato"` with `"source sans pro","Lato"` and `"Roboto Slab"` with `"source sans pro","Roboto Slab"` as it's not got any whitespace and was horrific to try and read/sort through.

ocdtrekkie:
Still had my test view visible on my other computer, thought a sample picture with the purple would be nice to have here.

image

zenhack:
LGTM

Quoting Jacob Weisz (2020-02-24 08:39:01)

PR #3245, opened 2020-02-27
Update docs theme

zdb999:
Updating to the latest version of the ReadTheDocs theme, which has a nested depth of 4 by default and can be configured for more.

The theme comes from [here](https://github.com/mkdocs/mkdocs/tree/master/mkdocs/themes/readthedocs).


roadmap/self-hosting/README.md
```diff
@@ -42,6 +40,10 @@ Post-install, configuration options can be modified by the administrator within
 - The URL of the main Sandstorm UI. (`BASE_URL`) Changing this also resets the OAuth configuration, since OAuth identity providers require that you specify the callback URL when creating the client key.
 - The IP address and port to which the server binds. (`BIND_IP` and `PORT`)
 
+### TODO(feature) Docker
```
ocdtrekkie: Two notes here:
1. I can't merge requests that amend the roadmap in addition to the docs.
2. This roadmap document probably shouldn't have items added to it via community contributors. It is more or less as Kenton wrote/envisioned it. We use issues/feature requests to track things we'd like to happen on the project.

roadmap/self-hosting/README.md
```diff
@@ -67,6 +69,8 @@ Every update has a build number, and updates are only allowed to increase the nu
 
 A new release branch starts out "beta" but is marked "stable" at such a point as we feel it has proven itself in the wild.
 
+Admins should be able to change their update settings if they wish from the graphical user interface, without a fresh install.
```
ocdtrekkie: Just as a note, while there's no way to configure the sandstorm.conf file from the GUI (it wouldn't make sense for many of the options in that file, as they control the ability to see/access the GUI), you don't need a fresh install to change it. Just a service restart after changing the line in the config file. I'd like to see auto-update settings make it into the GUI at some point, though right now we don't have different branches anyways, so it's mostly an on/off thing.

zdb999:
I apologize. I'm relatively new to git, and I accidentally collapsed multiple things I was working on into the same branch by mistake. I'll try to figure out how to remove those.

ocdtrekkie:
@zdb999 I am terrible at git, FWIW. I usually end up taking the files I want to commit to my Windows PC where I can play with GitHub Desktop's UI for it. So no worries.

It's actually pretty exciting to hear if the outcome of this entails us being able to multi-tier our docs without switching to a wholly different theme!

zdb999:
Also, I should have checked more carefully for the customization in the old version of the template. I thought the css changes were all in `extra.css`. I checked that file line by line and all of its updates seemed to work. However, I should have been more careful to check within `sandstorm_readthedocs_modified_theme`. I can try to make those changes. Alternatively, you can reject this pull request and I'll resubmit when I've made the fixes you've identified. (I don't know if I can rebase/change the history on an active pull request.)

ocdtrekkie:
I’ve seen people force push over pull requests, but for me personally I prefer to just make a new one than figure out how to do that.

The customizations aren’t super well organized. There’s the original theme, and then mkdocs’ modifications and then our modifications, some of which are in the right place and some of which I just hacked in. #3244 has a comment explaining how I modified the fonts over, since the CSS file I tampered with is minified.

PR #3246, opened 2020-02-28
Update docs theme to enable greater depth

zdb999:
Updating to the latest version of the ReadTheDocs theme, which has a nested depth of 4 by default and can be configured for more.

The theme comes from [here](https://github.com/mkdocs/mkdocs/tree/master/mkdocs/themes/readthedocs).

This is my second attempt at this change, hopefully addressing the issues identified in https://github.com/sandstorm-io/sandstorm/pull/3245.


ocdtrekkie:
This looks much cleaner! I will need to find a bit of time this weekend to set it up on my machine and review properly. But my at-a-glance through the changes doesn't spot anything.

ocdtrekkie:
So I got an error trying to run this about not being able to detect the navigation_depth variable: `jinja2.exceptions.UndefinedError: 'mkdocs.theme.Theme object' has no attribute 'navigation_depth'
`

When I added navigation_depth: 4 to mkdocs.yml below the custom_dir: entry (at the same indent level), it worked.

Otherwise, despite a little jankiness, it seems to work alright. Part of me wonders if we should be updating mkdocs to 1.1 (which requires updating our instructions to Python3, of course, and grabbing the updated theme outright rather than dumping so much into the "modified theme". But on the other hand, I do really want to do even more tweaks.

zdb999:
I just made the modification you suggested. MkDocs 1.1 doesn't require that explicitly -- that's what I have on my laptop, and what I used for development. It would be good to upgrade at some point, but I don't think it's worth much time and effort at the moment. Also, in the future we could un-jank it by using the stock theme and overriding it properly, but I think we have something that works for now.

ocdtrekkie:
mkdocs says it drops Python 2 support in the 1.1 changelog, and 1.1 wasn't an option when I tried changing from setting to 1.0.4 in the mkdocs serve instructions. I am a Python noob and didn't want to hack around it, but we probably need to adjust the README and the like for Python 3.

I agree that we can use what works for now though. I did have a font-awesome error of some kind in the browser console while testing it locally too. Are you able to see that as well? I am not sure what FontAwesome is being used for here, but it didn't seem to work.

ocdtrekkie:
(I'd really like to use mkdocs' built-in FontAwesome rather than us committing it to the repo here, so I'm hoping it's just a "pointed at the wrong folder" issue we can patch up.)

zdb999:
I just added the font awesome folder back in. I took it out with the initial commit because it seemed unnecessary on my local machine, but the theme actually came with it.

ocdtrekkie:
Now I can’t merge this, that commit has a ton of fonts. :/ There should be a way to have it pull FontAwesome from the base theme.

zdb999:
Which base theme? The MkDocs repo?

zdb999:
Should I just remove the folder I added?

ocdtrekkie:
Another commit to remove it wouldn't help. So the issue I'm trying to avoid here, is that as git keeps the entire history of the project, if you add stuff to the repo, it is kept forever. When you clone the repo, you download everything from every commit. This isn't a big deal for code, but it sucks for binaries like images, fonts, and compiled executables. We do have images and fonts committed to the repo, but we don't want to do it more than absolutely necessary. (As an example, all the images in the Active Directory page on Docs are hosted on a Davros grain instead of being in the repo in order to avoid adding them to the repo.) This wasn't something I thought about much before, but other members of the community highlighted it recently.

There are ways to remove commits from a branch, or force push over a branch, but I might also just make a branch off the commit prior to your font commit so it's excluded from what I merge, and then see if I can fix the font issue.

So, the way this should work, is that mkdocs ships with the readthedocs theme. Then it overwrites any files in that with anything we have in our modified theme folder. This is why we currently only have a small portion of the theme in the folder, right, the rest is inherited from mkdocs. So we shouldn't need to add FontAwesome to our repo to be able to use it: It's installed with mkdocs when generating the docs for publishing. But for some reason, this PR isn't seeing FontAwesome when I tested, which probably means something like the relative reference is broken, which should be a pretty easy fix.

I think I can pull your existing commits into a new branch, add a fix if I can find it, and then have a PR that includes your commits and mine. Give me a day or two to find time to play with it.

zdb999:
I just undid the folder add and force pushed it, so that should no longer be in the history.

ocdtrekkie:
You know what? I suspect it works fine if we deploy it. I think this is a case where mkdocs serve might be failing in a way the actual product won't? Let me see if I can figure out how to make a test grain for Docs.

ocdtrekkie:
Okay, so, apparently, the version of mkdocs we have only has the woff file, and it's trying to load the woff2, which is where the error comes from. This is fine, it loads the woff. It's good enough for now until we upgrade our scripts to use mkdocs 1.1.

Issue #3225, opened 2020-02-15
meteor-testapp seems to fail for me


kentonv:
Starting from a clean build, the spk I get seems to fail at startup:

```
** SANDSTORM SUPERVISOR: Starting up grain. Sandbox type: userns
** Starting Mongo...
2020-02-15T23:47:57.245+0000 I STORAGE  Engine custom option: log=(prealloc=false,file_max=200KB)
about to fork child process, waiting until server is ready for connections.
forked process: 10
child process started successfully, parent exiting
** Starting Meteor...
/programs/server/boot.js:356
}).run();
   ^

Error: The babel-runtime npm package could not be found in your node_modules 
directory. Please run the following command to install it:

  meteor npm install --save babel-runtime

    at meteorInstall.node_modules.meteor.babel-runtime.babel-runtime.js (packages/babel-runtime.js:37:9)
    at fileEvaluate (packages/modules-runtime.js:197:9)
    at require (packages/modules-runtime.js:120:16)
    at packages/babel-runtime.js:139:15
    at packages/babel-runtime.js:150:3
    at /programs/server/boot.js:303:34
    at Array.forEach (native)
    at Function._.each._.forEach (/programs/server/node_modules/underscore/underscore.js:79:11)
    at /programs/server/boot.js:128:5
    at /programs/server/boot.js:352:5
** HTTP-BRIDGE: App server exited with status code: 1
*** lost connection to grain (probably because it shut down) ***
```

cc @zenhack

zenhack:
I was able to fix this by putting a cal call to `meteor npm install` just above this line:

https://github.com/sandstorm-io/meteor-spk/blob/bc9121a049e7f30c3bf075f4ebda5dcbd4cf1038/meteor-spk#L71

I don't think I understand why there's a similar call in the server directory, and whether that's entirely redundant with this or covers some other case that's important.

It seems like the fix though is to add that and then update meteor-spk.

zenhack:
I'm tempted to suggest we junk meteor-spk for the purposes of the test app, and just use raw spk packaging. Reasoning:

[...]

PR #3249, opened 2020-03-01
Update meteor-testapp to Meteor 1.8.2 / meteor-spk 0.5.0

kentonv:
Another update will be needed to work with Meteor 1.9, but first Sandstorm itself needs to be updated, which requires node-capnp to be updated for Node 12.


ocdtrekkie:
This PR appears to exchange the "babel runtime" error documented in #3225 with:

> Error: Cannot find module "capnp"

meteor-testapp/package.json
```diff
@@ -5,6 +5,7 @@
     "start": "meteor run"
   },
   "dependencies": {
+    "@babel/runtime": "^7.8.4",
```
zenhack: Hm, should we have both `@babel/runtime` and `babel-runtime`? Not sure if there's actually a difference or if this is just a duplicate.

kentonv: IIRC if I tried removing either one, it broke... but I have no idea, feel free to play with it further.

kentonv:
> This PR appears to exchange the "babel runtime" error documented in #3225 with:
>
> > Error: Cannot find module "capnp"

WTF it was totally working on my machine, but now I'm seeing that too.

kentonv:
I fixed `meteor-spk` to put the `capnp` module where Meteor will find it, and also updated it to Meteor 1.10.1, and updated the testapp to use this version.

xet7:
@kentonv 

Have you uploaded that fixed meteor-spk? I get capnp errors when starting Wekan that uses meteor-spk 0.5.1.

kentonv:
@xet7 I just realized I hadn't pushed the git branch, did that now. But if you grabbed the 0.5.1 tarball, yes, that's the one that should work.

Issue #775, opened 2015-08-28
New UI: Send feedback really needs to be a form.


ocdtrekkie:
A lot of users do not have working `mailto:` due to webmail clients. On my work machine, my experience was seeing the briefest of popups on Edge which promptly disappeared. No indication of what was supposed to happen was visible.

ocdtrekkie:
Likely equally true for "Contact support".

kentonv:
I agree this should eventually be a form but we probably aren't going to be able to implement it for launch...

PR #3252, opened 2020-03-03
Have "send feedback" link to GitHub issues

ocdtrekkie:
We should have people open GitHub issues with problems/suggestions/feedback rather than emailing Kenton. As it was, I found this link irritating before (#775), because unexpected mailtos often launch unintended email clients in the days most people use webmail.

I chose to not use /new here like I did for Docs, because if you aren't logged into GitHub, that'll present a login screen. I think that's okay for my Docs PR, where the link text says "open an issue on GitHub", and hence, people clicking GitHub will expect to login. But for "Send feedback", sending them to the issues page is the cleanest solution.

docs.microsoft.com literally now just opens GitHub issues for any "feedback" on documentation pages, so this is a totally unheard of pattern, and the community can respond often faster than Kenton can.

### DIFF
--- a/shell/imports/client/accounts/login-buttons.html
+++ b/shell/imports/client/accounts/login-buttons.html
@@ -91,7 +91,7 @@ licenses, included in the LICENSES directory.
     {{/if}}
 
     {{#if showSendFeedback}}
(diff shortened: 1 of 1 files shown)


kentonv:
Very nice.

Not using `/new` is probably the right call anyway since an existing issue might answer the user's question.

PR #3253, opened 2020-03-04
Docs: Remove offers of services we don't offer

ocdtrekkie:
Since Asheesh is not here to help promote your app, Nena is not here to design your app icons, and Oasis is not here to provide free service to app developers, we should remove all of that. I think eventually community-sponsored initiatives can be added in here, but let's clean this out for now.


zenhack:
LGTM.

Quoting Jacob Weisz (2020-03-04 10:41:51)

PR #3256, opened 2020-03-06
powerbox-server: dedup code from async-helpers.

zenhack:
Per the (now-deleted) comment, this previously wasn't this simple
because powerbox-server.js was inside of a meteor package, but @zarvox
did a recent refactoring that changed that, so we can just delete the
duplicate and import now.


zarvox:
LGTM

Issue #3212, opened 2020-02-11
Developer docs should be sorted into more categories


ocdtrekkie:
Docs has a limitation of two levels of docs based on the theme. This may justify a theme switch someday. But for now I think we can use "Developing with..."

Right now there is stuff about vagrant-spk dev and spk dev and Powerbox APIs all in one big poorly-sorted category. I would rather break down into:

1. Developing with vagrant-spk
- vagrant-spk tutorial
- platform stacks
2. Developing with spk
- spk tutorial
- raw packaging
3. Developing with meteor-spk (though this might be too specific for top-level placement)
4. Developing with docker-spk (docs don't exist yet, maybe should?)
- docker-spk tutorial
5. Integrating apps with Sandstorm
- powerbox
- email from apps

ocdtrekkie:
I am not sure where #3213 would fit into either this or the existing docs. Platform-specific packaging info is a bit disparate between "raw Sandstorm hacking" info and vagrant-spk stack info. Perhaps we need another section about how to package apps written in specific languages/frameworks, including what strategies are available.

ocdtrekkie:
Perhaps as meteor-spk and docker-spk are less-used and less-documented, we should have an "Other tools" heading which has single-page documents on each.

zenhack:
i think I'd put 1-4 under a broader category of "Buidling and packaging apps."

I agree with having an "Other tools" section.

ocdtrekkie:
So, the problem is, there's about a half-dozen vagrant-spk-specific pages, and we can't do multiple layers with our current Docs theme. So we either have to dump all of our different dev tools' pages into one big category like we have now (I hate this), have a category for each dev tool (proposed above), or find an alternative theme that will let us have more levels of documentation (possibly worth investing time in).

zenhack:
Ah, that's annoying. I guess my take then would be to use your proposal
above for the short term, and maybe look into another theme in the long
term.

Quoting Jacob Weisz (2020-02-26 12:39:40)

zdb999:
MkDocs has updated their version of ReadTheDocs, and the new version allows for much deeper nesting. I pulled in the latest copy, and have verified locally that the custom css modifications still work correctly, and opened a pull request (https://github.com/sandstorm-io/sandstorm/pull/3245) to allow for greater nesting.

zdb999:
[...]

PR #3257, opened 2020-03-08
Reorganize docs and add back scroll bar

zdb999:
This is a doc reorganization attempt now https://github.com/sandstorm-io/sandstorm/pull/3246 has landed. It roughly follows https://github.com/sandstorm-io/sandstorm/issues/3212, but not exactly. This was just a first pass, and I am happy to make changes based on the community's feedback.

I also made some other small changes. In addition to the pure reorganization, I have:

* Modified `extra.css` to add back the scrollbar. Readthedocs has gotten rid of the scrollbar on their recent updates, but its really annoying not to have one give how long our TOC is.
* Enabled admonitions, a MkDocs markdown parsing feature that I think will be useful in the future.
* Rename the "Developer Handbook" "What makes a good Sandstorm app?" I felt that the old title was vague, and that the new one better describes what it actually does.


ocdtrekkie:
I like it. We can always iterate on this as we go.

PR #3258, opened 2020-03-08
Docs: Square off search box

ocdtrekkie:
Sandstorm.io and the App Market have a very angular look, as does the Sandstorm UI itself. A 4px border-radius is common on sandstorm.io buttons, and I think it makes Docs a bit more fitting with the overall branding.

### DIFF
--- a/docs/sandstorm_readthedocs_modified_theme/css/theme_extra.css
+++ b/docs/sandstorm_readthedocs_modified_theme/css/theme_extra.css
@@ -88,7 +88,7 @@ pre .cs, pre .c {
 
 form .search-query {
     width: 100%;
-    border-radius: 50px;
+    border-radius: 4px;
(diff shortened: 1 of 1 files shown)


zenhack:
With the caveat that I haven't actually built and looked at it, I like the idea behind this.

ocdtrekkie:
@zenhack So... neither have I... But I tested what it'd look like in Firefox dev console when I edited the only border-radius:50 to 4.

ocdtrekkie:
image

zenhack:
LGTM.

ocdtrekkie:
Hah. So it didn't work because GitHub can't find the border-radius:50px; that's buried in theme.css.

I did fix it on this search bar though: https://docs.sandstorm.io/en/latest/search.html?q=test

PR #3260, opened 2020-03-08
GitHub build workflow: don't build twice on prs.

zenhack:
Only build on prs and pushes to *master*, not arbitrary branches.

(I haven't actually tested this; we'll see what happens when I open the PR).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,5 @@
 name: build_and_test
-
-on: [push, pull_request]
-
+on: [push]
 env:
   CCACHE_DIR: ${{ github.workspace }}/.ccache"
   # TODO: remove this. See https://github.com/sandstorm-io/sandstorm/issues/3188


zenhack:
Hm, doesn't seem to be triggering the build at all. will investigate.

ocdtrekkie:
I guess the thing I thought about that concerns me with this approach is that when you push new commits to an open PR it won't run again automatically?

zenhack:
Hm. What if we *just* did it on push, not PR?

zenhack:
(Trying it)

zenhack:
Ah, I misunderstood the behavior: it only duplicates the run when you push to another branch on the sandstorm-io repo (and thus is not triggering for my new push since the branch is on my repo).

Maybe the solution is for you guys to just send prs from your own forks, rather than using branches on the main repo?

zenhack:
(actually, the reason it didn't run is because of an error in the config file; let's see how it behaves if I remove that).

zenhack:
Actions do seem to run, but they aren't showing up here as checks:

https://github.com/zenhack/sandstorm/actions/

I think I'm going to close this.

Issue #3218, opened 2020-02-12
Cleanup/refactor: use capnproto for spk's connection to sandstorm.


zenhack:
Right now the communication protocol between spk (for spk dev) and sandstorm is completely ad-hoc, which makes it difficult to extend. Now that the C++ implementation of capnproto supports fd passing, I don't *think* there are any reasons why this couldn't just use capnp rpc. At some point we should make this change.

If we had this in place now, it might make the implementation of #3215 a bit more straightforward, as we could just add an rpc method so spk could ask sandstorm for its base url, rather than having to do something ad-hoc like parse the config file (and be ready to fall back if it's not readable?).

PR #3261, opened 2020-03-09
Spec out a capnp version of the devmode protocol.

zenhack:
Per #3218.

Nothing is implemented yet, this specifies the protocol via a schema.

### DIFF
--- a/src/sandstorm/control-socket.capnp
+++ b/src/sandstorm/control-socket.capnp
@@ -1,0 +1,75 @@
+# Sandstorm - Personal Cloud Sandbox
+# Copyright (c) 2020 Sandstorm Development Group, Inc. and contributors
+# All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+@0xd4c5feffbd79e908;
+
+$import "/capnp/c++.capnp".namespace("sandstorm");
+
+using Util = import "util.capnp";
+
+interface Controller {
+  # Bootstrap interface offered by sandstorm on the control socket.
+  # This is used by some commands to control a running sandstorm
+  # server.
+  #
+  # This interface is not currently implemented. For now the commands
+  # referenced below use the `devmode` socket, and speak an ad-hoc
+  # protocol that provides similar functionality.
+  #
+  # The old protocol exists for historical reasons: when the devmode
+  # socket was first implemented, capnproto did not support passing
+  # file descriptors with a message. Now it does, so there is no
+  # need for a custom solution, and moving to capnproto will improve
+  # extensibility and maintainability.
+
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
I am going to put ready-for-review on this, because it needs review, even though it doesn't explicitly need to be merged at this juncture.

src/sandstorm/control-socket.capnp
```diff
@@ -0,0 +1,75 @@
...
+# limitations under the License.
+
+@0xd4c5feffbd79e908;
+
+$import "/capnp/c++.capnp".namespace("sandstorm");
+
+using Util = import "util.capnp";
+
+interface Controller {
+  # Bootstrap interface offered by sandstorm on the control socket.
```
kentonv: Could use a little more text explaining that this is a control protocol for the Sandstorm server itself.

src/sandstorm/control-socket.capnp
```diff
@@ -0,0 +1,75 @@
...
+
+  devShell @0 () -> (shellFds :List(FileDescriptor), cancel :Util.Handle);
+  # Stop the running sandstorm shell process, and get a list of file
+  # descriptors that the shell should inherit on startup.
+  #
+  # This is used by `sandstorm dev-shell`; the returned file descriptors
+  # will be passed to an instance of the shell launched from the developer's
+  # source tree, rather than from the installed sandstorm root.
+  #
+  # - `shellFds` is a list of file descriptors that the shell should inherit.
```
kentonv: When I read this my first question was: "What FD numbers should they be assigned? Like, does this include stdio FDs?" But looking at the code I see that's slightly complicated; it differs between dev mode and non-dev mode. Maybe worth a comment clarifying that these are sandstorm-specific FDs (not stdio) and the FD numbers to use depend on how you're invoking the shell.

src/sandstorm/control-socket.capnp
```diff
@@ -0,0 +1,75 @@
...
+  # socket was first implemented, capnproto did not support passing
+  # file descriptors with a message. Now it does, so there is no
+  # need for a custom solution, and moving to capnproto will improve
+  # extensibility and maintainability.
+
+  const socketPath :Text = "/var/sandstorm/socket/control";
+  # Path to the control socket, relative to the sandstorm install (normaly
+  # /opt/sandstorm).
+
+  devShell @0 () -> (shellFds :List(FileDescriptor), cancel :Util.Handle);
```
kentonv: I think handles used in this way are usually called `handle` elsewhere in Sandstorm protocols, rather than `cancel`. Not sure if it's worth being consistent.

kentonv:
Seems pretty reasonable, just some minor comment/naming nitpicks.

src/sandstorm/control-socket.capnp
zenhack: Hm, maybe it would be better to define a struct here that (roughly) corresponds to FdBundle, rather than just being an opaque list? The way I wrote this follows the current protocol closely, but the array feels like an implementation detail that on a second look I'm not sure makes sense anymore.

src/sandstorm/control-socket.capnp
zenhack: It does mention that right after this, though I've tweak the wording so hopefully it reads less ambiguously as "this is what this is for" rather than "these commands happen to use it for this."

src/sandstorm/control-socket.capnp
zenhack: Done.

src/sandstorm/control-socket.capnp
kentonv: Yeah that makes a lot of sense to me.

zenhack: Done.

PR #3262, opened 2020-03-10
Spk dev: display BASE_URL to the user.

zenhack:
Most of the diff (first commit) is just moving the config parsing into a separate file; the second patch just reads the value from the config and includes it in the message.

### DIFF
--- a/src/sandstorm/config.h
+++ b/src/sandstorm/config.h
@@ -40,6 +40,6 @@ struct Config {
 // out to id(1).  If false, we ignore SERVER_USER.
 Config readConfig(const char *path, bool parseUids);
 
-};
+}; // namespace sandstorm
 
 #endif
--- a/src/sandstorm/config.h
+++ b/src/sandstorm/config.h
@@ -1,0 +1,45 @@
+#ifndef SANDSTORM_CONFIG_H_
+#define SANDSTORM_CONFIG_H_
+
+#include <kj/string.h>
+
+namespace sandstorm {
+
+struct UserIds {
+  uid_t uid = 0;
+  gid_t gid = 0;
+  kj::Array<gid_t> groups;
+};
+
+struct Config {
+  kj::Maybe<uint> httpsPort;
+  kj::Array<uint> ports;
+  uint mongoPort = 3001;
+  UserIds uids;
+  kj::String bindIp = kj::str("127.0.0.1");
+  kj::String rootUrl = nullptr;
+  kj::String wildcardHost = nullptr;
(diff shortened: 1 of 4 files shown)


src/sandstorm/config.h
```diff
@@ -0,0 +1,45 @@
...
+  kj::Maybe<kj::String> stripePublicKey = nullptr;
+};
+
+// Read and return the config file from `path`.
+//
+// If `parseUids` is true, we initialize `uids` from SERVER_USER.  This requires shelling
+// out to id(1).  If false, we ignore SERVER_USER.
+Config readConfig(const char *path, bool parseUids);
+
+};
```
kentonv: Nitpick: Should be:

```
}  // namespace sandstorm
```

kentonv: Also in the `.c++` file.

src/sandstorm/run-bundle.c++
```diff
@@ -1689,128 +1511,12 @@ private:
   }
 
   Config readConfig() {
-    // Read and return the config file.
-    //
-    // If parseUids is true, we initialize `uids` from SERVER_USER.  This requires shelling
```
kentonv: So the code had a comment mentioning `parseUids`, but no such option actually existed?

But you've added this option in the factored-out code?

zenhack: Correct.

src/sandstorm/config.h
zenhack: Done.

src/sandstorm/run-bundle.c++
kentonv: Does the UID parsing not work in the `spk dev` use case, or did you only skip it because it's not needed?

zenhack:
Quoting Kenton Varda (2020-03-15 20:19:08)

>    Does the UID parsing not work in the spk dev use case, or did you only
>    skip it because it's not needed?

The latter. I don't have a strong opinion about this though.

src/sandstorm/config.h
kentonv: Looks like you left in the semicolons. FWIW the semicolon is redundant here -- though I'm not going to hold up merge on something like that.

kentonv:
> The latter. I don't have a strong opinion about this though.

OK, I don't have a strong opinion either, was just curious.

The CI failure is my fault, fixed by #3270. Interesting that the CI build for pull requests merges with master, rather than test your exact commit.

PR #3263, opened 2020-03-10
Mount a tmpfs at /dev/shm in the sandbox.

### LINKED ISSUE #3196: Consider putting a tmpfs at /dev/shm, to support shm_open and friends.
background: yesterday @orblivion hit an issue where postgres wouldn't start because it was trying to open something in `/dev/shm`, which doesn't currently exist in the sandbox. It may be possible to work around this, but it seems sensible to make a tmpfs available to grains at `/dev/shm` to generally improve compatibility with software that expects it.

zenhack:
...To get shm_open() and friends working.

Fixes #3196.

Note that the tests are currently *not* passing on my machine -- for some reason `shm_open` is returning `ENOSYS`. Based on the links @abliss posted at:

https://github.com/sandstorm-io/sandstorm/issues/3196#issuecomment-580909238

and the fact that there are two alternate implementations of `shm_open`, depending on build flags:

https://sourceware.org/git/?p=glibc.git;a=blob;f=rt/shm_open.c;h=b4623d3cdaf5a26e4fd795a866c12aa1b29eaa9a;hb=HEAD

...one possible explanation is that my system's glibc is built without support for this. So I'd be curious as to whether the tests pass for others. Grepping through the postgres source suggests it has some fallbacks if shm_open is not available, so another possibility is that the errors @orblivion saw re: `/dev/shm` weren't actually going through `shm_open`.

We should investigate and work out what's going on there before merging this.

### DIFF
--- a/src/sandstorm/supervisor.c++
+++ b/src/sandstorm/supervisor.c++
@@ -954,6 +954,12 @@ void SupervisorMain::makeCharDeviceNode(
   KJ_SYSCALL(mount(kj::str("/dev/", realName).cStr(), dst.cStr(), nullptr, MS_BIND, nullptr));
 }
 
+void mountTmpFs(const char *name, const char *dest) {
+    KJ_SYSCALL(mount(name, dest, "tmpfs",
+                     MS_NOSUID | MS_NODEV,
+                     "size=16m,nr_inodes=4k,mode=770"));
+}
+
 void SupervisorMain::setupFilesystem() {
   // The root of our mount namespace will be the app package itself.  We optionally create
   // tmp, dev, and var.  tmp is an ordinary tmpfs.  dev is a read-only tmpfs that contains
@@ -986,8 +992,7 @@ void SupervisorMain::setupFilesystem() {
     // 2) When we exit, the mount namespace disappears and the tmpfs is thus automatically
     //    unmounted.  No need for careful cleanup, and no need to implement a risky recursive
     //    delete.
-    KJ_SYSCALL(mount("sandstorm-tmp", "tmp", "tmpfs", MS_NOSUID,
-                     "size=16m,nr_inodes=4k,mode=770"));
+    mountTmpFs("sandstorm-tmp", "tmp");
   }
   if (access("dev", F_OK) == 0) {
     KJ_SYSCALL(mount("sandstorm-dev", "dev", "tmpfs",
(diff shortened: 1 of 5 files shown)


zenhack:
Looks like it's failing the same way in CI as it s on my machine:

https://github.com/sandstorm-io/sandstorm/pull/3263/checks?check_run_id=499117544#step:9:1407

kentonv:
According to the man page for `shm_open`, you need to link with librt (`-lrt`), which I think test-app currently isn't. Maybe you're seeing a stub function in the base glibc that would be overridden if you linked with librt?

zenhack:
I added `-lrt` in the Makefile (I got a linker error without it), so I don't think that's it, unless the test app uses different LDFLAGS than everything else? (but then I don't understand why adding in the makefile would have fixed anything, so that seems unlikely).

kentonv:
Oh woops, I thought I looked to see if you had added it and didn't see it, but given it's the very first thing in the diff I don't know how I could have missed it.

But... I dunno then.

ocdtrekkie:
As a note, this will also help get .NET Core working on Sandstorm: https://github.com/dotnet/runtime/issues/2534#issuecomment-602944041

zenhack:
Fixed. Just had a bogus leading slash in front of `dev/shm`. Derp.

zenhack:
@ocdtrekkie, feel free to tag this one ready-for-review.

src/sandstorm/supervisor.c++
```diff
@@ -997,6 +1002,18 @@ void SupervisorMain::setupFilesystem() {
...
+
+    // Create /dev/shm so shm_open() and friends work. Note that even though /dev
+    // is already a tmpfs, we need to mount a separate tmpfs for /dev/shm, because
+    // the former will be read-only.
+    //
+    // TODO: it might be nice to have /dev/shm and /tmp share the same partition,
+    // so we don't have to strictly separate their storage capacity. We could mount
+    // a single tmpfs somewhere invisible, create subdirectories, and then bind-mount
+    // them to their final destinations.
+    KJ_SYSCALL(mkdir("dev/shm", 0700));
```
kentonv, line 1014: It looks like the `/dev` filesystem itself is created with `0755` permission, any particular reason to use `0700` here? I guess it ultimately doesn't mater since there's only one user in the sandbox and they own everything?

zenhack:
Quoting Kenton Varda (2020-04-05 18:55:11)

>    It looks like the /dev filesystem itself is created with 0755
>    permission, any particular reason to use 0700 here? I guess it
>    ultimately doesn't mater since there's only one user in the sandbox and
>    they own everything?

I tend to default to being conservative with permissions, but yeah it
probably doesn't matter.

PR #3268, opened 2020-03-14
Update nl.i80n.json

FreekDankelman:
Dutch translation

### DIFF
--- a/shell/i18n/nl.i18n.json
+++ b/shell/i18n/nl.i18n.json
@@ -81,17 +81,17 @@
           "explanation": "Voor elke vriend die een deellink aanmaakt krijgt u 2 GB"
         },
         "isNotPaid": {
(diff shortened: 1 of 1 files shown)


xet7:
In my Finnish translation I did translate "Grains" to "Jyvät", because for Finnish speaking people it's easier to understand, and that word in English is not in common use. At Finland farmers use grains and land to grow some stuff. For me, grain is Sandstorm-style container, similar to Docker and Snap, with a difference that grains are only running when someone uses them.

ocdtrekkie:
@xet7 I think it may make sense to defer to a given translator on whether or not to translate certain terms, like Sandstorm and grain. Different locales may sway different ways on the matter.

xet7:
@ocdtrekkie 

Yes, it's a decision by the translator of course. It depends on culture, etc, what words are in use in each country.

kentonv:
Thanks!

PR #3274, opened 2020-03-20
Add basic eslint config, fix some issues.

zenhack:
These issues were caught by eslint and fixed:

- In a couple places were were setting global variables without
  explicitly declaring them. We now actually declare variables for
  these. One such case was of the form 'Foo = class Foo {', and is
  now just 'class Foo {', since the assignment was redundant anyway.
- A few files that used said globals did not explicitly import them; now
  they do.
- In once place we were using what is apparently deprecated syntax
  for octal literals. We now use the current 0oXXX syntax.

### DIFF
--- a/shell/.eslintrc.yml
+++ b/shell/.eslintrc.yml
@@ -1,0 +1,11 @@
+---
+parser: "babel-eslint"
+parserOptions:
+  ecmaVersion: 2020
+  sourceType: "module"
+
+  # There are a few places where we do if(Meteor.isServer) { import ... }.
+  # Ideally we'd refactor so this was unneccesary, but for now the import
+  # *has* to be conditional, so tell eslint to allow this:
+  allowImportExportEverywhere: true
+reportUnusedDisableDirectives: true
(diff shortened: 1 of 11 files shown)


zenhack:
Note that poking around in the source I think the set of files where global variables were being defined without declaration is much larger than what eslint actually caught here.

shell/.eslintrc.yml
```diff
@@ -0,0 +1,11 @@
+---
+parser: "babel-eslint"
```
zarvox: Any reason in particular for choosing `babel-eslint` over the default `espree` parser vs going straight to `@typescript-eslint/parser`?

zenhack: The default didn't work because there are a couple places where we're doing `if(Meteor.isServer) { import ... }` and it refuses to parse that (it's apparently not technically legal according to the spec).

I didn't try the typescript one; maybe I'll see how that goes.

zarvox: Ahhh, seems reasonable.  I'll merge this; feel free to switch to the typescript one in a followup patchset.

PR #3275, opened 2020-03-21
Actually start *checking* the typescript code

zenhack:
Apparently the meteor typescript package doesn't actually *check* the typescript source, it just translates `.ts` files to `.js` so the build works (grumble). This adds a tsconfig.json, runs tsc as part of CI, and fixes some things that cropped up.

Note that this is built on top of #3274, so the first commit is from that. I didn't want to do two parallel prs because they would conflict, since the lint target invokes both. That PR should be merged first.

### DIFF
--- a/shell/client/main.ts
+++ b/shell/client/main.ts
@@ -36,7 +36,7 @@
 // For more information on Meteor's load order, see
 // https://guide.meteor.com/structure.html#load-order
 
-import { Template } from "meteor/templating";
+import 'meteor/templating';
 
 // Load packages that the sandstorm shell depends on before sandstorm itself.
 
--- a/shell/client/main.ts
+++ b/shell/client/main.ts
@@ -36,8 +36,6 @@
 // For more information on Meteor's load order, see
 // https://guide.meteor.com/structure.html#load-order
 
-import 'meteor/templating';
-
 // Load packages that the sandstorm shell depends on before sandstorm itself.
 
 // sandstorm-identicons.
--- a/shell/package.json
+++ b/shell/package.json
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "lint": "eslint ."
+    "lint": "eslint . && tsc"
   },
   "author": "",
   "repository": {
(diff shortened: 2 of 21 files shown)


zenhack:
@zarvox, per the commit message, I removed the actual `Template` variable name but left the import of `meteor/templating` in place, but I *suspect* we can actually just remove it outright; it seems doubtful that module has side effects where we care about load order. Thoughts?

https://github.com/sandstorm-io/sandstorm/pull/3275/files#diff-0432ab27c4c846c40e8887d26bda0387L39-R39

shell/client/main.ts
```diff
@@ -36,7 +36,7 @@
 // For more information on Meteor's load order, see
 // https://guide.meteor.com/structure.html#load-order
 
-import { Template } from "meteor/templating";
+import 'meteor/templating';
```
zarvox: Yes, you can remove this line entirely.

shell/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "lint": "eslint ."
+    "lint": "eslint . && tsc"
```
zarvox: Your commit message claims you specify `--noEmit` here but I don't see it.  Am I missing something?

Separately: maybe put the `tsc` check before the `eslint` check; I anticipate it running faster.

shell/client/main.ts
zenhack: Done.

shell/package.json
zenhack: Re: noEmit, it's in the tsconfig.json file, which I apparently forgot to `git add`, oops. It's there now.

Re: ordering, I've observed the opposite: eslint is noticeably faster.

zarvox: If I understand correctly, eslint as configured right now is running very few checks on the codebase, which might explain the speed.  Happy to leave this as-is until there's an inflection point. :)

In due time, we'll probably want to enable a broader suite of eslint checks, like `extends: ["airbnb-base", "plugin:meteor/recommend", "plugin:import/errors", "plugin:import/warnings"]` to the eslintrc, which will do substantially more work/checks.  Doing so right now, however, will give us on the order of several tens of thousands of things to fix, and some of the "safe" automated changes (`--fix`) appear to break things in ways I haven't had a chance to investigate in full.  We can take this slowly.

zenhack: Yeah, that was my thinking: get it in place first and add checks incrementally.

PR #3276, opened 2020-03-23
Incremental frontend refactorings

zarvox:
I started going through the files imported in `main.ts` from the top and converting them from side-effect-ful imports to explicit imports and exports.  If you want to review this, going commit-by-commit is probably the way to go.  With the exception of the final commit (with a fix for what appears to be a long-standing bug), I believe this patch series does not change any behavior.

Identicons were first.  We largely used Identicon as a namespace; I've taken to flattening things out into just the bare functions when we're not instantiating a class or doing anything stateful.

I moved our dependency on fibers to be explicit in package.json, and moved several other things we used to pull in via `Npm.require` to `import` statements.  There's still a few `Npm.require`s floating around in conditional branches; I'm inclined to split things apart into separate client and server packages where appropriate.  There's whole swaths of methods on SandstormDb that are only implemented on client or on server, and plenty more that expose the same interface but have separate implementations.

I moved Capnp to explicit imports.  The style I've been trying to take for ordering imports is:

* imports from Meteor
* imports from npm packages, alphabetically
* imports from /imports

My position is that Capnp imports like `const Whatever = Capnp.importSystem(...)` should go after all the JS module imports.

Since almost all of our code uses double-quotes for strings today, I changed a couple instances of single-quotes introduced in recent patches to double-quotes for consistency.

I moved the side-effect-ful import of `db-deprecated.js` up from `00-startup.js` (where, before imports were available, we dumped things we needed to be loaded first for their side-effects) so that all side-effect-ful imports live in `main.ts`.

I moved `SandstormDb` to explicit imports and exports.

Lastly, I made some small cleanups to db.js including fixing what I believe is a bug in `refStaticAsset`, where we would incorrectly specify a query for the object for which we wanted to increment the refcount.

### DIFF
--- a/shell/client/main.ts
+++ b/shell/client/main.ts
@@ -38,11 +38,7 @@
 
 // Load packages that the sandstorm shell depends on before sandstorm itself.
 
-// sandstorm-identicons.
-import "../imports/sandstorm-identicons/identicon.js";
-import "../imports/sandstorm-identicons/helpers.js";
-
-// sandstorm-db.  Depends on sandstorm-identicons.
+// sandstorm-db.
 import "../imports/sandstorm-db/db.js";
 import "../imports/sandstorm-db/profile.js";
 
--- a/shell/client/main.ts
+++ b/shell/client/main.ts
@@ -138,6 +138,7 @@ import "../imports/client/grain/grainlist-client.js";
 import "../imports/client/setup-wizard/wizard.js";
 import "../imports/client/vendor/ansi-up.js";
 import "../imports/client/widgets/widgets-client.js";
+import "../imports/db-deprecated.js";
 import "../imports/client/00-startup.js";
 import "../imports/client/admin-client.js";
 import "../imports/client/demo-client.js";
--- a/shell/imports/blackrock-payments/client/billingPrompt.js
+++ b/shell/imports/blackrock-payments/client/billingPrompt.js
@@ -14,7 +14,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import { Meteor } from "meteor/meteor";
 import { Template } from "meteor/templating";
+import { SandstormDb } from "/imports/sandstorm-db/db.js";
 
 var idCounter = 0;
 
--- a/shell/imports/blackrock-payments/client/billingSettings.js
+++ b/shell/imports/blackrock-payments/client/billingSettings.js
@@ -15,6 +15,7 @@
 // limitations under the License.
 
 import { Template } from "meteor/templating";
+import { SandstormDb } from "/imports/sandstorm-db/db.js";
 
 var messageListener = function (showPrompt, template, event) {
   if (event.origin !== window.location.protocol + "//" + makeWildcardHost("payments")) {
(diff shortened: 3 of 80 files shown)


zenhack:
LGTM.

Probably unrelated, but I noticed looking at the CI output though, we seem to be hitting new errors:

https://github.com/sandstorm-io/sandstorm/pull/3276/checks?check_run_id=526835831#step:10:113

I feel like I've hit that once or twice when developing locally. Is phantomjs used anywhere outside the 'unit tests' that @kentonv disabled?

zarvox:
Looks like building `meteor-testapp.spk` is also reaching it.  I have very little context on that.

Does a nonzero exit code not translate to a failed github check?  Seems like having presubmit checks quietly fail with a pair of green checkmarks is not what we want.

zarvox:
https://timonweb.com/posts/how-to-fix-referenceerror-primordials-is-not-defined-error/ suggests that the error we're seeing is a result of using node 12 with something that depends on `graceful-fs@^3.0.0`.  That is to say: `sandstorm/tests` is probably currently incompatible with node 12, and probably needs to take a bunch of version upgrades.

Installing/running the tests themselves with an older version of npm/node appears to work locally for now, but it's probably a good idea to upgrade everything there too.

zenhack:
> Does a nonzero exit code not translate to a failed github check? Seems like having presubmit checks quietly fail with a pair of green checkmarks is not what we want.

Currently it does not, because unfortunately there are a handful of known failures. @ocdtrekkie put together a milestone:

https://github.com/sandstorm-io/sandstorm/issues?q=is%3Aopen+is%3Aissue+milestone%3A%22Fix+tests+%282020%29%22

meteor-testapp is broken for other reasons, see: https://github.com/sandstorm-io/sandstorm/issues/3225

phantomjs has been unmaintained for a few years now, so it wouldn't surprise me at all if it doesn't work with current node. `meteor-testapp` shouldn't actually need it; it runs the tests in chrome.

---

...but it seems like this is all unrelated to this pr.

zarvox:
This is still seeing two new test failures relative to master -- tests/appdemo and tests/schedule appear to be failing:

```
 ✖ tests/appdemo

   - Test appdemo link (43.904s)
   Timed out while waiting for element <.demo-startup-modal .start> to be present for 30000 milliseconds.  - expected "visible" but got: "not found"
       at Object.module.exports.Test appdemo link (/home/zarvox/git/sandstorm/tests/tests/appdemo.js:39:6)
       at processTicksAndRejections (internal/process/task_queues.js:79:11)

 ✖ apps/schedule

   - Test one-shot tasks (9.911s)
   Testing if element <.grainlog-contents > pre> contains text: "Running job oneshot".  - expected "Running job oneshot" but got: "...** SANDSTORM SUPERVISOR: Starting up grain. Sandbox type: userns"
       at common (/home/zarvox/git/sandstorm/tests/apps/schedule.js:60:13)
       at Object.module.exports.Test one-shot tasks (/home/zarvox/git/sandstorm/tests/apps/schedule.js:100:3)
```

So I probably have some investigation to do still.

zarvox:
Fixed those two new test failures (relative to current master).  Seems to have to do with trying to move away from the global `isSignedUpOrDemo()` to the instance method on `SandstormDb`, so I just unwound that line in that commit.

This should be good to merge now -- it passes the same set of tests that master does.

PR #3280, opened 2020-03-27
admin panel: add an explanation for network capabilities.

### LINKED ISSUE #2039: What are Network capabilities?
Might be worth adding a short description about what this page is for.

User tested said 

> its unclear what granted by and granted to refer to in network capabilities. Theres no data yet b/c it is a new server. Do users or apps grant network access?

image

@paulproteus proposes "(i) Sandstorm's <a href="...">security model</a> means that apps have no network access until granted to them. This page enables granting network access."

zenhack:
Fixes #2039

@kentonv mentioned on that issue that the page is likely to cover other types of capabilities at some point. I am of the opinion that it makes sense to wait until that actually happens to update the explanation; otherwise I think it will just seem overly abstract (especially without other examples).

This will need translation (cc @xet7, who else?). (It would be nice if we had a team for translators so when submitting prs that change UI text we could just cc @sandstorm-io/translators or something).

### DIFF
--- a/shell/i18n/en.i18n.json
+++ b/shell/i18n/en.i18n.json
@@ -1202,6 +1202,7 @@
     "networkCapabilities": {
       "name": "Network capabilities",
       "subtext": "Manage direct network access for grains.",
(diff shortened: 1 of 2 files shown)


ocdtrekkie:
As a note, I am 👍 on the bite size fix parade.

PR #3281, opened 2020-03-27
install.sh: on busybox systems, use adduser

### LINKED ISSUE #3272: Install script fails on Alpine Linux (no useradd command)
Apparently Alpine Linux doesn't have a `useradd` command, it has `adduser` instead, so the install script fails here if it is asked to create the sandstorm user:

https://github.com/sandstorm-io/sandstorm/blob/2131dcb7cc28177416980c427ff3304651e6cb93/install.sh#L1183

Ironcally, the comment above that says we're using usermod because adduser isn't generally available. Perhaps we could try usermod and fall back to adduser if it's not available.

Relevant discussion on IRC:

https://freenode.irclog.whitequark.org/sandstorm/2020-03-19#26600332;

zenhack:
...if useradd does not exist. See the comment.

Fixes #3272

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1174,13 +1174,28 @@ create_server_user_if_needed() {
 
   # OK! Let's proceed.
   #
-  # useradd comes from the passwd package, whereas adduser is a Debian-specific utility
-  # <https://packages.debian.org/adduser>, so we prefer useradd here. Per the man page for useradd,
-  # USERGROUPS_ENAB in /etc/login.defs controls if useradd will automatically create a group for
-  # this user (the new group would have the same name as the new user). On systems such as OpenSuSE
-  # where that flag is set to false by default, or on systems where the administrator has personally
-  # tuned that flag, we need to provide --user-group to useradd so that it creates the group.
-  useradd --system --user-group "$SERVER_USER"
+  # To create the server user, we first try `useradd`, which is widely available on most
+  # distros. If that isn't available, it's likely we're running on a busybox based system.
+  # busybox provides an `adduser` command, so if the busybox binary exists we try that
+  # instead.
+  #
+  # Note that debian provides an `adduser` command as well, but its usage is different.
+  # useradd is available on debian anyway, so we'll end up using that.
+  if which useradd >/dev/null; then
+    # Per the man page for useradd, USERGROUPS_ENAB in /etc/login.defs controls if useradd
+    # will automatically create a group for this user (the new group would have the same
+    # name as the new user). On systems such as OpenSuSE where that flag is set to false
+    # by default, or on systems where the administrator has personally tuned that flag,
+    # we need to provide --user-group to useradd so that it creates the group.
+    useradd --system --user-group "$SERVER_USER"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -1174,13 +1174,28 @@ create_server_user_if_needed() {
...
+  # Note that debian provides an `adduser` command as well, but its usage is different.
+  # useradd is available on debian anyway, so we'll end up using that.
+  if which useradd >/dev/null; then
+    # Per the man page for useradd, USERGROUPS_ENAB in /etc/login.defs controls if useradd
+    # will automatically create a group for this user (the new group would have the same
+    # name as the new user). On systems such as OpenSuSE where that flag is set to false
+    # by default, or on systems where the administrator has personally tuned that flag,
+    # we need to provide --user-group to useradd so that it creates the group.
+    useradd --system --user-group "$SERVER_USER"
+  elif which busybox >/dev/null; then
```
kentonv, line 1191: Hmm, `busybox` is commonly installed on normal Linux systems, so the existence of `busybox` doesn't necessarily mean that we're currently running under it. Is there a better way to detect this? Or maybe you really want to check for the presence of `adduser` and `addgroup`?

zenhack: Well, it should only fall back to this if it can't find `useradd`. I decided to check for busybox rather than `adduser`/`addgroup` since the latter might potentially be other (incompatible) commands with those names. Perhaps we should call `busybox adduser` to make sure we're not running some other command.

I think running the busybox versions when available should be ok, even if it's not the system default -- they should make the same modifications to `/etc/passwd` and friends, no?

kentonv: On my system, `busybox adduser` doesn't work:

```
$ busybox adduser
adduser: applet not found
```

And it looks like your usage here is indeed incompatible with the Debian adduser / addgroup commands, which have no short flags, only long flags.

Maybe you can detect busybox like this? https://stackoverflow.com/a/24884579

zenhack:
Quoting Kenton Varda (2020-04-06 10:30:32)

>    On my system, busybox adduser doesn't work:
> $ busybox adduser
> adduser: applet not found

My guess is the `adduser` command wasn't compiled in to the version of
busybox on your system; subcommands (applets) can be turned on or off at
compile time.

> Maybe you can detect busybox like this?
> https://stackoverflow.com/a/24884579

That talks about detecting if the currently running shell is busybox,
which is not what we want (and it never should be, since our script
starts with #!/bin/bash). But I guess we could add an extra check to
make sure that adduser is a link to busybox; something like:

    [ "$(basename $(readlink $(which adduser)))" = busybox ]

...but I think using `busybox adduser` should work if busybox includes
it. I guess this might result in better error messages?

kentonv:
Ah right, I forgot busybox is not just used as a shell, but also as a symlink target.

Yeah I guess checking if the symlink links to busybox sounds like the right approach.

zenhack:
Quoting Kenton Varda (2020-04-11 11:21:08)

> Yeah I guess checking if the symlink links to busybox sounds like the
> right approach.

Done.

PR #3285, opened 2020-03-29
Workaround node 8.x SSL issue

### LINKED ISSUE #3284: fix node SSL bug
The version of node bundled with sandstorm, 8.15.1, seems to have a bug in its SSL implementation. As a result, tiny-tiny-rss cannot use sandstrom-httpGet to fetch certain feeds (for example, ones on https://nitter.net).

Minimal example:
```
$ /opt/sandstorm/latest/bin/node -e "require('https').get('https://nitter.net', (e,r,b) => {console.log(e);});"                                         events.js:183
      ^

Error: write EPROTO 140122876737344:error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure:../deps/openssl/openssl/ssl/s23_clnt.c:802:

    at WriteWrap.afterWrite [as oncomplete] (net.js:868:14)
```

I'm not sure when exactly the bug was fixed. It might be https://github.com/nodejs/help/issues/1730 . My system node is v10.18.1 and it is able to fetch.

abliss:
Fixes #3284.

(I confirmed that applying this patch locally restores the ability of ttrss to fetch feeds from https://nitter.net .)

### DIFF
--- a/meteor-bundle-main.js
+++ b/meteor-bundle-main.js
@@ -14,6 +14,11 @@ var https = require('https');
 var net = require('net');
 var url = require('url');
 
+// fix ssl for some sites on node v8.x
+// TODO(abliss): remove this after upgrading node to v10.x
+// see https://github.com/nodejs/node/issues/16196#issuecomment-393091912
+require("tls").DEFAULT_ECDH_CURVE = "auto";
+
 var isDevShellMode = process.argv.length > 2;
 
 var firstInheritedFd = isDevShellMode ? 65 : 3;


ocdtrekkie:
As noted in #3284, Sandstorm is supposed to be on Node 12 with the latest version, I think?

kentonv:
Yes, we're on node 12 as of the last update (three weeks ago). Are you sure this is still needed? If so I guess the comment needs updating.

xet7:
@kentonv 

Does meteor-spk work on node 12 ?

kentonv:
@xet7 I haven't had a chance to update it yet. :/

abliss:
Nope, probably not still needed.

PR #3291, opened 2020-04-12
Fix many places we fail eslint's `no-undef` rule

zarvox:
I mostly went for things that were trivial to find/replace with `globalDb`, but
there's a few typos fixed, missing `const`s, and a couple of places where it
looks like we have been naming the wrong variable for some time.

This brings us down to 706 no-undef lint failures from >1000.  There are still
many failures that will require more in-depth refactoring, but at least this
helps reduce the noise so we can find them more easily.

I believe this should make no behavior changes, besides fixing existing
brokenness around the few typos.

### DIFF
--- a/shell/imports/blackrock-payments/server/payments-api-server.js
+++ b/shell/imports/blackrock-payments/server/payments-api-server.js
@@ -16,6 +16,7 @@
 
 import Crypto from "crypto";
 import { Meteor } from "meteor/meteor";
+import { check } from "meteor/check";
 
 import Capnp from "/imports/server/capnp.js";
 const PaymentsRpc = Capnp.importSystem("sandstorm/payments.capnp");
@@ -251,7 +252,7 @@ BlackrockPayments.registerPaymentsApi =
         if (user) {
           sendInvoice(this._db, user, this._invoice, this._config);
         } else {
-          console.error("Stripe charge didn't match any user: " + chrage.id);
+          console.error("Stripe charge didn't match any user: " + charge.id);
         }
       }, err => {
         // The error could be because the charge was already captured, but the only indication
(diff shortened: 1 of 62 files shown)


shell/imports/server/installer.js
```diff
@@ -508,7 +508,7 @@ getAllManifestAssets = (manifest) => {
 
   const handleLocalizedText = (text) => {
     if (text.defaultTextAssetId) {
-      result.push(defaultTextAssetId);
+      result.push(text.defaultTextAssetId);
```
zarvox, line 511: This appears to have been a bug

shell/imports/sandstorm-db/scheduled-jobs-db.js
```diff
@@ -160,7 +160,7 @@ SandstormDb.prototype.scheduledJobIncrementRetries = function (id) {
   check(id, String);
 
   this.collections.scheduledJobs.update(
-    { _id: job._id },
+    { _id: id },
```
zarvox, line 163: This appears to have also been a bug.

shell/imports/blackrock-payments/server/payments-api-server.js
```diff
@@ -251,7 +252,7 @@ BlackrockPayments.registerPaymentsApi =
         if (user) {
           sendInvoice(this._db, user, this._invoice, this._config);
         } else {
-          console.error("Stripe charge didn't match any user: " + chrage.id);
+          console.error("Stripe charge didn't match any user: " + charge.id);
```
zarvox, line 255: This was a bug, but a pretty immaterial one that would only get hit if we got a spurious Charge object from Stripe, which would be quite unusual.

zarvox:
Same 4 end-to-end tests failing as on master

PR #3293, opened 2020-04-15
Fix more `no-undef` eslint warnings.

zenhack:
This gets rid of about 70 of them. I also set the warning to `"warn"` in the config, so it won't fail CI but we can still see the output.

### DIFF
--- a/shell/imports/sandstorm-autoupdate-apps/autoupdate-apps-tests.js
+++ b/shell/imports/sandstorm-autoupdate-apps/autoupdate-apps-tests.js
@@ -23,6 +23,7 @@ import { Meteor } from "meteor/meteor";
 import { Accounts } from "meteor/accounts-base";
 
 import { SandstormDb } from "/imports/sandstorm-db/db.js";
+import { SandstormAutoUpdateApps } from "/imports/sandstorm-autoupdate-apps/autoupdate-apps.js";
 
 const globalDb = new SandstormDb();
 // TODO(cleanup): Use a lightweight fake (minimongo-based?) database here and construct a clean
--- a/shell/imports/sandstorm-autoupdate-apps/autoupdate-apps-tests.js
+++ b/shell/imports/sandstorm-autoupdate-apps/autoupdate-apps-tests.js
@@ -24,7 +24,7 @@ import { Accounts } from "meteor/accounts-base";
 import { HTTP } from "meteor/http";
 
 import { SandstormDb } from "/imports/sandstorm-db/db.js";
-import { SandstormAutoUpdateApps } from "/imports/sandstorm-autoupdate-apps/autoupdate-apps.js";
+import { SandstormAutoupdateApps } from "/imports/sandstorm-autoupdate-apps/autoupdate-apps.js";
 
 const globalDb = new SandstormDb();
 // TODO(cleanup): Use a lightweight fake (minimongo-based?) database here and construct a clean
--- a/shell/imports/sandstorm-autoupdate-apps/autoupdate-apps-tests.js
+++ b/shell/imports/sandstorm-autoupdate-apps/autoupdate-apps-tests.js
@@ -21,6 +21,7 @@ import Crypto from "crypto";
 
 import { Meteor } from "meteor/meteor";
 import { Accounts } from "meteor/accounts-base";
+import { HTTP } from "meteor/http";
 
 import { SandstormDb } from "/imports/sandstorm-db/db.js";
 import { SandstormAutoUpdateApps } from "/imports/sandstorm-autoupdate-apps/autoupdate-apps.js";
--- a/shell/imports/server/00-startup.js
+++ b/shell/imports/server/00-startup.js
@@ -24,6 +24,7 @@ import { migrateToLatest } from "/imports/server/migrations.js";
 import { ACCOUNT_DELETION_SUSPENSION_TIME } from "/imports/constants.js";
 import { onInMeteor } from "/imports/server/async-helpers.js";
 import { monkeyPatchHttp } from "/imports/server/networking.js";
+import { SandstormAutoUpdateApps } from "/imports/sandstorm-autoupdate-apps/autoupdate-apps.js";
 let url = require("url");
 
 process.on('unhandledRejection', (reason, p) => {
--- a/shell/imports/server/00-startup.js
+++ b/shell/imports/server/00-startup.js
@@ -15,6 +15,7 @@
 // limitations under the License.
 
 import { Meteor } from "meteor/meteor";
+import { HTTP } from "meteor/http";
 
 import { SandstormDb } from "/imports/sandstorm-db/db.js";
 import { globalDb } from "/imports/db-deprecated.js";
--- a/shell/imports/server/00-startup.js
+++ b/shell/imports/server/00-startup.js
@@ -25,7 +25,7 @@ import { migrateToLatest } from "/imports/server/migrations.js";
 import { ACCOUNT_DELETION_SUSPENSION_TIME } from "/imports/constants.js";
 import { onInMeteor } from "/imports/server/async-helpers.js";
 import { monkeyPatchHttp } from "/imports/server/networking.js";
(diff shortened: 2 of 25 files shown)


zenhack:
Hm, a couple extra tests failed. Will investigate.

zenhack:
I can't reproduce the failure for `apps/schedule` locally. For `tests/autoupdate`, it hangs waiting for the topbar to show up while installing an app. That's weird because the `installApp` helper is used in a ton of other tests and they're all fine. I tried reverting 6bb7013, since that touches the topbar, but it had no effect. I'll stare at it more later.

shell/imports/sandstorm-autoupdate-apps/autoupdate-apps-tests.js
```diff
@@ -23,6 +23,7 @@ import { Meteor } from "meteor/meteor";
 import { Accounts } from "meteor/accounts-base";
 
 import { SandstormDb } from "/imports/sandstorm-db/db.js";
+import { SandstormAutoUpdateApps } from "/imports/sandstorm-autoupdate-apps/autoupdate-apps.js";
```
zarvox, line 27: `SandstormAutoUpdateApps` should be `SandstormAutoupdateApps` (the U in update should be lowercase).

shell/imports/server/00-startup.js
```diff
@@ -24,6 +24,7 @@ import { migrateToLatest } from "/imports/server/migrations.js";
 import { ACCOUNT_DELETION_SUSPENSION_TIME } from "/imports/constants.js";
 import { onInMeteor } from "/imports/server/async-helpers.js";
 import { monkeyPatchHttp } from "/imports/server/networking.js";
+import { SandstormAutoUpdateApps } from "/imports/sandstorm-autoupdate-apps/autoupdate-apps.js";
```
zarvox, line 28: Ditto here.

shell/imports/shared/testing.js
```diff
@@ -16,6 +16,7 @@
 
 import { Meteor } from "meteor/meteor";
 import { globalDb } from "/imports/db-deprecated.js";
+import { SandstormAutoUpdateApps } from "/imports/sandstorm-autoupdate-apps/autoupdate-apps.js";
```
zarvox, line 19: And here.

zarvox:
`tests/autoupdate` failed due to an error in the Meteor method that got called:

```
Exception while invoking method 'fetchAppIndexTest' ReferenceError: SandstormAutoupdateApps is not defined
    at MethodInvocation.fetchAppIndexTest (imports/shared/testing.js:67:9)
    at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1771:12)
    at packages/ddp-server/livedata_server.js:719:19
    at packages/ddp-server/livedata_server.js:717:46
    at packages/ddp-server/livedata_server.js:715:46
    at new Promise (<anonymous>)
    at Session.method (packages/ddp-server/livedata_server.js:689:23)
    at packages/ddp-server/livedata_server.js:559:43
```

I've noted the capitalization mismatch that caused this in https://github.com/sandstorm-io/sandstorm/pull/3293/commits/f39bd342963c61690b52aec6fb4a2e846b125465

zenhack:
Wouldn't it be nice if we had a tool that would catch silly typos like that for us? :P

Pushed a fix, let's see if that fixes the tests.

zenhack:
Hm, still seeing failures on CI but I can't reproduce them locally.

zenhack:
I assume this is related:

https://github.com/sandstorm-io/sandstorm/pull/3293/checks?check_run_id=591993444#step:10:2413

I don't see that error in my log locally. Will stare some more later.

zarvox:
I agree that the hash mismatch explains the failure mode for `tests/autoupdate`.  I don't know what to attribute that to, other than a transient network flake or something. :/

I'm also seeing a failures for `apps/collections`, which appears to be in a place where (locally) I've observed losing a race condition against intro.js before (based on the screenshots).  I'm not certain that's what happened here, but it seems very plausible that we're colliding badly with the introjs overlay.  I've opened #3294 with a patch that I think might help.

One more thing that we could do infra-wise to get a better sense of what fails in these test runs is to put the failure screenshots somewhere.  They get dropped into `<GIT_ROOT>/tests/screenshots`; if we could subsequently browse that directory or just upload all of its contents somewhere, that would show us what the browser window looked like at the time of the failure, which can be *very* informative when trying to debug hard-to-reproduce failures.

ocdtrekkie:
If those screenshots are already produced on test run, it should be possible to have the runner upload them as test artifacts, yes, @abliss?

abliss:
Yep, it should be as easy as a new yml stanza here: https://github.com/sandstorm-io/sandstorm/blob/2ea90d2ef35148031b3b4a7672a518bdec3a6b17/.github/workflows/build.yml#L80

zenhack:
Pushed a commit to upload the screenshots, Maybe this will give us some insight on the schedule tests. If the autoupdate thing was a transient network problem this will also let us know.

zarvox:
In the most recent test run, I'm only seeing one novel failure: `apps/schedule`

```
 ✖ apps/schedule

   - Test one-shot tasks (9.752s)
   Testing if element <.grainlog-contents > pre> contains text: "Running job oneshot".  - expected "Running job oneshot" but got: "...** SANDSTORM SUPERVISOR: Starting up grain. Sandbox type: userns"
       at common (/home/runner/work/sandstorm/sandstorm/tests/apps/schedule.js:60:13)
       at Object.module.exports.Test one-shot tasks (/home/runner/work/sandstorm/sandstorm/tests/apps/schedule.js:100:3)
```

Indeed, the screenshot of the browser window shows only the supervisor line.  I have a hypothesis for how this happened:

The test code leads me to believe that there may be a race condition between client and server.  Under normal operation, we'd expect the following sequence of events:

1. the test's `.click(selector)` causes the JS in the test app to make some HTTP or websocket request to the grain backend.
2. the grain's backend makes some request over capnproto RPC to the sandstorm meteor backend to schedule a job to be run in 5 minutes
3. the sandstorm meteor backend's capnproto interface implementation inserts some record into the mongo database
4. the test's `.execute('Meteor.call("runDueJobsAt", ' + firstCheck.toString() + ');')` triggers a request to the sandstorm meteor backend
5. the meteor backend runs a db query for scheduled jobs due at a particular time, finds a record, and triggers it.

But there's actually no synchronization forcing 4 to come after 3, so there's another sequence of events that is also possible:

1. the test's `.click(selector)` causes the JS in the test app to make some HTTP or websocket request to the grain backend.
2. the grain's backend makes some request over capnproto RPC to the sandstorm meteor backend to schedule a job to be run in 5 minutes
3. the test's `.execute('Meteor.call("runDueJobsAt", ' + firstCheck.toString() + ');')` triggers a request to the sandstorm meteor backend
4. the meteor backend runs a db query for scheduled jobs due at a particular time, finds no records, and triggers nothing.
5. the sandstorm meteor backend's capnproto interface implementation, responding to the RPC in step 2, inserts some record into the mongo database.  In five minutes, it might fire, but our test framework has given up by then.

The solution is to enforce the intended ordering with a barrier of some sort.  One way to go about that would be:

* Make the test app grain backend return something in the request to schedule the job, but only once the sandstorm backend has acknowledged the request.
* Make the test app frontend change something in the page when that request completes successfully (adding an element with an ID like `#success-oneshot` would do)
* Make the test itself wait on that frontend change to appear before calling the `runDueJobsAt` Meteor method.

zenhack:
I pushed a commit that just does a `.pause(short_wait)`. `short_wait` = 5 seconds, so if the race condition is what's causing the failure, then that should fix it -- if it's taking that long for the job to hit the DB something else is wrong. It'd be nice to rework the test not to rely on `.pause` so heavily, just because it makes it slower, but that's a bit more work and this is already kindof orthogonal to what this pr is trying to do.

zenhack:
OK -- back down to the usual 4 failures.

ocdtrekkie:
Should "only four failures" be a :tada:? I'm going with it.

The screenshot upload change to the Action is going to be a nice long-term improvement for testing too. :)

zarvox:
Thanks.  Filed #3295 to track fixing this the "right" way

PR #3294, opened 2020-04-18
Fix flakiness from race condition in installApp test helper

zarvox:
When navigating to /apps, it's possible that the introjs overlay will make the
.app-list element not "visible" (because the overlay is blocking the entirety
of the page).

It's also possible that we'll win that race, but then once we navigate to the
grain list page, we'll lose a race against a different introjs overlay.

Additionally, the introjs overlay does not appear to get reactively removed
when we mark that it's no longer needed in the Session.  So we need to catch
things early.

Making the call to `disableGuidedTour()` *before* navigating to any of the
pages that feature introjs helpers should help ensure that they never render
and never cause test failures.

It's possible we should rethink how we manage introjs and testing -- perhaps we
should disable the overlays earlier for testing purpose, like in user login.
There's only two tests that actually bother interacting with the introjs
overlays, and perhaps they could use a different login method (or pass a
parameter to decline disabling all introjs overlays).

I believe this patch may fix the flakiness around `apps/collections` that we're
seeing in the checks for #3293.

### DIFF
--- a/tests/commands/installApp.js
+++ b/tests/commands/installApp.js
@@ -29,7 +29,16 @@ exports.command = function(url, packageId, appId, dontStartGrain, callback) {
   var ret = browser
     .init()
     .url(this.launch_url + "/install/" + packageId + "?url=" + url)
-    .waitForElementVisible("#step-confirm", very_long_wait)
+    .waitForElementVisible("#step-confirm", very_long_wait);
+
+  if (!dontStartGrain) {
+    ret = ret
+      // The introjs overlay often doesn't destroy itself fast enough and intercepts
+      // clicks that we don't want it to intercept. So we manually disable it here.
+      .disableGuidedTour();
+  }
+
+  ret = ret
     .click("#confirmInstall")
     .url(this.launch_url + "/apps")
     .waitForElementVisible(".app-list", medium_wait)
@@ -37,9 +46,6 @@ exports.command = function(url, packageId, appId, dontStartGrain, callback) {
 
   if (!dontStartGrain) {
     ret = ret
-      // The introjs overlay often doesn't destroy itself fast enough and intercepts
-      // clicks that we don't want it to intercept. So we manually disable it here.
-      .disableGuidedTour()
       .click(appSelector(appId))
       .waitForElementVisible(actionSelector, short_wait)
       .click(actionSelector)


zarvox:
Okay, I see 8 test failures now.  There are a few explanations I can come up with:

* this patch has somehow broken 4 additional tests, in addition to the 4 that are known to be failing?
  * possible, but it seems unlikely based on the change?  I changed one file to do the exact same set of total things, in a slightly different order.  I do not really understand how this could have changed behavior so broadly.
* those tests were already flaky?
  * possible, though we haven't seen them flake here before
* something is weird with the test framework?
  * might be nice to see screenshots or console logs for the failures
* the infra we're running this on fails to actually make the changes we'd expect before the deadline (limited CPU/weird scheduling constraints)
  * seems unlikely that weird timing issues would last _30 seconds_ like some of the failing tests...
* something is dropping events/failing to deliver clicks?
  * who knows

Anyone have any insights?

zenhack:
I can reproduce all of those failures locally, so it's not the infra, and probably not them just already being flaky.

FYI, a while back I added a feature to the test script so you can do:

```
make test SHOW_BROWSER=true
```

And it will actually open a visible browser window when running the tests, so you can watch, rather than just seeing the final screenshot.

I'm attaching a tarball with the screenshots from my local failed run, though you can probably reproduce them locally if need be. Maybe later today I'll find some time to stare at this.

[screenshots.tar.gz](https://github.com/sandstorm-io/sandstorm/files/4497160/screenshots.tar.gz)

zarvox:
Looks like there were 8 failures in one test run "(push") and 9 in the other "(pull request)".  So there are 5 new failures in play.

Going through these:

* `tests/grain`: did not fail in your set of screenshots.  failed in infra on both test runs. (???)
* `tests/restore-open-grains`: screenshot still shows viewing the grain, as though the click on `.navitem-create-grain>a` never happened.  I find this curious.
* `tests/trash`: failed in your set of screenshots, did not fail on the "push" check, but did fail on the "pull request" check.  Expecting to see the "Share access" modal, but not seeing it in the screenshot.  Could be some issue in the topbar?  Worth following up on.
* `apps/backgrounding`: Screenshot doesn't show the debug log in the other browser window, which is where our assertion is firing (can't find the `.grainlog-contents > pre` element)
* `apps/powerbox`: The screenshot shows the element and the `offer: success` text that the next line asserts should be there.  I don't understand why this test failed at all, unless it was a performance issue and somehow things converged to the expected state between test failure and the screenshot being taken.

zarvox:
FWIW, the existing `tests/account-settings` failure looks to be failing in the same way that `tests/restore-open-grains` is failing -- a click on `.navitem-create-grain>a` does not appear to be causing any visible change in the app.

zenhack:
When I run these tests locally with `SHOW_BROWSER=true`, I see that:

*  `apps/powerbox` fails because the grain debug log never opens
* `apps/backgrounding` fails intermittently.
* I was still unable to reproduce the failure for `tests/grain`, but I noticed a feature-introducing pop-up that stayed on the screen for a while, so it's possible that's taking longer in CI and causing a failure?
* I didn't get failures on the other two this time, so there's some non-determinism there too.

If you rebase on top of master, now that #3293 is merged, you should get screenshots as a build artifact to look at what's happening in CI.

ocdtrekkie:
Looks like this one indeed brings our recurring test failure list up to 8. I did see one test that had 9.

zenhack:
Yeah, this wasn't ready to merge... Maybe we should revert? //cc @kentonv

PR #3302, opened 2020-04-27
Use es6 modules in the Sandcats ACME.js plugin.

zenhack:
@kentonv, the magic seems to be to put `"type": "module"` in package.json, but I had to actually pop it open and try it to figure it out, so I figured I'd just send a pr rather than make you replicate it.

...it would be nice if it were possible for other people to run these tests somehow.

### DIFF
--- a/acme-dns-01-sandcats/index.js
+++ b/acme-dns-01-sandcats/index.js
@@ -1,8 +1,8 @@
 'use strict';
 
-const https = require("https");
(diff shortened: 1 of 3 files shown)


kentonv:
Thanks! And sorry the test is crappy... the test framework (provided by acme.js) is designed to run against a real live service and I didn't have the patience to write code to have it register a new sandcats account or whatever for testing.

PR #3308, opened 2020-05-01
Docs: Provide advice for AD-over-LDAP config

ocdtrekkie:
The default searchUsername populated by Sandstorm is `uid`, which is not populated in Active Directory. I figured we should add an implementation note here on what setting(s) someone should consider when trying to use LDAP with Active Directory. I believe both choices presented in this PR are potentially valid choices depending on the environment.


zenhack:
I have no experience with AD and haven't really dug into our own ldap support, so I can't really comment much on the substance of the docs, but looks reasonable?

ocdtrekkie:
I'll take a "looks reasonable" with the part where this is a tested implementation. Also, this information is valid for both my new Active Directory domain created this week and my work's Active Directory domain created sometime in the 1990s. So I have fairly high confidence in this being very true about Active Directory domains.

PR #3311, opened 2020-05-02
Revert "Fix flakiness from race condition in installApp test helper"

kentonv:
Reverts sandstorm-io/sandstorm#3294

This seems to have created more test failures.


kentonv:
(@zenhack and @ocdtrekkie told me this should be reverted.)

PR #3135, opened 2019-05-03
Changed doc to reflect that Let's Encrypt can now issue wildcard hosts

Hugogerva:

PR #3312, opened 2020-05-02
Docs: Quick fix for FAQ entry about Let's Encrypt

ocdtrekkie:
A lot more docs need to be adjusted for this release, but we might as well get this out of the way. This supersedes #3135.

### DIFF
--- a/docs/administering/faq.md
+++ b/docs/administering/faq.md
@@ -516,60 +516,10 @@ that we provide. Pay special attention to the `RewriteRule` stanzas.
 
 ## Can I use Let's Encrypt for adding HTTPS to Sandstorm?
 
-**Short answer:** No. We wish we could say yes. We are eagerly awaiting support of wildcard
-certificates by Let's Encrypt. For advice that will work, read our [docs about configuring
-HTTPS.](ssl.md)
-
-**Medium answer:** You can use Let's Encrypt for [static
-publishing](../developing/web-publishing.md) on personal or corporate domain names like
-`mysite.example.com`, but not your main Sandstorm server interface.
-
-Here's the long answer.
-
-To have working HTTPS, Sandstorm needs a wildcard certificate for all the reasons documented in [our
-documentation on wildcard DNS.](wildcard.md)
-
-One hypothetical alternative to a wildcard certificate is dynamic integration with Let's Encrypt.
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
As a note, eventually this will be obvious/a non-issue not worthy of being mentioned on the FAQ page at all.

docs/administering/faq.md
```diff
@@ -516,60 +516,10 @@ that we provide. Pay special attention to the `RewriteRule` stanzas.
...
-
-One approach is to use `*.example.sandcats.io` for wildcard hostnames, but use a `BASE_URL` of `example.com`
-where `example.com` is certified via Let's Encrypt. However, this would mean that `example.com` contains
-IFRAME elements pointing at `*.example.sandcats.io` and these IFRAMEs may not be able to set cookies within
-the subdomain due to the increasingly-common **third-party cookie-blocking** done by browsers such as Chrome
-and Firefox.
+Yes, support for Let's Encrypt was added in 0.263. To use Let's Encrypt, your Sandstorm server needs
+to be able to modify your DNS records. This is made possible by [the ACME.js library](https://git.rootprojects.org/root/acme.js) as long as
+you use a supported DNS provider (Sandcats.io, Cloudflare, Digital Ocean, DNSimple, Duck DNS, GoDaddy,
+Gandi, Namecheap, Name.com, AWS Route 53, or Vultr). You can now configure this in the admin panel.
```
zenhack, line 522: A couple thoughts:

- I don't think we need to call out what library we're using here.
- We should mention that it's possible to manually upload a cert if needed (useful if the user's DNS provider is not supported), though the user will need to keep it up to date themselves.

ocdtrekkie: My thought is that the library retains relevant regarding issues with or interest in the supported providers.

I didn't want to go crazy about things like manual upload options here because this is just an FAQ and I wanted to purge the suggestion that it can't be done.

Perhaps the right thing to do would be to come back and add a link here to the HTTPS docs page when that is revised as well?

zenhack: I think given that you also explicitly list the supported providers it's still not important. But otherwise I'm not too picky; I agree the most important thing is to remove the incorrect information.

ocdtrekkie: I am on mobile now, I will merge and then revise it again alongside the HTTPS page update and remove the extraneous information.

PR #3314, opened 2020-05-04
Remove redundant implementation of waitPromise.

zenhack:
This had been previously blocked on the fact that sandstorm-db was a
meteor package, and thus couldn't call into the server import.


kentonv:
Meteor must have a built-in way of doing this these days, I wonder what it is.

PR #3315, opened 2020-05-04
Use es6 modules in the Sandcats ACME.js plugin (again).

zenhack:
This apparently didn't land on master, because @kentonv merged #3299 into master before merging #3302 into the `lets-encrypt-sandcats` feature branch. So, new pr for the same patch.


kentonv:
Looks like what actually happened is I merged it, but then I forgot to pull the changes down to my local client, and then I did a rebase in my local client and force-pushed those up, so it got lost...

PR #3316, opened 2020-05-04
More eslint fixes.

zenhack:


shell/imports/db-deprecated.js
```diff
@@ -66,11 +66,6 @@ Misc = globalDb.collections.misc;
 Settings = globalDb.collections.settings;
 
 currentUserGrains = globalDb.currentUserGrains.bind(globalDb);
-isDemoUser = globalDb.isDemoUser.bind(globalDb);
-isSignedUp = globalDb.isSignedUp.bind(globalDb);
-isSignedUpOrDemo = globalDb.isSignedUpOrDemo.bind(globalDb);
```
zarvox, line 71: I vaguely recall attempting a similar refactor myself, and discovering that calling functions on `globalDb` that had implementations that referenced a `this` wound up with `this` not referring to `globalDb` but whatever was `this` in the calling scope, and that causing breakage in many, many places.

zenhack: /me looks at `db.js`. Yuk. We probably should hold off on this until and first refactor `SandstormDb` as a proper class instead of doing `_.extend(SandstormDb.prototype, ...)`...

zenhack:
I've updated this to just the stuff in blackrock-payments; we can defer the globalDb stuff for later. Hopefully the tests will pass now.

PR #3317, opened 2020-05-04
Docs: Add sniproxy guide to share TLS port

ocdtrekkie:
Okay, so, I rebased #1588 (and cleaned up the merge conflict with ssl.md) and then applied Asheesh's review notes. I have not adjusted the formatting of `sniproxy` where appropriate yet, nor added this doc to the sidebar, both of which I will do before merging this. I also may do a bit more review of my own, I haven't given it a clean read-through yet. I also removed the diagram because the link is dead.

@pjamar if there is any chance you still have that diagram somewhere, I would love to see it! <3

### DIFF
--- a/docs/administering/sniproxy.md
+++ b/docs/administering/sniproxy.md
@@ -2,36 +2,13 @@
 
 The purpose of this tutorial is to set up sniproxy so it’s possible to use regular https:// URLs for your main web server, like nginx or Apache, as well as Sandstorm, which has its own HTTPS handling. To make that work, they will need to share port 443.
 
-## Introduction
-
-The purpose of this tutorial is to set up sniproxy so it's possible to use Sandstorm-verified TLS encryption while coexisting with another web server that also uses TLS and share port 443 with it.
-
-The main reason is to allow users to connect with your Sandstorm instance in the standard HTTPS port (443) and keep using that port also for any other web apps.
-
-This action is going to change the BASE_URL of Sandstorm, presumably, so you'll end up erasing the OAuth configuration. To fix that, you will need to follow the "How do I log in, if there's a problem with logging in via the web?" section of [the FAQ](faq.md).
-
 This document assumes the server is running Debian or one of its derivatives (e.g. Ubuntu). If your server is not, you might need to adjust some steps. Note that there will be some downtime in this process so you might want to do it when your server is not very busy.
+
+We are also going to assume you are using the Sandcats DNS/HTTPS service in this example, however this will work with any Sandstorm-managed TLS setup, provided the pattern is distinct from your other web services on the server.
 
 ## Install sniproxy
 
-If you're lucky, the package `sniproxy` might be present on your GNU/Linux distro; otherwise, you'll have to install it yourself. Follow the instructions at [the sniproxy homepage](https://github.com/dlundquist/sniproxy) for your operating system and install it.
-
-In this example, on Ubuntu 10.04, the instructions were:
-
-```bash
-# Install required packages
-sudo apt-get install autotools-dev cdbs debhelper dh-autoreconf dpkg-dev gettext libev-dev libpcre3-dev libudns-dev pkg-config fakeroot git
-
-# Clone sniproxy repo from Github
-git clone https://github.com/dlundquist/sniproxy.git
-
-# Compile and create the package
-cd sniproxy
-./autogen.sh && dpkg-buildpackage
-
-# Install the package
-sudo dpkg -i ../sniproxy_*_*.deb
-```
+It is likely the package `sniproxy` is present on your Linux distro and you can install it through your package manager. Otherwise, follow the instructions at [the sniproxy homepage](https://github.com/dlundquist/sniproxy) for your operating system to install it manually.
 
 ## Setting up sniproxy
 
@@ -79,7 +56,15 @@ table https_hosts {
 
 ### Startup
 
-We'll have to ensure sniproxy starts when the system is rebooted. For that we'll need to ensure it is enabled in ```/etc/default/sniproxy``` (```ENABLED=1```). You may also need to run this command to make sure sniproxy will automatically start on boot up:
+We'll have to ensure sniproxy starts when the system is rebooted. For that we'll need to ensure it is enabled in ```/etc/default/sniproxy``` (```ENABLED=1```).
+
+The command to instruct your machine to start sniproxy on boot depends on your distro, but it is likely systemd-based, upon which you can run:
+
+```bash
+sudo systemctl enable sniproxy
+```
+
+Otherwise, if you have a sysvinit-based distribution, you may need to run:
 
 ```bash
 sudo update-rc.d sniproxy enable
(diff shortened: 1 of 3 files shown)


ocdtrekkie:
I un-italicized sniproxy everywhere, but put it in backticks where referred to as a package name. The biggest thing I decided to pull out in the last commit was the suggestion to take questions about this to Server Fault, as Sandstorm's community is small enough that pushing people out to other sites and services for help is probably going to get them lost.

docs/administering/ssl.md
```diff
@@ -21,10 +21,9 @@ You have a few options.
...
 
-- You [can install `sniproxy` to share port
-  443](https://xamar.sandcats.io/shared/Bqa9dftNbc1Ni06D-SgBdkFuM_iky8VHAlTw0Rk1lzN) between your
-  existing server and Sandstorm so that Sandstorm can manage (and autorenew) its own certificates.
-  This allows you to combine an **existing web server on port 443** with free sandcats.io HTTPS.
+- You [can install `sniproxy` to share port 443](sniproxy.md) between your existing server and Sandstorm so that
+  Sandstorm can manage (and autorenew) its own certificates. This allows you to combine an **existing
+  web server on port 443** with free sandcats.io HTTPS.
   
 - You [can follow this guide](https://juanjoalvarez.net/es/detail/2017/jan/12/how-set-sandstorm-behind-reverse-proxy-keeping-you/)
```
zenhack, line 28: This site appears to be down. Don't know whether that's permanent or transient. Not really related to this pr but I noticed while reviewing.

docs/administering/sniproxy.md
```diff
@@ -0,0 +1,146 @@
+# Sharing the HTTPS Port with sniproxy
+
+The purpose of this tutorial is to set up sniproxy so it’s possible to use regular https:// URLs for your main web server, like nginx or Apache, as well as Sandstorm, which has its own HTTPS handling. To make that work, they will need to share port 443.
+
+## Introduction
+
+The purpose of this tutorial is to set up sniproxy so it's possible to use Sandstorm-verified TLS encryption while coexisting with another web server that also uses TLS and share port 443 with it.
```
zenhack, line 7: This seems redundant with the paragraph above.

docs/administering/sniproxy.md
```diff
@@ -0,0 +1,146 @@
...
+
+The main reason is to allow users to connect with your Sandstorm instance in the standard HTTPS port (443) and keep using that port also for any other web apps.
+
+This action is going to change the BASE_URL of Sandstorm, presumably, so you'll end up erasing the OAuth configuration. To fix that, you will need to follow the "How do I log in, if there's a problem with logging in via the web?" section of [the FAQ](faq.md).
```
zenhack, line 11: I think this should be worded more conditionally; if a user is going from sandstorm facing the internet directly to moving it behind sniproxy, BASE_URL will not change. This seems to be making some unstated assumptions about the user's prior configuration.

We should probably have a section near the top that states explicitly what the starting point is.

docs/administering/sniproxy.md
```diff
@@ -0,0 +1,146 @@
...
+
+This document assumes the server is running Debian or one of its derivatives (e.g. Ubuntu). If your server is not, you might need to adjust some steps. Note that there will be some downtime in this process so you might want to do it when your server is not very busy.
+
+## Install sniproxy
+
+If you're lucky, the package `sniproxy` might be present on your GNU/Linux distro; otherwise, you'll have to install it yourself. Follow the instructions at [the sniproxy homepage](https://github.com/dlundquist/sniproxy) for your operating system and install it.
+
+In this example, on Ubuntu 10.04, the instructions were:
```
zenhack, line 19: Should we maybe update this for a new version? This is literally a decade-old distro. It looks like Debian and Ubuntu both ship packages for sniproxy now, so there's no real reason to build from source, and with the source method the user won't get automatic updates with the rest of their distro.

I imagine these steps can all be replaced by `sudo apt-get install sniproxy` or something.

docs/administering/sniproxy.md
```diff
@@ -0,0 +1,146 @@
...
+        priority notice
+    }
+}
+
+table https_hosts {
+    .*\.sandcats\.io 127.0.0.1:6443
+}
+```
+
+### Startup
```
zenhack, line 80: This section also looks bitrotten to me; probably a modern version would be something like `systemctl enable sniproxy`.

docs/administering/sniproxy.md
```diff
@@ -0,0 +1,146 @@
...
+
+```bash
+sudo service nginx stop
+sudo service sandstorm stop
+sudo service sniproxy start
+sudo service sandstorm start
+sudo service nginx start
+```
+
+You probably have to fix the Sandstorm administrator OAuth token. Run the following and follow the instructions that it prints out:
```
zenhack, line 136: I'm not sure I follow why this would be necessary. Also, the admin token has nothing to do with OAuth afaik.

docs/administering/ssl.md
ocdtrekkie: I'll look into it. Good catch.

docs/administering/sniproxy.md
ocdtrekkie: Yeah, I removed another paragraph like this. I am not sure how much of this happened when transitioning this from a blog post to a docs page or something. I'll clean it up.

docs/administering/sniproxy.md
ocdtrekkie: I think it might be fair to just pull this out entirely. You do make a good point, and the solution to that additional problem is elsewhere in the docs already.

docs/administering/sniproxy.md
ocdtrekkie: The paragraph above does specify it might be in your distro. I guess we should make that sound less "hopeful" and more likely. And then just remove this, as the sniproxy homepage linked above is stated to have the correct instructions if a manual install is needed. We shouldn't be bothering to keep updated instructions. (I noticed how old this distro was, but I didn't want to modify it from "the last thing someone knows worked". Getting rid of it simplifies this problem.)

docs/administering/sniproxy.md
ocdtrekkie: I think we should probably document both... Sandstorm continues to support sysvinit-based setup, see here: https://github.com/sandstorm-io/sandstorm/blob/master/install.sh#L1610

docs/administering/sniproxy.md
ocdtrekkie: I agree we can remove this, as we are removing the assumption the BASE_URL is being changed in this tutorial. (The reason for admin-token is presumably that if someone invalidated their Google/GitHub login, they need to run admin-token to get into the admin panel to correct their OAuth configuration.)

docs/administering/sniproxy.md
zenhack: SGTM.

ocdtrekkie:
@zenhack I agree in whole or in part with all of your review notes, thanks for the fresh set of eyes. My next commit will hopefully shorten this up a lot, and I'll also look at making it clearly this will work with any domain where Sandstorm manages TLS. Although I am largely okay with the use of Sandcats in examples, as long as we aren't claiming it only works with Sandcats.

docs/administering/sniproxy.md
zenhack: Ok, yeah, let's just get rid of it.

zenhack:
Quoting Jacob Weisz (2020-05-10 00:23:22)

>    I think we should probably document both... Sandstorm continues to
>    support sysvinit-based setup, see here:

Works for me.

docs/administering/ssl.md
ocdtrekkie: I'll swap it out for the archive.org link for now, and then revisit the entire topic (and the worthiness of the linked page) in my next PR.

ocdtrekkie:
Okay, this is now a lot shorter!

PR #3323, opened 2020-05-06
Fix broken use of Settings.

### LINKED ISSUE #3322: createAccountForCredential TypeError: Cannot read property 'findOne' of undefined
Hello,

on a new installation with latest-version 264 no other users can be created via LDAP Login. After login with LDAP the user sees this message: **Failed to create account. Internal server error (500)** surrounded in Sanstorm GUI. The Admin user works like expected and all apps are working too within this admin-user. Here is the Log: 

`stack: 709c10 7109e0
Exception while invoking method 'createAccountForCredential' TypeError: Cannot read property 'findOne' of undefined
    at Function.SandstormDb.fillInProfileDefaults (imports/sandstorm-db/profile.js:233:30)
    at MethodInvocation.createAccountForCredential (imports/server/accounts/credentials/credentials-server.js:193:17)
    at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1771:12)
    at packages/ddp-server/livedata_server.js:719:19
    at packages/ddp-server/livedata_server.js:717:46
    at packages/ddp-server/livedata_server.js:715:46
    at new Promise (<anonymous>)
    at Session.method (packages/ddp-server/livedata_server.js:689:23)
    at packages/ddp-server/livedata_server.js:559:43`

We use nginx reverse Proxy with wildcard HTTPS Certificates (on Admin user all works fine.) LDAP is the only active Auth-Provider enabled.

Thanks!

sandstorm

Edit: Enabling Email Authentication; Registering with Email and after that add LDAP to this Account is working - but that is only a bad workaround because each user must do this for himself...

zenhack:
This should fix #3322, but I don't have an actual LDAP setup to test with.
Help testing would be appreciated.

This was caused by the recent refactoring -- we added an explicit import
for `Settings`, but no explicit *export*.

This is actually the only remaining use of the variable, so instead of
simply revert the broken refactor, I just switched it over to using
`globalDb.collection.settings` (which is used all over the place), and
removed the `Settings` alias entirely.

### DIFF
--- a/shell/imports/sandstorm-db/profile.js
+++ b/shell/imports/sandstorm-db/profile.js
@@ -21,6 +21,7 @@ import { _ } from "meteor/underscore";
 
 import Identicon from "/imports/sandstorm-identicons/identicon.js";
 import { SandstormDb } from "./db.js";
+import { globalDb } from "/imports/db-deprecated.js";
 
 let makeIdenticon;
 let httpProtocol;
--- a/shell/imports/sandstorm-db/profile.js
+++ b/shell/imports/sandstorm-db/profile.js
@@ -21,7 +21,6 @@ import { _ } from "meteor/underscore";
 
 import Identicon from "/imports/sandstorm-identicons/identicon.js";
 import { SandstormDb } from "./db.js";
-import { Settings } from "/imports/db-deprecated.js";
 
 let makeIdenticon;
 let httpProtocol;
@@ -230,7 +229,7 @@ SandstormDb.fillInProfileDefaults = function (credential, profile) {
     profile.name = profile.name || "Demo User";
     profile.handle = profile.handle || "demo";
   } else if (services.ldap) {
-    const setting = Settings.findOne({ _id: "ldapNameField" });
+    const setting = globalDb.collections.settings.findOne({ _id: "ldapNameField" });
     const key = setting ? setting.value : "";
     profile.handle = profile.handle || services.ldap.username;
     profile.name = profile.name || services.ldap.rawAttrs[key] || profile.handle;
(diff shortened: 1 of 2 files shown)


shell/imports/sandstorm-db/profile.js
```diff
@@ -230,7 +229,7 @@ SandstormDb.fillInProfileDefaults = function (credential, profile) {
     profile.name = profile.name || "Demo User";
     profile.handle = profile.handle || "demo";
   } else if (services.ldap) {
-    const setting = Settings.findOne({ _id: "ldapNameField" });
+    const setting = globalDb.collections.settings.findOne({ _id: "ldapNameField" });
```
zarvox, line 232: strictly speaking, `globalDb` isn't in scope here, so you should import it.  (Yes, that results in a circular dependency, but that's the state of the world today.)

Better would be something like `this.collections.settings.findOne`, given that it's a method being attached to `SandstormDb`, but I think it might have to go on the prototype to get the `this` behavior correct.

In an ideal world, we'd have a real class for `SandstormDb` (and perhaps more precise implementations extending it as `ServerDb` and `ClientDb`) and we'd avoid tossing methods on `SandstormDb` from files other than the one that defines it.

zenhack: Added the import. And yes, we should refactor all of this to be more reasonable at some point.

shell/imports/sandstorm-db/db.js
```diff
@@ -24,6 +24,7 @@ import { Random } from "meteor/random";
 import { SHA256 } from "meteor/sha";
 import { HTTP } from "meteor/http";
 import { iconSrcForPackage } from "/imports/sandstorm-identicons/helpers.js";
+import { globalDb } from "/imports/db-deprecated.js";
```
zarvox, line 27: Ummm, I think I intended this import to go in the profile.js file, where `globalDb` is used.

zenhack:
Quoting Drew Fisher (2020-05-06 19:27:42)

>    Ummm, I think I intended this import to go in the profile.js file,
>    where globalDb is used.

So did I :/. fixed now.

ocdtrekkie:
It sounds like we have another user hitting this on IRC this morning.

ocdtrekkie:
They built from source and it did fix it. \o/

PR #3325, opened 2020-05-06
Add Google Cloud Platform acme support

abliss:
Also fixes the alphabetization.


ocdtrekkie:
I think we also need to add it to the About page, if it's a third party plugin for ACME.js (similar to Kenton adding the Cloudflare one to the About page). I think the rest of them are hosted with ACME.js by the original author, so they don't get their own entries.

ocdtrekkie:
Is this blocked on https://github.com/latacora/acme-dns-01-gcp/pull/2 as well?

abliss:
> Is this blocked on [latacora/acme-dns-01-gcp#2](https://github.com/latacora/acme-dns-01-gcp/pull/2) as well?

I don't think we need to block on it, but the experience will improve when that lands.

kentonv:
Hmm, looking at the changes you've submitted upstream, they seem pretty critical to using this in Sandstorm, right? I'd be inclined not to merge this until it can actually work reasonably.

I think npm supports specifying a git repo URL and commit hash instead of a version. Maybe you could specify your git branch?

ocdtrekkie:
I mostly left off the ready-to-review tag pending adding the credits for the plugin. But yeah, I noticed it was kinda sad that they haven't even responded to the PR yet, even though the last commit was only three days prior to Adam's PR.

abliss:
I don't mind pointing it to my own fork, but perhaps it would make more sense to have a fork under sandstorm-io/ and point there?

kentonv:
If we think we'll need to maintain the fork long-term, then maybe. If it's temporary then whatever. As long as you point to a specific commit hash then I don't really care where it's hosted... :)

ocdtrekkie:
In the realm of stupid PRs that took way more time to file than fix, I noticed a typo in this plugin's readme. But it was forked of course, from the ACME.js folks' stuff (and not really relevant to the GCP plugin). But the Root git site is cloned from coolaj86's git site. And in the end, I made it back to https://git.coolaj86.com/coolaj86/acme-dns-01-test.js/pulls/1

abliss:
The plugin author merged and deployed to npm, and I updated the About page, so this should be good to merge now.

PR #3327, opened 2020-05-06
Remove "Home" label from login page

### LINKED ISSUE #2051: Remove "Home" label on the top bar when not signed into Sandstorm
Context:
- Right now, the home page of every Sandstorm install shows the "top bar", which has two items in it - "Home" and "Sandstorm" - but clicking either of them doesn't really do anything.
- See e.g. https://rose.sandcats.io/ for a non-splash-page example.
- See e.g. https://oasis.sandstorm.io/ for a splash-page-ful example.

Therefore, @paulproteus plans to submit a PR removing the top bar from being visible on the homepage. As this isn't something I'm going to do this _second_, discussion is very welcome.

ocdtrekkie:
Fixes #2051 

Absolutely no testing has been done. But it's a language file only change. Basically, the label "Home" is solely displayed on the log in screen, where it does nothing and serves no informative purpose. I don't really see a good reason to invent a more complicated solution here.

Current:
image
Expected result (via Firefox dev console hacking):
image

I'm a little surprised the little divider also seems to go away (CSS is a mysterious beast), but I think it looks pretty clean that way.

### DIFF
--- a/shell/i18n/en.i18n.json
+++ b/shell/i18n/en.i18n.json
@@ -40,9 +40,6 @@
       "title": "Not Found",
       "text": "404 Not Found :("
     },
-    "root": {
-      "title": ""
-    },
     "grainLogContents": {
       "prefix": "..."
     },
--- a/shell/i18n/en.i18n.json
+++ b/shell/i18n/en.i18n.json
@@ -41,7 +41,7 @@
       "text": "404 Not Found :("
     },
     "root": {
-      "title": "Home"
+      "title": ""
     },
     "grainLogContents": {
       "prefix": "..."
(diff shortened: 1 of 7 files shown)


zenhack:
I think we should be able to just delete this:

https://github.com/sandstorm-io/sandstorm/blob/649c60718795d98af2474fb6421aeb7eda0a08d4/shell/imports/client/shell.html#L213

...and then remove those "root" sections from the localization files entirely.

ocdtrekkie:
I was trying to think if there was any particular reason one might want to leave it there, or handle it differently in a localization, but I can't really think of one. And the idea of deleting a whole *section* of localization is appealing.

ocdtrekkie:
Done.

PR #3344, opened 2020-05-22
Port some of the identicon stuff over to typescript.

zenhack:
Just identicon.js for now, the rest will come in future commits.

This was mostly just a matter of adding annotations in the right places,
but I also had to rename a local variable s (to `sat`), because it was
being re-assigned to a value of a different type (number to number[]).
There should be no functional impact.

The identicon module is interesting mostly because it is the only
dependency of sandstorm-db/db.js. The db stuff will require more
refactoring to get to a place where typescript will be happy with it,
but it will also have much bigger payoffs.

### DIFF
--- a/shell/imports/sandstorm-identicons/identicon.ts
+++ b/shell/imports/sandstorm-identicons/identicon.ts
@@ -9,10 +9,15 @@
  */
 
 // Trivially modified for Meteor context by Kenton Varda.
-// Later modified to produce an SVG rather than a PNG.
+// Later modified to produce an SVG rather than a PNG, and
+// to convert it to type script.
 
 class Identicon {
-  constructor(hash, size, margin) {
+  private hash: string;
+  private size: number;
+  private margin: number;
+
+  constructor(hash: string, size?: number, margin?: number) {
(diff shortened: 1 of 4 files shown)


shell/imports/sandstorm-identicons/identicon.ts
```diff
@@ -54,14 +59,14 @@ class Identicon {
...
     rectangles.push(`<rect x="${x}" y="${y}" width="${w}" height="${h}"/>`);
   }
 
   // adapted from: https://gist.github.com/aemkei/1325937
-  hsl2rgb(h, s, b) {
+  hsl2rgb(h: number, s: number, b: number) {
     h *= 6;
-    s = [
+    const sat = [
       b += s *= b < .5 ? b : 1 - b,
```
zarvox, line 70: I can't help but find these multiple operator-assignments on a single line terrifying

(no change needed, but yikes, would be nice if this wasn't quite so opaque)

zenhack: Yeah, I have no idea what's going on here :/

Issue #3295, opened 2020-04-19
`apps/schedule` test is racy


zarvox:
As discussed in #3293, `apps/schedule` can fail due to a race condition.  In the success case, the following sequence occurs:

1. the test's `.click(selector)` causes the JS in the test app to make some HTTP or websocket request to the grain backend.
2. the grain's backend makes some request over capnproto RPC to the sandstorm meteor backend to schedule a job to be run in 5 minutes
3. the sandstorm meteor backend's capnproto interface implementation inserts some record into the mongo database
4. the test's `.execute('Meteor.call("runDueJobsAt", ' + firstCheck.toString() + ');')` triggers a request to the sandstorm meteor backend
5. the meteor backend runs a db query for scheduled jobs due at a particular time, finds a record, and triggers the scheduled job.

But there's actually no synchronization forcing 4 to come after 3, so there's another sequence of events that is also possible:

1. the test's `.click(selector)` causes the JS in the test app to make some HTTP or websocket request to the grain backend.
2. the grain's backend makes some request over capnproto RPC to the sandstorm meteor backend to schedule a job to be run in 5 minutes
3. the test's `.execute('Meteor.call("runDueJobsAt", ' + firstCheck.toString() + ');')` triggers a request to the sandstorm meteor backend
4. the meteor backend runs a db query for scheduled jobs due at a particular time, finds no records, and triggers nothing.
5. the sandstorm meteor backend's capnproto interface implementation, responding to the RPC in step 2, inserts some record into the mongo database. In five minutes, it might fire, but our test framework has given up by then.

The solution is to enforce the intended ordering with a barrier of some sort. One way to go about that would be:

* Make the test app grain backend return something in the request to schedule the job, but only once the sandstorm backend has acknowledged the request.
* Make the test app frontend change something in the page when that request completes successfully (adding an element with an ID like #success-oneshot would do)
* Make the test itself wait on that frontend change to appear before calling the `runDueJobsAt` Meteor method.

(We've worked around this in the meantime with a five-second pause, but that increases the total runtime of the testsuite, and it'd be good to fix this the "right" way)

ocdtrekkie:
We can close this now, yes?

zarvox:
Yes, the race condition I identified was fixed by #3349, so I'll close this.

There's a much-less-pressing code health thing where we still call `.pause()` in a couple places when ideally we'd be checking for a condition until a deadline instead, but that can be a separate ticket if we want to write that down.

[...]

PR #3348, opened 2020-05-26
Some cleanup in db.js

zenhack:
This patch:

- makes `SandstormDb` a class, rather than just a prototype function.
- moves many method definitions from the call to `extend` to the class.

The comments had the latter as a TODO (though there are still some I haven't gotten). This is also a step towards making it easier to convert this module to typescript.

### DIFF
--- a/shell/imports/sandstorm-db/db.js
+++ b/shell/imports/sandstorm-db/db.js
@@ -981,28 +981,6 @@ const calculateReferralBonus = function (user) {
   }
 };
 
-function isAdmin() {
-  // Returns true if the user is the administrator.
-
-  const user = Meteor.user();
-  if (user && user.isAdmin) {
-    return true;
-  } else {
-    return false;
-  }
-}
-
-function isAdminById(id) {
-  // Returns true if the user's id is the administrator.
-
-  const user = Meteor.users.findOne({ _id: id }, { fields: { isAdmin: 1 } });
-  if (user && user.isAdmin) {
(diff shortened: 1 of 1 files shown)


zarvox:
I'm seeing an additional test failure, beyond the usual 4:

```
 ✖ apps/schedule

   - Test one-shot tasks (15.188s)
```

I forget -- have we seen that test flake before, on some other build?

zenhack:
Hm, it's passing consistently when I run it locally. That's the test #3295 is about; maybe we hit a hiccup on GitHub's infra and hit the race in spite of the excessive timeout?

Either way, I'll likely work on #3295 soonish.

zarvox:
Reran the tests after merging #3349 and we're back to the usual 4 failures, so I'm gonna merge this.

Longer term I think we should think about how to manage the client/server behavior split -- I'm inclined towards a `SandstormDb` base class, and a `ServerSandstormDb` and a `ClientSandstormDb` that extend `SandstormDb` with the appropriate functionality or implementations for their contexts, and avoiding doing `Meteor.isClient`/`Meteor.isServer` checks.

It's also likely that some of the things that currently live as methods on SandstormDb don't actually need to be placed there at all, and that they are convenience functions that once needed to be available early in the load order, and due to limitations in earlier versions of Meteor, we stuck them on SandstormDb, which was one of our first packages to be loaded.  Those I'd favor leaving off the SandstormDb class.

PR #3349, opened 2020-05-27
tests: apps/schedule: avoid a call to .pause()

zenhack:
...by implementing the proper fix for the race condition described in

As expected, this shaves about 5 seconds off of each of these tests, or
about 15 seconds off of the test suite as a whole.

Note that there are two other calls to pause() in there; these are a bit
harder to eliminate, since the events they're waiting for are triggered
by things outside of any http request, so we can't easily create an
element on the page and do waitForVisible or the like, and as far as I
can tell nightwatch doesn't give us an easy way to say "wait until this
element contains this text," which is what we really want here.


zarvox:
I'm happy to merge this, since it's a clear improvement to the status quo.  Thanks!

Some musings: one way we might be able to remove the other `pause()` calls is by changing the implementation of `runDueJobsAt` (the Meteor method only used for testing) to:

1. Return a list of the jobs that were run, and their statuses.
2. Not return until all the jobs due have finished running.

It looks like imports/server/scheduled-job.js already does 2.  I'm not sure if our usage of nightwatch blocks the test runner until the Meteor method has returned, though.

One other change we might want to make is to switch the callsites to use `executeAsync()` (so we can get a return value/block on the result of the Meteor method call) and override nightwatch's script timeout like we do in https://github.com/sandstorm-io/sandstorm/blob/master/tests/commands/loginDevAccount.js#L30-L31 to be whatever the short_wait deadline the test was going to tolerate anyway.

I like the general pattern of having deadlines for conditions to be met, rather than waiting a fixed amount of time and only after that pause, checking the condition.  Even if in practice nightwatch/selenium is polling under the hood, as the poll interval gets shorter, we asymptotically approach event-driven behavior.

Nightwatch appears to have a separate Chai Expect assertion library option which provides `.before` and `.after` deadlines on all the shapes of assertions -- perhaps something there would help us do deadline-driven tests rather than pause-driven tests? https://nightwatchjs.org/api/expect/#expect-before

PR #3350, opened 2020-05-28
Actually run the removeFeatureKeys migration.

zenhack:
Eslint caught this (with an unused-variable warning); apparently this
has been defined for a long time but never actually set to *run*.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,7 +71,7 @@ jobs:
           # the constant below.
           make lint | tee lint.log
           num_problems=$(cat lint.log | grep ' problems (0 errors, ' | awk '{print $2}')
-          [ "$num_problems" = 857 ]
+          [ "$num_problems" = 856 ]
       - name: test
         run: |
           set -o pipefail
--- a/shell/imports/server/migrations.js
+++ b/shell/imports/server/migrations.js
@@ -1217,6 +1217,7 @@ const MIGRATIONS = [
   onePersonaPerAccountPostCleanup,
   cleanupBadExpiresIfUnused,
   deleteReferralNotifications,
+  removeFeatureKeys,
 ];
 
 const NEW_SERVER_STARTUP = [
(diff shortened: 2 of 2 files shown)


zenhack:
@zarvox, vaguely off topic but: many of the unused-variable warnings are unused function parameters. In principle we could reliably drop these, because javascript allows you to call a function with more arguments that it declares, but I find the presence of these parameters to be somewhat self-documenting, and I'm not a fan of the idea of removing parameters that are nominally part of an interface, even if an implementation doesn't actually use them. But it might also be nice to catch "unintentional" unused arguments. Other linters I've seen let you prefix a variable with `_` to say "yes, I know this is unused, it's intentional."

This comes up a lot on the migrations module, since you have a large list of functions that accept a db and backend, and many of them don't use the latter. Setting `{"args": "none"}` for that warning gets rid of 275 warnings.

What are your thoughts?

kpreid:
Assuming https://eslint.org/docs/rules/no-unused-vars is the relevant documentation, there is an option called `argsIgnorePattern` which lets you set up the underscore or other convention for unused arguments. (I personally find unused-variable warnings very handy for catching dead code that's either deletable or should have been used, but I'm not currently contributing any code to Sandstorm so my opinion should have little weight.)

zenhack:
Quoting Kevin Reid (2020-05-27 22:52:30)

> Assuming [1]https://eslint.org/docs/rules/no-unused-vars is the
> relevant documentation, there is an option called argsIgnorePattern
> which lets you set up the underscore or other convention for unused
> arguments.

Aha, I knew something like that must exist! In that case I'd like to
just use that option so we can clearly distinguish "this implementation
doesn't happen to use this parameter, but it's part of the interface"
from genuine mistakes.

> (I personally find unused-variable warnings very handy for
> catching dead code that's either deletable or should have been used,
> but I'm not currently contributing any code to Sandstorm so my opinion
> should have little weight.)

Indeed; this PR is a fix for a real bug that that warning caught, so I
absolutely don't want to just turn it off.

zenhack:
Huh, tests are hanging, will have to investigate.

zarvox:
This is amazing and I find it hilarious.  Three-year old code that never ran and contains a bug.  :joy: 

On the `no-unused-vars` lint: I'm inclined to keep the warning about unused parameters, allow underscore-prefixing to squelch the warning, and convert unused parameters to use the underscored format.  My opinion is that the arity of a function really should match between definition and callsite unless you're explicitly doing intentional magic with `arguments`.

We're down to 3 regular test failures, looks like!  Time to recalibrate our expectations for future pull requests!

PR #3351, opened 2020-05-28
Make uploadMeteorTestApp.js navigate to a canonical URL

zarvox:
This appears to fix `tests/account-settings`.

Looks like `appHooks` passing in https://github.com/sandstorm-io/sandstorm/runs/715515605 was a transient success, and it's back to failing.

### DIFF
--- a/tests/commands/uploadMeteorTestApp.js
+++ b/tests/commands/uploadMeteorTestApp.js
@@ -20,7 +20,7 @@ const testappPath = require('path').resolve(__dirname + "/../assets/meteor-testa
 
 exports.command = function() {
   return this
-    .url('/apps')
+    .url(this.launch_url + '/apps')
     .waitForElementVisible('.upload-button', short_wait)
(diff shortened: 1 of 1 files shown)


zenhack:
LGTM

> Looks like appHooks passing in https://github.com/sandstorm-io/sandstorm/runs/715515605 was a transient success, and it's back to failing.

Yeah, I would have found it *very* surprising if adding that migration tests fixed the issue we were seeing there -- this makes much more sense. should probably reopen #3225.

ocdtrekkie:
Reopened as requested

zarvox:
I don't think #3225 was related to appHooks.  We should probably open a new ticket for the appHooks test failure, but #3225 was about meteor-testapp crashing on launch and being on an out-of-date version of Meteor, and that's a separate issue that has since been fixed.

zarvox:
Filed #3353 to track the appHooks failure.

PR #3352, opened 2020-05-28
Silence a bunch of unused-argument warnings

zenhack:
A migration is just a function that takes a SandstormDb and a backend,
but the majority of these functions don't actually use the backend.

This patch updates our eslint config to allow silencing an unused
argument warning by prefixing the argument with an underscore, then uses
this throughout migrations.js. This way we can start cutting down on
the noise.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,7 +71,7 @@ jobs:
           # the constant below.
           make lint | tee lint.log
           num_problems=$(cat lint.log | grep ' problems (0 errors, ' | awk '{print $2}')
-          [ "$num_problems" = 856 ]
+          [ "$num_problems" = 812 ]
       - name: test
         run: |
           set -o pipefail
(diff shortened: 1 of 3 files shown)


ocdtrekkie:
It looks like this failed out on linting as 812 problems, despite you setting the threshold to 813...

zenhack:
Should be fixed now.

ocdtrekkie:
Is it difficult to make that only fail when the number of problems is *higher* than the threshold? It seems weird that if someone was to fix some linting issues, the check would fail, if they also didn't know to adjust that number.

zenhack:
Quoting Jacob Weisz (2020-05-28 16:41:47)

>    Is it difficult to make that only fail when the number of problems is
>    higher than the threshold? It seems weird that if someone was to fix
>    some linting issues, the check would fail, if they also didn't know to
>    adjust that number.

See the comment right above that check; it would be easy to make it pass
if it's lower than expected, but the point is to make sure we don't
regress, so when we make improvements we should update the threshold so
that the improvements "stick."

zarvox:
LGTM, thanks!

Issue #1153, opened 2015-11-12
Menu cut off/does not scroll in landscape on Firefox Mobile 


ocdtrekkie:
When I'm have a grain open in landscape on my phone, four options on the hamburger menu are cut off, and the menu can not scroll or be dragged at all. I have Firefox on Android.

spollard:
I can confirm this is still happening. I may take  a look myself since it seems like a small change...

ocdtrekkie:
That'd be great! Note that I no longer have an Android, so I can't test this.

spollard:
I just made a PR. Can somebody test it? #3354
I determined the changes by fiddling with the css in Firefox. I didn't want to setup a whole sandstorm dev env for just this tiny change....

I tested this at mobile width without any changes to other css. 

I tested this at desktop width also by forcing the grain items to be tall: 
body > .topbar > .navbar li.navitem-grain {   height: 320px;  }

PR #3354, opened 2020-06-01
Allow scrolling the navbar

spollard:
When there are many grains open on desktop or just a few grains open on Firefox on Android, the user needs to be able to scroll the navbar menu down to see them. These changes should allow that without breaking anything else. It was a little difficult to trace all the curly brackets for scss, but here are the intended changes:

@media (max-width: 900px) {
    body > .topbar {
        overflow: hidden; /\* REMOVE THIS \*/
        overflow: scroll; /\* ADD THIS \*/
    }
}
@media (min-width: 901px) {
     body > .topbar > .navbar {
        overflow: scroll; /\* ADD THIS \*/
    }
}

### DIFF
--- a/shell/client/styles/_topbar.scss
+++ b/shell/client/styles/_topbar.scss
@@ -88,7 +88,7 @@ body>.topbar {
     //background-color: #eee;
     //color: black;
     transition: height 0.2s;
-    overflow: hidden;
+    overflow: scroll;
 
     &.expanded {
       height: 100%;
@@ -655,6 +655,7 @@ body>.topbar {
     }
     @media #{$mobile} {
       position: relative;
+      overflow: scroll;
       li {
         font-size: 14pt;
         line-height: 48px;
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
As noted in #1153, this needs a bit of testing in the actual codebase probably.

zenhack:
Thanks for the patch! A couple things are a little wonky though:

- This sets the topbar to be scrollable *even when collapsed*, which is weird if you happen to trigger it. From poking around in firefox's inspector, it looks like we already add an 'expanded' class to the topbar when it expands, so this could probably be fixed by selecting on that.
- When scrolling, the grain title scrolls with the menu, but the hamburger icon and the notifications icon do not (see attached screenshot), which is also weird. It's less obvious to me how to fix this easily though, I'd have to stare at things a bit longer.

Screen Shot 2020-06-02 at 21 35 15

spollard:
I just fixed the scrolling behavior to auto so the scroll bar isn't there all the time. I also made the grain title on mobile stay fixed to the top of the window when the rest of the menu is scrolled.

zenhack:
Tested, LGTM.

kentonv:
Thanks!

PR #3359, opened 2020-06-06
Automatically switch Sandcats users to Let's Encrypt over the next few weeks.

kentonv:
Each server will migrate at a random time between June 10 and June 23.

### DIFF
--- a/acme-dns-01-sandcats/index.js
+++ b/acme-dns-01-sandcats/index.js
@@ -1,8 +1,8 @@
 'use strict';
 
-import https from "https";
-import querystring from "querystring";
-import dns from "dns";
+const https = require("https");
+const querystring = require("querystring");
+const dns = require("dns");
 
 class Challenge {
   constructor(options) {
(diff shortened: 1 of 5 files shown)


kentonv:
@zenhack I had to revert the conversion of acme-dns-01-sandcats to ES module syntax. It turns out this broke it, because we dynamically `require()` these modules, and you can't `require()` an ES module, you have to import it. I tried using dynamic `import()`, but that didn't work either, saying the module wasn't found. I think Meteor may have some restrictions around dynamic imports, but instead of trying to figure it out, I reverted to what had been tested before.

It's tough to test this ACME stuff. I wonder if we could set up an integration test that uses the Let's Encrypt staging service somehow...

ocdtrekkie:
Should we file an issue to delete all this code after the cutover is complete?

kentonv:
@ocdtrekkie If you want, but deleting code is my favorite part, I'm not gonna forget about it. :)

kentonv:
Also I definitely call dibs on the code deletion. :)

PR #3361, opened 2020-06-08
Makefile: make sure the meteor bundle is present before running ekam.

### LINKED ISSUE #3360: Build Error: installing from source on ubuntu
error from `make` on fresh digital ocean ubuntu box

`uname -a`
```
Linux sandstorm2 4.15.0-66-generic #75-Ubuntu SMP Tue Oct 1 05:24:09 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

prep
```
sudo apt-get update
sudo apt-get install build-essential libcap-dev xz-utils zip \
    unzip strace curl discount git python zlib1g-dev \
    golang-go cmake
curl https://install.meteor.com/?release=1.8.2 | sh
```

error message
```
✘ install: node-capnp/node-capnp.ekam-manifest
    node-capnp/capnp.node: not found
✘ jstest: node-capnp/capnp-test.js
    full log: tmp/node-capnp/capnp-test.js.log
✘ compile: node-capnp/capnp.cc
    /ekam-provider/canonical/node-capnp/capnp.cc:29:10: fatal error: 'node.h' file not found
    #include <node.h>
             ^~~~~~~~
    PLEASE submit a bug report to https://crbug.com and run tools/clang/scripts/process_crashreports.
      py (only works inside Google) which will upload a report and include the crash backtrace, 
      preprocessed source, and associated run script.
    Stack dump:
    0.	Program arguments: /root/sandstorm/deps/llvm-build/Release+Asserts/bin/clang++ -I/ekam-
      provider/c++header -std=c++1z -Wall -Wextra -Wglobal-constructors -Wno-sign-compare -Wno-unused
      -parameter -O2 -Wall -g -DSANDSTORM_BUILD=0 -DKJ_HAS_OPENSSL -DKJ_HAS_ZLIB -DKJ_HAS_LIBDL -
      pthread -fPIC -I/include/node -DKJ_STD_COMPAT -c /ekam-provider/canonical/node-capnp/capnp.cc -
      o node-capnp/capnp.o 
    1.	/usr/lib/gcc/x86_64-linux-gnu/7.5.0/../../../../include/c++/7.5.0/bits/hashtable_policy.h:88:
      5: current parser token '{'
    2.	/usr/lib/gcc/x86_64-linux-gnu/7.5.0/../../../../include/c++/7.5.0/bits/hashtable_policy.h:36:
      1: parsing namespace 'std'
    3.	/usr/lib/gcc/x86_64-linux-gnu/7.5.0/../../../../include/c++/7.5.0/bits/hashtable_policy.h:48:
      1: parsing namespace 'std::__detail'
    4.	/usr/lib/gcc/x86_64-linux-gnu/7.5.0/../../../../include/c++/7.5.0/bits/hashtable_policy.h:86:
      5: parsing struct/union/class body 'std::__detail::__is_noexcept_hash'
    #0 0x000000000239ac78 (/root/sandstorm/deps/llvm-build/Release+Asserts/bin/clang+++0x239ac78)
    clang: error: clang frontend command failed due to signal (use -v to see invocation)
    clang version 11.0.0 (https://github.com/llvm/llvm-project/ 4e0d9925d6a3561449bdd8def27fd3f3f1b3f
      b9f)
    Target: x86_64-unknown-linux-gnu
    Thread model: posix
    InstalledDir: /root/sandstorm/deps/llvm-build/Release+Asserts/bin
    clang: note: diagnostic msg: Error generating preprocessed source(s).
Makefile:265: recipe for target 'tmp/.ekam-run' failed
make: *** [tmp/.ekam-run] Error 1
```

zenhack:
Otherwise building node-capnp fails if the user has never run `meteor`,
due to not being able to find the development headers. This doesn't
trigger if the build is run as the same user that ran meteor's
`curl ... | sh` installer, which is why we haven't previously noticed
it.

Note that if the bundle is not present ./find-meteor-dev-bundle.sh will
pull it in, but although we define a variable in terms of this at the
top of the Makefile, its value is computed lazily, so it will not be run
until it is actually needed.

It might work to just use := instead of = when assigning the variable,
but conceptually having the build process depend on node seems like the
right thing (and it should also ensure a rebuild if the version of
node/the bundle is updated).

Fixes #3360

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -261,7 +261,7 @@ tmp/ekam-bin: tmp/.deps
 	    (cd deps/ekam && $(MAKE) bin/ekam-bootstrap && \
 	     cd ../.. && ln -s ../deps/ekam/bin/ekam-bootstrap tmp/ekam-bin)
 
-tmp/.ekam-run: tmp/ekam-bin src/sandstorm/* tmp/.deps deps/boringssl/build/ssl/libssl.a deps/libsodium/build/src/libsodium/.libs/libsodium.a | deps/llvm-build
+tmp/.ekam-run: $(NODEJS) tmp/ekam-bin src/sandstorm/* tmp/.deps deps/boringssl/build/ssl/libssl.a deps/libsodium/build/src/libsodium/.libs/libsodium.a | deps/llvm-build
 	@$(call color,building sandstorm with ekam)
 	@CC="$(CC)" CXX="$(CXX)" CFLAGS="$(CFLAGS2)" CXXFLAGS="$(CXXFLAGS2)" \
 	    LIBS="$(LIBS2)" NODEJS=$(NODEJS) tmp/ekam-bin -j$(PARALLEL)


kentonv:
> Note that if the bundle is not present ./find-meteor-dev-bundle.sh will pull it in

Hmm, I don't think that's correct. `find-meteor-dev-bundle.sh` seems to fail with no output (on stdout) if the bundle isn't found. Unfortunately `make` ignores the failure in the context of initializing a variable. There's no code in `find-meteor-dev-bundle.sh` which would install meteor (or a specific version of meteor) if it's missing. Maybe there should be? Though installing meteor as a build step might disturb some people.

> although we define a variable in terms of this at the top of the Makefile, its value is computed lazily

TIL `make` doesn't evaluate a variable until it runs a rule that references the variable (though it does seem to evaluate the variable before executing any lines of the rule, even if the variable is not referenced until later on).

That said, I don't think that's the issue here. `$(NODEJS)` is referenced before executing Ekam.

I think your change has the effect that if meteor is not installed, then `tmp/.ekam-run` ends up listing `/bin/node` as a dependency. `/bin/node` doesn't exist on most systems, but in theory it could, which would lead to some confusion here.

What we really want is that if `find-meteor-dev-bundle.sh` fails, `make` should fail. Maybe the answer is to change the definition of `METEOR_DEV_BUNDLE to:

    METEOR_DEV_BUNDLE=$$(./find-meteor-dev-bundle.sh)

This would have the effect that `find-meteor-dev-bundle.sh` is executed just-in-time every place where it is used. That is probably fine since it caches its result anyway.

kentonv:
Actually that won't work either, because a failure inside `$()` doesn't propagate to the outer shell.

```
$ echo $(false)

$ echo $?
0
```

Ugh.

zenhack:
Hm, and now this patch doesn't actually cause it to error out early for me anymore; not sure what I was doing differently before.

In any case it looks like you're right; the problem is that we're ignoring the failure from that script entirely. Closing, since this is the wrong solution.

zenhack:
We could probably handle this by making the assignment strict and
checking the status explicitly in the Makefile:

METEOR_DEV_BUNDLE := $(shell ./find-meteor-dev-bundle.sh)
ifneq '$(.SHELLSTATUS)' '0'
$(error Failed to find dev bundle)
endif

I haven't been able to find a way to convince make to just fail if any
$(shell ...) call fails, which seems like the sensible thing.

PR #3362, opened 2020-06-08
First pass at modernizing ssl.md

ocdtrekkie:
So here's my first attempt at this. I basically give people a shunt over to sandcats-https.md right away if they are using Sandcats, then explain the other ways Sandstorm can configure HTTPS directly. I shoved reverse-proxy and self-signed tutorials out to an "Additional options" header, since that's now less ideal. I also tweaked the sniproxy bullet to let people know it's not just for Sandcats.io users here.

Paging @zenhack for opinions.

### DIFF
--- a/docs/administering/ssl.md
+++ b/docs/administering/ssl.md
@@ -1,9 +1,8 @@
 ## Sandstorm and HTTPS
 
-Sandstorm can now terminate TLS connections for most conventional configurations you may wish to employ.
-This includes automatic certificate renewal of certificates, if you utilize a supported DNS provider, as well
-as the ability to manually upload your own certificates, either self-signed or from a well-trusted certificate
-vendor.
+Sandstorm can terminate TLS connections for most conventional configurations you may wish to employ. This
+includes automatic certificate renewal of certificates if you use a supported DNS provider, as well as the
+ability to manually upload your own certificates, either self-signed or from a well-trusted certificate vendor.
 
 ### How to get HTTPS on your Sandstorm install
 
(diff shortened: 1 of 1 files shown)


docs/administering/ssl.md
```diff
@@ -1,29 +1,38 @@
 ## Sandstorm and HTTPS
 
-If you are using a hostname like `example.sandcats.io`, then you likely already have working HTTPS
-(SSL) for your hostname. This page provides details on a variety of options for setting up HTTPS,
-including the `sandcats.io` free certificates.
+Sandstorm can now terminate TLS connections for most conventional configurations you may wish to employ.
```
zenhack, line 3: I think we should leave out the word "now" here; the fact that this didn't used to be the case isn't super relevant to folks reading these docs (as opposed to release notes or such).

docs/administering/ssl.md
```diff
@@ -1,29 +1,38 @@
...
+This includes automatic certificate renewal of certificates, if you utilize a supported DNS provider, as well
```
zenhack, line 4: nitpick: "utilize" is an unnecessarily big word for "use," and IMO it scans poorly here. Same thing in another place or two below (I'm not too inclined to die on this hill though...).

docs/administering/ssl.md
```diff
@@ -1,29 +1,38 @@
...
-You have a few options.
+If you are using a hostname like `example.sandcats.io`, then you likely already have working HTTPS for your
+hostname. [This page](sandcats-https.md) provides help and advice for enabling and troubleshooting HTTPS for
+Sandcats-based installs.
 
-- Use [sandcats.io free HTTPS](sandcats-https.md), a free service of the Sandstorm.io company. Read
-  details, including how to enable/disable, on that page. Sandstorm automatically renews these
-  certificates, and they are valid in virtually all browsers.
+If you utilize a supported DNS provider (Sandcats.io, Cloudflare, Digital Ocean, DNSimple, Duck DNS, GoDaddy,
+Gandi, Namecheap, Name.com, AWS Route 53, or Vultr), you can create an ACME account from "SSL/TLS Certificates"
```
zenhack, line 15: I think I would break this list out into a proper bullet list below; this sentence scans poorly with the huge parenthetical in the middle of it. Maybe end this paragraph with something like "The currently supported DNS providers are:" and then follow with a bullet list.

(This will also make it a bit nicer to edit if/when we add more providers).

docs/administering/ssl.md
ocdtrekkie: I also feel the first comma here is extraneous. And so shortening the sentence fragment I wish to merge into the other sentence fragment is nice.

docs/administering/ssl.md
ocdtrekkie: Agreed

docs/administering/ssl.md
ocdtrekkie: It feels like... really long to do that this way... but I did it. What do you think?

PR #3366, opened 2020-06-10
Improve the "unknown host" error page.

zenhack:
The main thing this does is re-order some of the information. From the
questions we get it seems like 99% of the time when a user sees this
page it's because they visited the wrong domain when trying to go to
Sandstorm's main interface, but the old error message de-emphasizes this
possibility, and we've seen a lot of confusion from users.

This patch puts the most likely possibility front-and-center, so users
who don't read the full wall of text still see it. Static publishing
is a less common cause of this error, so move that further down.

This also structures the page a bit more clearly, with each possible
source of errors being under a large heading indicating "If you were
tring to do X," to make finding information generally easier.


zenhack:
Just noticed I'd accidentally committed a bunch of updates to the submodules with the second patch on this. Fixed & force pushed.

PR #3369, opened 2020-06-19
Put grains in cgroups.

### LINKED ISSUE #3186: Creating a backup while the grain is running could cause corruption.
@orblivion brought this up on the mailing list, and this is an issue I was aware of but I don't think we have an open ticket for it so here it is:

If we take a .zip backup while a grain is running, there is a possibility of data corruption of the grain is modifying the filesystem while we are copying it. We should shut down the grain while taking the backup to ensure integrity.

zenhack:
This patch adds rudimentary support for cgroups (v2). Currently, we create a cgroup `grains` under the main sandstorm cgroup, and then put each grain in a cgroup within that.

Right now this doesn't really achieve anything useful, as we're not setting any policy on the cgroups, but this will enable future changes including:

- Making sandstorm robust to DoS attacks by grains; right now it's entirely possible to bog down a sandstorm box using e.g. a fork bomb inside a grain, but using cgroups we can place appropriate limits to prevent this kind of thing.
- Atomically pausing grains when taking backups, which would solve #3186.

Since in some cases (notably if we're not running as root) we may not be able to gain the ability to manage cgroups, this gracefully degrades to the status quo if we are unable to do so. Note that this is going to be the case when running the test suite (since it doesn't start as root).

Also worth noting is that on most modern systems, systemd starts up and assigns most of the cgroup controllers to the legacy cgroup v1 hierarchies, so on such a system we won't be able to actually impose limits (though sorting into cgroups still works). Users can address this by supplying a kernel command line parameter, but we probably can't make it automatically work very easily. Hopefully distros will gradually migrate away from this setup and eventually it will "just work" on most systems.

I tested this both on my local dev setup and thelibrary.sandcats.io, where it appears to work. I would appreciate others testing on their setups (possibly running different distros from me and such).

### DIFF
--- a/src/sandstorm/cgroup2.c++
+++ b/src/sandstorm/cgroup2.c++
@@ -4,36 +4,36 @@
 #include <kj/debug.h>
 
 namespace sandstorm {
-  Cgroup::Cgroup(kj::StringPtr path)
-    : dirfd(raiiOpen(path, O_DIRECTORY|O_CLOEXEC))
-  {}
+Cgroup::Cgroup(kj::StringPtr path)
+  : dirfd(raiiOpen(path, O_DIRECTORY|O_CLOEXEC))
+{}
 
-  Cgroup::Cgroup(kj::AutoCloseFd&& dirfd)
-    : dirfd(kj::mv(dirfd))
-  {}
+Cgroup::Cgroup(kj::AutoCloseFd&& dirfd)
+  : dirfd(kj::mv(dirfd))
+{}
 
-  Cgroup Cgroup::getOrMakeChild(kj::StringPtr path) {
-    int status;
-    do {
-      errno = 0;
-      status = mkdirat(dirfd.get(), path.cStr(), 0700);
-    } while(status != 0 && errno == EINTR);
+Cgroup Cgroup::getOrMakeChild(kj::StringPtr path) {
+  int status;
+  do {
+    errno = 0;
+    status = mkdirat(dirfd.get(), path.cStr(), 0700);
+  } while(status != 0 && errno == EINTR);
 
-    if(status < 0) {
-      KJ_REQUIRE(errno == EEXIST);
-    }
-
-    return Cgroup(raiiOpenAt(dirfd.get(), path, O_DIRECTORY|O_CLOEXEC));
+  if(status < 0) {
+    KJ_REQUIRE(errno == EEXIST);
   }
 
-  void Cgroup::removeChild(kj::StringPtr path) {
-    KJ_SYSCALL(unlinkat(dirfd.get(), path.cStr(), AT_REMOVEDIR));
-  }
+  return Cgroup(raiiOpenAt(dirfd.get(), path, O_DIRECTORY|O_CLOEXEC));
+}
 
-  void Cgroup::addPid(pid_t pid) {
-    auto procsfd = raiiOpenAt(dirfd.get(), "cgroup.procs", O_WRONLY);
-    auto pidStr = kj::str(pid);
-    auto cStr = pidStr.cStr();
-    KJ_SYSCALL(write(procsfd.get(), cStr, strlen(cStr)));
-  }
+void Cgroup::removeChild(kj::StringPtr path) {
+  KJ_SYSCALL(unlinkat(dirfd.get(), path.cStr(), AT_REMOVEDIR));
+}
+
+void Cgroup::addPid(pid_t pid) {
+  auto procsfd = raiiOpenAt(dirfd.get(), "cgroup.procs", O_WRONLY);
+  auto pidStr = kj::str(pid);
+  auto cStr = pidStr.cStr();
+  KJ_SYSCALL(write(procsfd.get(), cStr, strlen(cStr)));
+}
 };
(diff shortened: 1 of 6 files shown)


zenhack:
Ack, noticed the way I phrased the commit message is going to trigger #3186 being closed when this is merged, which is not what we want. We'll have to re-open it.

src/sandstorm/cgroup2.h
```diff
@@ -0,0 +1,35 @@
+#ifndef SANDSTORM_CGROUP2_H_
+#define SANDSTORM_CGROUP2_H_
+
+#include <sys/types.h>  // For pid_t
+#include <kj/io.h>
+
+namespace sandstorm {
+  class Cgroup {
```
kentonv, line 8: Nitpick: We don't normally indent the body of a namespace.

(Also in the .c++ file.)

src/sandstorm/cgroup2.c++
```diff
@@ -0,0 +1,39 @@
...
+  Cgroup::Cgroup(kj::AutoCloseFd&& dirfd)
+    : dirfd(kj::mv(dirfd))
+  {}
+
+  Cgroup Cgroup::getOrMakeChild(kj::StringPtr path) {
+    int status;
+    do {
+      errno = 0;
+      status = mkdirat(dirfd.get(), path.cStr(), 0700);
+    } while(status != 0 && errno == EINTR);
```
kentonv, line 20: Maybe use `KJ_SYSCALL_HANDLE_ERRORS`, which takes care of EINTR handling and will let you handle EEXIST...

src/sandstorm/run-bundle.c++
```diff
@@ -1296,6 +1296,8 @@ private:
       unshareUidNamespaceOnce();
     }
 
+    KJ_SYSCALL(unshare(CLONE_NEWCGROUP));
```
kentonv, line 1299: What happens to people running older kernels that don't understand `CLONE_NEWCGROUP`? Granted, this flag is quite old now, but I'd hate for someone who has had a sandstorm server running on an ancient kernel just fine up until now find it suddenly breaks when they auto-update.

src/sandstorm/run-bundle.c++
```diff
@@ -1370,6 +1372,31 @@ private:
     // in, copy its contents to /etc/resolv.conf.host-initial so we can use that if needed.
     backupResolvConf();
 
+    // Mount the cgroup2 filesystem at run/cgroup2:
```
kentonv, line 1375: What happens if the host system is running cgroups v1? Can a namespace still use v2?

zenhack: You can have both mounted simultaneously, although they can't be managing the same controllers. So if v1 has already claimed the cpu, memory, etc controllers, then mounting here will still work, and we can still do things like free the cgroup, but we'll need to gracefully handle the case where controllers we want to use aren't available. But right now we don't try to use any of them anyway, so it's fine.

src/sandstorm/cgroup2.c++
zenhack: Aha, knew something like that had to exist. Will do.

src/sandstorm/run-bundle.c++
zenhack: I suppose we could treat this as non fatal and just disable use of cgroups if it does fail.

zenhack: Pushed a commit that changes this; if unshare returns EINVAL it just skips cgroup setup, which should do the right thing on older kernels.

zenhack:
@kentonv, I believe I've addressed each of your comments.

kentonv:
I think before we can merge this we need to actually test it on:
* A very old Linux kernel (we claim to support all the way back to 3.10).
* A kernel that has CLONE_NEWCGROUP but does not have cgroupsv2 -- if there is any such thing. (Can cgroupsv2 support be turned off at compile time?)
* The oldest kernel version that supports CLONE_NEWCGROUP and cgroupsv2. I worry about the possibility that the very first release of cgroupsv2 didn't work for some reason.

(Alternatively, we could actually check the kernel version directly, and only attempt to use cgroups on kernels that we are confident are new enough. Though I still wonder what happens if cgroupsv2 is not compiled in / not enabled.)

zenhack:
I tested this on a CentOS 7 vm, which ships 3.10, and it works.

From digging around in the kernel tree/history: it looks like cgroups2 is older than cgroup namespaces, and I only see `CONFIG_CGROUPS`; there doesn't appear to be an option to disable only v2. So I don't think we have to worry about systems that support cgroup namespaces but not cgroups2. However, digging through the implementation, it looks like `CONFIG_CGROUPS` is disabled entirely, `CLONE_NEWCGROUP` will just silently noop, rather than returning EINVAL, so we probably need to be ready for the subsequent call to mount() to fail, and handle that gracefully; I'll adjust.

I'll also try to test on a kernel with very early cgroup ns support.

zenhack:
Made the change to mount() handling. Earliest version with cgroup namespaces appears to be 4.6, will find time to test soonish.

zenhack:
Ok, tested on a 4.6 kernel, made some minor adjustments (see the last commit). It seems to not break sandstorm on that version, though it's not actually creating the cgroups for reasons I haven't dug into. I'm not sure how much energy I want to spend making sure this does more than graceful degradation on older systems; I think I'm willing to settle for this. On earlier kernel versions unshare() returns EINVAL and the whole thing is skipped, and per above I tested on a 3.10 kernel and sandstorm still seems to work.

PR #3374, opened 2020-06-24
Add `-> stream` to a couple capnp methods.

### LINKED ISSUE #3373: Use new capnp 'streaming' functionality in our schema.
Capnp 0.8 introduced a new "streaming" feature:

https://capnproto.org/news/2020-04-23-capnproto-0.8.html#multi-stream-flow-control

There are some interfaces in our schema that would benefit from this; we should add the `-> stream` where appropriate.

zenhack:
...to take advantage of the new flow-control stuff in capnp 0.8

This also requires some mechanical changes to the C++ uses of these,
but it's mostly just a matter of dropping some calls to ignoreResult()
and the like.

Closes #3373

### DIFF
--- a/src/sandstorm/web-session.capnp
+++ b/src/sandstorm/web-session.capnp
@@ -463,7 +463,7 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
   }
 
   interface WebSocketStream {
-    sendBytes @0 (message :Data);
+    sendBytes @0 (message :Data) -> stream;
     # Send some bytes.  WARNING:  At present, we just send the raw bytes of the WebSocket protocol.
     # In the future, this will be replaced with a `sendMessage()` method that sends one WebSocket
     # datagram at a time.
--- a/shell/imports/server/drivers/ip.js
+++ b/shell/imports/server/drivers/ip.js
@@ -36,6 +36,10 @@ class ByteStreamConnection {
   }
 
   write(data) {
+    // TODO: try to apply some backpressure? Node docs say that write()
+    // always succeeds and just buffers the data if needed, so we could
+    // end up burning a bunch of memory if the sender is sending too
+    // fast.
     this.connection.write(data);
   }
 
--- a/src/sandstorm/app-index/app-index.c++
+++ b/src/sandstorm/app-index/app-index.c++
@@ -88,7 +88,7 @@ public:
         auto promises = kj::heapArrayBuilder<kj::Promise<void>>(3);
         auto req1 = stream.writeRequest();
         req1.setData(content.getContent());
-        promises.add(req1.send().then([](auto&&) {}));
+        promises.add(req1.send());
         promises.add(stream.doneRequest().send().then([](auto&&) {}));
         promises.add(stream.getResultRequest().send().then([](auto&&) {}));
 
--- a/src/sandstorm/sandstorm-http-bridge.c++
+++ b/src/sandstorm/sandstorm-http-bridge.c++
@@ -503,7 +503,7 @@ private:
   bool aborted = false;
 
   kj::Maybe<kj::Own<kj::PromiseFulfiller<void>>> writeReady;
-  capnp::Request<ByteStream::WriteParams, ByteStream::WriteResults> nextWrite = nullptr;
+  capnp::StreamingRequest<ByteStream::WriteParams> nextWrite = nullptr;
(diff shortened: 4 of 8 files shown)


src/sandstorm/web-session.capnp
```diff
@@ -463,7 +463,7 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
   }
 
   interface WebSocketStream {
-    sendBytes @0 (message :Data);
+    sendBytes @0 (message :Data) -> stream;
```
kentonv, line 466: Hmm, I wonder if this is going to be problematic since there's no `end()` or `done()` method. That goes against the recommended pattern and I haven't thought a whole lot about what exactly happens in this case.

I _think_ that if you queue up a bunch of streaming calls and then drop the capability, all calls will be delivered before the destination object is destroyed...

You might not find out if an exception was thrown on the server side, but I guess that's not expected anyway for WebSocket... the application has to detect problems by replying.

So......... it might work?

kentonv:
Did you look at all the callers of these methods to verify that they wait for the returned promise to resolve before sending the next call?

I think in some places we had hacks where we'd try to have multiple writes in-flight at once. Those hacks are no longer necessary with streaming.

zenhack:
You're right there's a bunch of logic in WebSocketPipe that (if we decide it's safe to use this for sendBytes()) is now obsolete. 

There's also this loop, which looks like we could almost just replace it with a call to pump():

https://github.com/sandstorm-io/sandstorm/blob/master/src/sandstorm/supervisor.c%2B%2B#L2311-L2331

...except that FdInputStream isn't a subclass of AsyncInputStream. It looks like kj has support for wrapping an fd in something that is an instance, but we'd have plumb the lowlevelprovider down to it.

Other than those two, it does look like everything waits for the previous write to complete before starting the next one (there are were bunch of comments about adding paralellism in the future).

I'm pushing another couple commits that address everything but that loop in the supervisor; I'll stare at things a bit more later. Though if you're not confident that doing this with sendBytes() is correct maybe we should back that part out.

kentonv:
> except that FdInputStream isn't a subclass of AsyncInputStream.

That code is reading from a file, not a socket. File reads are always synchronous (ignoring AIO which is complicated and broken). LowLevelAsyncIoProvider actually can't wrap disk files. But you don't need it to, anyway. You could write an implementation of `kj::AsyncInputStream` that wraps `kj::InputStream` and lets all reads() execute synchronously. Then you could use pumpTo(). That said, it might actually be easier to directly write a pump loop here, I dunno.

I think the `sendBytes()` stuff should work, and if it doesn't, probably we should make it work.

zenhack:
I converted the loop into a recursive function that waits on the intermediate promises.

I poked around a bit looking at the JavaScript side of things too, and there are a number of places where we implement ByteStream servers and the write() method won't provide any backpressure, which is possibly a DoS vector, which we should address, but this patch doesn't make it any worse I think. I added a comment in one such place.

I think this PR is good to go, BUT: I did hit a segfault at one point while testing. I haven't been able to reproduce it. It seems likely that it's not related to this PR, but I wanted a second opinion. Some things I noticed.

- Some of the line numbers in the stack trace seem off. I told addr2line to interpret the trace relative to the sandstorm binary, but I have a hunch that the stack trace may actually be coming from somewhere else, maybe node (see below). Unfortunately, I'd lost track of the original list of addresses before I noticed this.
- The stack trace seems to originate inside realloc, and seems to be going through the capnp dynamic API (hence my hypothesis that this is actually happening in node).

I see two possibilities:

1. This is just some existing bug that I got unlucky and hit once, but isn't related to this change.
2. `node-capnp` has a problem with the streaming schema.

(1) seems much more likely; from what I understand the streaming should be un-obtrusive for existing implementations that don't know about it. But maybe the dynamic API is treating it specially in a way that causes node-capnp to barf?

@kentonv, interested in your thoughts.

[stack.txt](https://github.com/sandstorm-io/sandstorm/files/4866975/stack.txt)

kentonv:
I'm pretty sure a stack trace from Node would not be decodeable at all because the addresses would be from a dynamically-loaded library which are usually loaded in a totally different address range from static binaries.

But the stack trace is clearly nonsense. It looks like what often happens when you decode a stack trace against a slightly different version of the binary than the one that created it.

Anyway... I think we should just merge.

PR #3377, opened 2020-06-26
Add a 'clone grain' button.

zenhack:
This adds a "clone grain" button to the top bar on the grain view, right next to the download backup button. It was pretty simple to implement; internally it just creates a backup and then immediately restores it as a new grain. There are a couple loose ends, so this is WIP:

- [x] In the build system we're doing something with the topbar icons to change the colors from black to white, but it doesn't work for copy.svg, which seemed like it made the most sense for this button (we weren't previously using the icon). Need to fix this, as right now it's a black icon on a dark background and isn't really visible.
- [x] There's no progress indicatior, so for larger grains where the backup & restore might actually take some time, the user won't see feedback.
- [x] We don't do anything in response to errors other than log a console message. We should provide better feedback here too.

...but it works!

I'll try to find some time to tie up the loose ends soonish.


ocdtrekkie:
It's hard for me to express how excited I am that this exists.

zenhack:
Alright, I pushed some commits to fix the icon and provide a bit of user feedback. I'd call this ready for review.

The way I wrote the code it's *supposed* to also include a spinner in the dropdown that it displays during the clone, but it doesn't, just the 'Cloning Grain...' text, and I can't figure out why. Open to insights. I'm also not really wedded to having a spinner there in the first place, though it seems sensible.

An actual progress indicator would be nicer, but we don't have any way to actually get that information without a much more invasive change.

kentonv:
Looks like there are conflicts... in the number of linter errors. Doh.

Care to post a screenshot?

zenhack:
Fixed the conflicts. Screenshot attached.

Screen Shot 2020-07-04 at 14 05 17

zenhack:
Fixed the conflicts, again.

Issue #3370, opened 2020-06-20
Instructions for extracting Sandcats certs no longer valid


ocdtrekkie:
In https://docs.sandstorm.io/en/latest/administering/ssl/ we link to an archived tutorial (and script) that extracts the Sandcats certs from the filesystem. After ACME support, the certs are no longer on the file system, and this tutorial doesn't work.

 18:27 <kentonv> you can extract the tls keys from mongo
18:27 <kentonv> `db.settings.findOne({_id: "tlsKeys"}) `
 18:30 <kentonv> this hacky terminal command should work for extracting the key: `echo 'db.settings.findOne({_id: "tlsKeys"})' | sudo sandstorm mongo`
18:31 <kentonv> so if you had a cron job that does that once a day and saves the output to the right places you should be good 

That tutorial already came from a dead site, so we should remove it, and replace it with current instructions.

ocdtrekkie:
Fixed by #3378 and #3382

PR #3378, opened 2020-06-30
Updated cron script for extracting certificates

Michael-S:
The original script to extract Sandcats certificates no longer works,
per Sandstorm issue #3370.  This is an attempt at a new script that
appears to work properly.

The new script could also be copied to a repository owned by the
Sandstorm project or one of the committers, and it's trivial shell
commands so I'm happy to relicense it upon request.

### DIFF
--- a/docs/administering/ssl.md
+++ b/docs/administering/ssl.md
@@ -44,10 +44,11 @@ To share port 443 with other services on the same machine:
 - You [can install `sniproxy` to share port 443](sniproxy.md) between your existing server and Sandstorm so that
   Sandstorm can manage (and autorenew) its own certificates. This allows you to combine an **existing
   web server on port 443** with Sandstorm.
-  
+
 - You [can follow this guide](https://web.archive.org/web/20190922195059/https://juanjoalvarez.net/es/detail/2017/jan/12/how-set-sandstorm-behind-reverse-proxy-keeping-you/)
-  that explains how to use a [cron script](https://github.com/Michael-S/sandstorm_certs_extract_cron)
-  to extract the certificates from your Sandstorm-managed TLS Sandstorm installation. Please note that the
-  guide itself references an older cron script that does not work with Sandstorm versions newer than
-  June 2020. The extracted certificates will be a format where your reverse proxy can use them so it can
-  serve Sandstorm by HTTPS along with any other services on your server.
+  that explains how to use a cron script to extract the certificates from your installation with
+  Sandstorm-managed TLS. Please note that the guide itself references an older cron script that
+  does not work with Sandstorm versions newer than June 2020, but there is an updated script
+  [here](https://github.com/Michael-S/sandstorm_certs_extract_cron). The extracted certificates
+  will be formatted so your reverse proxy can use them to serve Sandstorm by HTTPS
+  along with any other services on your server.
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
I know I'd really like to get the guide itself brought in-house as a Docs page at some point, since I think linked to an archive.org post is a little sad. I am not inherently opposed to linking to external repositories, particularly if they're hosted on GitHub or another fairly stable web presence.

My default inclination is to merge, as it is better than what we have. But I would like to run it by @zenhack as well.

If I had one bit of feedback, it would be that technically this should work for any certificate renewed via ACME in Sandstorm, not just Sandcats.io, although that is the most likely reason someone would use this rather than operating their own certbot. We do want to strive to detach assumptions in the docs that Sandstorm's HTTPS is only for Sandcats.

ocdtrekkie:
The term I started using in my previous docs edits was "Sandstorm-managed TLS" to refer to any scenario where the certificate was served/renewed by Sandstorm itself as opposed to a reverse proxy or the like.

Michael-S:
That all makes sense to me.  I don't want to rewrite the archived blog post on a reverse proxy because that's not what I'm extracting the certificates for so.  That means I can't test it.  I run my own syncthing (file synchronization service: https://github.com/syncthing/syncthing ) server on the same machine that runs Sandstorm, and I extract the certificates from Sandstorm for that. 

I will try to update the wording.

ocdtrekkie:
That's a big part of why I haven't tackled rewriting the blog post yet: It's definitely something I wouldn't have personally tested.

zenhack:
Minor, but I might change:

- Make the 'cron script' text not a link.
- At "... guide itself references an older cron script that does not work with Sandstorm versions newer than
  June 2020.", maybe add "but there is an updated script [here]" with a link.

I like that a bit better because it makes it harder to go fetch the script without taking in the context; you have to read the thing warning you that the post has some bitrot.

+1 for better than what we have.

docs/administering/ssl.md
```diff
@@ -46,7 +46,8 @@ To share port 443 with other services on the same machine:
   web server on port 443** with Sandstorm.
   
 - You [can follow this guide](https://web.archive.org/web/20190922195059/https://juanjoalvarez.net/es/detail/2017/jan/12/how-set-sandstorm-behind-reverse-proxy-keeping-you/)
-  that explains how to use a [cron script](https://github.com/juanjux/sandstorm-sandcats-cert-installer) 
-  to extract the certificates from your (sandcats.io enabled) Sandstorm installation to a location and 
-  format where your reverse proxy can use them so it can serve Sandstorm by HTTPS, keeping your 
-  sandcats.io domain and free auto-renewable certificates, along with any other services on your server.
+  that explains how to use a [cron script](https://github.com/Michael-S/sandstorm_certs_extract_cron)
+  to extract the certificates from your Sandstorm-managed TLS Sandstorm installation. Please note that the
```
ocdtrekkie, line 50: Possibly pedantic, but "installation with Sandstorm-managed TLS" would read better and be less repetitive.

Michael-S:
Thanks, both of those suggestions make sense.  I also rewrote the ending sentence in an attempt to be clearer.

PR #3380, opened 2020-07-02
Clarify installer downloads over HTTPS

ocdtrekkie:
I read this in the Docs, and I was briefly like "I sure hope we don't download the bundle using HTTP". Which, of course, we don't. It was a quote from text in the installer script, so I fixed it in both places.


kentonv:
FWIW, since we verify signatures, technically downloading over HTTP would be OK.

ocdtrekkie:
I do recall that, but since we do use HTTPS, we shouldn't give anyone a reason to complain.

PR #3382, opened 2020-07-03
Source web page was updated with the new script

juanjux:


ocdtrekkie:
Awesome! And your blog is back! I will cc @Michael-S and @zenhack here.

One quick thing about the blog post is that:
> Double check that the files are on [sandstorm_dir]/data/var/sandcats/https/yourname.sandcats.io (they are like timestamps with .csr or response-json extensions or no extension, with the timestamp being the expiration date)
is a step that I think is no longer valid for the current way this works.

ocdtrekkie:
I am going to go ahead and merge this. Thanks!

PR #3383, opened 2020-07-04
Add CLI for configuring ACME and use it in installer script

kentonv:
Currently the CLI doesn't support directly setting certificates but that would be easy for someone to add. I'm out of time this weekend, though.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -281,7 +281,7 @@ USE_DEFAULTS="no"
 USE_EXTERNAL_INTERFACE="no"
 USE_SANDCATS="no"
 SANDCATS_SUCCESSFUL="no"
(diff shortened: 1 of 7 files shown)


kentonv:
Once this is released, new installs will no longer use globalsign. After a short period to let things settle, we can remove all the globalsign code.

kentonv:
Whoops I broke the build because I apparently fixed a linter error. Will correct shortly.

kentonv:
@ocdtrekkie We should now update the docs to tell people how to configure TLS from the command line, for non-sandcats.

Would also be nice to update the installer to automate some of that, perhaps.

ocdtrekkie:
Yeah, I have this thread in my inbox still so I can review it and look at next steps for docs and such. 👍

PR #3386, opened 2020-07-06
Remove stray HTTPS = sandcats.io in installer text

ocdtrekkie:
No longer true just for Sandcats.io.


kentonv:
Technically the installer still only supports HTTPS via sandcats, but seems fine to change this in anticipation I suppose.

PR #3393, opened 2020-07-23
Fix SVGs in App Index review interface

### LINKED ISSUE #3392: App index doesn't display SVGs in review interface
This is a tiny literally-only-affects-me bug because it's in app market review. But I spent some time looking at if I can fix it, and the fix is probably very, very simple:

image

This is an escaping bug, in that when encodeURIcomponent is applied to the data portion of the SVG, such that the SVG data looks like `%3Csvg width='44'`, it displays fine.

image

The code governing the display of SVGs in this interface is here: https://github.com/sandstorm-io/sandstorm/blob/master/src/sandstorm/app-index/review.html#L93 and I am wondering if replacing the "data" on that line with `encodeURIComponent(data)` will fix it?

ocdtrekkie:
Fixes #3392

This is _completely untested_ but it appears to be a standard built-in of JavaScript, so I'm not _expecting_ it to cause a problem. In my dev console, URI encoding the SVG data fixed display. As SVGs currently do not display at all, I can't imagine this change being _worse_.

### DIFF
--- a/src/sandstorm/app-index/app-index.capnp
+++ b/src/sandstorm/app-index/app-index.capnp
@@ -165,8 +165,8 @@ const pkgdef :Package.PackageDefinition = (
 
   manifest = (
     appTitle = (defaultText = "Sandstorm App Index"),
-    appVersion = 3,
-    appMarketingVersion = (defaultText = "2016-02-22"),
+    appVersion = 4,
+    appMarketingVersion = (defaultText = "2020-08-04"),
 
     actions = [
       ( title = (defaultText = "New Sandstorm App Index"),
--- a/src/sandstorm/app-index/review.html
+++ b/src/sandstorm/app-index/review.html
@@ -90,7 +90,7 @@
       var fieldHandlers = {
         svg: function (data) {
           var img = document.createElement("img");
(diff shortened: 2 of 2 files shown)


ocdtrekkie:
I don't know the deployment process for this fix, but I wonder if we should/need to increment: https://github.com/sandstorm-io/sandstorm/blob/master/src/sandstorm/app-index/app-index.capnp#L168

zenhack:
You should probably bump the appVersion, yes.

This looks reasonable to me. Obviously you should build the package and test it though.

kentonv:
Yeah please bump the app version.

I think to deploy, I have to build the spk and install it on alpha...

ocdtrekkie:
I'll do that. I haven't done so yet because I am considering trying to fix one more extreme peeve I have with the review interface too, but I haven't had time to look at it this past week.

ocdtrekkie:
I'm just going to nudge the version so we're review ready. I'm not sure I'm adequately equipped to make the fix I want to make, which is basically to truncate the changelog field to a reasonable amount of data (or, for extra points, make it an expandable field or something cool). Wekan's changelog, for instance, contains all 400+ versions, and makes the review interface difficult to scroll through.

PR #3398, opened 2020-08-02
Freeze grains during backup

zenhack:
...using cgroups, if available.

This addresses #3186 IF the system supports cgroups. I'm not sure it's worth the trouble to try to solve it otherwise.

This works on thelibrary.sandcats.io and my local dev setup. I'm marking it as a draft because there's more testing I want to do, but modulo any bugs shaken out by further testing, this is good to go.

In particular:

- [x] I still need to test it on some older kernels to make sure it doesn't break anything if things aren't supported.
- [ ] I also want to write a few automated tests esp. for code the paths that are hard to test manually (because they only come up if the grain is not already running when a backup is taken, which doesn't happen as part of normal workflows, though I believe it can be triggered by the meteor apis).


src/sandstorm/backend.c++
```diff
@@ -491,11 +520,32 @@ kj::Promise<void> BackendImpl::deletePackage(DeletePackageContext context) {
...
+    // Not running. Mark the grain as backing up, so nobody tries to boot it
+    // until we're done:
+    auto paf = kj::newPromiseAndFulfiller<void>();
+    supervisors.insert(std::make_pair(grainId, BackingUpGrain{paf.promise.fork()}));
+    backupFulfiller = kj::mv(paf.fulfiller);
+  } else {
+    // If cgroups is available, freeze the corresponding cgroup
+    // during the backup:
+    KJ_IF_MAYBE(cg, cgroup) {
+      auto grainCgroup = cg->getOrMakeChild(grainId);
```
kentonv, line 542: This `getOrMakeChild()` should always find an already-made cgroup, right? Maybe worth having a `getChild()` method that throws if it doesn't exist?

kentonv:
LGTM, just a minor comment.

src/sandstorm/backend.c++
zenhack: Correct, done (also merged changes back in from master). Still need to do some more testing though.

zenhack: Correct, done (also merged some changes back in from master). Still need to do some more testing though.

kentonv:
This is still marked draft. Is it still a draft?

zenhack:
Yeah, I still want to do the testing mentioned above, this just fell of my radar. Will probably find time for it soon.

zenhack:
Alright, I did some testing and am now comfortable with this being merged. Everything works fine on older kernels, and I tested this merged with the grain list backup pr to confirm that things work correctly when backing up grains that aren't running already.

Automating the tests seems difficult, and wouldn't run in CI anyway since the automated test suite doesn't run as root, which is required for cgroup support, so I'm tempted to just punt on that...

Issue #3192, opened 2020-01-24
Replace HackSessionContext with powerbox APIs, and remove it.


zenhack:
We've talked about this, but I can't find an open issue tracking it, so I'm creating one.

Now that the powerbox exists, we should start working towards replacing the functionality of `HackSessionContext`, so that we can remove it.

Things that need doing.

- [ ] Right now receiving email and static publishing both make use of a "public id." Instead:
  - [ ] To receive email, apps should make powerbox offers for `EmailSendPort`.
  - [ ] static publishing should also use some powerbox interface. There's an interface called `WebSite` that's speced out in `web-publishing.capnp`, but it does not appear to be referenced from anywhere else in the source tree.
- [ ] The method `getUserAddress` should be replaced. For most use cases, this can probably be replaced with a powerbox request for `EmailSendPort`.
- [ ] `httpGet` needs to be removed. Apps should use the powerbox with `ApiSession` instead.

@kentonv suggested having a grandfathering process for older apps that use these APIs, so we don't have to fix every existing app before removing an interface. We could start a new capnp file analogous to `appid-replacements.capnp`, in which we'd whitelist apps permitted to continuing using these APIs. 

For each API that we want to remove, once we have implemented the alternative functionality, we should identify popular legacy apps that use these APIs and, if porting is non-trivial, add them to the whitelist.

kentonv:
It would be cool if, for compatibility purposes, we could implement some sort of wrapper that actually runs inside the sandbox that implements HackSessionContext in terms of powerbox requests. sandstorm-supervisor could spawn this wrapper when it detects that it's loading an old app, based on "API version" in the metadata. In this case there would be no need to grandfather old apps.

However, I haven't thought through whether that would actually be practical to re-implement all the old functionality this way. Certainly it's likely that there would be some quirks caused by powerbox requests being injected where they didn't happen before, but maybe we can find a way to do it that works acceptably with all current known apps?

zenhack:
Quoting Kenton Varda (2020-01-26 04:16:01)

>    It would be cool if, for compatibility purposes, we could implement
>    some sort of wrapper that actually runs inside the sandbox that
>    implements HackSessionContext in terms of powerbox requests.
>    sandstorm-supervisor could spawn this wrapper when it detects that it's
>    loading an old app, based on "API version" in the metadata. In this
>    case there would be no need to grandfather old apps.
>
>    However, I haven't thought through whether that would actually be
[...]

Issue #3399, opened 2020-08-02
Mongo somtimes fails to start during CI.


zenhack:
I've seen this several times on PRs in the recent past, most recently:

https://github.com/sandstorm-io/sandstorm/runs/936456067?check_suite_focus=true#step:11:393

It looks like sometimes on startup mongo fails to elect itself as the master node (of its one-node cluster...), and thus sandstorm fails to start. I haven't ever seen this outside of CI.

ocdtrekkie:
I am going to close this because I haven't seen it in years, and we replaced our MongoDB version extensively in the `dev` branch.

PR #3407, opened 2020-08-11
Disable HTTP access through hack-session.

zenhack:
Per the comments in db.js, it is still possible to re-enable this for
now, through an un-documented database setting that users will have to
ask us about to learn of. In the comments for #3192 we'd planned on
making this an option in sandstorm.conf, but this is a bit easier to
implement and given that we're intentionally avoiding documenting this,
requiring people who need it to pop open `sandstorm mongo` seems fine.

When we finally remove this, we'll want to add a database migration that
deletes the setting.

---

It's up for discussion *when* we merge this; we could have it land in the next release if folks thing that's reasonable (that gives folks a few weeks to get the ttrss update). I could be talked into putting it off for another release if others think that's too soon.

We should also decide how long we're going to wait between this landing and actually outright removing the functionality.

### DIFF
--- a/shell/imports/sandstorm-db/db.js
+++ b/shell/imports/sandstorm-db/db.js
@@ -1131,6 +1131,23 @@ class SandstormDb {
     };
   }
 
+  allowLegacyHackSessionHttp() {
+    // Returns true iff the legacy hack-session http-related methods have been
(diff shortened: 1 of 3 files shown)


zenhack:
Looks like failure is #3399 again.

PR #3409, opened 2020-08-15
Content security policy for grains.

zenhack:
This blocks grains from loading external resources, with the exception
of images and media (per the comment, there are legitimate uses for
these that we want to allow; eventually we'll want the user to
specifically approve these).

---

While the pr itself is complete, before this lands we need to go through and find apps that this breaks, and maybe fix/update them. So far I've lightly tested:

- Davros
- uWiki
- Etherpad
- Ethercalc
- Draw.io
- Rocket.Chat
- Contact Otter
- ttrss

Of these, the following problems arose:

- [ ] Draw.io seems to be trying to pull in MathJax from a cdn, so I assume rendering some formulae will break
- [ ] As @orblivion pointed out on the mailing list, uWiki pulls in some third party resources, @ocdtrekkie is already working on it.
- [ ] Contact otter tries to pull in styles and scripts from elsewhere. At least the styles should be pulled in to the package. Google analytics should of course remain broken, as that's the whole point of all this.

The others shook out some bugs in the policy but trigger no violations with the current iteration. We should go through other apps to see what else breaks.

### DIFF
--- a/src/sandstorm/web-session-bridge.c++
+++ b/src/sandstorm/web-session-bridge.c++
@@ -1287,20 +1287,20 @@ kj::Promise<void> WebSessionBridge::handleResponse(
           tables.hContentSecurityPolicy,
           kj::str(
             "default-src 'none'; "
-#define UNSAFE "'unsafe-inline' 'unsafe-eval' data:; "
+#define UNSAFE "'unsafe-inline' 'unsafe-eval' data: blob:; "
             "img-src * " UNSAFE
             "media-src * " UNSAFE
             "script-src 'self' " UNSAFE
             "style-src 'self' " UNSAFE
             "child-src 'self' " UNSAFE
             "worker-src 'self' " UNSAFE
+            "font-src 'self' " UNSAFE
 
             // frame-src needs to allow references to BASE_URL, because
             // we allow apps to pull the content of offer-iframes from
             // there:
             "frame-src 'self' ", baseHttpHost, " ", UNSAFE
 #undef UNSAFE
-            "font-src 'self'; "
 
             // 'self' alone does not allow websocket connections; see:
             // https://github.com/w3c/webappsec-csp/issues/7
--- a/src/sandstorm/web-session-bridge.c++
+++ b/src/sandstorm/web-session-bridge.c++
@@ -100,12 +100,16 @@ WebSessionBridge::Tables::Tables(kj::HttpHeaderTable::Builder& headerTableBuilde
 
 WebSessionBridge::WebSessionBridge(
     kj::Timer& timer, WebSession::Client session, kj::Maybe<Handle::Client> loadingIndicator,
-    const Tables& tables, Options options)
+    const Tables& tables, Options options,
+    kj::Maybe<kj::String>&& host,
+    kj::Maybe<kj::String>&& baseHost)
     : timer(timer),
       session(kj::mv(session)),
       loadingIndicator(kj::mv(loadingIndicator)),
       tables(tables),
-      options(options) {}
+      options(options),
+      host(kj::mv(host)),
+      baseHost(kj::mv(baseHost)) {}
 
 void WebSessionBridge::restrictParentFrame(kj::StringPtr parent, kj::StringPtr self) {
   KJ_REQUIRE(!options.isApi, "can't apply frame restriction to API endpoint");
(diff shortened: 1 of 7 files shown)


ocdtrekkie:
This is super exciting to see. In my case I didn't even realize my package was loading external resources. I'd rather we just break them up front.

zenhack:
I pushed another commit that adds a config option to disable this, the idea being the same as for the db setting in #3407; if someone comes to us and needs a window in which they can keep an internal app working while porting it, we can tell them about this (otherwise undocumented) option. After a little while we'll remove it.

The warning statement isn't printing anything to my logs, which I don't quite understand, but otherwise this seems to work.

ocdtrekkie:
draw.io removed their Sandstorm packaging files from the repo some time ago. I believe Contact Otter just dropped Sandstorm support, though the .sandstorm directory is still there in the repo. (The site is dead though.)

I imagine both would accept a pull request (or fork), but presumably someone would have to do the work.

I think there's likely a lot of odd and ends this will break, but it has to be done at some point. With some sort of recovery path, as you've added, my hope is that's suitable to give time for someone to prod us if there's an app breakage that causes someone pain.

zenhack:
Ok, so I just went through and did at least basic testing of every app in the app market to see what we're dealing with. Most stuff worked, nothing was what I judged to be show-stoppingly broken. In detail:

- Several apps tried to load fonts through `data:` or `blob:` URLs. I've adjusted the policy to allow that since we do it for other resources and it's not actually a problem.
- A number of apps tried to load external fonts; no big deal, as they degrade gracefully (I wouldn't have noticed without looking at the js console):
  - TextEdit
  - Permanote
  - Giftr
  - NodeBB
  - Biandang
  - ShareLaTeX
  - Ghost
- Some apps try to load stylesheets from cdns. They "look wrong," but the functionality is still basically usable:
  - Brainstorm
  - Giftr
- Dillinger tries to load [katex](https://katex.org/) from a cdn, so presumably typesetting math will not work correctly (similarly to draw.io).
- For the Swagger editor, I see in the console:

```
Content Security Policy: The page’s settings blocked the loading of a resource at http://generator.swagger.io/api/gen/servers (“connect-src”). angular.js:12011
Content Security Policy: The page’s settings blocked the loading of a resource at http://generator.swagger.io/api/gen/clients (“connect-src”). angular.js:12011
```

And at the top of the page it displays:

```
Server code generator is unreachable! Client code generator is unreachable!
```

The basic editing functionality appears to work. I'm not familiar enough with what you're supposed to do with this app to figure out if anything is actually broken.

---

So, I'm comfortable with that level of breakage. We'll leave the flag in there in case somebody's got a private app that is horribly broken by this.

abliss:
On Sun, Aug 16, 2020, 02:57 Ian Denhardt <notifications@github.com> wrote:

>
>    - For the Swagger editor, I see in the console:
>
> Content Security Policy: The page’s settings blocked the loading of a resource at http://generator.swagger.io/api/gen/servers (“connect-src”). angular.js:12011
>
> Content Security Policy: The page’s settings blocked the loading of a resource at http://generator.swagger.io/api/gen/clients (“connect-src”). angular.js:12011
>
>
> And at the top of the page it displays:
>
> Server code generator is unreachable! Client code generator is unreachable!
>
>
> The basic editing functionality appears to work. I'm not familiar enough
> with what you're supposed to do with this app to figure out if anything is
> actually broken.
>

I believe this is used to do code-generation of a client and a server-stub
based on an openAPI definition. I think it all happens on the backend, and
this code was probably not packaged with the sandstorm app (which is likely
just the static html/js bundle?). So if you wanted to use this
functionality, there'd be no reason not to do it at the swagger.io page.

Nice work!
--Adam

>

ocdtrekkie:
Okay, I went ahead and filed notification issues with all of the apps with notable breakage. I am not too worried about the font-only ones that gracefully degrade. I am not sure how many notifications will turn into fixed packages, but since we are pre-Kenton-reviewing-this, we have some time. Since it's best practice for Sandstorm not to utilize client-side script loading from CDNs and the like, regardless of the rollout timeline and strategy, it is good to for people to know these apps are doing this.

src/sandstorm/web-session-bridge.c++
```diff
@@ -1248,6 +1254,59 @@ kj::Promise<void> WebSessionBridge::handleResponse(
       headers.set(tables.hContentSecurityPolicy, "default-src 'none'; sandbox");
 
       headers.set(tables.hAccessControlExposeHeaders, kj::strArray(exposedHeaders, ", "));
+    } else if(!allowLegacyRelaxedCSP) {
+      // Disallow loading of remote resources. Note the following:
+      //
+      // - Currently there are still exceptions for images and media, as these have
+      //   some legitimate use cases (e.g. embedding images in feeds in ttrss) and
```
orblivion, line 1261: Not a priority for now, but are there any ideas for how to prevent information leaking when getting images? A malicious application could put something private in a GET string, but sometimes GET strings are legitimately required to return an image (requested size, etc).

zenhack: This has been talked about on IRC a bit, but what I want to try is:

- By default, block images & media.
- Use CSP's report-uri/report-to mechanisms to detect when grains try to load external images & media.
- When that happens, in the Sandstorm UI, pop up a thing that says "Images & media have been blocked for your privacy [load images & media]."
- If the user clicks the load images button, we refresh the contents of the iframe with a relaxed CSP.

This is an approach that I think will be familiar to people from experience with email clients.

kentonv:
I think this might break the app index review app, which fetches app author profile info from keybase using client-side fetches. I guess I can enable the fallback option on alpha until someone has a chance to update the app index...

kentonv:
If there are reports of breakages and we need to push a quick release that reverts this (probably by changing the default value of the setting, rather than rolling back the code), let me know.

ocdtrekkie:
Reported on the mailing list: "ShareLatex seems to be broken. All buttons disappeared...  The functionality is still there, if you know, where the buttons were."

PR #3413, opened 2020-08-22
Remove the custom node-capnp import logic.

zenhack:
This removes the custom logic to import node-capnp, and instead
installs deps/node-capnp as an npm dependency of the shell directly.
To make this work, we also need to get the schema into meteor's
node_modules, so we create a package for those too.

---

No functional change, just cleans up one of the gnarly bits of our build system.

I also have a local change to node-capnp that skips creating the extra node_modules directory. When this lands I'll send a PR for that too.


kentonv:
Does this mean that the module gets built against the installed version of libcapnp, rather than the one built in-tree?

zenhack:
See my comment above: yes, but that was not my intent, which is why this is marked as a draft.

kentonv:
> See my comment above:

lol I can't read

zenhack:
Alright, I've gotten it to use the in-tree capnp now (and only build capnp.node once). This required changes in node-capnp's repo, so https://github.com/capnproto/node-capnp/pull/65 will have to be merged for this to work.

ocdtrekkie:
I'll hold on the update to node-capnp being merged, and then I'll mark this ready for review.

PR #3421, opened 2020-08-29
Some improvements to our eslint config

zenhack:
- Scan typescript files
- Be a bit more restrictive re: global environment; this could catch
  undefined variable errors that the old config would have missed.


ocdtrekkie:
Looks like we're up to like 9 test failures here?

ocdtrekkie:
The previous PR has most of them too. I'm wondering where these slid in a bit.

PR #3431, opened 2020-09-17
app-index: Use powerbox/ApiSession for keybase API

zenhack:
...instead of making requests client-side, as the latter will stop
working once we block it with CSP.

This adds a button the UI to request the keybase endpoint, which then
triggers a powerbox request. The resulting capability is used
server-side to query the keybase API.

### DIFF
--- a/src/sandstorm/app-index/keybase.c++
+++ b/src/sandstorm/app-index/keybase.c++
@@ -12,9 +12,7 @@
 #include <sandstorm/app-index/app-index.capnp.h>
 #include <sandstorm/util.h>
 
-namespace sandstorm {
-namespace appindex {
-namespace keybase {
+namespace sandstorm::appindex::keybase {
 
 kj::String getPowerboxDescriptor() {
   capnp::MallocMessageBuilder msg;
@@ -123,6 +121,4 @@ kj::Promise<kj::Own<LookupResults::Reader>> Endpoint::lookupFingerPrint(kj::Stri
   });
 }
 
-};
-};
 };
--- a/src/sandstorm/app-index/keybase.c++
+++ b/src/sandstorm/app-index/keybase.c++
@@ -108,7 +108,7 @@ kj::Promise<kj::Own<LookupResults::Reader>> Endpoint::lookupFingerPrint(kj::Stri
               capnp::MallocMessageBuilder msg;
               auto results = msg.initRoot<LookupResults>();
               json.handleByAnnotation<LookupResults>();
-              json.decode(bytesToString(bytes.asBytes()), results);
+              json.decode(kj::str(bytes.asChars()), results);
               return capnp::clone(results.asReader());
             }
             case WebSession::Response::Content::Body::STREAM:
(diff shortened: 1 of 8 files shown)


ocdtrekkie:
Should also bump the appVersion and appMarketingVersion on the PR.

zenhack:
Done.

Quoting Jacob Weisz (2020-09-17 10:58:17)

src/sandstorm/util.c++
```diff
@@ -940,4 +940,13 @@ void TwoPartyServerWithClientBootstrap::taskFailed(kj::Exception&& exception) {
   KJ_LOG(ERROR, exception);
 }
 
+kj::String bytesToString(kj::ArrayPtr<const kj::byte> bytes) {
```
kentonv, line 943: I think this is equivalent to:

    kj::str(bytes.asChars())

src/sandstorm/app-index/keybase.h
```diff
@@ -0,0 +1,53 @@
...
+#ifndef SANDSTORM_APPINDEX_KEYBASE_H_
+#define SANDSTORM_APPINDEX_KEYBASE_H_
+
+#include <kj/string.h>
+#include <capnp/serialize.h>
+#include <sandstorm/api-session.capnp.h>
+#include <sandstorm/app-index/keybase-api.capnp.h>
+#include <sandstorm/app-index/app-index.capnp.h>
+
+namespace sandstorm {
```
kentonv, line 26: In modern C++ (either 14 or 17, can't remember) you can do `namespace sandstorm::appindex::keybase {`.

I'm not sure if we're currently compiling with C++17 but if not we should start. Or maybe even C++20...

src/sandstorm/util.c++
zenhack: Good call; I'd missed `.asChars()`. Pushed  patch that uses that instead.

src/sandstorm/app-index/keybase.h
zenhack: Done. Seems to work with our existing build flags (`-std=c++1z`).

kentonv: 1z is 17. Cool.

src/sandstorm/app-index/keybase.c++
```diff
@@ -0,0 +1,124 @@
...
+              auto bytes = body.getBytes();
+              capnp::JsonCodec json;
+              capnp::MallocMessageBuilder msg;
+              auto results = msg.initRoot<LookupResults>();
+              json.handleByAnnotation<LookupResults>();
+              json.decode(kj::str(bytes.asChars()), results);
+              return capnp::clone(results.asReader());
+            }
+            case WebSession::Response::Content::Body::STREAM:
+              KJ_FAIL_ASSERT("TODO(now)");
```
kentonv, line 113: Hmm looks like you meant to handle this?

zenhack: Ack, yes, I did. I'll find some time in the next couple days, but looking at ExternalWebSession it looks like this isn't ever actually used currently, so no need to block on it I guess...

Issue #3410, opened 2020-08-15
Allow apps to opt-in to more restrictive CSP


zenhack:
#3409 adds a content-security-policy for grains running in a ui session, which (except for some holes that we'll plug in the future) prevents grains from leaking info to the outside world from the client via any mechanism csp can control.

However, the policy does allow some things that csp can block, and that might be useful to block as an app-bug mitigation measure: things like inline scripts and use of `eval()`. 

We can't just hard-block these unconditionally, because many apps actually use them, but it would be nice if apps that don't could opt in to the extra protection offered by a CSP that blocks these things.

My notion of what this would look like is that we'd add some fields to the package definition indicating whether to enable these extra protections. The default in the schema file would be off, so that it doesn't change the behavior for existing apps, but `spk init` should generate package definitions that explicitly turn these on, so they're the default for "new" apps, unless the packager opts out of them.

mnutt:
If there was an easy way to just diff the CSP headers an app sends and apply whichever is more restrictive that'd probably be the easiest to explain to developers and the most flexible. I also have a few routes I want to be more restrictive than others (serving untrusted content and setting `sandbox` attributes) and it would be nice to differentiate. But I could get by with an app-wide package definition.

zenhack:
I guess we could add the information to the response types in web-session.capnp, and then have the bridge handle it.

ocdtrekkie:
Isn't the bridge inside the app though? Couldn't one then implement a bridge with a less restrictive CSP?

zenhack:
Maybe I wasn't super clear. The idea would be, we'd first add fields to the `Response` struct:

https://github.com/sandstorm-io/sandstorm/blob/93d80ac4bf2ca721f239fb58395a32970f92e953/src/sandstorm/web-session.capnp#L229

...for each of the csp sources we want apps to be able to use. So we might do something like:

```capnp
enum CspSrc {
  self @0;
  dataUrl @2;
  blobUrl @3;
  eval @4;
  # ...
}

struct Response {
  # ...
  defaultSrc :CapnpSrc;
  scriptSrc :CapnpSrc;
  # ...
}
```

Importantly, this type would *not* include variants for anything that we don't want apps to be able to allow. This is the same thing we do with everything else in the HTTP protocol; you'll note many status, headers & such do not have representations in web-session, which means that the grain does not even have a way of telling sandstorm it would like to set them, since the web-session protocol can't even represent them.

[...]

PR #3432, opened 2020-09-17
Spec out API for apps to set further CSP restrictions.

zenhack:
Per #3410

Nothing is actually implemented yet; I'd like feedback on the design.

### DIFF
--- a/src/sandstorm/web-session.capnp
+++ b/src/sandstorm/web-session.capnp
@@ -226,9 +226,53 @@ interface WebSession @0xa50711a14d35a8ce extends(Grain.UiSession) {
     qValue @1 :Float32 = 1;
   }
 
+  struct ContentSecurityPolicy {
+    # A Content-Security-Policy header. By default, Sandstorm blocks all
+    # third party resources, but allows resources that do not let an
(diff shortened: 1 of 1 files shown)


kentonv:
Realistically what kinds of restrictions do we think apps are going to want to apply to themselves? I can see the use case for disallowing inline and eval, but beyond that, is there a strong use case for other restrictions? It seems like only those two bits are interesting -- I'm not sure what legitimate security concerns are solved by restricting the others (beyond what Sandstorm already intends to do by default).

If apps are only really likely to use those two bits then I'd say we should define two flags and leave the rest out. Otherwise I think the logic to merge the app's CSP with Sandstorm's policy is going to be sort of complicated without adding that much value.

zenhack:
@mnutt, have an opinion (since IIRC you suggested this on IRC)?

I don't feel that strongly; unless others see a reason to keep the full set I'm fine removing it -- it'll be easy to add later if we want it.

ocdtrekkie:
Whatever magical algorithm decides who I can assign to what has said I can't assign @mnutt to this.

PR #3434, opened 2020-09-21
Remove obsolete messges re: oasis.

zenhack:
Both of these events have come to pass; we don't need the notification
code for them any more.

### DIFF
--- a/shell/imports/client/apps/applist.html
+++ b/shell/imports/client/apps/applist.html
@@ -3,21 +3,6 @@
   {{#sandstormTopbarItem name="title" priority=5 topbar=globalTopbar }}{{_ "apps.title"}}{{/sandstormTopbarItem}}
   {{#if isSignedUpOrDemo}}
   <div class="app-list">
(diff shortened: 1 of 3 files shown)


ocdtrekkie (approved):

kentonv:
Should we take this a step further and remove all the code supporting those flags? Might even make sense to rip out all the billing code entirely since it was never really designed for anything other than Oasis, at this point it depends on a very old version of the Stripe API, and in general we're not maintaining it...

zenhack:
I removed other references to those flags. I don't really have an opinion on the billing code -- it seems like it might be nice to have in there if someone decides they want to offer Sandstorm hosting, but I'm not familiar with the code and don't know how close it is to being useful, and I don't know how different the stripe API is likely to be at this point.

ocdtrekkie:
Yeah, I kinda imagine it's useful as a "if you want to run with this, start here". Especially since Sandstorm hosting interest does stop by from time to time.

PR #3435, opened 2020-09-21
Add a backup button to the grain list.

zenhack:
...to allow multiple grains to be backed up by checking their boxes on this page.

There are some sharp edges currently, which we should fix:

- [ ] There is no feedback given to the user when they click the button, so
  if there is a delay between clicking it and the first backup being
  ready to download, they may be confused as to why nothing has
  happened.
- [x] The button is greyed out of you've selected grains that you don't own,
  since it isn't possible to back those up. But there's no indication
  given to the user as to why the button is disabled and this makes it
  annoying to back up all of your grains if others have shared grains with
  you, since you can't just use the select all button.

...so I'm marking this as a draft for now.


shell/imports/client/grain/grainlist-client.js
```diff
@@ -270,6 +271,28 @@ SandstormGrainListPage.bulkActionButtons = function (showTrash) {
...
+          // Users can only back up their own grains, so disable if
+          // they've selected anyone else's. TODO: give the user more
+          // feedback on this somehow.
+        },
+
+        onClicked: function (ownedGrainIds, _sharedGrainIds) {
+          ownedGrainIds.forEach((grainId) => {
+            // TODO: use the grain title, not the grain id, as the suggested
+            // file name.
+            makeAndDownloadBackup(grainId, grainId);
```
zenhack, line 292: I don't understand why, but in testing this actually does use the grain's title as the file name -- so perhaps the browser is ignoring the hint we have here anyway?

ocdtrekkie:
Can we add hover text to the disabled button? Similar to this:

image

zenhack:
Good idea, done.

kentonv:
Interesting. I wonder if this should be treated as a background task... request a backup be made, wait a while for Sandstorm to create the zip in the background, then download it when it's ready?

zenhack:
Updated this wrt. changes on master, and also did i18n for the new error message.

Re: feedback: we should probably just do the same thing we do for individual grain backups, i.e. display a popup noting that a backup is happening.

zenhack:
So I keep circling back to this and trying to get a popup displaying for in progress backups, but the control flow for making that happen is actually really gnarly, since it requires some action at a distance, and everything I come up with either ends up being major surgery or would pile on even more tech debt.

Meanwhile, most grains aren't so huge that the delay between clicking the button and the zip file being ready is actually that significant, and this feature has been sitting here unmerged for _three months_. I'm wondering if it makes sense to punt on that bit for now so that we can ship this, and figure out how to make doing this kind of thing less messy asynchronously.

Thoughts?

ocdtrekkie:
I am not opposed, but it's up to Kenton probably, as he noted it here.

zenhack:
I'm going to mark this as not-a-draft; I should double check that it hasn't bit-rotten, but it's been sitting here for over a year at this point and I think the perfect is the enemy of the good, per my prior comment.

zenhack:
I've merged master back in, and confirmed that this still works as expected.

PR #3445, opened 2020-10-03
Fix #3443

### LINKED ISSUE #3443: Unhandled exception in Promise: TypeError: Cannot read property 'catch' of undefined
Not sure if this causes me any actual problems, but it's spamming in my System Log, I thought it was at least potentially novel and interesting?

```
Unhandled exception in Promise:  TypeError: Cannot read property 'catch' of undefined
    at imports/server/backend.js:91:14
    at imports/server/async-helpers.ts:24:17
    at imports/server/async-helpers.ts:12:3
    at runWithEnvironment (packages/meteor.js:1286:24)
    at packages/meteor.js:1299:14
    at imports/server/async-helpers.ts:22:5
    at new Promise (<anonymous>)
    at inMeteor (imports/server/async-helpers.ts:21:10)
    at SandstormBackend.useGrain (imports/server/backend.js:90:12)
    at imports/server/scheduled-job.js:89:23
    at packages/meteor.js:550:25
    at runWithEnvironment (packages/meteor.js:1286:24)
```

...It'd still be nice if the log had timestamps so I had some idea when this happened, lol.

zenhack:
The comment for useGrain() says that the callback is supposed to return
a promise, but we're instead waiting for the promise, and then not
returning a result.

Per my comment in #3443, I assume we're seeing these messages hourly, since that's the interval at which ttrss's job runs. I'm going to deploy this to my server and confirm that it does in fact solve the problem

### DIFF
--- a/shell/imports/server/scheduled-job.js
+++ b/shell/imports/server/scheduled-job.js
@@ -87,7 +87,7 @@ export const runDueJobs = (nowMillis) => {
 
       intervalHandle = Meteor.setInterval(() => {
         globalBackend.useGrain(job.grainId, (supervisor) => {
-          waitPromise(supervisor.keepAlive());
+          return supervisor.keepAlive();
         });
         db.updateScheduledJobKeepAlive(job._id);
       }, KEEP_ALIVE_INTERVAL_MILLIS);


zenhack:
Can confirm this fixes the problem.

PR #3447, opened 2020-10-04
Fix a path traversal vulnerability involving static publishing.

zenhack:
N.B. I originally sent this to security@sandstorm.io out of an abundance of caution, but as @kentonv points out it's very unlikely any real apps are affected in a way that matters, and there are some deisgn questions regarding the solution, so per his request I'm opening a normal PR so we can discuss publicly.

---

Previously, getWwwFileHack() would unconditionally follow symbolic
links.  An attacker able to place symlinks in a grains /var/www could
abuse this to perform a path traversal attack, allowing regular files
visible to the supervisor but outside of the static publishing directory
to be accessed via static publishing.

This is limited to the grain's mutable storage, since the supervisor
self-sandboxes, but notably it does include the parent directory of the
sandbox, which is not normally accessible to the grain. This directory
contains the grain log.

The damage is mostly confined to the grain itself, as the supervisor has
access to little else besides the grain's storage, but this could still
be used to extract confidential information from a grain that allows
less-privileged users of the grain to upload files for static
publishing, if the upload mechanism allows symlinks.

With this patch, we avoid following symlinks automatically and instead
resolve them manually, refusing to follow any link that points outside
of www.

---

The design issue @kentonv raised via email:

> As for the ability for symlinks to refer to other app-visible files, I think it's debatable whether it should be the app's responsibility vs. Sandstorm's to validate symlinks. I can see an argument that Sandstorm should do it to defend the app, but I can also see an argument that it's useful for the app to be able to create symlinks that point outside of www, and maybe it should be the app's responsibility to make sure they don't point anywhere bad?

As written, the patch takes the more conservative approach of only following symlinks that stay inside the static publishing directory. My inclination is to keep it this way; so far I don't believe we've seen any use cases that would greatly benefit from the more relaxed approach, and it seems worthwhile avoid app authors needing to write fiddly validation code if we can.

### DIFF
--- a/src/sandstorm/util-test.c++
+++ b/src/sandstorm/util-test.c++
@@ -239,5 +239,71 @@ KJ_TEST("SubprocessSet") {
   promiseCat.wait(io.waitScope);
 }
 
+KJ_TEST("raiiOpenAtIfExistsContained") {
+  {
+    char tempdir[] = "/tmp/sandstorm-test.XXXXXX";
+    KJ_REQUIRE(mkdtemp(tempdir) != nullptr);
+    KJ_DEFER(KJ_SYSCALL(rmdir(tempdir)));
+
+    auto dir = raiiOpen(tempdir, O_DIRECTORY);
+
+    raiiOpenAt(dir.get(), "file", O_CREAT | O_RDWR);
+    KJ_DEFER(KJ_SYSCALL(unlinkat(dir.get(), "file", 0)));
+
+    KJ_SYSCALL(symlinkat("file", dir.get(), "link-to-file"));
+    KJ_DEFER(KJ_SYSCALL(unlinkat(dir.get(), "link-to-file", 0)));
+
+    KJ_SYSCALL(symlinkat("..", dir.get(), "link-to-parent"));
+    KJ_DEFER(KJ_SYSCALL(unlinkat(dir.get(), "link-to-parent", 0)));
+
+    KJ_SYSCALL(symlinkat("/", dir.get(), "link-to-root"));
+    KJ_DEFER(KJ_SYSCALL(unlinkat(dir.get(), "link-to-root", 0)));
+
+    KJ_SYSCALL(mkdirat(dir.get(), "subdir", 0700));
+    KJ_DEFER(KJ_SYSCALL(unlinkat(dir.get(), "subdir", AT_REMOVEDIR)));
+
+    KJ_SYSCALL(symlinkat("..", dir.get(), "subdir/link-to-parent"));
+    KJ_DEFER(KJ_SYSCALL(unlinkat(dir.get(), "subdir/link-to-parent", 0)));
+
+    KJ_SYSCALL(symlinkat("../..", dir.get(), "subdir/link-to-grandparent"));
+    KJ_DEFER(KJ_SYSCALL(unlinkat(dir.get(), "subdir/link-to-grandparent", 0)));
+
+    KJ_SYSCALL(symlinkat("/", dir.get(), "subdir/link-to-root"));
+    KJ_DEFER(KJ_SYSCALL(unlinkat(dir.get(), "subdir/link-to-root", 0)));
+
+
+    auto expectSucceed = [&](kj::StringPtr path) -> kj::AutoCloseFd {
+      KJ_IF_MAYBE(fd, raiiOpenAtIfExistsContained(dir.get(), path, O_RDONLY)) {
+        return kj::mv(*fd);
+      } else {
+        KJ_FAIL_ASSERT("Opening ", path, " should have succeeded.");
+      }
+    };
+
+    auto expectFail = [&](kj::StringPtr path) {
+      auto maybeExn = kj::runCatchingExceptions([&]() {
+        raiiOpenAtIfExistsContained(dir.get(), path, O_RDONLY);
+      });
+      KJ_IF_MAYBE(exn, maybeExn) {
+      } else {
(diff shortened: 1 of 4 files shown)


ocdtrekkie:
If we don't think any current apps need laxer restrictions, I would be inclined to say we should be more restrictive until someone justifies otherwise.

src/sandstorm/util.c++
```diff
@@ -95,6 +96,47 @@ kj::Maybe<kj::AutoCloseFd> raiiOpenAtIfExists(
   }
 }
 
+// scratch buffer for file paths; used by raiiOpenAtIfExistsContained:
+thread_local char path_buf[PATH_MAX+1];
```
kentonv, line 100: I'd allocate this on the stack instead of `thread_local`. `thread_local`'s implementation is rather complex, especially for large values like this.

src/sandstorm/util.c++
```diff
@@ -95,6 +96,47 @@ kj::Maybe<kj::AutoCloseFd> raiiOpenAtIfExists(
...
+      case ENOENT:
+        return nullptr;
+      case ELOOP:
+        {
+          if(symlink_limit == 0) {
+            KJ_FAIL_SYSCALL("openat()", error);
+          }
+          symlink_limit--;
+
+          memset(&path_buf, 0, sizeof path_buf);
```
kentonv, line 123: IMO it'd be cleaner to use `readlinkat()`'s return value to determine where to put the NUL terminator, rather than zero-initialize the buffer upfront and rely on that. That way does less work and also gives you a chance to check if the link data turned out to be longer than PATH_MAX (which is very possible, there's no actual limit on path lengths, PATH_MAX exists for historical reasons).

You could also use the KJ filesystem API's readlink() which automatically grows the buffer as needed to fit the full path.

src/sandstorm/util.c++
```diff
@@ -95,6 +96,47 @@ kj::Maybe<kj::AutoCloseFd> raiiOpenAtIfExists(
...
+          KJ_SYSCALL(readlinkat(parentFd, part, path_buf, PATH_MAX));
+          kj::Path nextPath = path.clone().parent().eval(path_buf);
```
kentonv, line 125: Hmm should this actually be `path.slice(0, i).eval(path_buf)`?

src/sandstorm/util-test.c++
```diff
@@ -239,5 +239,71 @@ KJ_TEST("SubprocessSet") {
   promiseCat.wait(io.waitScope);
 }
 
+KJ_TEST("raiiOpenAtIfExistsContained") {
```
kentonv, line 242: Can you add tests covering the case of following "foo/some-symlink/..", verifying that the `..` goes to the symlink target's parent, not back to `foo`?

I _think_ your approach does this "correctly" (where "correctly" means "matching the traditional kernel behavior even if it's confusing"), since you evaluate each symlink strictly after canonicalizing the preceding path so that it contains no symlinks. But this is a common point of confusion with symlink resolution so worth a test.

src/sandstorm/util.c++
zenhack: Ah, I'd remembered `PATH_MAX` as being the maximum size the kernel will accept as an argument to a syscall (obviously with the way mount() works there's no way to have a limit on full paths), but looking it up again it looks like that's only for syscalls that take a nul-terminated string and no buffer length, so it doesn't apply here. good catch.

src/sandstorm/util.c++
```diff
@@ -95,6 +96,44 @@ kj::Maybe<kj::AutoCloseFd> raiiOpenAtIfExists(
...
+        return nullptr;
+      case ELOOP:
+        {
+          if(symlink_limit == 0) {
+            KJ_FAIL_SYSCALL("openat()", error);
+          }
+          symlink_limit--;
+
+          memset(&path_buf, 0, sizeof path_buf);
+          KJ_SYSCALL(readlinkat(parentFd, part, path_buf, PATH_MAX));
```
zenhack, line 121: Just noticed when re-reading this: we never actually re-assign parentFd after its initial assignment to dirfd, so this is clearly broken. I'll find some time soonish to beef up the test suite and fix things.

src/sandstorm/util-test.c++
zenhack: As it turns out, this *doesn't* work, because we pass the path to `kj::Path::parse` to split it up into segments -- which does the resolution in path space before we start walking the segments ourselves.

zenhack: ...I suppose we could split the path ourselves, but it feels a bit sad to introduce extra complexity here when the supervisor already rejects paths with `..` in them anyway. Maybe we should just change the API so it accepts a PathPtr instead of a string, and move the call to `Path::parse` out into the supervisor code.

src/sandstorm/util.c++
zenhack: Added a bunch more tests, fixed this and one or two other issues.

src/sandstorm/util-test.c++
zenhack: Ok, I went ahead and changed the function to accept a `kj::Path`, which avoids this case (which we don't actually need).

src/sandstorm/util.c++
zenhack: Done.

zenhack:
@kentonv, I believe I've addressed all of your comments.

PR #3449, opened 2020-10-10
Add deSEC DNS provider

rs22:
deSEC (https://desec.io/) is a non-profit organization offering a free DNS service that can be configured using a REST API (which, by the way, is also [open source](https://github.com/desec-io/desec-stack)).

In order to add support for them as a DNS provider in Sandstorm, I've created an [acme-dns-01-desec](https://www.npmjs.com/package/acme-dns-01-desec) npm package, to which this PR adds a reference.

An example options json looks like this:
```json
{
  "token": "<auth token>"
}
```
The deSEC docs describe how to obtain the token using their API: https://desec.readthedocs.io/en/latest/auth/tokens.html
There is no UI for that yet.

During development, I've added the npm package directly into this repository, so feel free to stash the commits.


ocdtrekkie (changes requested):
DNS providers should be listed alphabetically. Also, I think ACMS.js plugins that aren't first party to ACME.js or included in the Sandstorm repo directly are credited in third party works used, so you probably should add this plugin there too.

ocdtrekkie:
Awesome to see this (and cool to learn about this service). I left a couple notes, but I imagine this should be really easy to get in the next release.

rs22:
Thanks for your feedback! I hope everything looks good now.

ocdtrekkie (approved):

ocdtrekkie:
It looks good to me!

kentonv:
This looks great. Thanks for moving it out into its own npm package -- that seems like the right thing to do. (The only reason sandcats is in-tree is because sandcats isn't designed to be used by anything other than Sandstorm...)

Also thanks for keeping the code short and dependency-free. :)

Issue #2571, opened 2016-09-20
Any chance at some other social authentication?


bluenevus:
There is google and github which is nice.  Any chance at some of the other popular ones?

Facebook
Twitter
Microsoft
Yahoo
Linkedin
OpenID

![image](https://cloud.githubusercontent.com/assets/3675043/18683542/ac17303e-7f3e-11e6-8f80-15f5c650765c.png)

bluenevus:
BTW...absolutely love what you all have done here.  We previously had many of these hosted on servers and had to update each with a custom webui to launch each from one panel.  Now its all in one place.  Brilliant.

kentonv:
We do want to add some of these at some point. One problem is that we need to make sure that if someone logs in with facebook one day and twitter the next, we somehow detect that they are (or, might be) the same person rather than dropping them into an empty account and leaving them confused. This is already a problem with google/github/email options, but becomes worse as we add more. Probably we should match on e-mail address and suggest: "Hey, it looks like you previously logged in with Facebook. Please do that again now and we'll link your accounts."

fungi:
Though if a deployer wants to lock an instance down to a single authentication source, having the ability for them to pin it to, say, a particular OpenID provider could prove convenient (especially if it's a provider they're already relying on for authenticating other unrelated services, or perhaps even one they operate themselves). In a single-source auth scenario, the need to link accounts is mostly irrelevant (I won't say nonexistent, sometimes someone might inadvertently have more than one ID even within the same provider).

PR #3452, opened 2020-10-15
OpenID Connect Login Provider

rs22:
This adds OpenID Connect as another login provider. The authentication is handled by the `switch:accounts-oidc` meteor package.

I suppose there are some things to discuss:

 - The referenced package follows the meteor conventions nicely (i.e. with regards to naming and URLs). However, it requires to manually configure the identity provider API endpoints (`authorizationEndpoint`, `tokenEndpoint`, `userinfoEndpoint`) and does not support autodiscovery, which I think is one of the selling points of OIDC. Also, there is a bug where no `?` is inserted into the redirect URL to separate the path from the query string, which [hasn't been fixed](https://github.com/switch-ch/meteor-accounts-oidc/pull/20) since May.
- I've mostly copied large code sections from the GitHub login provider in a number of (already large) source files. I guess refactoring this code would make sense -- but maybe that's out of the scope of this PR.
- I wasn't sure how to deal with missing translations for the new strings, so I've just run the `sync-langs.py` script for all other languages.

Issue reference: #2571


ocdtrekkie:
This is a pretty exciting thing to see! It probably needs to be reviewed and tested carefully though, login providers are referenced in a lot of places in the code, and nobody's added one in years. I am not too worried about the i18n, we usually leave new features in English until people come in and fix their respective translations. I appreciate you going the distance and using the i18n framework up front though!

zenhack:
Thanks for putting this together!

Does the bug you mention affect Sandstorm's use here? Can we work around it? This worries me a little bit, as it seems like the package might not be very well maintained. It looks like that pr you linked fixes the problem, so I suppose we could fork and ship a vendored version with the patch if we needed to.

shell/imports/client/admin/login-providers.js
```diff
@@ -58,6 +61,16 @@ const idpData = function (configureCallback) {
         configureCallback("ldap");
       },
     },
+    {
+      id: "oidc",
+      label: "OpenID Connect",
+      icon: "/ldap.svg", // Or use identicons
```
ocdtrekkie, line 67: The ldap.svg graphic seems plausible for LDAP and SAML because they almost always refer to SSO for internal organizations. OpenID Connect can presumably be any OpenID Connect provider (including both internal SSO and external service providers), so I feel like we should try to get an openid.svg here, especially since OpenID has a logo.

rs22:
Thanks for your remarks! I've added the OpenID logo in two variants (one is flat, the other one is colored -- according to the existing UI conventions).

The bug in `switch:accounts-oidc` does affect our use here: The authorization endpoint needs to be manually suffixed with a `?`, otherwise you'll get a 404 after the redirect. This could be easily worked around in the Sandstorm code, but I'd prefer the library to be fixed instead.

I have created an auth0 account so you can try out the login provider on local.sandstorm.io:6080 using the following configuration:

- Identity Server URL: `https://sandstorm-test.us.auth0.com`
- Authorization Endpoint: `/authorize?`
- Token Endpoint: `/oauth/token`
- User Info Endpoint: `/userinfo`
- Client ID: `o9ub08mHb5mpGe8ODdcty6HeqiOpob4h`
- Client Secret: `oPEN23dWlY0aLunHJCQqu88IrM_VEgLMKZJnjxmPFN0rGoaDYw1kuU29Amfe_U2y`

You can then log in using the credentials `r.schmid@outlook.com` / `Sandstorm!`. The auth0 account should work for the next 20 days...

shell/imports/server/migrations.js
```diff
@@ -550,6 +554,9 @@ const consolidateOrgSettings = function (db, _backend) {
     ldap: {
       enabled: orgLdap ? orgLdap.value : false,
     },
+    ldap: {
```
ocdtrekkie, line 557: Is this supposed to be for OIDC here? It looks like it's just the above three lines duplicated.

ocdtrekkie:
It looks like the linter has jumped up a bit too: ✖ 809 problems (1 error, 808 warnings) with your latest push. I think we only allow bringing that number down, so we need to see if we can resolve some of these.

ocdtrekkie:
Looking pretty good, IMHO. @zenhack should probably review this yet. I am not sure what is wrong with the build process too, I am not sure if that's something with the PR or with the Action.

rs22:
Thanks for looking at this already! I'll just summarize what I've changed over the last few days:

 - First, I inlined the `switch:accounts-oidc` package to fix the previously mentioned bug. Most of that code was copied from the meteor implementations of various login providers (e.g. [github-oauth](https://github.com/meteor/meteor/tree/devel/packages/github-oauth), [accounts-github](https://github.com/meteor/meteor/tree/devel/packages/accounts-github)), so although the OpenID endpoints could be dynamically configured, this wasn't a full implementation of an OpenID Connect client.
- Thus, I went for including the `openid-client` npm package instead, which provides a full (server side) implementation of the OpenID standard and even is [certified](https://openid.net/developers/certified/). Now, in most cases, it is sufficient to provide the server url and let the autodiscovery take care of the remaining configuration.
- I tested it with the Auth0 IDP and another one from my organization. That second one required me to manually override the 'client authentication method', which I turned into a configurable parameter as a result. ([Others](https://github.com/auth0/express-openid-connect/pull/93) have this problem, too.)
- I also implemented Organization support

Adding the npm package pulled in quite a bunch of additional dependencies -- and there's one thing that I don't like in particular: The `clean-stack` package (required through `openid-client` -> `p-any` -> `p-some` -> `aggregate-error` -> `clean-stack`) makes a call to node's `os.homedir()` method when loading [[source](https://github.com/sindresorhus/clean-stack/blob/v3.0.0/index.js#L7)]. 

That call failed in the context of the shell, which I have now worked around by setting the `HOME` environment variable -- as also described in the [docs](https://docs.sandstorm.io/en/latest/developing/raw-python/) for packaging python apps. Please let me know if there's a better workaround.

zenhack (changes requested):
This looks broadly reasonable to me. I tested the basic login and it seems to work, though I didn't test stuff like the organization support. I've made a couple comments in-line. Additionally:

- Do we need to use a meteor package to define the auth backend, or can this be a misc. js module? The latter is generally preferable as it's less machinery and doesn't restrict the dependency graph (we did a significant refactoring a while back to go from meteor packages -> modules, and it's let us get rid of some duplicated code...).
- I would kindof like to move in the direction of fewer global-scope side effects, to make it easier to isolate bits of the code for testing purposes -- the prevalence of this kind of thing is one of the bigger categories of tech debt we have. I guess I don't have any specific action items here, but maybe something to think about when we end up refactoring this. There is indeed what feels like a lot of copypasta and we should do something bout that at some point.
- It's kindof sad that this is both (1) a singleton provider -- you can only point a sandstorm instance to one IDP iiuc, and (2) doesn't dynamic discovery of providers, which seems like the exciting bit of OpenID. But those are things that I think can be follow-on items.

shell/packages/accounts-oidc/package.js
```diff
@@ -0,0 +1,21 @@
+const { Package } = require("meteor/tools");
```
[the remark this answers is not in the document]
rs22: The linter complained about `Package` not being defined, so I tried to fix it with this line. Now the package.js is gone entirely.

rs22:
I've now moved the code into the `shell/client` and `shell/server` subdirectories -- so the separate package is now gone.

At least the client-side `loginWithOidc` method is not attached to the global `Meteor` object anymore. Still, most of the code is copy-pasted and I agree with your concerns.

As far as I know, there are not many IDPs yet that support the dynamic discovery protocol (i.e. users just provide an email address, the corresponding IDP is automatically detected, and the relying party registers itself). It would be cool, if Sandstorm could become such an IDP for federated logins... But that's a topic for the next PR, I guess.

zenhack:
Since the two modules are tightly coupled, I think it would make a bit more sense to keep the them together, maybe under `shell/imports/oidc`; you'll note we have a few other directories that are split off from the `client/` and `server/` hierarchies.

Other than that, LGTM; I think everything else can be subsequent prs.

ocdtrekkie:
I'll go ahead and mark this ready, but feel free to make the change Ian recommends there.

zenhack (approved):

kentonv:
Awesome!

It looks like a minor update to the dutch translation that I just merged conflicts with this, oops. :(

Speaking of translations, I noticed your translation files add a lot of lines where the English text is prefixed with `*_` like:

```
        "title": "*_OpenID Connect login configuration",
```

What does this mean? AFAICT this `*_` prefix doesn't appear anywhere in our current translation files. Is this some sort of special directive that the i18n framework will handle in some way, or is it going to appear with that prefix in the UI?

My understanding is that if a particular string is missing from a specific translation, we'll fall back to the English version, so maybe it's not necessary to copy the strings over?

I realize that this was probably all done by `sync-langs.py`, but I don't really have any idea how that script is meant to be used. Hoping that someone can try enabling some other languages and see exactly what is happening here, to verify we aren't introducing any ugliness for non-english users (aside from the ugliness of having to see some English until someone actually translates it).

ocdtrekkie:
@kentonv My understanding is that sync-langs.py copies in the English into other language files, so that you can update them. *_ presumably being something you can search for to find fields you need to update. I think it'd show in the UI.

I agree it's probably best to just omit these lines if not translating them, but with any luck our various translators will hopefully step in when they see new features that need i18n. (It'd still be good to establish a way to notify all of them when we add new strings...)

ocdtrekkie:
It looks like the only conflict is the "acceptable number of problems" value in the Action.

ocdtrekkie:
Ian changed 796 -> 795, and Richard changed 796 -> 791, so I'm betting the happy merge conflict is 790. But I'm re-running the checks to be sure. =)

kentonv:
Yeah I don't think we should be throwing a bunch of weird `*_` into people's UIs, so let's just revert all the non-english files...

ocdtrekkie:
791 appears to still be the magic number for resolving the conflict on the linter. And yeah, I agree wrt lang files.

ocdtrekkie (approved):

kentonv:
Thanks!

PR #3455, opened 2020-10-22
Fix getGrainTitle example

### LINKED ISSUE #3430: "Using the grain's title in your app" example from documentation does not work
The docs contain an example ["Using the grain's title in your app"](https://docs.sandstorm.io/en/latest/developing/path/#using-the-grains-title-in-your-app):
```javascript
var getGrainTitleRpcId = 0;

// Sandstorm will reply via postMessage, so we need to set up a handler:
window.addEventListener('message', function(event) {
  if(event.source !== window.parent) {
    // SECURITY: ignore messages not from the parent.
    return;
  }
  if(event.data.rpcId === getGrainTtileRpcId) {
    console.log("The grain's title is: ", event.data.grainTitle);
  }
})

// Now make the request:
window.postMessage({
  getGrainTitle: {},
  rpcId: getGrainTitleRpcId,
  // If subscribe is true, sandstorm will push future updates to the
  // grain's title. If it is false or absent, the app will not be
  // notified of updates.
  subscribe: true,
}, '*')
```

I get one `message` event which contains no grain title and no follow-up events when the grain title is changed.  Is there a way to run a debugger inside Meteor inside Sandstorm?  This sounds complicated.

troyjfarrell:
Fixes #3430

### DIFF
--- a/docs/developing/path.md
+++ b/docs/developing/path.md
@@ -94,13 +94,13 @@ window.addEventListener('message', function(event) {
     // SECURITY: ignore messages not from the parent.
     return;
   }
-  if(event.data.rpcId === getGrainTtileRpcId) {
+  if(event.data.rpcId === getGrainTitleRpcId) {
     console.log("The grain's title is: ", event.data.grainTitle);
   }
 })
 
 // Now make the request:
-window.postMessage({
+window.parent.postMessage({
   getGrainTitle: {},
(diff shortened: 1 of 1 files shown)


ocdtrekkie (approved):
Seems quite sensible! I see other documented postMessages for Sandstorm features are documented as directed at window.parent as well, and it makes sense.

PR #3456, opened 2020-10-22
Correctly handle responses to HEAD requests that use Content-Length

### LINKED ISSUE #3454: Server responds to HEAD requests with HTTP 500
I'm trying to package a Docker [registry](https://github.com/docker/distribution) server as a Sandstorm app. When pushing images, the client makes HEAD requests with the blob ids of individual layers to check if they already exist on the server. The server initially responds with a 404, however Sandstorm turns this response into a 500 internal server error.

This applies to all 404 responses to HEAD requests where the Content-Length header is set on the response. It can be reproduced using the following node.js express HTTP server:

```js
const express = require('express')
const app = express()
const port = 8080

app.get('/empty_response', (req, res) => {
  res.status(404).end()
})

app.listen(port, () => {
  console.log(`Example app listening at http://localhost:${port}`)
})
```

The issue can be observed like this:
```
$ curl --head http://api-<id>.local.sandstorm.io:6080/path_does_not_exist -H 'authorization: Basic <auth>'
HTTP/1.1 500 Internal Server Error
Connection: close
Content-Length: 278
Content-Type: text/plain
```
The following entry appears in the Sandstorm logs:
```
capnp/rpc.c++:146: info: returning failure over rpc; exception = kj/async-inl.h:1230: failed: PromiseFulfiller was destroyed without fulfilling the promise.
stack: 4761a0 677e60
```

This is what the expected response looks like:
```
$ curl --head http://localhost:8080/path_does_not_exist
HTTP/1.1 404 Not Found
X-Powered-By: Express
Content-Security-Policy: default-src 'none'
X-Content-Type-Options: nosniff
Content-Type: text/html; charset=utf-8
Content-Length: 159
Date: Wed, 21 Oct 2020 09:55:48 GMT
Connection: keep-alive
```

The following works fine (no `Content-Length` set in the response):
```
$ curl --head http://api-<id>.local.sandstorm.io:6080/empty_response -H 'authorization: Basic <auth>'
HTTP/1.1 404 Not Found
Content-Type: text/html; charset=UTF-8
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: ETag
Content-Security-Policy: default-src 'none'; sandbox
Vary: Authorization
```

I'm not sure where this can be fixed. Is this a problem of the sandstorm-http-bridge or the builtin HTTP server?

zenhack:
Previously, if the bridge saw a Content-Length header, but the length of
the body did not match, it would throw an error. However, for HEAD
requests this is incorrect, so with this patch the check only applies to
non-HEAD requests; for HEAD requests we will expect an empty body.

This also changes WebSessionBridge to ignore the body of HEAD requests

Fixes #3454

---

@rs22,  want to test?

### DIFF
--- a/src/sandstorm/web-session-bridge.c++
+++ b/src/sandstorm/web-session-bridge.c++
@@ -140,9 +140,10 @@ kj::Promise<void> WebSessionBridge::request(
     case kj::HttpMethod::HEAD: {
       auto req = session.getRequest();
       req.setPath(path);
-      req.setIgnoreBody(method == kj::HttpMethod::HEAD);
+      bool ignoreBody = method == kj::HttpMethod::HEAD;
+      req.setIgnoreBody(ignoreBody);
       auto streamer = initContext(req.initContext(), headers);
-      return handleResponse(req.send(), kj::mv(streamer), response);
+      return handleResponse(req.send(), kj::mv(streamer), response, ignoreBody);
     }
 
     case kj::HttpMethod::POST: {
@@ -814,7 +815,10 @@ class WebSessionBridge::ByteStreamImpl final: public ByteStream::Server {
 public:
   ByteStreamImpl(uint statusCode, kj::StringPtr statusText,
                  kj::HttpHeaders&& headers,
-                 kj::HttpService::Response& response) {
+                 kj::HttpService::Response& response,
+                 bool ignoreData = false)
+  : ignoreData(ignoreData)
+  {
     state.init<NotStarted>(NotStarted { statusCode, statusText, kj::mv(headers), response });
   }
 
@@ -842,10 +846,14 @@ public:
   }
 
   kj::Promise<void> write(WriteContext context) override {
-    auto fork = queue.then([this,context]() mutable {
+    auto fork = queue.then([this,context]() mutable -> kj::Promise<void> {
       auto& stream = ensureStarted(nullptr);
-      auto data = context.getParams().getData();
-      return stream.write(data.begin(), data.size());
+      if(ignoreData) {
+        return kj::READY_NOW;
+      } else {
+        auto data = context.getParams().getData();
+        return stream.write(data.begin(), data.size());
+      }
     }).fork();
     queue = fork.addBranch();
     return fork.addBranch();
@@ -902,6 +910,7 @@ private:
   kj::Maybe<kj::Own<kj::PromiseFulfiller<void>>> doneFulfiller;
   kj::Promise<void> queue = kj::READY_NOW;
   kj::Maybe<Aborter&> aborter;
+  bool ignoreData = false;
 
   kj::AsyncOutputStream& ensureStarted(kj::Maybe<uint64_t> size) {
     if (state.is<NotStarted>()) {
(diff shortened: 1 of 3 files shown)


rs22:
Thanks a lot! This solves my problem :rocket: 

I noticed that the Content-Length header in HEAD responses still is not passed on to the requesting client by Sandstorm. This is no problem in my particular case, but other applications might depend on this?

The RFC says that it's optional:

> A server MAY send a Content-Length header field in a response to a HEAD request

from https://tools.ietf.org/html/rfc7230#section-3.3.2

> The server SHOULD send the same header fields in response to a HEAD request as it would have sent if the request had been a GET, except that the payload header fields MAY be omitted.

from https://tools.ietf.org/html/rfc7231#section-4.3.2

zenhack:
Yeah, that's a bit fiddly to do since right now the WebSession API infers the Content-Length from the actual size of the body -- I think it would *work* to use the streaming api and just call expectSize() without actually writing the data afterwards, but that's a bit gross because the docs for ByteStream suggest that that should be considered an error... probably a "cleaner" solution would be to tweak the interfaces somehow, but I'm inclined to say that should probably be a follow-on issue then.

src/sandstorm/web-session-bridge.c++
```diff
@@ -814,7 +815,10 @@ class WebSessionBridge::ByteStreamImpl final: public ByteStream::Server {
 public:
   ByteStreamImpl(uint statusCode, kj::StringPtr statusText,
                  kj::HttpHeaders&& headers,
-                 kj::HttpService::Response& response) {
+                 kj::HttpService::Response& response,
+                 bool ignoreData = false)
+  : ignoreData(ignoreData)
+  {
```
kentonv, line 821:
```suggestion
      : ignoreData(ignoreData) {
```

src/sandstorm/web-session-bridge.c++
```diff
@@ -842,10 +846,14 @@ public:
   }
 
   kj::Promise<void> write(WriteContext context) override {
-    auto fork = queue.then([this,context]() mutable {
+    auto fork = queue.then([this,context]() mutable -> kj::Promise<void> {
       auto& stream = ensureStarted(nullptr);
-      auto data = context.getParams().getData();
-      return stream.write(data.begin(), data.size());
+      if(ignoreData) {
```
kentonv, line 851:
```suggestion
      if (ignoreData) {
```

src/sandstorm/web-session-bridge.c++
```diff
@@ -1352,8 +1362,12 @@ kj::Promise<void> WebSessionBridge::handleResponse(
           case WebSession::Response::Content::Body::BYTES: {
             auto data = body.getBytes();
             auto stream = out.send(status.getId(), status.getTitle(), headers, data.size());
-            auto promise = stream->write(data.begin(), data.size());
-            return promise.attach(kj::mv(stream), kj::mv(in));
+            if(ignoreSuccessBody) {
```
kentonv, line 1365:
```suggestion
            if (ignoreSuccessBody) {
```

kentonv (commented):
I _think_ the changes to `web-session-bridge.c++` are redundant as KJ's HTTP implementation will already ignore the body for a HEAD request. If so can we revert that part?

zenhack:
> I think the changes to web-session-bridge.c++ are redundant as KJ's HTTP implementation will already ignore the body for a HEAD request. If so can we revert that part?

Looks like you're right, done.

PR #3458, opened 2020-10-25
cgroups: don't fail to back up if cgroup.freeze doesn't exist.

### LINKED ISSUE #3457: Error 500 on grain backup
Tested both on my server and Alpha, here's from my log:

```
capnp/rpc.c++:146: info: returning failure over rpc; exception = sandstorm/util.c++:66: failed: openat(dirfd, name.cStr(), flags, mode): No such file or directory; name = cgroup.freeze
stack: 51f46d 6db3ac 49d276 49d33a 5661a5 565d7d 5280d1 70c170
Exception while invoking method 'backupGrain' Error: remote exception: openat(dirfd, name.cStr(), flags, mode): No such file or directory; name = cgroup.freeze
```

@orblivion noted it first. cc @zenhack

zenhack:
Apparently this functionality is configurable independently from
the rest of cgroups2, so we can't rely on it necessarily being
there.

Fixes #3457

---

@ocdtrekkie, @orblivion, can you guys test this on your own systems to make sure it fixes the problem? Once CI completes there should be a build you can download, or you can build from source.

### DIFF
--- a/src/sandstorm/cgroup2.c++
+++ b/src/sandstorm/cgroup2.c++
@@ -38,10 +38,13 @@ void Cgroup::addPid(pid_t pid) {
   KJ_SYSCALL(write(procsfd.get(), cStr, strlen(cStr)));
 }
 
-Cgroup::FreezeHandle Cgroup::freeze() {
-  auto freezeFd = raiiOpenAt(dirfd.get(), "cgroup.freeze", O_WRONLY);
-  KJ_SYSCALL(write(freezeFd.get(), "1\n", 2));
-  return Cgroup::FreezeHandle(kj::mv(freezeFd));
+kj::Maybe<Cgroup::FreezeHandle> Cgroup::freeze() {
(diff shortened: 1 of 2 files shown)


ocdtrekkie:
I am not super confident in tampering with my production server, but I'll take a backup in case it is a hold up to getting this released.

ocdtrekkie:
Based on this: https://kernelnewbies.org/Linux_5.2#Add_freezer_controller_for_cgroups_v2 which was released in July 2019, it seems assuming cgroups2 has freeze is wrong because freeze was added after the release of cgroups2.

Ubuntu 18.04.3, for example, is based on the 5.0 kernel, which would have cgroups2 but not `cgroup.freeze`.

ocdtrekkie (approved):
I've tested this PR. It works.

orblivion:
Thanks. Backup works for me again.

PR #3461, opened 2020-11-04
Install docs: add strace to the dependency list.

zenhack:
@troyjfarrell pointed out on the mailing list that this is used by
`make-bundle.sh`, but not mentioned in the docs.

`gcc` is also missing as a dependency, but probably we should solve that
one by changing the script to use the same C compiler as the rest of the
build.


ocdtrekkie (approved):

ocdtrekkie:
This is a documentation fix, so I will merge it! :D

PR #3465, opened 2020-11-11
Finish removing legacy HTTP access via HackSession

zenhack:
This has been disabled for a few months and nobody's complained, so I think we can safely remove this.

This patch removes:

- The implementations of the obsolete methods on `HackSession` itself (they now just throw exceptions)
- Support code used to check whether to enable the functionality
  - Also includes a database migration to remove the setting.
- Some code paths that were only used via `HackSession`, and are not relevant when the powerbox is used to obtain HTTP access.


ocdtrekkie:
<3 code deletion

kentonv:
Awesome.

PR #3473, opened 2020-11-23
Fix some regressions re: the recent static publishng symlink fix.

### LINKED ISSUE #3472: Web publishing appears to be broken
EDIT: This appears to be the case on both Alpha and my own server. My guess is Ian's security fix for static publishing is it.

Static publishing appears broken from the two things I've tested (Davros grains and docs.sandstorm.io):

A Davros grain:
```
ERROR: The server threw an exception. Details:

(remote):0: failed: remote exception: remote exception: open: Too many levels of symbolic links; name = www
stack: 70d6f0 49a390 48bc30 52ad90 605f50 609c60
```

docs.sandstorm.io:
```
ERROR: The server threw an exception. Details:

(remote):0: failed: remote exception: remote exception: expected part.findFirst('/') == nullptr; '/' character in path component; did you mean to use Path::parse()?; part = en/latest/index.html
stack: 70d6f0 49a390 48bc30 52ad90 605f50 609c60
```

zenhack:
There are two problems:

1. Davros apparently makes /var/www itself a symlink to another
   directory
2. We need to call Path::parse() on the path. Apparently I didn't test
   this thoroughly enough after changing raiiOpenAtIfExistsContained
   to take a kj::Path rather than a string.

We fix (1) by allowing /var/www itself to be a symlink, and then only
enforcing that symlinks within that directory do not point outside of
it.

Fixes #3472

---

@ocdtrekkie sent me a backup of his davros grain, and I've confirmed that this fixes that case. I have not tested the issue affecting docs.sandstorm.io, but from the error message I'm fairly confident that this is the issue.

I will probably send a follow pr making sure we have test coverage for all this, but this should at least fix the issue.

### DIFF
--- a/src/sandstorm/supervisor.c++
+++ b/src/sandstorm/supervisor.c++
@@ -2209,31 +2209,32 @@ public:
       }
     }
 
-    auto fullPath = kj::str("sandbox/www/", path);
-    KJ_IF_MAYBE(fd, raiiOpenIfExists(fullPath, O_RDONLY)) {
-      struct stat stats;
-      KJ_SYSCALL(fstat(*fd, &stats));
+    auto sandboxDir = raiiOpen("sandbox", O_RDONLY);
+    KJ_IF_MAYBE(wwwDir, raiiOpenAtIfExists(sandboxDir.get(), "www", O_RDONLY|O_NOFOLLOW)) {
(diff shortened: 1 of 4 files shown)


zenhack:
Since @kentonv just reverted the original fix to get something out quickly, this created merge conflicts with this pr, so I've reverted the reversion and forced-pushed.

PR #3481, opened 2020-12-22
Makefile: add missing dependency meteor-testapp-dev -> meteor-spk

zenhack:


ocdtrekkie:
Does this help alleviate test failures?

zenhack:
No, this target isn't used by the test suite. It runs the test app in dev mode, which is useful for experimentation. I noticed this when trying to use it that way, since it should have set up meteor-spk and didn't.

PR #3492, opened 2021-01-19
Fix tests/grain.js tests using Chromium 80 and Debian 10

troyjfarrell:
Use the old wire protocol with ChromeDriver (set w3c to false).  This fixes some Chrome (mis)behaviors, including failing to switch browser windows.  Tested with Chromium 80 on Debian 10.

### DIFF
--- a/tests/nightwatch.json
+++ b/tests/nightwatch.json
@@ -28,7 +28,10 @@
       "desiredCapabilities": {
         "browserName": "chrome",
         "javascriptEnabled" : true,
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
Looks like only three tests are failing on this run... three we usually expect to fail. But we used to have more failures... does this fix some of our known failing tests?

zenhack:
I'd noticed the number had gone up on my local machine, this fixes a number of tests that I'd been meaning to investigate more closely. Thanks!

PR #3493, opened 2021-01-20
Fix some linter warnings

zenhack:
Explicitly import/export updateStripeData, and deal with some unused
variable/parameter warnings.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,7 +72,7 @@ jobs:
           # the constant below.
           make lint | tee lint.log
           num_problems=$(cat lint.log | grep ' problems (0 errors, ' | awk '{print $2}')
-          [ "$num_problems" = 790 ]
+          [ "$num_problems" = 776 ]
       - name: test
         run: |
           set -o pipefail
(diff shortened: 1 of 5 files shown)


ocdtrekkie:
I imagine this falls into "unused code" territory at this point, but smaller warning numbers is better.

zenhack:
Yeah, I was more or less just starting at the top, and `b` is pretty early in the alphabet...

zenhack:
Looks like selenium failed to start in CI... that seems like probably an unrelated transient issue; is there some way to retry the run?

zenhack:
Actually, scratch that, looks like this might be a persisent (but unrelated) issue, and I can reproduce it locally on master.

```
  value: {
    message: 'session not created: This version of ChromeDriver only supports Chrome version 88\n' +
      'Current browser version is 87.0.4280.88 with binary path /usr/bin/google-chrome\n' +
      "Build info: version: '3.141.59', revision: 'e82be7d358', time: '2018-11-14T08:25:53'\n" +
      "System info: host: 'fv-az67-957', ip: '10.1.0.4', os.name: 'Linux', os.arch: 'amd64', os.version: '5.4.0-1032-azure', java.version: '1.8.0_275'\n" +
      'Driver info: driver.version: unknown\n' +
      'remote stacktrace: #0 0x55706ae44199 <unknown>\n',
    error: 'session not created'
```

...Though there's an update form Chromium in the archlinux repos already, which I am downloading. So maybe just wait for ubuntu to update to the latest chrome, since it just came out? Not sure how long that will be.

PR #3497, opened 2021-01-29
Be less permissive about when to fail CI.

zenhack:
See the comment. I've been forgetting to look at CI a lot, so we really
should make this yell at us if unexpected stuff fails.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,9 +76,18 @@ jobs:
       - name: test
         run: |
           set -o pipefail
-          # Currently these tests never pass.
-          # TODO: When they are fixed, remove the || true
+          # Currently we have some failing tests, and while most
+          # test frameworks have an "expected to fail" feature, I(zenhack) can't
+          # find this for nightwatch... so for now we use the same trick as above
+          # to catch unexpected increases.
+          #
+          # TODO: When they are fixed, remove the extra logic and just fail if the
+          # whole test doesn't pass.
           (make test |& tee testlog.txt; echo "Result: $?") || true
+          num_failures=$(cat testlog.txt \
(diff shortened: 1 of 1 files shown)


zenhack:
(I would not be surprised if this fails the first CI run, since I obviously couldn't test it locally... will debug as we go).

ocdtrekkie:
Really glad to see this. Being pickier about linter warnings than test failures was a bit silly. ;) And getting this down to zero was one of our big milestones we set, so we probably shouldn't be letting this slip.

zenhack:
Looks like the logic for this is okay now (after a few iterations), but some of the failures are intermittent, so the last run failed because one of the usual failures didn't happen...

I suppose I could just specify an acceptable ranage (2-4?), but that makes me uneasy because it'd be easy to miss an unexpected failure if it coincides with one of the flaky tests happening to pass. I really wish I had a way to mark specific individual tests as expected to fail.

zenhack:
Also, I'm seeing five failures on my local machine:

- `tests/account`
- `tests/appHooks`
- `tests/trash`
- `apps/powerbox`
- `apps/web-publishing`

We have issues open for 4 of those, but not `tests/trash` -- but I have a hazy memory of being told that at one point or another that test was flaky -- @kentonv, does that ring a bell?

zenhack:
Actually, I am mistaken -- we don't have an issue for `apps/powerbox` either, but we do have one for `apps/schedule`, which didn't fail locally -- another flaky one.

ocdtrekkie:
Do we know why these tests are flaky? Can we maybe shore up the intermittent ones a bit before merging this change?

zenhack:
I don't think we know why they're flaky, no -- figuring that out is going to take some debugging.

zenhack:
I'm just going to close this; I plan on actually pushing through and fixing all the tests soon, at which point this will be moot.

PR #3501, opened 2021-02-21
Add post-2017 system calls to the seccomp filter.

zenhack:
Add post-2017 system calls to the seccomp filter.

Per the comment above, it would be better to just block all calls above
a certain syscall number, but in the immediate term let's just add these
explicitly (I have some WIP on a proper blacklist on another branch).


zenhack:
Rebased on top of master, got rid of the ekam update

kentonv:
Hmm, I thought there were a lot more new syscalls, but I'll go ahead and merge this as a start.

Issue #2989, opened 2017-10-20
Switch from libseccomp to BPF assembly


zenhack:
From <https://github.com/sandstorm-io/sandstorm/blob/5ee73e6022e8d3620d48e082a08a70a725795c9d/src/sandstorm/supervisor.c%2B%2B#L1145-L1147>:

```c++
// TODO(security): We should probably disallow everything except SOCK_STREAM and SOCK_DGRAM but
//   I don't totally get how to write such conditionals with libseccomp. We should really dump
//   libseccomp and write in BPF assembly, which is frankly much easier to understand.
```

I didn't see an open issue for this, so thought I'd add one. It's seems like most of the comments in that section refer to things that wouldn't be hard to deal with using raw BPF asm.

kentonv:
Yeah, I've been working with BPF asm for Cloudflare Workers, and it's really quite nice.

zenhack:
Yeah, I've played around with it and it *is* fairly nice.

What tools do you use to actually assemble it? There's an assembler
under tools/ in the linux source tree, but I don't know what's standard.

kentonv:
Someone at Cloudflare extracted the tools so that they're independently buildable (before I joined, actually!): https://github.com/cloudflare/bpftools

I added some additional preprocessing so that I could #include C headers and evaluate constant expressions, but it's a pretty terrible hack, not sure if it's even worth releasing.

zenhack:
A while back I fooled around with trying to do this to get a sense of what it might look like.
What I ended up doing was just piping the assmebly source through cpp, and #including the output of this program from the assembly:

https://gist.github.com/zenhack/9e7c27b00893fff116b168ae85056ff9

...which got run as part of the build. It didn't seem worth the effort to try to automate it any further.

zenhack:
Just a note, I have some work on this on a branch, which also turns the filter into a whitelist (I started just working on the latter, but found it difficult enough to encode certain things that I decided to just go all-in on this). Still needs some debugging & testing though.

PR #3502, opened 2021-02-23
Add new experimental seccomp whitelist, using bpf_asm

zenhack:
This adds a new, experimental seccomp filter, which:

* Is deny-by-default; the current filter requires us to stay on top of new kernel features and add them to the list, which is obviously bad.
* Is written using bpf assembler, rather than libseccomp. This makes writing certain kinds of conditionals easier; see #2989.

The new filter is disabled by default, and can be enabled via a config file option. It is currently too restrictive; while several apps that I've tested appear to work, some others still do not. Of course, even after fixing the issues I'm aware of, there are likely to be other problems I haven't caught.

The migration plan is as follows:

* We will ship this feature disabled by default at first, and ask users to test and report issues.
* Once we have gained some confidence in the new filter, we will enable it by default.
* After some time has passed, we can remove the old filter (and the dependency on libseccomp).

...which is more or less what we did when @kentonv rewrote the gateway in C++.

Note that this bundles the bpf_asm tool from the linux kernel source tree; it unfortunately isn't generally available as its own package. The source for the tool is unmodified from the noted commit, but I've contacted upstream about improving its error checking; see https://lore.kernel.org/bpf/c964892195a6b91d20a67691448567ef528ffa6d.camel@linux.ibm.com/T/#t

I will continue to fiddle with the filter itself, but this is representative of the design and probably could be merged as-is, since I don't expect it will be quite right when it lands anyway.


ocdtrekkie:
I think it's probably safe to mark this ready for review, since it's behind a config flag anyways.

Out of curiosity, is there a good way we could convince the tests to attempt to run against the new filter during this transition time? I know the tests run tests on several apps.

zenhack:
@ocdtrekkie, the config file for the test run is generated by `tests/run-local.sh`, so we could make it happen there. It would probably make sense to do this for the csp option as well.

zenhack:
Ack, the build failed. It looks like some of the constants I'm using post-date the headers in the VM running the build. Perhaps we should just vendor those, since the values are never going to change anyway.

zenhack:
Rather than vendoring the headers, I decided to just remove the offending flag from the allow list, making the filter more restrictive -- presumably if the flag is new enough to not be in the headers of some systems, then it is new enough for apps not to assume it exists.

zenhack:
One other thing to note is: the current filter uses more relaxed rules for ptrace when an app is in dev mode (whereas in prod it's blocked entirely). The new filter doesn't currently do that, in part because it's less clear how to feed a config option into the `bpf_asm` filter, though probably we could come up with some way of doing this. It's possible to kick this down the road a bit though, esp. since for the time being developers can use the old filter if they want strace and such to work in dev.

zenhack:
Mostly as a reminder to myself when I get back to this: the current hangup is with some apps (including rocket.chat and hacker slides) a rejected call to ioctl() is causing them to fail to start.

ocdtrekkie:
I am not sure if @amluto is still interested, but I wanted to draw his attention to this PR as he wrote much of the original seccomp filter.

zenhack:
(Slightly more up to date status: i fixed the ioctl() issue, the next hangup is with Mongo calling setsockopt() with some options I haven't whitelisted. I'm tempted to just say to heck with it and not look at arguments for setsockopt() -- we don't for the current filter -- but I want to more carefully verify that that's really safe first).

kentonv:
Instead of vendoring the bpfasm code, can we use https://github.com/cloudflare/bpftools ? (That's what the Cloudflare Workers runtime uses, FWIW...)

src/sandstorm/backend.h
```diff
@@ -33,10 +33,13 @@ namespace sandstorm {
...
-  BackendImpl(kj::LowLevelAsyncIoProvider& ioProvider, kj::Network& network,
+  BackendImpl(kj::LowLevelAsyncIoProvider& ioProvider,
+              kj::Network& network,
               SandstormCoreFactory::Client&& sandstormCoreFactory,
-	      kj::Maybe<Cgroup>&& cgroup,
-              kj::Maybe<uid_t> sandboxUid);
+              kj::Maybe<Cgroup>&& cgroup,
+              kj::Maybe<uid_t> sandboxUid,
+              bool useExperimentalSeccompFilter,
+	      bool logSeccompViolations);
```
kentonv, line 42: Nitpick: Seems to be a tab here.

src/sandstorm/seccomp-bpf/clean-header.ekam-rule
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+this_dir="sandstorm/seccomp-bpf"
+
+if test $# = 0; then
+  echo "trigger canonical:${this_dir}/gen-clean-h"
```
kentonv, line 8: For a rule that is intended to be executed only once, not on a per-input basis, you can just do the logic when the rule first executes rather than registering a trigger.

src/sandstorm/seccomp-bpf/gen-clean-h.c
```diff
@@ -0,0 +1,152 @@
+// This program prints out a "clean" header for use in bpf assembly, defining
+// constants we need from various system headers. `bpf_asm` will choke on the
+// originals, for two reasons:
+//
+// 1. They contain C code
```
kentonv, line 5: In the Workers Runtime, my solution to this was to have a special marker in the BPF file to mark where the code actually started. Then, I had an extra preprocessing step -- after running the C preprocessor, I use a `while read LINE` loop to in bash to remove everything up to the marker.

src/sandstorm/seccomp-bpf/gen-clean-h.c
```diff
@@ -0,0 +1,152 @@
...
+// 2. The #defines use expressions, which `bpf_asm` doesn't understand.
```
kentonv, line 6: In the Workers Runtime I have a hack that actually finds parenthesized expressions and evaluates them using bash's `$(())`...

```
EXPRESSION_REGEX="^([^(]*)\((.*)\)([^)]*)$"

# ... later on, in the line-reading loop ...
      elif [[ "$LINE" =~ $EXPRESSION_REGEX ]]; then
        PREFIX=${BASH_REMATCH[1]}
        EXPRESSION=${BASH_REMATCH[2]}
        SUFFIX=${BASH_REMATCH[3]}

        # Remove unsigned/long suffixes.
        EXPRESSION=$(echo "$EXPRESSION" | tr -d UuLl)

        VALUE=$(( EXPRESSION ))

        echo "$PREFIX$VALUE$SUFFIX"
```

It mostly works, believe it or not.

kentonv: You don't have to use either of these suggestions though, if you'd rather keep it this way that's up to you. Just wanted to bring up what worked for me.

zenhack:
Re: the bpftools repo, I'd be open to it if that can be kept up to date; it looks like its copy of `bpf_asm` hasn't been updated in 7 years. I actually started with the version of `bpf_asm` in that repo, but updated to linux master when I discovered that the older version doesn't even issue a warning if you overflow the jump target field in conditional branch instructions -- would have saved me an afternoon if I'd started off using the newer version :/. I also pushed a patch into linux upstream to make this a hard error, which has been merged by the appropriate subsystem maintainer, but I don't think has found its way into linus's tree yet. If we're going to be on our own re: updates though, I'd prefer to vendor so we can handle that ourselves.

src/sandstorm/seccomp-bpf/gen-clean-h.c
zenhack: I think I'll stick with the existing approach.

PR #3515, opened 2021-04-19
Docs: Update security practices for April 2021

ocdtrekkie:
I found the Beta Notice very out of date, as we are technically not "in beta", nor does the HTTP hole still exist. The Powerbox exists, so I removed that note, and I updated the CSP confinement note to specify that it's implemented.

### DIFF
--- a/docs/using/security-practices.md
+++ b/docs/using/security-practices.md
@@ -66,7 +66,7 @@ while allowing statistics gathering when the user consents to it.
 By our analysis, Sandstorm automatically protected users from over 95% of the publicly disclosed
 security vulnerabilities discovered in apps on the Sandstorm app market, before the vulnerabilities
 were even disclosed. We also mitigated most Linux kernel security issues. See [Security
-non-events](security-non-events) for examples of security problems which were mitigated by
+non-events](security-non-events.md) for examples of security problems which were mitigated by
 Sandstorm.
 
 ## Strategies
@@ -143,7 +143,9 @@ _capabilities_ via covert channels.
 
 _**Notice:** As of this writing (April 2021), Sandstorm retains limited
 incoming and outgoing SMTP to allow email clients to work. This will be
-removed once email support via the Powerbox is implemented._
+removed once email support via the Powerbox is implemented. Additionally,
+there are limitations to our client-side sandboxing at this time. See
+the related section on this document for details._
 
 ### Capability-based Usable Security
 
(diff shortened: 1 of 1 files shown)


docs/using/security-practices.md
```diff
@@ -308,6 +295,6 @@ off, which is still a big win.
 Sandstorm will soon employ the `Content-Security-Policy` header to
 prevent an app from communicating with other origins without
 permission, in order to implement full confinement. As of this writing
-(May 2015), this has not yet been put in place, mostly because
-server-side confinement is not complete (as described earlier) which
-makes client-side confinement largely moot for the moment.
+(April 2021), this has been implemented, and can be enabled with a
+configuration option, but is pending wider rollout on ensuring existing
+apps are updated to be compatible with this change.
```
zenhack, line 300: Note that the version that's implemented is still not quite full confinement. In particular:

- Images & media are still allowed through; I have some WIP that rigs up csp reporting to trigger permission prompt rather than just letting it through.
- It is still possible to use webrtc, and as of right now there's no way to tell the browser to disallow this -- I need to get back to bugging w3c folks about this.

ocdtrekkie: Newest commit should address this.

docs/using/security-practices.md
```diff
@@ -141,17 +141,9 @@ channel by leaking plain bits -- in technical terms, _capabilities_ in
...
-opened some intentional holes in our confinement model. For example,
-we have allowed outgoing HTTP to arbitrary servers in order to permit
-the TinyTiny RSS app to fetch RSS feeds, and we have allowed incoming
-and outgoing SMTP (with certain restrictions) to allow email clients
-to work. These holes will be closed as soon as the Powerbox UI and
-drivers make them obsolete, but in the meantime Sandstorm does not yet
-implement true confinement._
+_**Notice:** As of this writing (April 2021), Sandstorm retains limited
+incoming and outgoing SMTP to allow email clients to work. This will be
+removed once email support via the Powerbox is implemented._
```
zenhack, line 146: I do think we should still more strongly flag the fact that the network isolation is not fully implemented here as well, perhaps noting that the CSP feature still needs to be finished.

ocdtrekkie: I think my latest commit addresses this. I didn't want to go into detail twice in the document, so I referred to the latter comment. I also (think) I fixed a link which was reported broken on IRC.

zenhack:
LGTM

PR #3519, opened 2021-05-02
Grain settings page

zenhack:
This adds a per-grain settings page, which can be accessed by clicking on a new gear icon in the top-bar on the grain page. Right now, it displays any scheduled jobs for the grain and that's it. I don't consider this ready yet, because the design is really *that* bad. I'd like to make this more presentable before we ship it. I also probably want to add a button to cancel background tasks.

The first commit is an unrelated slight refactoring that I was planning on using and then didn't end up needing in the final design -- but I think it's a bit cleaner so I decided to keep it.

Screenshot:

Screen Shot 2021-05-02 at 15 32 39


zenhack:
Another gripe I have is that right now the gear icon still shows up if you've got the settings open, but it just doesn't do anything. Maybe it should go back to the grain, just so there's some kind of feedback?

ocdtrekkie:
I'd agree, the buttons for sharing and webkeys tend to toggle on and off if you click on them.

The other high priority visibility item I'd look for here is the granted Powerbox requests, and the ability to revoke them.

Users don't authorize the scheduled tasks, so I imagine we'd just have an on-off toggle to block the app from using that functionality here?

zenhack:
I implemented the toggle behavior on the gear icon.

I agree wrt powerbox requests & revoking scheduling access altogether, but I also think it's ok to ship a first version of this with limited functionality and then add stuff to it.

Looks like there are some new linter errors I need to hunt down.

ocdtrekkie:
As far as the styling question, Can we lift the same style sheets used in the admin panel?

The headings and table styling here looks like what we want:
image

zenhack:
I think that would make sense.

ocdtrekkie:
I would also replace "Open Grain" with "Return to Grain" or similar, unless we expect people to navigate straight to the settings. People may feel like they already opened the grain, so I think the language choice there is weird.

zenhack:
I think I was anticipating adding a button on the grain list, but I'll change it to 'return' at least until I actually do that.

zenhack:
Alright, I added some basic styling, and also a couple more columns to the table. Screenshot:

Screen Shot 2021-05-02 at 21 39 51

I'm going to mark this as not a draft; there's more we could do but this is no longer at the level where I would be embarrassed if we shipped this.

ocdtrekkie:
That looks really good. And I'd probably find the next run information pretty useful for troubleshooting and testing.

shell/i18n/en.i18n.json
```diff
@@ -256,6 +260,26 @@
...
+      "scheduledTasks": {
+        "title": "Scheduled Tasks",
+        "headings": {
+          "description": "Description",
+          "period": "Period",
+          "next": "Runs next at",
+          "created": "Created at"
+        },
+        "periodNames": {
+          "anually": "anually",
```
ocdtrekkie, line 275: s/anually/annually x2 here

ocdtrekkie (changes requested):

ocdtrekkie (approved):

zenhack:
I made two other tweaks:

- The return to grain button is now styled.
- I added cancel buttons for the tasks in the scheduled tasks table. It felt super weird to me to have a page called settings where you couldn't *do* anything, so I felt like I should implement at least one thing.

Screenshot of the result:

Screen Shot 2021-05-06 at 18 40 01

zenhack:
Per discussion with @ocdtrekkie on #3522, I backed out the cancel button feature. I'll likely add the toggle as a separate pr stacked on top of this one.

zenhack:
Merged in changes from master. @kentonv, did you miss this one when making the last release? Would appreciate a review.

kentonv:
Hmm I could have sworn this wasn't marked ready yet last time, but the history seems to disagree with me. Sorry about that.

Looks great to me!

PR #3520, opened 2021-05-02
Use typescript friendly linting rules.

zenhack:
eslint's no-unused-vars warning was flagging variables in typescript
source that were only used in types (but not at the value level). This
patch changes our eslint config to use typescript friendly
plugins/rules.

This also skips a few .js files that don't make sense to lint and were
causing spurrious warnings, e.g. minifed outputs in public/.

The new plugins also add some more warnings, a couple of which I fixed.
The patch nets us -1 warning.

---

I hit this while working on #3519

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,7 +72,7 @@ jobs:
           # the constant below.
           make lint | tee lint.log
           num_problems=$(cat lint.log | grep ' problems (0 errors, ' | awk '{print $2}')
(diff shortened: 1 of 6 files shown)


zenhack:
I'm not sure this is actually working yet, so marking it as a draft for now.

zenhack:
Ok, yeah, this doesn't actually fix the core problem I was trying to fix. Closing for now, I may do another pass at this later.

PR #3524, opened 2021-05-21
Makefile: Made certain variables overridable from environment variables

nniro:
These variables are now overridable/settable : CC, CXX, CFLAGS and CXXFLAGS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,14 +16,14 @@
 
 # You may override the following vars on the command line to suit
 # your config.
-CC=$(shell pwd)/deps/llvm-build/Release+Asserts/bin/clang
-CXX=$(shell pwd)/deps/llvm-build/Release+Asserts/bin/clang++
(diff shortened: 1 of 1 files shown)


zenhack (commented):
Hm, I thought these were already overridable; it looks like it only works as-is if you do:

`make CXX=/path/to/other/clang++`

As opposed to:

`CXX=/path/to/other/clang++ make`

+1 for making it work both ways. Perhaps want to do the same for the remaining variables in this block as well?

kentonv:
Thanks for contributing, but TBH I think I prefer it as it was -- the variables are meant to be overridden explicitly by passing parameters to `make`. Is there some particular scenario where passing arguments doesn't work well?

ocdtrekkie:
I am going to go ahead and close this one, as Kenton declined it without further explanation for why it was preferred, and no further explanation has been given in a couple of intervening months.

PR #3528, opened 2021-06-05
Opt out of FLoC

### LINKED ISSUE #3517: Sandstorm should probably opt servers out of FLoC
We probably should defend our users, as much as possible, against hostile web browsers. GitHub Pages has already decided to automatically opt all users out of FLoC, and WordPress is discussing doing the same by default. https://github.com/WICG/floc

`Permissions-Policy: interest-cohort=()`

zenhack:
Fixes #3517

---

I'll likely do a cleanup/documentation pass in the next day or two, but this works.

### DIFF
--- a/src/sandstorm/gateway.c++
+++ b/src/sandstorm/gateway.c++
@@ -101,8 +101,7 @@ GatewayService::GatewayService(
       tables(tables), baseUrl(kj::Url::parse(baseUrl, kj::Url::HTTP_PROXY_REQUEST)),
       wildcardHost(wildcardHost), termsPublicId(termsPublicId), tasks(*this),
       allowLegacyRelaxedCSP(allowLegacyRelaxedCSP),
-      defaultHeaders(kj::HttpHeaders(tables.headerTable))
-{
+      defaultHeaders(kj::HttpHeaders(tables.headerTable)) {
   // Tell chrome not to involve us in its spying on its users:
   defaultHeaders.set(tables.hPermissionsPolicy, "interest-cohort=()");
 }
--- a/src/sandstorm/gateway.c++
+++ b/src/sandstorm/gateway.c++
@@ -20,6 +20,7 @@
 #include <kj/encoding.h>
 #include <sandstorm/mime.capnp.h>
 #include "util.h"
+#include "util/http.h"
 #include "smtp-proxy.h"
 #include <sys/socket.h>
 #include <arpa/inet.h>
@@ -84,6 +85,7 @@ GatewayService::Tables::Tables(kj::HttpHeaderTable::Builder& headerTableBuilder)
       hDav(headerTableBuilder.add("Dav")),
       hLocation(headerTableBuilder.add("Location")),
       hOrigin(headerTableBuilder.add("Origin")),
+      hPermissionsPolicy(headerTableBuilder.add("Permissions-Policy")),
       hUserAgent(headerTableBuilder.add("User-Agent")),
       hWwwAuthenticate(headerTableBuilder.add("WWW-Authenticate")),
       hXRealIp(headerTableBuilder.add("X-Real-IP")),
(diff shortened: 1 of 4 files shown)


src/sandstorm/util/http.c++
```diff
@@ -0,0 +1,35 @@
...
+  kj::HttpHeaders newHeaders = headers.cloneShallow();
+  extraHeaders.forEach(
+    [&](kj::HttpHeaderId id, kj::StringPtr value) {
+      newHeaders.set(id, value);
+    },
+    [&](kj::StringPtr name, kj::StringPtr value) {
+      newHeaders.add(name, value);
+    }
+  );
+  extraHeaders = kj::mv(newHeaders);
```
kentonv, line 20: It's not necessary to store the new headers as a class member, because the contract of `Response::send()` states that the headers can be destroyed as soon as the method returns. So you could put the new object on the stack instead.

I guess it doesn't actually make a difference in the end, but feels kind of weird to modify the class state like this.

src/sandstorm/gateway.c++
```diff
@@ -98,7 +100,12 @@ GatewayService::GatewayService(
     : timer(timer), shellHttp(kj::newHttpService(shellHttp)), router(kj::mv(router)),
       tables(tables), baseUrl(kj::Url::parse(baseUrl, kj::Url::HTTP_PROXY_REQUEST)),
       wildcardHost(wildcardHost), termsPublicId(termsPublicId), tasks(*this),
-      allowLegacyRelaxedCSP(allowLegacyRelaxedCSP) {}
+      allowLegacyRelaxedCSP(allowLegacyRelaxedCSP),
+      defaultHeaders(kj::HttpHeaders(tables.headerTable))
+{
```
kentonv, line 105:
```suggestion
      defaultHeaders(kj::HttpHeaders(tables.headerTable)) {
```

src/sandstorm/util/http.h
```diff
@@ -0,0 +1,24 @@
...
+    kj::Own<kj::AsyncOutputStream> send(
+        kj::uint statusCode, kj::StringPtr statusText, const kj::HttpHeaders& headers,
+        kj::Maybe<uint64_t> expectedBodySize = nullptr);
+    kj::Own<kj::WebSocket> acceptWebSocket(const kj::HttpHeaders& headers);
+
+  private:
+    void updateHeaders(const kj::HttpHeaders& headers);
+
+    kj::HttpService::Response& origResponse;
+    kj::HttpHeaders extraHeaders;
```
kentonv, line 22:
```suggestion
public:
  ExtraHeadersResponse(kj::HttpService::Response& orig, kj::HttpHeaders&& extraHeaders);

  kj::HttpHeaders& headers();

  // Implements Response
  kj::Own<kj::AsyncOutputStream> send(
      kj::uint statusCode, kj::StringPtr statusText, const kj::HttpHeaders& headers,
      kj::Maybe<uint64_t> expectedBodySize = nullptr);
  kj::Own<kj::WebSocket> acceptWebSocket(const kj::HttpHeaders& headers);

private:
  void updateHeaders(const kj::HttpHeaders& headers);

  kj::HttpService::Response& origResponse;
  kj::HttpHeaders extraHeaders;
```

src/sandstorm/gateway.c++
```diff
@@ -174,6 +181,15 @@ bool isAllowedBasicAuthUserAgent(kj::StringPtr ua) {
 }
 
 kj::Promise<void> GatewayService::request(
+    kj::HttpMethod method, kj::StringPtr url, const kj::HttpHeaders& headers,
+    kj::AsyncInputStream& requestBody, Response& origResponse) {
+
+  auto response = kj::heap<util::http::ExtraHeadersResponse>(origResponse, defaultHeaders.cloneShallow());
```
kentonv, line 187: If `ExtraHeadersResponse` kept `defaultHeaders` as a const reference, then the clone wouldn't be needed here. It would be safe since `GatewayService` definitely outlives the request/response.

src/sandstorm/util/http.c++
zenhack: Aha, I'd missed that; that does make this much simpler.

zenhack:
I've incorporated your suggestions.

PR #3531, opened 2021-06-14
Add PowerDNS plugin for ACME.js

### LINKED ISSUE #3530: Add support for passing ACME challenges to PowerDNS-based DNS services
Sandstorm is all about self-hosting. However, all DNS services which currently are supported by Sandstorm to pass ACME challenges (with the purpose of creating TLS certificates) cannot be self-hosted. PowerDNS Authorative Server is a software allowing to self-host a DNS service. An operator of a Sandstorm instance should be able to configure Sandstorm to use their own PowerDNS Authorative Server instance, akin

```
sandstorm configure-acme-challenge powerdns '{ "foo": "bar" }'
```

However, currently, this fails:

```
server:~ # sandstorm configure-acme-challenge powerdns '{ "foo": "bar" }'
sandstorm configure-acme-challenge: powerdns: unknown module: powerdns
Try 'sandstorm configure-acme-challenge --help' for more information.
```

It fails because, currently, the Sandstorm distribution does not contain the acme-dns-01-powerdns NodeJS package (see https://www.npmjs.com/package/acme-dns-01-powerdns or https://git.y.gy/firstdorsal/acme-dns-01-powerdns).
Therefore, this package should be included in the Sandstorm distribution, like the other acme-dns-01-* NodeJS package are already included.

It seems that adding a dependency on this package is a trivial task.

ocdtrekkie:
Fixes #3530 as requested by @riesenrad.

I have not at all tested this, but this PR's changes should be very parallel to #3449, and, unless something is horribly wrong with this package, shouldn't impact anyone who isn't trying to use PowerDNS.

### DIFF
--- a/shell/imports/client/admin/certificates-client.js
+++ b/shell/imports/client/admin/certificates-client.js
@@ -39,6 +39,7 @@ const DNS_PROVIDERS = {
   "gcp": "Google Cloud Platform",
   "namecheap": "NameCheap",
   "namedotcom": "Name.com",
+  "powerdns": "PowerDNS",
   "route53": "Route53 (AWS)",
   "vultr": "Vultr",
 };
--- a/shell/imports/client/shell.html
+++ b/shell/imports/client/shell.html
@@ -525,6 +525,7 @@ limitations under the License.
         <li><a href="https://github.com/nodecraft/acme-dns-01-cloudflare">ACME.js Cloudflare plugin by Nodecraft</a> (MIT {{_ "shell.about.dependencies.license"}})</li>
         <li><a href="https://github.com/rs22/acme-dns-01-desec">ACME.js deSEC plugin by Robert Schmid</a> (MIT {{_ "shell.about.dependencies.license"}})</li>
         <li><a href="https://github.com/latacora/acme-dns-01-gcp">ACME.js GCP plugin by Latacora</a> (MPL)</li>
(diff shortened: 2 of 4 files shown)


ocdtrekkie:
Okay, so, I botched this, and also discovered a potential problem. In the about page, I failed to correctly update the source link when I copypasted, so I need to update that. Then I found the source for 1.0.4 here: https://git.y.gy/firstdorsal/acme-dns-01-powerdns but that's two versions old, so we probably shouldn't merge this with old source available.

ocdtrekkie:
I've messaged the contact email on NPM to ask about the updated source.

zenhack:
Before we merge this, I'd like to skim the implementation of whatever version we actually end up using to make sure there's not anything ridiculous happening there.

In the longer term, it would be nice to use lavamoat or such to isolate dependencies like this. I don't like the fact that this architecture has us adding random packages for each provider that execute with full authority to the entire sandstorm box. But we don't need to block on that.

ocdtrekkie:
Agreed. This is a very small package (<150 lines), so glancing it over shouldn't be too bad. But the fact that we're just throwing in relatively unknown packages isn't ideal.

ocdtrekkie:
I am planning to fix up this PR tonight. https://git.y.gy/firstdorsal/acme-dns-01-powerdns has been updated with the latest source and had it's metadata cleaned up.

zenhack:
Skimming the implementation it looks reasonable.

PR #3532, opened 2021-06-15
Add PowerDNS plugin for ACME.js

### LINKED ISSUE #3530: Add support for passing ACME challenges to PowerDNS-based DNS services

ocdtrekkie:
Fixes #3530 as requested by @riesenrad.

I have not at all tested this, but this PR's changes should be very parallel to #3449, and, unless something is horribly wrong with this package, shouldn't impact anyone who isn't trying to use PowerDNS.

### DIFF
--- a/shell/imports/client/admin/certificates-client.js
+++ b/shell/imports/client/admin/certificates-client.js
@@ -39,6 +39,7 @@ const DNS_PROVIDERS = {
   "gcp": "Google Cloud Platform",
   "namecheap": "NameCheap",
   "namedotcom": "Name.com",
+  "powerdns": "PowerDNS",
   "route53": "Route53 (AWS)",
   "vultr": "Vultr",
 };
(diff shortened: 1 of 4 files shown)


ocdtrekkie:
Okay, I decided to just replace #3531 instead of screwing with branch pushes. This should be a good PR.

zenhack (approved):

PR #3539, opened 2021-06-30
CSP: Set worker-src to none

### LINKED ISSUE #3538: Disallow grain service workers
Currently, the CSP policy (even the more restrictive one) does not disallow service workers. This may be a security problem; while discussing the webrtc stuff, it was pointed out that service workers can muck with response headers:

https://github.com/w3c/webrtc-extensions/pull/81#issuecomment-871563136

...so we probably need to block grains from installing service workers. I don't think any existing apps actually use service workers, and if they did they'd probably break a lot for the same reason local storage doesn't really work in Sandstorm, so I don't think there's a downside to getting rid of them.

This is a one-line patch; I'll find time for it soonish.

zenhack:
Fixes #3538


zenhack:
Hm, did github recently make their path checking more strict?

PR #3543, opened 2021-07-28
Update doc link to allowed non-web-browser User-Agent strings

garrison:
This updates the documentation following the change in 4edb2aeb1788542a78d1a3f85884848c45c09d0c (#3038).

### DIFF
--- a/docs/developing/http-apis.md
+++ b/docs/developing/http-apis.md
@@ -226,7 +226,7 @@ the browser that the other can read.
 There is also a generic API hostname that allows all API tokens. However, if you make a request to
 the generic API hostname using HTTP Basic auth, then those requests are subject to a [whitelist of
 non-web-browser `User-Agent`
-strings](https://github.com/sandstorm-io/sandstorm/search?utf8=%E2%9C%93&q=BASIC_AUTH_USER_AGENTS). Therefore,
+strings](https://github.com/sandstorm-io/sandstorm/search?utf8=%E2%9C%93&q=isAllowedBasicAuthUserAgent). Therefore,
 it is vastly easier to configure HTTP clients to use the token-specific hostname.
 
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
I'm prepared to accept this simply because the old one no longer works at all, and the new one does work... but I am wondering if something like `%22bool%20isAllowedBasicAuthUserAgent%22` might be a better search term, as it only presents one result, at the place where the necessary information is located?

I find it frustrating that in either case, the link doesn't present the necessary information unless you click into the code page itself, but I assume the previous author of this page chose to use a search rather than highlighting line numbers to maintain accuracy despite code changes over the years.

garrison:
> ... but I am wondering if something like `%22bool%20isAllowedBasicAuthUserAgent%22` might be a better search term

That is an alternative.  Here are the two links to compare:

- [isAllowedBasicAuthUserAgent](https://github.com/sandstorm-io/sandstorm/search?utf8=%E2%9C%93&q=isAllowedBasicAuthUserAgent)
- ["bool isAllowedBasicAuthUserAgent"](https://github.com/sandstorm-io/sandstorm/search?utf8=%E2%9C%93&q=%22bool%20isAllowedBasicAuthUserAgent%22)

A few notes:

- Each search option cuts off the function definition at the same point -- there is no additional information in the preview for the second option.
- For some reason, the `bool` gets highlighted in the preview for the second option, while the actual most important part of the string does not.
- The first option provides additional context regarding where the function is used, and it's fairly obvious just from looking at the preview where the function definition is, arguably even for those not proficient in C++.

> I assume the previous author of this page chose to use a search rather than highlighting line numbers to maintain accuracy despite code changes over the years.

That is my suspicion too.

ocdtrekkie:
Makes sense to me.

Issue #3427, opened 2020-09-02
Provide non-interactive method for production install in install.sh


garrison:
I recently created a [one-click install script](https://github.com/garrison/sandstorm-linode-stackscript) that deploys Sandstorm on a new Linode instance (see also: [this thread](https://groups.google.com/g/sandstorm-dev/c/A8qCNUJXMOs) on sandstorm-dev).  The existing [install.sh](https://github.com/sandstorm-io/sandstorm/blob/master/install.sh) script allows a development instance to be created without interaction by passing the `-d` flag, but setting up a production server currently requires interaction with the script.  I managed to work around this for now [using empty-expect](https://github.com/garrison/sandstorm-linode-stackscript/blob/4950cc143d0d5115216e3a6e829d801cd184e58a/sandstorm-linode-stackscript.bash#L55-L59), but ideally install.sh should have direct support for configuring a new instance using environment variables alone.

ocdtrekkie:
This is also related to #3379 (which needs this *and* #3395)

garrison:
Looking at 6b5351c4a58c29e9bf3ad6d16bae27895aeaad38, it seems I may actually be able to get part way there by setting both `CHOSEN_INSTALL_MODE` and `DESIRED_SANDCATS_NAME`.  However, there appears to be no way to set `SANDCATS_REGISTRATION_EMAIL` from outside the script, but fixing this would require only a minor change.  I may experiment with this in a few days.

ocdtrekkie:
As we discussed in Office Hours, I also think it'd be really great if we could figure out a way to configure a valid SMTP server/account, and properly mark the registration email as the admin user, so that we can ideally skip past the setup wizard for non-interactive installs. This might fall outside of install.sh's purview though, since it'd entail database configuration.

PR #3544, opened 2021-07-29
Allow non-interactive full server install via environment variables

garrison:
This pull request allows users to achieve a production install without interaction by providing environment variables to `install.sh`, thus fixing #3427.  I tried to be as conservative in my changes as possible so as not to introduce bugs, but comments on the approach are welcome.

Specifically, the following environment variables should be passed for a noninteractive install:

- `CHOSEN_INSTALL_MODE=1` (production server, not development server)
- `ACCEPTED_FULL_SERVER_INSTALL=yes` (skip the prompt for confirmation)
- `DESIRED_SANDCATS_NAME` 
- `SANDCATS_REGISTRATION_EMAIL`
- `ACME_EMAIL` (for Let's Encrypt)

A few caveats:
- If the provided `DESIRED_SANDCATS_NAME` is invalid or already taken, the install script will then become interactive to try to remedy the situation, rather than erroring out.
- The existing `install.sh` includes logic involving a `SANDCATS_DOMAIN_RESERVATION_TOKEN` (see e.g. 6b5351c4a58c29e9bf3ad6d16bae27895aeaad38), but I don't believe it is publicly documented how to get such a token.
- If run as an ordinary user, the script restarts itself with `sudo` access, which causes some of the above environment variables to be lost, and the script becomes interactive again. I expect this can be fixed by adding these variables to the call to `rerun_script_as_root`. Once I fix and test this, I will change this from a "draft" to a regular PR.


ocdtrekkie:
This is really exciting to see work on! Over the years we've occasionally had people ask about ways to do this, but the answer has always been kind of lacking. It'll be nice to have a better solution for them.

My only question would be about CHOSEN_INSTALL_MODE. That's different than how we presently set dev mode, isn't it?

ocdtrekkie:
I guess I am familiar only with the "passing -d" method of setting dev install, lol. But since we need to set a pile of variables for this to work, I don't think the production install needs an equivalent.

ocdtrekkie:
It probably would be good to add the information from your PR text into https://docs.sandstorm.io/en/latest/administering/install-script/ as well, since that's where people will look for it.

ocdtrekkie:
Re: reservation token, I guess @paulproteus might remember, but I think there was going to be some sort of Sandcats pre-registration service for that which never materialized.

It's possible that's code that should be deleted, to be honest.

zenhack:
Thanks for doing this!

I think if we're going to expose `CHOSEN_INSTALL_MODE` as a documented interface, I'd prefer to use clear names for the values, rather than numbers; maybe just `development` and `production`.

Other than that this looks broadly sensible, though I haven't tested it.

ocdtrekkie:
Without having tested all of the various install paths, everything *looks* good. Some of the cleanup is definitely needed. I am not sure whether or not we should be concerned about any breakage with any existing scripts... I know we've helped a few people play with that over the years, but my guess is that not many or none have built a production setup out of it.

ocdtrekkie:
Is there any blockers on moving this from a draft into ready for Kenton?

garrison:
I haven't seriously tested the most recent version of this. Will try to do so soon.

garrison:
I have made the changes in the cleanup commit more conservative (b2d3851896637856d4ca591d7a9f2da377e65574) now that I realized that `SANDCATS_GETCERTIFICATE` is essentially unused.  I don't expect any more major changes, but I still want to test this on a full server install and will do so later today.

garrison:
One implication of `SANDCATS_GETCERTIFICATE` having no effect in the current install script [see footnote] is that passing the `-i` option to withhold requesting a certificate is currently broken; it has no effect either.  Instead of setting a variable with no effect, the current version of this patch series replaces this option with a `TODO(soon)` to either fix or remove this option.  Which path should we take?   What are the intended use cases?  Servers behind a reverse proxy?  Servers on a local area network, where certificates cannot be fetched by Let's Encrypt?  This has been broken for over a year, since 9ac03603119b68674284e6e3418aa294a37d82b8, so I want to make sure it is useful before I fix it.

[footnote]:  other than [here](https://github.com/sandstorm-io/sandstorm/blob/0811bcdc7d84ccf3c005554a7b269f984934e816/install.sh#L829-L831); note that the [other branch](https://github.com/sandstorm-io/sandstorm/blob/0811bcdc7d84ccf3c005554a7b269f984934e816/install.sh#L1979-L1984) is in a function that is never called

ocdtrekkie:
Have you had a chance to test this? Should I mark it ready for Kenton to review?

The install script has not gotten a significant amount of love in the Sandstorm-as-a-community-project years. The Let's Encrypt fixes were needed to maintain basic functionality, but not sure anyone's really focused on tuning the potential use cases of install script features and robustly checked every branch. Your judgment call is probably better than most of ours on what should or shouldn't be done with it, if you've spent some time hacking on it recently.

garrison:
I just tested this
- with development defaults
- as a development server in interactive mode
- as a production server via environment variables, with the script invoked as a regular user -- the only interaction required was to enter a sudo password, as expected
- as a production server partially specified by environment variables, with the remaining things entered in interactive mode
- as a production install in interactive mode, triggering every code path I could

I noticed a few bugs in the interactive code paths, which are now fixed (I missed a few places where I needed to `unset` a variable before calling `sandcats_register_name` recursively).

I think it's now ready for review.

PR #3548, opened 2021-08-11
Fix #3536, graphviz on "How it Works" page

### LINKED ISSUE #3536: Documentation page is broken
The [documentation page relating to "How it Works"](https://docs.sandstorm.io/en/latest/using/how-it-works/) is broken.

It shows an error of 

`Error : a bytes-like object is required, not 'str' : ['dot', '-Tsvg']`

but the documentation source page on github [appears](https://github.com/sandstorm-io/sandstorm/edit/master/docs/using/how-it-works.md) and previews fine.

It looks like the problem relates to an image that was intended to be inserted between some {% %} tags.

sandcats-documentation-error-screenshot_20210623_093931

garrison:
This fixes #3536, except the unrelated https://github.com/sandstorm-io/sandstorm/issues/3536#issuecomment-866981913

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -12,7 +12,7 @@ sudo apt-get install -y graphviz virtualenv
 cd ~/projects/sandstorm
 virtualenv tmp/docs-virtualenv
 tmp/docs-virtualenv/bin/pip install mkdocs==1.0.4
-tmp/docs-virtualenv/bin/pip install markdown-inline-graphviz==1.0
+tmp/docs-virtualenv/bin/pip install markdown-inline-graphviz-extension==1.1
 tmp/docs-virtualenv/bin/mkdocs serve
 ```
 
--- a/docs/generate.sh
+++ b/docs/generate.sh
@@ -50,7 +50,7 @@ assert_dependencies() {
     if [ ! -x "tmp/docs-virtualenv/bin/mkdocs" ] ; then
       mkdir -p tmp
       virtualenv tmp/docs-virtualenv
-      tmp/docs-virtualenv/bin/pip install mkdocs==1.0.4 markdown-inline-graphviz==1.0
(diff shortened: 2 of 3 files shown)


ocdtrekkie (approved):

ocdtrekkie:
Thanks!

garrison:
I cannot tell why, but much of the text which is meant to be in the diagrams at https://docs.sandstorm.io/en/latest/using/how-it-works/ is mistakenly outside them.  I cannot reproduce this locally (e.g. using `docs/generate.sh`), and looking at https://github.com/sandstorm-io/sandstorm/blob/master/.github/workflows/doc-prod.yml provides no additional insight ...

The source of my local build result is different from that page, however.  Is there an extra processing step going on before deployment?

ocdtrekkie:
Shouldn't be anything different than what you see in the repo. GitWeb is just statically hosting the output. 🤔

garrison:
Another idea: I will try to reproduce locally using the `ubuntu:latest` docker image, which is used in the action. Previously I have run `docs/generate.sh` from outside a container, but there are some differences, including a different minor version of python.

garrison:
I was able to reproduce locally.  Pretty sure the different version of graphviz is to blame: https://gist.github.com/garrison/c7c9dcff2f60c8eeb0e7959dde85ae27

Things look correct with graphviz 2.40.1 (ubuntu 18.04) but incorrect with version 2.43.0 (`ubuntu:latest`).

ocdtrekkie:
https://pypi.org/project/mkdocs-markdown-graphviz/ appears to be a superior fork to the previous two, and allegedly adds fixes to rendering.

garrison:
That looks promising.  Changing the graphviz version alone did not fix it.  All these extra `<p>` tags must be coming from somewhere else.

PR #3551, opened 2021-08-12
Artifacts from failing tests get valid filenames

### LINKED ISSUE #3550: Artifact path is not valid (tests failing)
`Artifact path is not valid: /tests/grain/Dev-Account:-Test-grain-frame_FAILED_Aug-12-2021-210941-GMT+0000-(Coordinated-Universal.png. Contains character: ":". Invalid characters include: ",:,<,>,|,*,?.`

This is interesting because:

- Tests are failing badly right now: `TEST FAILURE:  19 assertions failed, 271 passed. (7m 55s)`
- Our use of the colon in some test names, means when those tests fail, the test framework errors on uploading screenshots of the failure.

ocdtrekkie:
Fixes #3550, I think. (Well, it won't fix that tests are failing left and right, but it might give us screenshots again.)

### DIFF
--- a/tests/utils.js
+++ b/tests/utils.js
@@ -45,7 +45,7 @@ module.exports = {
       if (count === 0) {
         test = wrapLoginDev(test);
       }
-      newTests['Dev Account: ' + name] = test;
+      newTests['Dev Account- ' + name] = test;
       ++count;
     }
 


ocdtrekkie:
This does cause our test suite to successfully run. We are up to like 19 test failures because something appears to have broken horribly that I don't understand. But now we can see photo evidence of all of that failure.

zenhack (approved):
LGTM

zenhack:
Ug, I wish I knew of a way to mark individual tests as expected to fail -- I forget to actually check the test log a lot of the time, and it seems like we're backsliding. I think we really need to find a way to have CI flag *new* test failures for us, or this is just going to get worse and worse.

Sadly, #3497 doesn't quite work since some of the failures are intermittent.

I suppose we could just put a `return` at the top of the failing tests, with some greppable string in a comment so we can find the things that still need fixing. But this is somewhat sad.

@zarvox, I don't suppose you know of some functionality in nightwatch for this that I don't?

PR #3564, opened 2021-09-17
Update documentation on installing from source

garrison:
- `flex` and `bison` are required for the build.
- `locales` is required under Debian/Ubuntu; otherwise the build fails because `/usr/share/i18n/charmaps` is missing.
- bullseye is now the latest Debian stable release.  The source build fails on stretch because `SECCOMP_FILTER_FLAG_LOG` is undeclared, unless (presumably) the kernel headers were to be upgraded.  (See also 2021-09-06 discussion on IRC.)  The `contrib-` is no longer necessary because the `vboxsf` kernel module is in `debian/bullseye64`.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -135,6 +135,8 @@ Please install the following:
 * `xz`
 * `zip`
 * `unzip`
+* `flex`
+* `bison`
 * `strace`
 * `curl`
 * `python`
@@ -149,7 +151,7 @@ On Debian or Ubuntu, you should be able to get all these with:
 
     sudo apt-get install build-essential libcap-dev xz-utils zip \
         unzip strace curl discount git python zlib1g-dev \
-        golang-go cmake strace
+        golang-go cmake strace flex bison locales
     curl https://install.meteor.com/?release=1.8.2 | sh
 
 On Fedora 27 you should be able to get them with (as root):
@@ -157,13 +159,13 @@ On Fedora 27 you should be able to get them with (as root):
     dnf install make libcap-devel libstdc++-devel libstdc++-static \
        glibc-headers glibc-static glibc-locale-source xz zip \
        unzip strace curl discount git python2 zlib-devel \
-       golang cmake strace
+       golang cmake strace flex bison
     curl https://install.meteor.com/?release=1.8.2 | sh
 
 If you have trouble getting the build to work on your distro, we recommend trying in a virtual
 machine running the latest stable Debian release. This is easy to set up using Vagrant, like:
 
-    vagrant init debian/contrib-stretch64
+    vagrant init debian/bullseye64
     vagrant up
     vagrant ssh
 


docs/install.md
```diff
@@ -149,21 +151,21 @@ On Debian or Ubuntu, you should be able to get all these with:
 
     sudo apt-get install build-essential libcap-dev xz-utils zip \
         unzip strace curl discount git python zlib1g-dev \
-        golang-go cmake strace
+        golang-go cmake strace flex bison locales
     curl https://install.meteor.com/?release=1.8.2 | sh
 
 On Fedora 27 you should be able to get them with (as root):
```
zenhack, line 157: Fedora is up to version 34 now; are we sure the seccomp flag isn't an issue on this version as well?

I'm tempted to suggest that we should test this and probably bump the version here in the docs to something somewhat recent. I won't block on this though.

zenhack (approved):
Thanks!

zenhack:
Hm, I just tried to build sandstorm inside a fresh bullseye container (for completely unrelated reasons; trying to debug test failures), and got an error due to a missing `node.h`. In most situations, the solution is "install nodejs-dev," but we should be using the version of meteor shipped with node, so I don't think just adding that to the instructions is the correct solution.

In fact, I wonder if the test failures we're seeing are because we're building node-capnp against the system node headers, and it works on Kenton's machine because he's running bullseye, which is using node v12.x, and doesn't on mine, because I'm running Arch which has node v16.x, and v12 doesn't happen to cause problems.

Anyway, not going to block on sorting that out either; I think the right solution is to make sure we're *not* building against system headers, not to change the documentation.

garrison:
Hi Ian.  My efforts to build sandstorm actually started on fedora, but I have not gotten them to succeed yet.  I was planning a second PR once they do.  My working Debian Dockerfile and in-progress Fedora Dockerfile are at https://gist.github.com/garrison/f025090ce914641f0d1e492612dcbd4e

garrison:
> we should be using the version of meteor shipped with node

I'm confused by this statement.  Aren't all versions of meteor shipped with (some version of) node?  My first guess was that you meant to say [Node older than 12.x](https://github.com/sandstorm-io/sandstorm/blob/0811bcdc7d84ccf3c005554a7b269f984934e816/Makefile#L292-L295), and this is what my `Dockerfile` does, following the current build instructions.  But the actual Sandstorm build needs the version of meteor specified in `shell/.meteor/release`.  I would expect the 1.8.2 Meteor to be only used for a single purpose, but it seems that the Sandstorm build is pulling `node.h` from there, at least if the instructions are followed as given.  So maybe this old header that does not match the Node version in use is then causing issues?

zenhack:
Quoting Jim Garrison (2021-09-17 06:17:48)

> > we should be using the version of meteor shipped with node
> 
> I'm confused by this statement.

Was supposed to be "we should be using the version of node shipped with
meteor," so not surprising you were confused. But it's moot; once I gave
the VM enough ram for `find-meteor-dev-bundle.sh` to not crash, it found
the correct headers. So, complete red herring, sorry for the
distraction.

zenhack:
Quoting Jim Garrison (2021-09-17 05:55:51)

>    Hi Ian. My efforts to build sandstorm actually started on fedora, but I
>    have not gotten them to succeed yet. I was planning a second PR once
>    they do.

SGTM.

ocdtrekkie:
So after all this discussion... those dependencies should still be added to the docs, yes?

Second issue: We generally have used contrib- Debian boxes because of vagrant-spk, specifically. Do we know why there's no contrib-bullseye64 yet?

zenhack:
Quoting Jacob Weisz (2021-09-17 11:12:33)

>    So after all this discussion... those dependencies should still be
>    added to the docs, yes?

Yes.

> Second issue: We generally have used contrib- Debian boxes because of
> vagrant-spk, specifically. Do we know why there's no contrib-bullseye64
> yet?

Looking at the descriptions for the contrib images on
https://app.vagrantup.com/debian, I see:

> debian/contrib-buster64 v10.20210829.1
> Debian 10 "buster" with vboxsf contrib module for synced folders

Per @garrison's comment:

> The contrib- is no longer necessary because the vboxsf kernel module is in debian/bullseye64.

I assume this also means that the entire use case for those images is
gone, not just the reason we needed it, as it sounds like they were the
same?

ocdtrekkie:
It would've helped if I had actually read @garrison's comment about such, so that's a big shame on me.

garrison:
>> The contrib- is no longer necessary because the vboxsf kernel module is in debian/bullseye64.
>
> I assume this also means that the entire use case for those images is gone, not just the reason we needed it, as it sounds like they were the same?

I believe the only difference between a "contrib" image and the others is whether this module is included.  This seems consistent with the descriptions prior to bullseye at https://app.vagrantup.com/debian.

Issue #3029, opened 2018-01-29
Custom App Market URL


morgan8ter:
Sandstorm Version - 0.224
OS - Fedora 26 (Server Edition)

Sandstorm is working well except for an issue I'm experiencing with admin/app-sources

I would like to use a custom App market and have set the custom App market URL to https://apps.example.local and it sets the Install... from app market link/button correctly, however following that link/button goes to https://apps.sandstorm.io/?host=https://sandstorm.example.local instead of https://apps.example.local/?host=https://sandstorm.example.local

I can't see any issues in the system log and I haven't found any other posts related to this. Has anyone else seen this or have any suggestions to resolve. Let me know if there is any other info that I can provide that will help.

ocdtrekkie:
The redirect link back to your Sandstorm server is located in the App Market code, not Sandstorm itself, and the App Market code isn't customized by your Sandstorm server's settings. If you are hosting a version of https://github.com/sandstorm-io/sandstorm-app-market on your domain, you may have to find that in the code and change it yourself.

ocdtrekkie:
Ah, there's a settings.json for some of the app market URL settings, documented in the app market's readme. And I think I misread where you're having the issue. :/

There's a good chance there's something hardcoded somewhere though, not a lot of people run their own app market, so it probably hasn't been super well tested in that regard.

kentonv:
Whoa... Indeed, after changing the app market URL to something random, if I go to the apps page and hover over the "Install" button, the status bar shows the URL I set, but if I click on it, I always end up at apps.sandstorm.io instead. This seems to be a bug. In fact, it looks like several hard-coded copies of the link have shown up in the source code, even though I thought we had carefully replaced all of them.

morgan8ter:
Happy that you were able to reproduce this issue. Just noticed that it appears to occur only when the "Install" button is clicked, but works as expected if you right click the button and choose open in a new tab.

ocdtrekkie:
Clearly, we have the solution: Tell everyone to always right click and open in a new tab! :D

I think this should be pretty easy to find and fix before the next Sandstorm release, I might dig around a bit tomorrow or the next day.

ocdtrekkie:
There are plenty of references that are like, specific app links in documentation or even the admin panel that should remain, but these are the ones I think need to be changed, @kentonv, to resolve this for @morgan8ter. I will probably avoid doing my own PR on this one since each one likely has a little different solution, they're all in different types of files in the code. 

[...]

PR #3568, opened 2021-09-21
Fix bug #3029 - remove hardcoding of the app-market url.  

gischer:
Additionally, I replaced the deprecated globalDb with SandstormDb in two of the files I touched.

I'm uncertain about what testing I should attempt.

### DIFF
--- a/shell/imports/client/apps/applist-client.js
+++ b/shell/imports/client/apps/applist-client.js
@@ -8,7 +8,6 @@ import { TAPi18n } from "meteor/tap:i18n";
 
 import { introJs } from "intro.js";
 import { SandstormDb } from "/imports/sandstorm-db/db.js";
-import { globalDb } from "/imports/db-deprecated.js";
 
 SandstormAppList = function (db, quotaEnforcer) {
   this._filter = new ReactiveVar("");
@@ -173,7 +172,7 @@ Template.sandstormAppListPage.helpers({
   },
 
   appMarketUrl: function () {
-    const appMarket = globalDb.collections.settings.findOne({ _id: "appMarketUrl" });
+    const appMarket = SandstormDb.collections.settings.findOne({ _id: "appMarketUrl" });
     if (!appMarket) {
       return "#";
     }
@@ -202,7 +201,12 @@ Template.sandstormAppListPage.events({
   "click .install-button": function (event) {
     event.preventDefault();
     event.stopPropagation();
-    window.open("https://apps.sandstorm.io/?host=" +
+    const appMarket = SandstormDb.collections.settings.findOne({ _id: "appMarketUrl" });
+    let url = "https://apps.sandstorm.io/?host=";
+    if (appMarket) {
+      url = appMarket.value + '?host=';
+    }
+    window.open(url +
         document.location.protocol + "//" + document.location.host, "_blank");
   },
 
--- a/shell/imports/client/signup-client.js
+++ b/shell/imports/client/signup-client.js
@@ -20,13 +20,22 @@ import { Meteor } from "meteor/meteor";
 import { Template } from "meteor/templating";
 import { Router } from "meteor/iron:router";
 import { DEFAULT_SIGNUP_DIALOG } from "/imports/client/personalization.js";
-import { globalDb } from "/imports/db-deprecated.js";
+import { SandstormDb } from "/imports/sandstorm-db/db.js";
 
 Template.signup.helpers({
   signupDialog: function () {
-    const setting = globalDb.collections.settings.findOne("signupDialog");
+    const setting = SandstormDb.collections.settings.findOne({_id: "signupDialog"});
     return (setting && setting.value) || DEFAULT_SIGNUP_DIALOG;
   },
+
+  appMarketUrl: function() {
+    let url = "https://apps.sandstorm.io/";
+    const appMarketUrl = SandstormDb.collections.settings.findOne({ _id: "appMarketUrl" });
+    if (appMarketUrl) {
+      url = appMarketUrl.value;
+    }
+    return url;
+  }
 });
 
 Router.map(function () {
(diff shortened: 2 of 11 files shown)


shell/imports/client/apps/applist-client.js
```diff
@@ -173,7 +172,7 @@ Template.sandstormAppListPage.helpers({
   },
 
   appMarketUrl: function () {
-    const appMarket = globalDb.collections.settings.findOne({ _id: "appMarketUrl" });
+    const appMarket = SandstormDb.collections.settings.findOne({ _id: "appMarketUrl" });
```
zenhack, line 175: This is not correct; SandstormDb is a class, and `collections` is only assigned in its constructor, so this is going to throw an exception. `globalDb` is an *instance* of that class.

The right refactor is to do this the way you did it in `appMarketHost()`; use a db instance available via the data context.

Also, creating separate prs for logically separate changes in the future would make reviewing them easier.

Thanks for taking the initiative to clean up some of this.

zenhack (changes requested):
Thanks for the patch!

One major issue jumped out at me, pointed out in-line (it shows up in a couple places).

Otherwise this looks mostly reasonable to me, though:

1. Obviously it needs testing. Did you try this out manually, and were you able to successfully point Sandstorm at an alternate app market URL?
2. It would be nice if this setting were available in the admin panel or at least sandstorm.conf; I'm not a huge fan of adding features that require you to manually modify the database in order to use. I won't block on this, but if we merge without adding this we should at least open an issue for adding a UI.

@ocdtrekkie, looks like CI needs someone to approve running it since this is @gischer's first pr; do you have access to that?

ocdtrekkie:
Build in progress. Note that once you fix the issue Ian pointed out, you can actually download the output of the GitHub Actions run, and install it on your Sandstorm server for testing. Though I assume you'll want to address the issue Ian pointed out and then let it re-run first.

shell/imports/client/apps/applist-client.js
gischer: I have fixed the issue you highlighted.  Thanks for the information.

As to multiple prs, that makes a lot of sense, and something I work for on my own checkins.  And sometimes my ADD tendencies get in the way, I get excited and just check it all in.

gischer:
ok, I just pushed the changes suggested by @zenhack  I have built sandstorm, run it, and verified that changing the app market url in the right subpanel of the admin panel does, in fact, change what is available and it appears to be shown correctly.  

I'm a bit confused about @zenhack 's comment about having to change the database, since there is a place in the admin panel to change both the app market and the app index urls.

I guess the business about actions is that is a different branch of the repo I can download and perform the same tests on?  Do I understand that correctly?

ocdtrekkie:
Yep, if you go to Checks in GitHub, and artifacts for the test run, you can download the Sandstorm tarball built from your PR. But if you built it yourself to test with it's unnecessary.

I know changing the app market URL is already a setting in the UI, so O also don't understand Ian's comment there. You are just fixing the functionality resulting from that setting, which previously did not work correctly,

zenhack:
@ocdtrekkie, I still see the "1 workflow awaiting approval" message.

ocdtrekkie:
I assumed once I approved it once... apparently until after this is merged, any given run will require my approval.

shell/imports/client/signup-client.js
```diff
@@ -20,13 +20,22 @@ import { Meteor } from "meteor/meteor";
 import { Template } from "meteor/templating";
 import { Router } from "meteor/iron:router";
 import { DEFAULT_SIGNUP_DIALOG } from "/imports/client/personalization.js";
-import { globalDb } from "/imports/db-deprecated.js";
+import { SandstormDb } from "/imports/sandstorm-db/db.js";
 
 Template.signup.helpers({
   signupDialog: function () {
-    const setting = globalDb.collections.settings.findOne("signupDialog");
+    const setting = SandstormDb.collections.settings.findOne({_id: "signupDialog"});
```
zenhack, line 27: Looks like you missed this file when addressing my previous comment.

zenhack (changes requested):
@gischer, my mistake re the UI/database question; I'd mistakenly assumed since this was hard-coded it wasn't configurable; I hadn't realized that this already had a setting for it.

I'm going to try to find time this weekend to actually take this for a spin.

shell/imports/client/signup-client.js
gischer: Yep.  Just fixed it, and built.  A great example of how more eyes is good.

zenhack:
@ocdtrekkie re-run the tests?

ocdtrekkie:
Running!

shell/imports/client/signup-client.js
```diff
@@ -41,7 +52,7 @@ Router.map(function () {
     },
 
     data: function () {
-      const keyInfo = globalDb.collections.signupKeys.findOne(this.params.key);
+      const keyInfo = SandstormDb.collections.signupKeys.findOne(this.params.key);
```
zenhack, line 55: One more spot you missed, also in this file. I think this is the last one. (Also, to head it off, I suspect once you fix this the linter will tell you that `SandstormDb` is an unused import

zenhack:
(Vaguely related, I am somewhat dismayed that our test suite did not catch the issue I spotted...)

shell/imports/client/signup-client.js
gischer: Happy to. One thing though.  This is not in a template helper, it's in a route.  I'm not at all sure that Template.instance() works here, and I'm also concerned that if it doesn't it won't show up in my testing.

Do you have a suggestion?

zenhack: We could just punt and keep using globalDb here for now.

gischer: I spent 20 minutes trying to find out how the "data" variable gets set on a template, but I was unsuccessful.  I'm taking your advice and punting.

Pushed the changes.

gischer:
Well, strange things can happen, but its hard to see how that failure has anything to do with what I checked in.  Please point out what I missed.

ocdtrekkie:
@gischer I don't think the failing checks are on you. Master is failed as well. (Kenton builds and tests locally before deployment, which isn't always subject to the same weird test failures.)

PR #3569, opened 2021-09-24
Set Referrer-Policy: same-origin

### LINKED ISSUE #824: Add Content-Security-Policy: referrer no-referrer; to grain iframe
References:

* https://github.com/sandstorm-io/sandstorm/wiki/Security-non-issues
* https://w3c.github.io/webappsec/specs/referrer-policy/

I'd just do this right now but I don't think I have time to do it _and_ test it this very second.

garrison:
This sets `Referrer-Policy: same-origin` according to the discussion at https://github.com/sandstorm-io/sandstorm/issues/824#issuecomment-870142865, and would thus close #824 if merged.  Currently, I've chosen to set this header for all requests, including API requests (it may not be strictly necessary there, but should not hurt anything either) and regardless of whether `allowLegacyRelaxedCSP` is true (the legacy escape hatch should not be necessary for this header).  In the end, I chose to set the header to `same-origin` rather than `no-referrer` after reading this comment in `gateway.c++`:

https://github.com/sandstorm-io/sandstorm/blob/a0fd3f8884d72fc96bd9ade43e8283886f93490e/src/sandstorm/gateway.c%2B%2B#L341-L353

I did not actually test whether `Referrer-Policy: no-referrer` causes the same issue with `Origin: null`, but I would assume it does, since the meta tag is just [another way of setting the same policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy#integration_with_html).  I've so far been unable to find consistent information about whether the header or the meta tag takes precedence if both are specified and conflict with one another.  If it were to turn out that the header takes precedence, then that would allow us to remove the `Origin: null` workaround I quoted above without having to change any apps.

I've tested this a bit on my local machine, including loading some feeds with images in tinytinyRSS and otherwise verifying that the response header is being set.

### DIFF
--- a/src/sandstorm/web-session-bridge.c++
+++ b/src/sandstorm/web-session-bridge.c++
@@ -77,6 +77,7 @@ WebSessionBridge::Tables::Tables(kj::HttpHeaderTable::Builder& headerTableBuilde
       hETag(headerTableBuilder.add("ETag")),
       hIfMatch(headerTableBuilder.add("If-Match")),
       hIfNoneMatch(headerTableBuilder.add("If-None-Match")),
+      hReferrerPolicy(headerTableBuilder.add("Referrer-Policy")),
       hSecWebSocketProtocol(headerTableBuilder.add("Sec-WebSocket-Protocol")),
       hVary(headerTableBuilder.add("Vary")),
       hXFrameOptions(headerTableBuilder.add("X-Frame-Options")),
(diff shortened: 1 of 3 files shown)


ocdtrekkie (approved):

src/sandstorm/web-session-bridge.c++
```diff
@@ -1315,6 +1316,13 @@ kj::Promise<void> WebSessionBridge::handleResponse(
       );
     }
 
+    // Set Referrer-Policy: same-origin. Otherwise, external sites linked
+    // from a grain can learn its randomized hostname from the Referer
+    // header. This knowledge could then potentially be used to launch an
+    // XSRF attack. We avoid using "no-referrer" here to prevent Chrome from
+    // sending "Origin: null" (see gateway.c++).
```
zenhack, line 1323: I suspect if the code in `gateway.c++` ever gets updated, there's like a 98% chance we'll forget to update this comment. I'm not sure how much value the last sentence is really adding. If we're going to mention this wrinkle here, I'd prefer to word it in a way where it's not likely to bitrot.

zenhack (commented):
LGTM, modulo one comment in-line.

I think once we merge this, it be worthwhile doing some testing to see what current browsers actually do re: header/tag conflicts.

src/sandstorm/web-session-bridge.c++
garrison: I've improved this such that the comment now references a commit id, which will not change.

zenhack (approved):

PR #3571, opened 2021-09-28
Vagrantfile improvements

garrison:
This patch series contains a number of improvements to the `Vagrantfile`.  Each commit contains a detailed description of the change.  Something I forgot to mention: One of the side effects of changing to a non-`contrib` image (1d50c1ec59a3f67e1d5b12abad5a99461c0217ec) is that `libvirt` is now available as a second virtualization provider.  My desire to get `libvirt` actually working is what led to 7b265d6426fc446ac066de4df2b05c6760fbf091.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -45,7 +45,7 @@ EOF
 
   # Use NFS for the /vagrant shared directory, for performance and
   # compatibility.
-  config.vm.synced_folder ".", "/vagrant", type: "nfs"
+  config.vm.synced_folder ".", "/vagrant", type: "nfs", nfs_udp: false
 
   # Calculate the number of CPUs and the amount of RAM the system has,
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
I think all of this seems reasonable. I also think all of these changes would probably be excellent and similarly applicable for our vagrant-spk Vagrantfile as well. (Though changing the base box to Bullseye needs to be held off on for various other fixes with vagrant-spk stacks, most likely.)

zenhack (approved):
These all look reasonable to me. Thanks!

garrison:
I just realized that the upgrade to bullseye broke NFS (only) under Virtualbox.  I've pushed one more commit to fix this.  I've tested this final version successfully on both libvirt and Virtualbox.

PR #3572, opened 2021-10-06
Debugging CI; don't mind me.

zenhack:
Do not merge; I'm trying to figure out why CI is failing and so need to
trigger it.


zenhack:
Looks like this is specifically a problem with the meteor 1.8.2 that we pull in for the fonts; if I leave the `?release=1.8.2` off of the meteor install, I don't hit that error (but then we need to deal with the font problem. re-running the script replaces the old install, so that doesn't solve it).

It looks like grub-webfont is completely unmaintained at this point. But the libraries that do the heavy lifting for it are another story; the easiest way through this might just be to write the relevant glue code ourselves, rather than relying on grub-webfont. I may give this a whack if I'm in the mood soonish.

griff:
See #3576

PR #3576, opened 2021-10-09
Fix the icon font generation

griff:
This switches to the maintained `grunt-webfonts` and sets options to fix the font generation by normalizing the icon sizes.

The font is not exactly the same as can be seen here:
image

Compared to the original font:
image

The problem is that the SVGs are not the same size and while the old version of `grunt-webfont` would normalize the sizes automatically. In the last versions of the package you had to specify explicitly that you wanted the icons normalized and the algorithm for doing that seems to have changed slightly.

If we want to preserve the font as it was we need to normalize the SVGs manually in an editor.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -289,17 +289,11 @@ tmp/.shell-env: tmp/.ekam-run $(IMAGES) shell/imports/client/changelog.html shel
 	@cd shell/ && PATH=$(METEOR_DEV_BUNDLE)/bin:$$PATH $(METEOR_DEV_BUNDLE)/bin/npm install
 	@touch tmp/.shell-env
 
-# grunt-webfont 0.4.8 does not work on Node 12. The latest version of grunt-webfont, OTOH, produces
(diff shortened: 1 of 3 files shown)


ocdtrekkie (approved):

garrison:
Assuming this is the agreed-upon solution, it will make sense to revert ddfb84fd724 and edit to `.github/workflows/build.yml` accordingly too.

griff:
@zenhack the biggest icon is `plus.svg` with a size of 128x128px so the "correct" size is at least that big.

`grunt-webfonts` defaults to a `fontHeight` of 512 but gives a warning when fontHeight is not at least 1000 when normalizing.

kentonv:
Yeah, I think all of the icons were made by Nena. Earlier on we had some icons from some open source icons project but I think we replaced them all.

OK, trusting @zenhack that this seems to work, gonna merge.

PR #3583, opened 2021-11-10
Make use of Pipe::make() helper from utils

zenhack:
While working on something else, I noticed that the `Pipe` struct from util.h is duplicated in the test suite, so I deleted the one from the test suite to use the one from the library. The second commit also makes use of this helper at a spot in run-bundle where it makes sense. There may be other spots that make sense too, but this was the most straightforward.


zenhack:
(There should be no functional change here, this is just cleanup)

PR #3591, opened 2021-11-28
Update doc: Installing from source now requires python3

garrison:
Recently, the clang update script [was changed](https://chromium.googlesource.com/chromium/src/tools/clang.git/+/5c0e867150ea239d3f0f0316370e71672c77b91a) to require python3, specifically.

I discovered this change when attempting to build Sandstorm using a [Dockerfile](https://gist.github.com/garrison/f025090ce914641f0d1e492612dcbd4e#file-dockerfile-debian) that worked previously.  I have verified that this change alone fixes the build under Debian; it seems that python2 is not also required (nor is `/usr/bin/python`, for that matter).

In Fedora, "python" already refers to python3, so no change was needed to that install line.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
             echo "The github runner seems to have some version of golang already; installing golang-go breaks."
             which go || sudo apt-get install golang-go
 
-            sudo apt-get install -y build-essential libcap-dev xz-utils zip unzip strace curl discount git python zlib1g-dev cmake ccache
+            sudo apt-get install -y build-essential libcap-dev xz-utils zip unzip strace curl discount git python3 zlib1g-dev cmake ccache
       - name: install meteor
         run: |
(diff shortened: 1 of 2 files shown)


ocdtrekkie:
Hmmm... Our CI installs `python` on Ubuntu. I'm definitely not opposed to updating this... but we should make sure our docs and our CI match here. Do you want to update this line as well, and see if we still pass tests: https://github.com/sandstorm-io/sandstorm/blob/master/.github/workflows/build.yml#L19

garrison:
Oh yes, of course CI should be updated to match. But it looks like the CI has continued to work all along because python3 is already in the base image:

> python3 is already the newest version (3.8.2-0ubuntu2).

ocdtrekkie:
I figured that was the case. I mostly just want the CI to verify it does not fail when we stop telling it to install python2. I will merge this when it completes. Thanks!

ocdtrekkie (approved):

Issue #3512, opened 2021-04-08
Grains terminate when fs.inotify.max_user_watches exceeded


jfrederickson:
I think I must have ended up with too many files in my grains, because I'd started to exceed the sysctl option `fs.inotify.max_user_watches` (which defaulted to 8192). This manifested as the grain I was working with terminating (with my nginx reverse proxy showing a 502 in the grain's iframe), and this showing up in the grain's debug logs:

```
2 3 4 5 6 7 8 | ** SANDSTORM SUPERVISOR: Starting up grain. Sandbox type: userns *** Uncaught exception *** sandstorm/supervisor.c++:232: overloaded: inotify_add_watch: No space left on device; path = sandbox/www_1611581607845905071/2010/06/the-open-order-of-lockpicks-toool-visits-hive stack: 420cf2 420775 420446 41f65d 415b66 414f04 41cc79 6da447 ** SANDSTORM SUPERVISOR: Starting up grain. Sandbox type: userns *** Uncaught exception *** sandstorm/supervisor.c++:232: overloaded: inotify_add_watch: No space left on device; path = sandbox/www_1611581607845905071/2010/06/summer-hackaton-2010 stack: 420cf2 420775 420446 41f65d 415b66 414f04 41cc79 6da447
-- | --
```

Increasing the value of the above sysctl key solved my problem. I'd been troubleshooting this in #sandstorm:

```
[7:19:06 pm] <isd> Yeah, that'd be the supervisor trying to keep track of how much storage a grain is using. Could you open an issue? Would be nice to handle this more gracefully.
```

zenhack:
Perhaps we should just set this ourselves on startup if it is below a certain reasonable value?

zenhack:
(It would be nice to find a more generally robust way to track space usage, but I don't know of one that wouldn't be major surgery. In general, I do not like the design of inotify).

zenhack:
(Fwiw, the value on my laptop is 524288, having not changed it. On my live sandstorm box it is 14280. I don't know how these values are normally chosen).

ocdtrekkie:
I did a bit of poking around, this topic has cropped up in various projects. This Stack Overflow felt valuable: https://stackoverflow.com/questions/535768/what-is-a-reasonable-amount-of-inotify-watches-with-linux

> I know from Lsyncd (maybe you want to check that out?) people who have a million watches. It just eats a Gigabyte of memory.

[...]

PR #3593, opened 2021-11-29
Docs: Provide fix for inotify limit issue

ocdtrekkie:
This adds an FAQ entry we can refer to about #3512 

I'm not happy with where in Docs it's located, but other troubleshooting pages are presently for new installs or app development. We should probably break a selection of FAQ entries out into a more general troubleshooting section, but the key part for me is that Docs is searchable and this places it somewhere it can be found and referred to.

I am not positive I correctly described the nature of the inotify setting or it's behavior from a "How Linux works" standpoint, and of course, any notes on writing style would be appreciated.

### DIFF
--- a/docs/administering/faq.md
+++ b/docs/administering/faq.md
@@ -642,4 +642,4 @@ In some cases, grains with a large number of files may fail to start, and the er
 
 You can check the setting by executing the following command: `cat /proc/sys/fs/inotify/max_user_watches`
 
-To increase this limit, add a line like `fs.inotify.max_user_watches=32768` to `/etc/sysctl.conf` and then reboot the Sandstorm server.
+To increase this limit, add a line like `fs.inotify.max_user_watches=32768` to `/etc/sysctl.conf` and then execute the command `sysctl -p` to load the configuration.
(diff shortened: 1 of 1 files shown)


docs/administering/faq.md
```diff
@@ -635,3 +635,11 @@ domain name that **should** be served from this Sandstorm server.
...
 
 This should show you the user's profile info including e-mail address, etc.
+
+## A very large grain fails to start. How can I fix this issue?
+
+In some cases, grains with a large number of files may fail to start, and the error `sandstorm/supervisor.c++:232: overloaded: inotify_add_watch: No space left on device;` is logged in the grain log. This is caused by a Linux setting which limits the maximum number of inotify watches a given user account (in this case, `sandstorm`) may request. Due to how Sandstorm monitors storage usage, this value acts as the limit of the number of files which can exist in simultaneously open grains. Starting with the Linux kernel 5.11, the default value of this setting is dynamically adjusted based on memory, but on older kernels, it may be lower than warranted. The previous default was 8,192.
+
+You can check the setting by executing the following command: `cat /proc/sys/fs/inotify/max_user_watches`
+
+To increase this limit, add a line like `fs.inotify.max_user_watches=32768` to `/etc/sysctl.conf` and then reboot the Sandstorm server.
```
zenhack, line 645: It should work to just do `sysctl -p` rather than rebooting the machine.

zenhack (commented):
One minor thing, otherwise this looks reasonable to me. I agree the location leaves something to be desired.

zenhack (approved):

PR #3594, opened 2021-12-01
Make the grain close button bigger on mobile.

zenhack:
...Filling the entire vertical space alotted for it.

I have found this difficult to press at its current size on mobile; this
essentially doubles the size, so it should be better.

Aesthetically it would be nice to also enlarge the grain icon for
symmetry, but just applying the same rule ends up with it overflowing
the bounds significantly.

I am trying to resist the temptation to redo this whole section with
flexbox & grid right now, but I might find time to actually do that later,
as I think it would make all this easier to reason about.

---

# Screenshots

## Before

Screen Shot 2021-12-01 at 17 37 58

## After

Screen Shot 2021-12-01 at 17 36 54

---

It looks the same on desktop as it did before.

### DIFF
--- a/shell/client/styles/_topbar.scss
+++ b/shell/client/styles/_topbar.scss
@@ -766,6 +766,10 @@ body>.topbar {
           right: 0px;
           width: 24px;
           height: 24px;
+          @media #{$mobile} {
+            height: 44px;
+            width: 44px;
+          }
           margin: 2px;
           display: block;
           background-image: url("/close.svg");
--- a/shell/client/styles/_topbar.scss
+++ b/shell/client/styles/_topbar.scss
@@ -741,6 +741,11 @@ body>.topbar {
           left: (($navbar-width-desktop-shrunk - 24px) / 2);
           width: 24px;
           height: 24px;
+          @media #{$mobile} {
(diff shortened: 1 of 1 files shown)


zenhack:
I messed with it a bit more, new after screenshot below. Things are properly symmetrical now.

Screen Shot 2021-12-01 at 17 51 51

zenhack:
I notice the gogs icon in particular is a bit stretched. Perhaps we should just center these but leave them small? Not sure what to do there.

shell/client/styles/_topbar.scss
```diff
@@ -741,6 +741,11 @@ body>.topbar {
           left: (($navbar-width-desktop-shrunk - 24px) / 2);
           width: 24px;
           height: 24px;
+          @media #{$mobile} {
+            left: (($navbar-width-desktop-shrunk - 40px) / 2);
```
zenhack, line 745: I don't quite understand the intended high-level meaning of this constant.

ocdtrekkie:
I agree this button can be irritating to click on mobile. Very +1 on the fix here.

shell/client/styles/_topbar.scss
kentonv: TBH your guess is as good as mine at this point...

PR #3597, opened 2021-12-13
Quiet TLS connection error log spam.

### LINKED ISSUE #3596: Suppress noisy OpenSSL errors in system log?
A large portion of my system log contains the below exceptions from the gateway. I am not sure these should be logged on a production server. I get the impression they represent failed connection attempts from browsers (or likely bots) which do not support a new enough TLS version. Public servers are often subjected to a deluge of random connection attempts: If we're confident these aren't "Sandstorm errors", I think it might make sense to stop logging them.

```
sandstorm/gateway.c++:1072: error: exception = kj/compat/tls.c++:66: failed: OpenSSL error; message = error:100000f7:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER
stack: 58f2a4 596986 596fd5 544fa1 597a60 591de0 5fe0c0
sandstorm/gateway.c++:1072: error: exception = kj/compat/tls.c++:66: failed: OpenSSL error; message = error:1000009c:SSL routines:OPENSSL_internal:HTTP_REQUEST
stack: 58f2a4 596986 596fd5 544fa1 597a60 591de0 5fe0c0
sandstorm/gateway.c++:1072: error: exception = kj/compat/tls.c++:66: failed: OpenSSL error; message = error:100000f7:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER
stack: 58f2a4 596986 596fd5 544fa1 597a60 591de0 5fe0c0
sandstorm/gateway.c++:1072: error: exception = kj/compat/tls.c++:66: failed: OpenSSL error; message = error:100000f7:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER
stack: 58f2a4 596986 596fd5 544fa1 597a60 591de0 5fe0c0
sandstorm/gateway.c++:1072: error: exception = kj/compat/tls.c++:66: failed: OpenSSL error; message = error:100000f7:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER
stack: 58f2a4 596986 596fd5 544fa1 597a60 591de0 5fe0c0
sandstorm/gateway.c++:1072: error: exception = kj/compat/tls.c++:66: failed: OpenSSL error; message = error:100000b8:SSL routines:OPENSSL_internal:NO_SHARED_CIPHER
stack: 58f2a4 596986 596fd5 544fa1 597a60 591de0 5fe0c0
sandstorm/gateway.c++:1072: error: exception = kj/compat/tls.c++:66: failed: OpenSSL error; message = error:100000f0:SSL routines:OPENSSL_internal:UNSUPPORTED_PROTOCOL
stack: 58f2a4 596986 596fd5 544fa1 597a60 591de0 5fe0c0
sandstorm/gateway.c++:1072: error: exception = kj/compat/tls.c++:66: failed: OpenSSL error; message = error:100000b8:SSL routines:OPENSSL_internal:NO_SHARED_CIPHER
stack: 58f2a4 596986 596fd5 544fa1 597a60 591de0 5fe0c0
sandstorm/gateway.c++:1072: error: exception = kj/compat/tls.c++:66: failed: OpenSSL error; message = error:100000f7:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER
stack: 58f2a4 596986 596fd5 544fa1 597a60 591de0 5fe0c0
sandstorm/gateway.c++:1072: error: exception = kj/compat/tls.c++:66: failed: OpenSSL error; message = error:100000f7:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER
stack: 58f2a4 596986 596fd5 544fa1 597a60 591de0 5fe0c0
sandstorm/gateway.c++:1072: error: exception = kj/compat/tls.c++:66: failed: OpenSSL error; message = error:100000f7:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER
stack: 58f2a4 596986 596fd5 544fa1 597a60 591de0 5fe0c0
```

zenhack:
Fixes #3596

---

Note, I haven't actually checked for a difference in log spam on a prod machine yet, though this seems straightforward -- info is not logged by default. If we wanted to, we could add a config option to enable verbose logging, but I'm inclined to punt that until it is deemed actually useful.

### DIFF
--- a/src/sandstorm/gateway.c++
+++ b/src/sandstorm/gateway.c++
@@ -1024,6 +1024,12 @@ kj::Promise<void> GatewayTlsManager::listenLoop(kj::ConnectionReceiver& port) {
       tasks.add(tls->tls.wrapServer(kj::mv(stream))
           .then([this](kj::Own<kj::AsyncIoStream>&& encrypted) {
         return server.listenHttp(kj::mv(encrypted));
+      }, [](kj::Exception&& e) {
+        // On a production server, this happens rather a lot, since random bots
+        // will try to connect, and (due to old TLS implementations) fail.
+        // If we just let this bubble up to our top-level error handler, it
+        // results in a lot of log spam, so instead log it at info level here:
+        KJ_LOG(INFO, "Failed to initialize TLS connection: ", e);
       }).attach(kj::mv(tls)));
     } else {
       KJ_LOG(ERROR, "refused HTTPS connection because no TLS keys are configured");
--- a/src/sandstorm/gateway.c++
+++ b/src/sandstorm/gateway.c++
@@ -1021,8 +1021,8 @@ kj::Promise<void> GatewayTlsManager::listenLoop(kj::ConnectionReceiver& port) {
   return port.accept().then([this, &port](kj::Own<kj::AsyncIoStream>&& stream) {
     KJ_IF_MAYBE(t, currentTls) {
       auto tls = kj::addRef(**t);
-      tasks.add(tls->tls.wrapServer(kj::mv(stream))
-          .then([this](kj::Own<kj::AsyncIoStream>&& encrypted) {
+      auto onTls = tls->tls.wrapServer(kj::mv(stream));
+      tasks.add(onTls.then([this](kj::Own<kj::AsyncIoStream>&& encrypted) {
         return server.listenHttp(kj::mv(encrypted));
       }, [](kj::Exception&& e) {
         // On a production server, this happens rather a lot, since random bots


zenhack:
(NB. the second commit is just formatting changes; I found the per-existing indentation of this code really hard to follow. Now it at least agrees with the AST. The functional change is all in the first commit).

kentonv:
Does it make sense to use `TlsContext.Option.acceptErrorHandler` to customize this? (Happy to accept "no" as an answer, just wanted to check if you'd considered it. This is a fairly new feature, actually.)

> (NB. the second commit is just formatting changes; I found the per-existing indentation of this code really hard to follow. Now it at least agrees with the AST. The functional change is all in the first commit).

The pre-existing style was standard KJ style for lambdas: We indent the body of the lambda exactly one stop more than the surrounding function body.

The abstract argument for this is that lambdas, especially when used with `.then()`, are more like control flow than like expressions. The fact that a lambda is technically an expression is a bit of a weird inconsistency in the language, and as a result the AST does not match the higher-level semantic understanding. Traditionally, in imperative languages like C++, there's a strong hierarchy that goes: global declarations -> statement blocks -> expressions. Lambdas screw with that by nesting statement blocks inside expressions.

As a practical matter, if we actually indent lambda bodies according to the AST, they tend to get excessively deeply indented too quickly in KJ async code.

In any case, I do in fact prefer the KJ style. It may take some getting used to, but I do think it is easier to read once it's familiar. When lambdas are nested deep inside expressions I now find it hard to see that control flow is occurring.

Perhaps a different way to make the code clearer without deviating from KJ style would be to assign the promise to a local variable, and then do `tasks.add(kj::mv(promise))` as a separate statement.

zenhack:
Ok, I went with factoring out the promise.

In the general case I wouldn't mind making use of async/await now that it exists in C++20, though I don't think kj::Promise is compatible with it yet?

zenhack:
Re: setting the error handler in the tls context: I hadn't considered it. Given that there's also an smtp proxy using this, we probably should; I'll find time to swap that in soonish, but probably not today.

zenhack:
/me actually greps. Ah, I see some integration did get added at some point. I will have to look into that in the future.

kentonv:
> I wouldn't mind making use of async/await now that it exists in C++20, though I don't think kj::Promise is compatible with it yet?

It is, actually! As long as you're building with `-std=c++20 -fcoroutines-ts`. We could totally turn that on for the Sandstorm build.

zenhack:
Good to know.

Hm, after actually testing this out on my prod server I noticed messages were still being logged at info level. Looks like we're actually turning up the log level on startup since capnproto exceptions don't get logged otherwise. So this approach isn't going to work. I'm going to close, will have to reconsider how to deal with this.

zenhack:
Fwiw, I also did try setting an error handler, with this patch:

```
diff --git a/src/sandstorm/gateway.c++ b/src/sandstorm/gateway.c++
index 36e113c9..b5361499 100644
--- a/src/sandstorm/gateway.c++
+++ b/src/sandstorm/gateway.c++
@@ -973,6 +973,13 @@ void GatewayTlsManager::setKeys(kj::StringPtr key, kj::StringPtr certChain) {
   kj::TlsContext::Options options;
   options.useSystemTrustStore = false;
   options.defaultKeypair = keypair;
+  options.acceptErrorHandler = kj::Function<void(kj::Exception&& e)>([](auto&& e) {
+    // On a production server, TLS errors happen rather a lot, since random bots
+    // will try to connect, and (due to old TLS implementations) fail. If we
+    // just let this bubble up to our top-level error handler, it results in a
+    // lot of log spam, so instead log it at info level here:
+    KJ_LOG(INFO, "Failed to initialize TLS connection: ", e);
+  });
 
   currentTls = kj::refcounted<RefcountedTlsContext>(kj::mv(options));
   readyFulfiller->fulfill();
```

...but the message still gets logged at error level. This is true even if I take the logging statement out of the body of that function entirely. So I'm not quite sure what needs to happen to get that to override the main handler.

Issue #3240, opened 2020-02-22
apps/web-publishing test is broken


ocdtrekkie:
Also tracking this test failure so we can get it cleaned up.

`<kentonv> and apps/web-publishing which has been broken ever since I rewrote the HTTP handling into C++`

ocdtrekkie:
I suspect this one is super simple? One assertion is failed, which may just be a different choice of language.

```
✖ Testing if element <#result> contains text: "Shutdown success".  - expected "Shutdown success" but got: "Success"
    at Object.<anonymous> (/home/runner/work/sandstorm/sandstorm/tests/apps/web-publishing.js:69:17)
```

kentonv:
It's not super-simple, or I would have fixed it at the time. :)

It's been a while since I looked at it, but IIRC the problem is that the test tries to verify that web publishing still works while the grain is not running. But it was using some trick to verify that the grain was not running which doesn't work anymore -- with the new gateway code, whatever it was doing actually has the effect of starting the grain back up. Which is actually a good thing in general, but it poses a challenge for this test specifically. At the time I couldn't think of a simple way to change the test such that it actually tested what it was supposed to be testing, but I didn't want to delete the test because I felt that it the thing it was trying to test was actually important...

zenhack:
Picking through this, it seems that the problem is that `getUiBridge()` wraps the capability for the app in a `CapRedirector` that restarts it on disconnect, and because the app just does an `exit(0)` in its request handler, it's going to disconnect before returning.

The comments on `CapRedirector` suggest that it was originally intended to do this only for the sandstorm frontend, rather than grains, so I assume what changed when you wrote the gateway is that we started using this there too, whereas the old js version would just let the grain stay dead.

Since the CapRedirector only detects disconnects when making method calls, It seems like if we somehow deferred the exit() until after that method has returned, then it shouldn't get restarted unless another request is made to the grain. I will have to try this.

Issue #3239, opened 2020-02-22
tests/account test is broken


ocdtrekkie:
Important that we don't lose these, so making tracking issues, so we can get the tests running clean again.
`<kentonv> Also broken, but my fault: tests/account because I removed the ability to upgrade demo accounts,`

zenhack:
I tried just patching out the bit at the beginning that upgrades a demo account, and instead logs in to a fresh dev account:

```diff
diff --git a/tests/tests/account.js b/tests/tests/account.js
index 538221d5..109bf240 100644
--- a/tests/tests/account.js
+++ b/tests/tests/account.js
@@ -37,21 +37,15 @@ module.exports["Test link credentials"] = function (browser) {
   browser
     .init()
 
-    // Upgrade a demo account to a real account by linking an credential.
-    .url(browser.launch_url + "/demo")
-    .disableGuidedTour()
-    .waitForElementVisible(".demo-startup-modal .start", medium_wait)
-    .click(".demo-startup-modal .start")
-    .disableGuidedTour()
-    .waitForElementPresent(".main-content>.app-list", medium_wait)
-    .click(".login>button.show-popup")
+    // Sign up with a new dev account.
+    .url(browser.launch_url)
     .waitForElementVisible(".login-buttons-list", short_wait)
     .click(".login-buttons-list button.dev")
     .waitForElementVisible("input[name=name]", short_wait)
     .setValue("input[name=name]", devName1)
     .submitForm(".login-buttons-list form.dev")
     .waitForElementVisible("form.account-profile-editor", short_wait) // confirm profile
-    // Profile defaulted to new credential's name, not "Demo User".
+    // Profile defaulted to new credential's name.
     .assert.attributeContains("form.account-profile-editor input[name=nameInput]", "value", devName1)
     .submitForm("form.account-profile-editor")
     .execute(function () { return Meteor.user().profile.identicon; }, [], function (response) {
```

But, in this case the error message re: linking an in-use credential to another account never appears -- it goes through. Experimenting manually shows that this only fails for dev accounts entered in the "custom" box; if you try to log in as dave and then link alice it fails.

PR #3603, opened 2022-01-25
Rework the way updates work to fix #3584

### LINKED ISSUE #3584: Permissions error causing test failures
I finally spent a bit of time debugging the batch of new (~19) test failures we started seeing a month or two ago. For reference, I've attached a full log from `make test TESTCASE=tests/backup-restore.js`, which is one of the relevant tests; the many of the other tests fail with the same error.

The interesting bit is this, error, which comes when setting up the sandbox for grains during test:

```
kj/io.c++:365: failed: miniposix::write(fd, pos, size): Operation not permitted; fd = 3
```

The stack trace pointed to this call:

https://github.com/sandstorm-io/sandstorm/blob/0d53a6646ab99c2cd4e54d2fc2d04d69f1e18366/src/sandstorm/supervisor.c%2B%2B#L936

The error results in grains failing to start, and therefore causing the overall test to fail. I haven't ever seen this happen on an actual install (even a dev one), only when running the test suite.

I suspect something changed in the kernel that is affecting this (nothing changed on our end), but I don't really understand why this fails, or exactly what the relevant difference is between test & prod.

I would be interested to hear from folks wrt:

- Whether you can reproduce the error on your system.
- What distro & kernel version you're running.

Probably the next step is to see if we can an find a system that still works, and see if we can get it to fail on a newer kernel, and if so bisect.

[t1.log](https://github.com/sandstorm-io/sandstorm/files/7535005/t1.log)

zenhack:
Rather than having the update monitor itself exec the newly downloaded
version of sandstorm, it simply exits (with status 0), and its parent
process takes this as an indictation that it should re-spawn it (with
the new version). This avoids the need to map ourselves to UID 0 in the
user namespace, which avoids the EPERM error we were getting when
later hiding the real UID from a grain.

Per discussion on the issue, the old design no longer works as of Linux
5.12, after which spawning the grain we would need CAP_SETFCAP, which we
do not want to have.

---

This seems to basically work, but I'm marking it as a draft for now for a couple reasons:

- [x] It touches the update monitor, so I want to test it very thoroughly before merging; an update that broke updates would be very bad.
- [x] There are still *some* test failures, and while I expected that (there were known failures before we started having this problem) I want to stare at them a bit to make sure I understand what's going on.
- [x] There is a minor issue where when stopping sandstorm, a message is spuriously printed stating that sandstorm exited abnormally, even when it did not. Doesn't really cause problems, but I should fix it.

### DIFF
--- a/src/sandstorm/run-bundle.c++
+++ b/src/sandstorm/run-bundle.c++
@@ -536,44 +536,21 @@ public:
       KJ_SYSCALL(ftruncate(pidfile, 0));
     }
 
-    if (!runningAsRoot) unshareUidNamespaceOnce();
-
-    // Unshare PID namespace so that daemon process becomes the root process of its own PID
-    // namespace and therefore if it dies the whole namespace is killed.
-    KJ_SYSCALL(unshare(CLONE_NEWPID));
-
-    // Daemonize ourselves.
-    pid_t mainPid;  // PID of the main process as seen *outside* the PID namespace.
-    {
-      auto pipe = Pipe::make();
-
-      KJ_SYSCALL(mainPid = fork());
-      if (mainPid != 0) {
-        // Tell the child process its own PID, since being in a PID namespace its own getpid() will
-        // unhelpfully return 1.
-        pipe.readEnd = nullptr;
-        kj::FdOutputStream(kj::mv(pipe.writeEnd)).write(&mainPid, sizeof(mainPid));
-
-        // Write the pidfile before exiting.
-        {
-          auto pidstr = kj::str(mainPid, '\n');
-          kj::FdOutputStream((int)pidfile).write(pidstr.begin(), pidstr.size());
-        }
-
-        // Exit success.
-        context.exitInfo(kj::str("Sandstorm started. PID = ", mainPid));
-        return true;
+    // Daemonize ourselves:
+    pid_t mainPid;
+    KJ_SYSCALL(mainPid = fork());
+    if(mainPid != 0) {
+      // Write the pidfile before exiting.
+      {
+        auto pidstr = kj::str(mainPid, '\n');
+        kj::FdOutputStream((int)pidfile).write(pidstr.begin(), pidstr.size());
       }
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
These (found in your Action results) existed prior to the large number of account failures, so I think we are good there, at least in getting back to where we used to be.

 ✖ tests/account #3239 
 ✖ apps/web-publishing #3240 

You can see there are two other tests which also fail intermittently, all tracked here: https://github.com/sandstorm-io/sandstorm/milestone/19

zenhack:
Ok, yeah, it looks like we're good then. I was still seeing the appHooks failure locally and I couldn't remember if that one was one of the existing ones.

zenhack:
Might make sense to add this/#3584 to that milestone.

ocdtrekkie:
Done.

zenhack:
I can't seem to reproduce the error on stop anymore; maybe I fixed it and forgot, or maybe I was running a stale version when I last tested. In any case, it appears to work.

I still want to do a bit more testing; I patched the URLs used for updates and checked a bunch of stuff manually, both for a system wide install and just running the test configuration as an unprivileged user:

- The manual `sandstorm update` command works as expected.
- We still check for and download updates 10m after startup
- We check for an update after an hour if a previous check fails.

I still want to check:

- [ ] A valid, correctly signed update goes all the way through.
- [ ] Afterwards, update checks happen after 24 hours.

I'll need to futz around with the signing code to check this, and then it's mostly a matter of backgrounding it and checking back.

I'm being way more careful about this patch than I normally would be; I'm pretty confident it's fine, but breaking automatic updates would be very very bad, so I'm being really pedantic about this.

At some point I'd like to have some actual automated tests for this stuff, but that will have to wait.

zenhack:
Alright, finished with the above checks and everything seems to work; marking read for review.

zenhack:
(2) does seem much simpler. Implemented in #3609, closing.

PR #3604, opened 2022-01-26
Use Custom App Market URL as shown in href

gischer:
As of the prior patch, when using a custom app-market URL, the URL in the href of this <a> tag is correct, so we should just use it under the DRY principle. This requires setting
the target attribute to "_blank" to preserve current behavior of opening the
link in a new tab.

This addresses #3029

### DIFF
--- a/shell/imports/client/apps/applist-client.js
+++ b/shell/imports/client/apps/applist-client.js
@@ -199,19 +199,6 @@ Template.sandstormAppListPage.helpers({
 });
 
 Template.sandstormAppListPage.events({
-  "click .install-button": function (event) {
-    event.preventDefault();
-    event.stopPropagation();
-    const ref = Template.instance().data;
-    const appMarket = ref._db.collections.settings.findOne({ _id: "appMarketUrl" });
-    let url = "https://apps.sandstorm.io/?host=";
-    if (appMarket) {
-      url = appMarket.value + '?host=';
-    }
-    window.open(url +
-        document.location.protocol + "//" + document.location.host, "_blank");
-  },
-
   "click .upload-button": function (event, instance) {
     const input = instance.find(".upload-button input");
     if (input == event.target) {
--- a/shell/imports/client/apps/applist-client.js
+++ b/shell/imports/client/apps/applist-client.js
@@ -213,6 +213,14 @@ Template.sandstormAppListPage.events({
     });
   },
 
+  "click .install-button": function(event) {
+    event.preventDefault();
+    event.stopPropagation();
+    const url = event.currentTarget.getAttribute("href");
+    window.open(url +
+          document.location.protocol + "//" + document.location.host, "_blank");
+  },
+
   "click .uninstall-action": function (event) {
     const db = Template.instance().data._db;
     const appId = this.appId;
--- a/shell/imports/client/apps/applist.html
+++ b/shell/imports/client/apps/applist.html
@@ -43,7 +43,7 @@
     <h2 class="all-apps">{{_ "apps.appList.allApps"}}</h2>
     {{/if}}
 
-    <a class="app-button install-button {{#if searching}}hide{{/if}}" href="{{appMarketUrl}}">
+    <a class="app-button install-button {{#if searching}}hide{{/if}}" href="{{appMarketUrl}}" target="_blank">
       <div class="app-icon pseudoapp install-icon"></div>
       <span class="action-title">{{_ "apps.appList.install"}}</span>
       <span class="action-text">{{_ "apps.appList.fromMarket"}}</span>


ocdtrekkie:
I am a bit confused why the duplication exists here in the first place. I'd love to know exactly why it originally was built this way, and if there's any other functionality we somehow will break by making this a simple web link.

zenhack:
Maybe @zarvox or @kentonv knows. This looks ok to me though. (although I haven't tested it).

ocdtrekkie:
@gischer may be interested in upgrading his Sandstorm to use this PR, depending how urgent the fix is, the CI created a bundle here: https://github.com/sandstorm-io/sandstorm/suites/5075275864/artifacts/151084581

---

If you're willing to test, once the build is finished running you can install it by:

  - Download sandstorm-0.tar.xz from the Checks tab on this pr.
  - edit your /opt/sandstorm/sandstorm.conf, setting UPDATE_CHANNEL=none
  - run sandstorm update sandstorm-0.tar.xz.

When you're ready to go back to using the mainline releases, you'll want to undo the change to sandstorm.conf in order to turn auto-update back on.

gischer:
My best guess is that the html in question once upon a time used buttons 
<button> rather than links <a>.  For style reasons, they were changed to 
links.   With my last PR, I filled in the href of the link correctly.

Meanwhile minimal changes were made to the code that handles the button 
click, just transferring it to an anchor click.

Mind you, this is just a guess, but there are things in the code that 
call it a “button” that support that guess.

I will look into upgrading using the new bundle.

gischer:
Ok, I just downloaded and tested the build above and the App Market  
“button” (which is really a link) works in default mode and works with a 
custom App Market, too.

------ Original Message ------
From: "Jay L Gischer" ***@***.***>
To: "sandstorm-io/sandstorm" 
***@***.***>; 
"sandstorm-io/sandstorm" ***@***.***>
Cc: "Mention" ***@***.***>
Sent: 1/27/2022 9:46:54 AM
Subject: Re[2]: [sandstorm-io/sandstorm] Use Custom App Market URL as 
shown in href (PR #3604)

kentonv:
The manual use of `window.open()` might have had something to do with wanting the new tab to have `window.opener` point back to the Sandstorm tab. We take advantage of that such than when you click to install an app, we can actually close the new tab and instruct the original tab to perform the installation, rather than end up with multiple Sandstorm tabs.

Does that still work when using `target="_blank_"`?

gischer:
Hmm.  Well, I just found this:  
https://stackoverflow.com/questions/15670252/does-window-opener-work-with-target-blank-anchor-tags/15707103 
  which suggests also setting rel='opener' on the same anchor tag, then 
it works on all browsers.

OR, I could just put back in the action, grab the href value and use it 
in a window.open call, and avoid the recalculation, which seems a 
problem.

-j

------ Original Message ------

gischer:
I just pushed an update that has an explicit event that grabs the current target and uses its href attribute in a call to `window.open`  This should maximize browser compatibility, while keeping things very DRY.

ocdtrekkie:
That makes sense to me, at least. I will mention the nitpick that when you added the event back, you put it in a different spot, so the diff doesn't work as well as it could/should.

shell/imports/client/apps/applist-client.js
```diff
@@ -226,6 +213,14 @@ Template.sandstormAppListPage.events({
     });
   },
 
+  "click .install-button": function(event) {
+    event.preventDefault();
+    event.stopPropagation();
+    const url = event.currentTarget.getAttribute("href");
+    window.open(url +
```
kentonv, line 220: Don't you need to add `?host=` in here?

kentonv (commented):
Sorry for the slow response, I tend to look at code reviews only when I'm putting together the monthly release. :/

I'm a little unclear on the purpose of this change. Is this fixing a bug, or is it just a cleanup?

Have you been able to test the change locally?

shell/imports/client/apps/applist-client.js
ocdtrekkie: I don't think so, I think he is grabbing the URL wholesale from the HTML side of this.

ocdtrekkie:
My understanding is the event failed to work for custom app markets prior to this patch. He did download the test build built by GitHub Actions and said that fixed the issue.

shell/imports/client/apps/applist-client.js
kentonv: OK, then isn't this adding a second copy of the Sandstorm server URL, since the `?host=https://example.com` is already in there? You'd end up with `?host=https://example.comhttps://example.com`...

kentonv:
Sorry again for being very slow to reply. Feel free to ping me on IRC for faster turn-arounds.

I'm a bit confused here as the button seems to be working fine for me, without this change. When I set a custom app market URL and then click on the install button in the apps list, I get sent to the right place. @gischer it seems like you're seeing something different? Is it possible you accidentally tested on a version that was missing your fixes from #3568? I can't see how it would end up opening the wrong URL after that change, and I don't understand how this change would make a difference. I'd like to make sure we understand the bug we're fixing here, and not just trying things until it "works".

gischer:
I just did some testing.  There are more mysteries.

My installation is mansia.net  I am running v0.287 (2021-09-04). I have the App Market URL set to "https://app-market.mansia.net"  When I mouse over the "Install from App Market" button, it says "https://app-market.mansia.net"  When I click on this button, it takes me to "apps.sandstorm.io".  This is precisely the behavior described in #3029 

AND, when I check "https://app-market.mansia.net" I see that it does not support HTTPS, only HTTP, which is clearly something I need to address. 

My current thesis is that Sandstorm notices this and sends you to the default when it happens, though I certainly didn't notice any code for that.  Can that possibly be what's happening?

I'm going to go get a cert and add HTTPS capability to my app-market server.  With positive confirmation that Sandstorm switches, I'll drop this patch, and recommend #3029 be closed as fixed.

shell/imports/client/apps/applist-client.js
ocdtrekkie: Ahhhh, I see what you mean now. Yeah, this would probably cause that bug here.

gischer:
Ok, further investigation provides yet more mysteries.

I had a problem with my own app market site.  I resolved those problems.  You can see it at (https://app-market.mansia.net)

On my installation of Sandstorm, I have it set as the custom app market.  The correct URL shows in the rollover.  And yet, when I click that button it sends me to apps.sandstorm.io

ocdtrekkie:
I can't replicate that issue when setting your app market URL as my custom app market.

zenhack:
Commit logs indicate @kentonv merged your previous patch on 2021-10-23, which is after the release you said you're running (I take it you have automatic updates disabled?)

Given that, the obvious question is: does it work if you update to the latest sandstorm version?

ocdtrekkie:
Whoa, good catch, @zenhack. I saw $VERSION and *assumed* it was $LATEST_VERSION. But we are on 0.296, not 0.287.

gischer:
Huh.

The relevant line in my sandstorm.conf is

UPDATE_CHANNEL=dev

zenhack:
hm, if you manually run `sandstorm update` from the cli, what happens?

gischer:
Curiouser and curiouser

```
myprompt:~$ sandstorm update
Checking for updates on channel dev...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (60) server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none
More details here: http://curl.haxx.se/docs/sslcerts.html
curl performs SSL certificate verification by default, using a "bundle"
 of Certificate Authority (CA) public keys (CA certs). If the default
 bundle file isn't adequate, you can specify an alternate file
 using the --cacert option.
If this HTTPS server uses a certificate signed by a CA represented in
 the bundle, the certificate verification probably failed due to a
 problem with the certificate (it might be expired, or the name might
 not match the domain name in the URL).
If you'd like to turn off curl's verification of the certificate, use
 the -k (or --insecure) option.
*** Uncaught exception ***
sandstorm/run-bundle.c++:299: failed: curl failed; url = https://install.sandstorm.io/dev?from=287&type=manual; exi
tCode = 60
stack: 618fda 617d9b 63128e 6311ac 50665b 50624b
```

ocdtrekkie:
Is your curl perhaps out of date? Maybe the Sandstorm.io cert requires newer immediates or roots than your curl understands? It seems like there's a URL to download the updated certificate bundle.

gischer:
Ok.  My installed curl was the latest version (7.47.0) available for Ubuntu 16.04  (But as we see, not recent enough!)

My solution, cribbed from Ubuntu forums, was to get a statically linked curl that is more recent and use that.

This worked.  I updated Sandstorm.  I tested the App Market  button/link, and it works correctly now.

Feel free to ignore this PR, and close #3029

zenhack:
Glad we were able to help sort this out.

PR #3605, opened 2022-01-29
Fix a bug in the update tool's key generation.

zenhack:
I encountered this while testing #3603; as written, the `update-tool add`
subcommand does not print out the correct public key for the keypair
that it generates; instead, it prints out the seed used to generate the
keypair. This patch changes the code to actually generate the keypair
from the seed, and then print out the public key that should be added
to update-tool.capnp

I assume this code was not used to generate the current contents of that
file, since if it was loading the keyring would fail (which is the
problem I encountered).


kentonv:
weird

PR #3606, opened 2022-01-31
Fix web-publishing test

zenhack:
N.B. this includes the commits from #3603, which should be reviewed & merged first.

This fixes the `apps/web-publishing.js` test that has been failing since forever.

It folds the functionality from the https://github.com/jparyani/sandstorm-test-app/tree/web-publishing into the test app we maintain in-tree, and then implements the solution I suggested [here](https://github.com/sandstorm-io/sandstorm/issues/3240#issuecomment-1024863154), i.e. defer the exit() until the return message has been sent.

### DIFF
--- a/src/sandstorm/test-app/test-app.c++
+++ b/src/sandstorm/test-app/test-app.c++
@@ -137,10 +137,12 @@ public:
                  sandstorm::SessionContext::Client context,
                  sandstorm::WebSession::Params::Reader params,
                  kj::Promise<sandstorm::SandstormApi<>::Client>& api,
+                 capnp::TwoPartyVatNetwork::Connection& conn,
                  bool isPowerboxRequest = false)
       : isPowerboxRequest(isPowerboxRequest),
         sessionContext(kj::mv(context)),
-        api(api) {}
+        api(api),
+        conn(conn) {}
 
   kj::Promise<void> get(GetContext context) override {
     // HTTP GET request.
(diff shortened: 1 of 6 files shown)


zenhack:
Rebased this on top of #3609, though I'm still a bit stumped re: what to do next on that one. Conceptually this could be merged separately (if I rebased on top of master), but there's not much point given it's fixing a test and the whole suite is failing still.

zenhack:
Rebased again, now on top of the version of the other pr that actually works.

src/sandstorm/test-app/test-app.c++
```diff
@@ -151,6 +154,47 @@ public:
...
+      content.setMimeType("text/plain");
+
+      // After sending the return message, shut down the connection and then
+      // exit.
+      //
+      // We need to wait for the return message to actually be sent,
+      // otherwise the supervisor will try to re-connect in order to retry
+      // the method call, which will start us back up again.
+      //
+      // Note that we need to call evalLater *twice* to make this happen,
```
kentonv, line 173: Is this a good use case for `evalLast()`?

zenhack: Good call, didn't know about that one. Done, and rebased on top of master.

zenhack:
Hm, I got another failure for this on #3616 (which builds on this branch). But it didn't fail CI here, and I can't reproduce it locally...

PR #3614, opened 2022-04-03
WIP: make upgrading demo accounts available via a config option.

### LINKED ISSUE #3239: tests/account test is broken

zenhack:
This fixes #3239 by providing a config option to re-enable the ability
to upgrade demo accounts to full accounts, which was removed in order to
allow alpha to act as the demo server when oasis was shut down.

Unfortunately, this seems to break *other* tests. The handful I examined
resulted in permission errors trying to upload the test app; presumably
when this is enabled our test accounts are somehow getting marked as
not allowed to upload their own apps.

---

Accordingly, I'm marking this as a draft until that can be debugged.

### DIFF
--- a/src/sandstorm/config.c++
+++ b/src/sandstorm/config.c++
@@ -241,6 +241,8 @@ Config readConfig(const char *path, bool parseUids) {
       config.allowDemoAccounts = value == "true" || value == "yes";
     } else if (key == "ALLOW_DEV_ACCOUNTS") {
       config.allowDevAccounts = value == "true" || value == "yes";
+    } else if (key == "ALLOW_UNINVITED") {
+      config.allowUninvited = value == "true" || value == "yes";
     } else if (key == "IS_TESTING") {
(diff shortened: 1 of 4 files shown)


zenhack:
(Note that this includes the commits from #3609, since those are necessary for *any* of the tests to pass).

ocdtrekkie:
Does ALLOW_UNINVITED permit users who sign in without an invite to create grains? I am not sure if the config option is self-explanatory enough. How does this interact with the permission to allow demo accounts, and what happens if one but not the other is set?

...Which is to say, I maybe might suggest a better config variable name, and definitely would like to see the docs updated in the PR.

zenhack:
Yeah, I should definitely update the docs before this is merged, and I just copied the name from the meteor setting, so could probably be improved.

PR #3616, opened 2022-04-10
Mark known-failing tests as "expected to fail," disabling them by default. 

zenhack:
I'd like to actually *fix* these eventually, but in the meantime I'd like not to have to go look at the CI logs manually to know if we're OK. So, this patch:

- generalizes the code used to disable some demo related tests from *way* back, to cover the other outstanding test failures as well.
- documents this feature of the test suite.
- no longer permits test failures in CI.

This also includes the fix from #3606, since otherwise I'd have to disable that test as well, which seems silly. That PR should be merged first.


zenhack:
N.B. I'm having some problems building meteor-testapp.spk locally (the resulting app segfaults on startup) but they seem to be unrelated to this patch, as I can reproduce them on master. The upshot though is I'm waiting on CI to confirm that everything actually passes...

ocdtrekkie:
So this one looks like it still failed just the web publishing test, despite the web publishing test fix being included?

kentonv:
Needs conflict resolution.

zenhack:
Fixed the conflicts (it actually merged cleanly when I did it locally; weird).

ocdtrekkie:
Is this something we could still merge in even if v.299 is already published (@kentonv git push), as it's not production code?

PR #3622, opened 2022-05-20
Allow multiple values for HTTPS_PORT

### LINKED ISSUE #3619: Allow listening on multiple HTTPS_PORTs
So the first port vs. alternate port behavior for HTTP_PORT is documented here: https://docs.sandstorm.io/en/latest/administering/config-file/ and HTTP_PORT can be specified as a comma-separated list. However, HTTPS_PORT is specified as a single port, and specifies that the HTTPS_PORT will be treated as the first port if enabled. However, there is no support for HTTPS alternate ports.

The problem is that since Oasis shut down, I have operated my Sandcats server on a nonstandard port (6080, but using HTTPS), because my ISP had an explicit terms of use prohibiting exposing any server software on a well-known port (such as 443). However, that ISP recently sold my geographic area to another ISP with different terms which do not prohibit using well-known ports. Moving my server would solve a lot of problems, like letting me host websites cached via Cloudflare (and other external services which expect normal ports for things). However, that would break all of my existing HTTPS sharing links and API keys out in the world, which would be very sad and frustrating.

zenhack:
Fixes #3619

---

@ocdtrekkie, feel like testing? I have verified it doesn't break my box, but haven't actually tried adding a second port, mainly because I'd have to set up port forwarding and it's late enough that I'm wanting to put this down for the night.

### DIFF
--- a/src/sandstorm/run-bundle.c++
+++ b/src/sandstorm/run-bundle.c++
@@ -2509,15 +2509,22 @@ private:
             KJ_FAIL_REQUIRE("backend died; gateway aborting too");
           }));
 
+      auto isHttpsPort = [&](uint port) -> bool {
+        for(uint httpsPort : config.httpsPorts) {
+          if (httpsPort == port) {
+            return true;
+          }
+        }
+        return false;
+      };
+
       // Listen on main port.
       if (config.ports.size() > 0) {
         auto port = config.ports[0];
         auto listener = fdBundle.consume(port, *io.lowLevelProvider);
-        bool isHttps = false;
-        KJ_IF_MAYBE(p, config.httpsPort) {
-          isHttps = port == *p;
-        }
-        auto promise = isHttps ? tlsManager.listenHttps(*listener) : server.listenHttp(*listener);
+        auto promise = isHttpsPort(port)
+          ? tlsManager.listenHttps(*listener)
+          : server.listenHttp(*listener);
         promises = promises.exclusiveJoin(promise.attach(kj::mv(listener)));
       }
 
@@ -2530,7 +2537,9 @@ private:
         altPortServer = altPortServer.attach(kj::mv(altPortService));
         for (auto port: config.ports.slice(1, config.ports.size())) {
           auto listener = fdBundle.consume(port, *io.lowLevelProvider);
-          auto promise = altPortServer->listenHttp(*listener);
+          auto promise = isHttpsPort(port)
+            ? tlsManager.listenHttps(*listener)
+            : altPortServer->listenHttp(*listener);
           promises = promises.exclusiveJoin(promise.attach(kj::mv(listener)));
         }
         promises = promises.attach(kj::mv(altPortServer));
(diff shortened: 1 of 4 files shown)


src/sandstorm/run-bundle.c++
```diff
@@ -2611,10 +2620,6 @@ private:
     KJ_SYSCALL(setenv("HTTP_GATEWAY", "local", true));
 
     KJ_SYSCALL(setenv("PORT", kj::str(config.ports[0]).cStr(), true));
-    KJ_IF_MAYBE(httpsPort, config.httpsPort) {
-      // TODO(cleanup): At this point, all this does is tell Sandcats to refresh certs.
```
zenhack, line 2615: I'm fairly sure at this point this doesn't do *anything*; the sandcats logic doesn't look for this variable, and grepping around for it didn't come up with any uses.

ocdtrekkie:
I will definitely test this weekend (partially because if this works I do not want to wait for a release). Running a server backup now, which I always do before updating to experimental code.

ocdtrekkie:
Okay, so, we are close, because Sandstorm is definitely listening on the port! However, it looks like that host name recognition behavior may look at ports, because I get this (I have 443 as an HTTPS alternate port for now):

image

zenhack:
Interesting. And this doesn't happen with an extra port added to `PORT`?

I will have to actually test this myself soonish.

ocdtrekkie:
I don't have any HTTP ports exposed to the outside world, so I don't know, to be honest. I'd probably go for setting up a VM to play with it if I want to start opening those.

What I did test (because I was getting the unrecognized hostname on my alternate port) was trying to set up a static web publishing on the alternate port, and that also did not work.

src/sandstorm/run-bundle.c++
```diff
@@ -2530,7 +2537,9 @@ private:
         altPortServer = altPortServer.attach(kj::mv(altPortService));
         for (auto port: config.ports.slice(1, config.ports.size())) {
           auto listener = fdBundle.consume(port, *io.lowLevelProvider);
-          auto promise = altPortServer->listenHttp(*listener);
+          auto promise = isHttpsPort(port)
+            ? tlsManager.listenHttps(*listener)
+            : altPortServer->listenHttp(*listener);
```
ocdtrekkie, line 2542: Could the issue be here? If I can parse this, it means the altPortServer is only being reached if it's an HTTP port?

zenhack:
Good catch. Just pushed an attempt to fix this; give it a try when the
build finishes.

ocdtrekkie:
This definitely worked for the main URL, but it appears to not work for anything else.

https://ocdhost.sandcats.io/shared/X8HYTaml-nrKIHORyo-C9AwuCzZ0FUoT_-OEXHUw10f even has the same exact hostname, as the main URL, but it gives the unrecognized hostname error. Web publishing didn't work either.

zenhack:
Ok. I'm going to mark this as a draft for now; at some point I'll set up port forwarding myself and mess with it so we can do fewer round trips.

PR #3623, opened 2022-05-20
Docs: Update graphviz extension

ocdtrekkie:
The author of this mkdocs extension renamed it and relicensed it as GPLv3 from MIT.

https://github.com/rodrigoschwencke/mkdocs-markdown-graphviz to
https://gitlab.com/rodrigo.schwencke/mkdocs-graphviz

And he took down the original, which is why the Docs CI check on Ian's PR failed.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,2 +1,2 @@
 mkdocs==1.0.4
-mkdocs-markdown-graphviz==1.3
+mkdocs-graphviz==1.4.4
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,2 +1,3 @@
+jinja2==3.0.3
 mkdocs==1.0.4
-mkdocs-graphviz==1.4.4
+mkdocs-graphviz==1.3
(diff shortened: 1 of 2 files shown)


ocdtrekkie:
So also, a dependency of mkdocs also broke before version 1.2.4, unless we pin it to an older version, so I pinned that for now too: https://github.com/mkdocs/mkdocs/issues/2799

We probably should "just" update mkdocs to the latest, but I forget why we didn't before.

ocdtrekkie:
I am starting to think this graphviz chart thing is not worth the price of admission.

Test build:
image

Production:
image

ocdtrekkie:
Now I'm just annoyed. So I pulled back the mkdocs_graphviz to 1.3, which should be the same code as it was before the rename, but it's still broken in the test. And the jinja2 dependency was for mkdocs, not mkdocs_graphviz, so it seems unlikely that should've been it, so I am a bit baffled here.

ocdtrekkie:
So like... for the two graphs on the how it works page, is it worth the burden of constantly fighting the bitrot of this extension? Or should I plaster the finished SVGs from the current docs page into the page, remove the extension entirely, and maybe leave the graphviz language in a code comment somewhere so people can regenerate them outside of the docs if needed?

zenhack:
I'm fine with gutting it.

ocdtrekkie:
Superseded by #3647

PR #3635, opened 2022-06-11
Fix typescript errors

zenhack:
This fixes two errors currently being flagged by tsc in CI: https://github.com/sandstorm-io/sandstorm/runs/6845088479?check_suite_focus=true#step:10:12

The first of these we just ignore; I previously tried to follow the suggestion and change the extension to `.js`, but that seems to break the build, at least for @kentonv.

The second we just add a type annotation.

### DIFF
--- a/shell/imports/client/backups.ts
+++ b/shell/imports/client/backups.ts
@@ -1,4 +1,6 @@
 import downloadFile from '/imports/client/download-file.js';
+
+// @ts-ignore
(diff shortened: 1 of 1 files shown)


zenhack:
Superseded by #3636; that seems to fix the second error without adding the type annotation, and doesn't risk silencing real issues by mistake. I tested the built bundle carefully this time; I thought I had before but that broke the build for @kentonv so :shrug:. Hopefully this one will work better.

PR #3637, opened 2022-06-26
Apply backpressure when installing packages.

### LINKED ISSUE #3429: Excessive memory usage when downloading base apps
I've noticed that just after an initial install, when sandstorm is downloading the built-in apps, memory usage is very high; I was testing something on a VPS with 1 GiB of memory and with nothing else running on the box (and no grains) it was swapping.

I have a hunch this is because we're failing to apply back-pressure in some places and as a result buffering too much of the packages in memory, but would have to dig in a bit more to be certain.

zenhack:
This should fix #3429, and when installing apps I generally see stable memory usage now, though I haven't actually tried an install with all of the usual apps going at once.

This reworks the logic for uploading and downloading packages so that it applies some backpressure, instead of letting stuff build up in memory. It possibly applies *too much* backpressure, as it waits for each call to ByteStream.write(), but since it's generally reading from a network connection and writing to a local socket which is much faster I suspect this won't have much impact in practice. The "right" solution is probably to expose capnp-c++'s streaming support via node-capnp.

Along the way I did a few minor cleanups of nearby code, and fixed some warnings.

### DIFF
--- a/shell/imports/server/installer.js
+++ b/shell/imports/server/installer.js
@@ -19,7 +19,6 @@ import { pipeline, Writable } from "stream";
 
 import { Meteor } from "meteor/meteor";
 import { _ } from "meteor/underscore";
-import { Random } from "meteor/random";
 
 import { inMeteor, waitPromise } from "/imports/server/async-helpers.ts";
 import { ssrfSafeLookupOrProxy } from "/imports/server/networking.js";
@@ -294,17 +293,11 @@ class AppInstaller {
     console.log("Downloading app:", this.url);
     this.updateProgress("download");
 
-    this.uploadStream = globalBackend.cap().installPackage().stream;
-    return this.doDownloadTo(this.uploadStream);
-  }
-
-  doDownloadTo(out) {
     inMeteor(this.wrapCallback(function () {
       const safe = ssrfSafeLookupOrProxy(globalDb, this.url);
 
       let bytesExpected = undefined;
       let bytesReceived = 0;
-      const hasher = Crypto.createHash("sha256");
       let done = false;
       const updateDownloadProgress = _.throttle(this.wrapCallback(() => {
         if (!done) {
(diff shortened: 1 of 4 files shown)


zenhack:
Ack, looks like CI is failing due to the thing #3636 is supposed to fix. I'm hesitant to just stack this on top of that one because of the sheer size of that patch (even though it's not doing much conceptually).

shell/imports/server/installer.js
```diff
@@ -131,46 +131,42 @@ if (!Meteor.settings.replicaNumber) {
...
-      } catch (err) {
-        reject(err);
-      }
-    });
-
-    stream.on("error", (err) => {
-      // TODO(soon):  This event does't seem to fire if the user leaves the page mid-upload.
-      try {
-        backendStream.close();
+        backendStream.write(chunk).then(() => {
```
kentonv, line 153: Need to handle exceptions from `backendStream.write()`.

shell/imports/server/installer.js
```diff
@@ -131,46 +131,42 @@ if (!Meteor.settings.replicaNumber) {
...
+      if(err) {
         reject(err);
-      } catch (err2) {
-        reject(err2);
+        return;
       }
+      backendStream.done();
+      const packageId = hasher.digest("hex").slice(0, 32);
+      backendStream.saveAs(packageId)
+        .then((info) => resolve({info, packageId}), reject)
```
kentonv, line 166: Callback-oriented code always makes me nervous that errors might not be handled. You could convert more of this into async/await like so:

```suggestion
      resolve((async () => {
        backendStream.done();
        const packageId = hasher.digest("hex").slice(0, 32);
        let info = await backendStream.saveAs(packageId);
        return {info, packageId};
      })());
```

zenhack:
I've incorporated your suggestions. Should be good now.

PR #3639, opened 2022-07-09
release.sh: automatically push to master

zenhack:
Per discussion on IRC, the script currently pushes the tag, but does not
push to the master branch. This patch adds that logic to the script, so
hopefully Kenton won't have to remember to do this going forward.

Note that this is untested.


ocdtrekkie (approved):

zenhack:
Hm, web publishing test is failing -- looks like that fails intermittently now, rather than just always. By next release I'll either try to fix that or mark it as xfail too.

PR #3643, opened 2022-07-13
Google Apps for Work is Workspace

ocdtrekkie:
Trying to keep up with Google's rebranding is probably a pointless endeavor, but our lang files didn't even all use the same Google branding, as two of them referred to G Suite, and the rest as Google Apps for Work. This PR updates our language in both the setup wizard and the admin panel that this is Google Workspace now.

### DIFF
--- a/shell/i18n/fi.i18n.json
+++ b/shell/i18n/fi.i18n.json
@@ -710,12 +710,12 @@
             "subtext": "Käyttäjät joiden sähköpostiosoitteena on tämä domain tulevat olemaan tämän palvelimen organisaation jäseniä."
           },
           "gappsToggle": {
-            "label": "Käyttäjät jotka tunnistautuvat G Suiten kautta",
+            "label": "Käyttäjät jotka tunnistautuvat Google Workspace kautta",
             "subtext": "Huom: poistettu käytöstä koska Google kirjautuminen ei ole määriteltynä."
           },
           "gappsDomain": {
             "label": "Domain:",
-            "subtext": "Käyttäjät joilla on G Suite tili tällä domainilla tulevat olemaan tämän palvelimen organisaation jäseniä."
+            "subtext": "Käyttäjät joilla on Google Workspace tili tällä domainilla tulevat olemaan tämän palvelimen organisaation jäseniä."
           },
           "ldapToggle": {
             "label": "Käyttäjät tunnistautuvat LDAPin kautta",
@@ -1685,12 +1685,12 @@
           "subtext": "Käyttäjät joilla on tämän domainin sähköpostiosoite tulevat olemaan tämän palvelimen organisaation jäseniä."
         },
         "gappsToggle": {
-          "label": "G Suiten kautta kirjautuneet käyttäjät",
+          "label": "Google Workspace kautta kirjautuneet käyttäjät",
           "subtext": "Huom: poistettu käytöstä koska Google kirjautumista ei ole määritelty."
         },
         "gappsDomain": {
           "label": "Domain:",
-          "subtext": "G Suiten kautta kirjautuneet tämän domainin käyttäjät tulevat olemaan tämän palvelimen organisaation jäseniä."
+          "subtext": "Google Workspace kautta kirjautuneet tämän domainin käyttäjät tulevat olemaan tämän palvelimen organisaation jäseniä."
         },
         "ldapToggle": {
           "label": "LDAPin kautta kirjautuneet käyttäjät",
--- a/shell/i18n/fi.i18n.json
+++ b/shell/i18n/fi.i18n.json
@@ -710,7 +710,7 @@
             "subtext": "Käyttäjät joiden sähköpostiosoitteena on tämä domain tulevat olemaan tämän palvelimen organisaation jäseniä."
           },
           "gappsToggle": {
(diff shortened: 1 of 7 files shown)


shell/i18n/fi.i18n.json
```diff
@@ -710,12 +710,12 @@
             "subtext": "Käyttäjät joiden sähköpostiosoitteena on tämä domain tulevat olemaan tämän palvelimen organisaation jäseniä."
           },
           "gappsToggle": {
-            "label": "Käyttäjät jotka tunnistautuvat G Suiten kautta",
+            "label": "Käyttäjät jotka tunnistautuvat Google Workspace kautta",
```
zenhack, line 713: @xet7, what's the right translation here?

I wonder how much effort we should put into finding the other translators vs. just letting these get a bit stale. For ones where we can't track down the translators, maybe it's better to use the old names than use the new name in the wrong language?

zenhack (commented):
Minor thought in-line re: i18n, but otherwise looks good.

shell/i18n/fi.i18n.json
ocdtrekkie: My understanding (although I imagine it can be language-specific) is that generally brand names should not be translated, which is largely backed up by the contents of our current translations. (I also did run this string through Bing Translate and got a sane result, though that isn't authoritative.) I do think people today may not have any idea what Google Apps for Work is, and "G Suite" may not only suffer that, but may not obviously refer to Google.

I think in cases where we do have translated text, we should do our best to use Translate and similar tools to get close to accurate, and be very open to PRs that adjust i18n strings.

shell/i18n/fi.i18n.json
```diff
@@ -710,12 +710,12 @@
...
             "subtext": "Huom: poistettu käytöstä koska Google kirjautuminen ei ole määriteltynä."
           },
           "gappsDomain": {
             "label": "Domain:",
-            "subtext": "Käyttäjät joilla on G Suite tili tällä domainilla tulevat olemaan tämän palvelimen organisaation jäseniä."
+            "subtext": "Käyttäjät joilla on Google Workspace tili tällä domainilla tulevat olemaan tämän palvelimen organisaation jäseniä."
```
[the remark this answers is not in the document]
xet7: @ocdtrekkie 

Right.

shell/i18n/fi.i18n.json
```diff
@@ -1685,12 +1685,12 @@
           "subtext": "Käyttäjät joilla on tämän domainin sähköpostiosoite tulevat olemaan tämän palvelimen organisaation jäseniä."
         },
         "gappsToggle": {
-          "label": "G Suiten kautta kirjautuneet käyttäjät",
+          "label": "Google Workspace kautta kirjautuneet käyttäjät",
```
ocdtrekkie, line 1688: Actually, while we have @xet7 here.... the text in the setup wizard and the admin panel on the English version are identical, see above in the PR, but they differ slightly in the translation. Should they be the same in the Finnish version?

shell/i18n/fi.i18n.json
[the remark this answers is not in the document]
xet7: I would presume that most translations are similar, that if some charater has been added to word `G Suite` then same should be also at end of `Google Workplace`

shell/i18n/fi.i18n.json
xet7: Yes, when it's ` .... kautta` it's like `via Google Workplace` and requires it to be `Google Workspacen`

shell/i18n/fi.i18n.json
```diff
@@ -1685,12 +1685,12 @@
...
           "subtext": "Huom: poistettu käytöstä koska Google kirjautumista ei ole määritelty."
         },
         "gappsDomain": {
           "label": "Domain:",
-          "subtext": "G Suiten kautta kirjautuneet tämän domainin käyttäjät tulevat olemaan tämän palvelimen organisaation jäseniä."
+          "subtext": "Google Workspace kautta kirjautuneet tämän domainin käyttäjät tulevat olemaan tämän palvelimen organisaation jäseniä."
```
xet7, line 1693: `Google Workspacen`

shell/i18n/fi.i18n.json
ocdtrekkie: I mean, shouldn't line 713 and line 1688 be the same? And shouldn't line 718 and 1693 be the same?

In the English version, line 759 and 1739 are identical, and line 764 and 1744 are identical.

xet7: @ocdtrekkie 

Yes, everywhere should be `G Suiten kautta => Google Workplacen kautta`

Other language use word `via` or similar in that language. Finnish uses `n`

ocdtrekkie:
@xet7 How does this look? I think that commit might help explain what I was asking about "I mean, shouldn't line 713 and line 1688 be the same? And shouldn't line 718 and 1693 be the same?"

ocdtrekkie:
Okay, awesome. No other translation had localized the brand name except for the NL translation, so I think we're mostly good across the board then. I'm going to add a commit updating this in Docs too though.

zenhack:
LGTM

PR #3645, opened 2022-07-29
Docs: Fix GPG export instructions

ocdtrekkie:
I attempted to follow these instructions this evening, and gpg gave me an error claiming that "--export-options is not considered an option", and I had a fairly difficult time getting a search result to explain why. (I assumed it was removed/renamed/replaced.) As it turns out, an answer about the same error for a different option revealed... gpg is sort of picky about the order of arguments, and so I tried flipping the order here, and it worked without incident.


zenhack (approved):
I can confirm the command as written doesn't work for me either, though the fixed version does.

PR #3647, opened 2022-07-31
Docs: delete graphviz

ocdtrekkie:
This replaces our graphviz dependency which keeps being renamed/moved/broken with static SVGs, as we have not changed these graphs in several years anyways, and graphviz is solely used on one content page. I left the graphviz instructions in as comments so that we have a reference for regeneration if needbe. I also had to pin some other dependencies to stick with our old mkdocs for now, until someone goes and tests bringing everything current.

The CI build looks right, my browser doesn't like redirecting to index.html on local folder references so it doesn't all link together right for me on this Windows PC, but I don't think it's a problem for production, as the relative links all look right.


docs/using/how-it-works.md
```diff
@@ -22,7 +22,10 @@ and the grain occur over the Cap'n Proto WebSession format.  With existing
 applications, the Sandstorm HTTP bridge is used to translate between Cap'n
 Proto and HTTP.
 
-```graphviz dot communication_overview_http_app.svg
+![Diagram with HTTP Bridge](communication_overview_http_app.svg)
+
+<!--
```
zenhack, line 27: Might want to add a comment here explaining why this is commented out and how it relates to the svg.

zenhack (commented):
Minor thing, but otherwise LGTM.

docs/using/how-it-works.md
ocdtrekkie: Done.

PR #3653, opened 2022-07-31
Minor spelling fix

spollard:


ocdtrekkie:
Hmmm... interestingly enough, this could've be "once", or it could be "since"... I suspect the latter is a better fit but I'm okay either way.

zenhack:
I agree; since is probably better but it doesn't seem super important.

spollard:
Since sounds better. :)

ocdtrekkie:
Okay, since everyone went with "since", I'm going to close this. @spollard if you want to send a PR for "since", I'll merge it, or if you don't, I will submit my own and coauthor you at some future point when I remember.

spollard:
You got it. I don't really want to make a whole new PR...

zenhack:
I approved @ocdtrekkie's other pr, though in the future a force-push is probably the right thing to do in this situation; a separate pr does feel a bit excessive.

ocdtrekkie:
Post-mortem question: Should I have waited to close the PR longer? I definitely would've accepted a force-push to this PR and reopened/merged it. I also read @spollard's comment today as saying I should submit the fix, because he didn't want to, but that could've been incorrect in reading it that way.

I want to make sure I'm not being discouraging of contribution, if there's a different way I should've addressed this one, by all means call me out.

zenhack:
I'll let @spollard weigh in if he wants on how he took that. But it usually doesn't make sense to open a new PR for changes, unless it's a large-ish patch that's a completely different approach to something.

ocdtrekkie:
Indeed, though you cannot force push from the web UI, which is how I assume (perhaps erroneously, I grant) a "patch-5" branch was created. And I'd rather a fresh PR than a second commit in one PR.

spollard:
You were right! I didn't really care how the fix got in.

PR #3657, opened 2022-08-25
Override shouldResolveBeforeRedirecting to true in capnp membranes.

dom96:
I assumed that all membranes should have this enabled, but maybe that's not the case. Happy to change if so.

The reason this is necessary is because the default for this is being changed: https://github.com/capnproto/capnproto/pull/1530

### DIFF
--- a/src/sandstorm/app-index/app-index.c++
+++ b/src/sandstorm/app-index/app-index.c++
@@ -269,6 +269,10 @@ private:
     }
 
     kj::Own<MembranePolicy> addRef() override { return kj::addRef(*this); }
+
+    bool shouldResolveBeforeRedirecting() override {
+      return true;
+    }
   };
 };
 
--- a/src/sandstorm/sandstorm-http-bridge.c++
+++ b/src/sandstorm/sandstorm-http-bridge.c++
@@ -902,6 +902,10 @@ public:
   kj::Own<capnp::MembranePolicy> addRef() override {
     return kj::addRef(*this);
   }
+
+  bool shouldResolveBeforeRedirecting() override {
+    return true;
+  }
 private:
 };
 
--- a/src/sandstorm/supervisor.c++
+++ b/src/sandstorm/supervisor.c++
@@ -1494,6 +1494,10 @@ public:
           kj::mv(parent)
         }) {}
 
+  bool shouldResolveBeforeRedirecting() override {
+    return true;
+  }
+
   kj::Maybe<capnp::Capability::Client> inboundCall(
       uint64_t interfaceId, uint16_t methodId, capnp::Capability::Client target) override {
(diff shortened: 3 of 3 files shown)


ocdtrekkie:
I am not sure what this is or does or how we use it, but as you have discussed this with Kenton, going to mark it for him to review for the next release! Thanks for keeping us in mind!

zenhack:
Thanks for keeping us in the loop!

In places where we turn this back on I'd like to document *why* it's necessary; I don't want to just carry that because we don't understand our code.

For the membrane policy that I wrote at least (`SaveMembranePolicy`) I remember being unaware of this behavior, so strongly suspect it's not needed. The other policies I'm not familiar enough with to know off hand. I'll likely stare at them a bit in the near future, but also maybe @kentonv can impart some insight.

kentonv:
@dom96 Don't worry abound responding here, I'll take care of it. This will get merged one way or another before we next update the Cap'n Proto dep.

@zenhack The "resolve before redirecting" behavior is important in cases of membranes that are "lossy", that is, an RPC call that passes over the membrane and then back again might lose data compared to if the call never crossed the membrane. In Sandstorm, this happens with `save()` calls that enter a grain. In particular I believe I ran into a problem where the system would request a capability from a grain, and immediately make a promise-pipelined `save()` call on it. It turned out the grain actually returned a capability to a shell-implemented object, i.e. not an object implemented inside the grain. In that case, `save()` is supposed to construct an API token of type `frontendRef`, but once the `save()` call passed over the supervisor membrane into the app, it could no longer properly reflect back to the shell to produce this. My solution was to make it so when `save()` was called on a promise, the call would be held at the membrane until the promise resolved, so that if the promise resolved back out of the membrane, the `save()` call could be forwarded along without losing scope.

Unfortunately I built this behavior into `MembranePolicy` itself, thinking it was universal. Later on, in Cloudflare Workers, we had a case where this forced an unnecessary network round trip as a membrane on the client end which was intended to rewrite a certain kind of RPC call would wait patiently for the capability to resolve before it was willing to perform the rewrite.

There's also potentially a much worse interaction with objects that implement `shortenPath()`, which allows them to resolve themselves to some other capability at an arbitrary time. Such capabilities always look like promises, but do not necessarily resolve quickly, or ever. The "resolve before redirecting" policy could cause outright hangs when interacting with these.

It's clear that "resolve before redirecting" is not a good default, as it solves a specific, unusual problem in Sandstorm that doesn't apply to most use cases. It got to be a pain having to remember always to override this method in Workers use cases, so @dom96 and I agreed we should change the default. I told him to update all MembranePolicies in Sandstorm to override the method so that we know none of them will change behavior. It may indeed be the case, though, that some of the policies do not need the behavior and should simply use the new default.

PR #3659, opened 2022-08-29
Partially document ShellCli.

zenhack:
This patch:

- Partially documents the ShellCli interface in backend.capnp
- Exposes the path to the socket as a constant in the capnp schema.

There are two methods left un-documented, since I don't fully understand
what they do, and was hoping Kenton could shed some light on them, since
he wrote the code. In particular:

- I don't know what the directory argument to createAcmeAccount does.
- I'm generally fuzzy about how setAcmeChallenge is used.

If I can get clarification on those, I'll rebase and finish the job.

### DIFF
--- a/src/sandstorm/backend.capnp
+++ b/src/sandstorm/backend.capnp
@@ -217,9 +217,18 @@ interface GatewayRouter {
 }
 
 interface ShellCli {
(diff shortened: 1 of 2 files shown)


zenhack:
I'm fairly certain the ipnetwork test failure has nothing to do with this patch.

PR #3661, opened 2022-09-06
Docs: vagrant-spk update pass

ocdtrekkie:
This is some various/small tweaks and updates to the docs around vagrant-spk. It depends on changes in https://github.com/sandstorm-io/vagrant-spk/pull/331

### DIFF
--- a/docs/developing/troubleshooting.md
+++ b/docs/developing/troubleshooting.md
@@ -110,9 +110,9 @@ ImportError: No such module requests
 **How to fix the problem:** Our recommendations are as follows.
 
 - For Python apps, **add the app's virtualenv to alwaysInclude.** To do that, look for the
-  `alwaysInclude` list and add `opt/app/env`. See the [ContactOtter sandstorm-pkgdef.capnp for an
+  `alwaysInclude` list and add `opt/app-venv`. See the [ContactOtter sandstorm-pkgdef.capnp for an
   example.](https://github.com/phildini/logtacts/blob/27ac05f88896778baf5da155afa6c733d3d6a264/.sandstorm/sandstorm-pkgdef.capnp#L137)
-  It is sometimes helpful to also add `usr/lib/python2.7` (or the corresponding path for your
+  It is sometimes helpful to also add `usr/lib/python3.9` (or the corresponding path for your
   version of Python).
 
 - For nodejs apps, **add node_modules to alwaysInclude.** See the [Duolodo sandstorm-pkgdef.capnp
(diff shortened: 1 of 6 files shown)


docs/developing/troubleshooting.md
```diff
@@ -110,9 +110,9 @@ ImportError: No such module requests
 **How to fix the problem:** Our recommendations are as follows.
 
 - For Python apps, **add the app's virtualenv to alwaysInclude.** To do that, look for the
-  `alwaysInclude` list and add `opt/app/env`. See the [ContactOtter sandstorm-pkgdef.capnp for an
+  `alwaysInclude` list and add `opt/app-venv`. See the [ContactOtter sandstorm-pkgdef.capnp for an
```
zenhack, line 113: I wonder if there's another app we could point to here that's closer to being maintained? Also I see some potential  for confusion here since (I assume) that app pulls in `opt/app/env`.

docs/vagrant-spk/customizing.md
```diff
@@ -94,7 +94,7 @@ Other things you probably want to do in this script include:
 
 - Building folder structures in `/var`.  `/var` is the only non-tmpfs folder mounted R/W, and when a grain is first launched, it will start out empty.  It will persist between runs of the same grain, but be unique per app instance.
 - Preparing a database and running migrations.  You can also manually generate some tables once, place them somewhere under `/opt/app`, and copy them to `/var/lib/mysql` if your app takes a while to do migrations, at the potential cost of producing a larger `.spk`.
-- Launching other daemons that your app uses (`mysqld`, `redis-server`, `php5-fpm`, `uwsgi`, etc.)
+- Launching other daemons that your app uses (`mysqld`, `redis-server`, `php7-fpm`, `uwsgi`, etc.)
```
zenhack, line 97: Maybe it would be better to just say `php-fpm`, to make this a bit more evergreen? The executable name varies from distro to distro and I don't think it's that important that it mach exactly what's currently on debian -- especially since the lesp/lemp stacks are going to have that as an example anyway.

zenhack (approved):
A couple thoughts, neither of which I think are important enough to block.

docs/developing/troubleshooting.md
ocdtrekkie: The problem is there are no example apps currently showing that location, since that's a new placement as of today's merge. So I could drop the example entirely, but then I don't have an example for the formatting of that field in the pkgdef, which I feel has value.

I think I will leave it for the moment, but I have a strong desire to revisit this in future passes.

docs/vagrant-spk/customizing.md
ocdtrekkie: I agree here. I will plan to force push this before merging, as a commit to delete a digit would feel sad.

PR #3666, opened 2022-10-07
Docs: Add JS to target external links

ocdtrekkie:
This improves our advice on modifying apps to open links in a new tab with a bit of JS that will only affect external links, while leaving relative links alone.

### DIFF
--- a/docs/developing/troubleshooting.md
+++ b/docs/developing/troubleshooting.md
@@ -22,13 +22,27 @@ docs about the `vagrant-spk enter-grain` command](debugging.md).
 ## Clicking a link in the app does not open the link
 
 Sandstorm apps cannot navigate the user away from the app. Therefore, app
-authors should set `target="_blank"` on links within the app.
+authors should set `target="_blank"` on external links within the app.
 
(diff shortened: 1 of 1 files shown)


docs/developing/troubleshooting.md
```diff
@@ -31,6 +31,13 @@ the following HTML to your document's `<head>`:
 <base target="_blank">
 ```
 
+However, if you have both internal and external links, you can add a script
+such as the following which will only set the target on external links:
+
+```js
+window.onload = function(){ var a = document.getElementsByTagName('a'); for (var i=0; i<a.length; i++){ if(a[i].hostname != location.hostname) {a[i].setAttribute('target', '_blank'); } } }
```
zenhack, line 38: Let's add some line breaks & indentation, rather than cramming all this in on one line.

ocdtrekkie: My original thought was to keep things small, and make this more of a magic "just add this line" thing, since it's part of a larger page, but I have expanded it out and force-pushed the branch.

zenhack (approved):

PR #3668, opened 2022-10-13
Typecheck: Disable strictNullChecks

ocdtrekkie:
This is a *potential* fix for the failure of typecheck-ts. (We could also locked TypeScript down to 4.7.x, or fix the issues in question, of course...

I believe the new errors were caused due to: https://devblogs.microsoft.com/typescript/announcing-typescript-4-8/#unconstrained-generics-no-longer-assignable-to and this is an example of another project fixing their compatibility with 4.8: https://github.com/prisma/prisma/pull/15072 which I feel may be useful reference if someone wants to tackle this "correctly".

### DIFF
--- a/shell/tsconfig.json
+++ b/shell/tsconfig.json
@@ -12,7 +12,7 @@
     /* Strict Type-Checking Options */
     "strict": true,
     "noImplicitAny": true,
-    "strictNullChecks": true,
+    "strictNullChecks": false,
 
     /* Additional Checks */
     "noUnusedLocals": true,


ocdtrekkie:
Hah, and bypassing this got me to the actual error Kenton found when he was trying to test: https://github.com/sandstorm-io/sandstorm/actions/runs/3239675622/jobs/5309339670#step:12:436

Whole onion of test failures. Gotta get past the intermittent node-capnp to find the typecheck issue, gotta get past the typecheck issue to get child_process, etc.

zenhack:
I'd rather not fix it this way; strict null checking is the sort of thing that has the potential to catch a lot of real bugs.

ocdtrekkie:
Entirely fair: I opened this to document to abuse the CI for testing and leave notes. :)

PR #3669, opened 2022-10-20
Report install failures to Alpha

### LINKED ISSUE #3625: Install script reports errors to an Oasis URL
https://github.com/sandstorm-io/sandstorm/blob/master/install.sh#L92 seems to indicate we are reporting errors to https://api.oasis.sandstorm.io/api which I presume probably no longer works. Not sure if this data collector is something we can set up on Alpha or not, but it looks like a connectivity error when someone tries to report here and it's probably caused by the endpoint not existing.

ocdtrekkie:
Fixes #3625 

I don't totally want to shut this off, because this completely opt-in telemetry tells us A. if there's people installing Sandstorm at some vague rate (or at least trying and failing) and B. what issues they are running into. However, I could not rebuild the J(A)SON Datastore app, because it's a bitrotted meteor-spk app. https://github.com/sandstorm-io/sandstorm-error-collector is significantly simpler and more straightforward.

I've shared the grain I've pointed this at to both Kenton and Ian so everyone knows where it is.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -88,8 +88,8 @@ fail() {
     if USE_DEFAULTS=no prompt-yesno "Hmm, installation failed. Would it be OK to send an anonymous error report to the sandstorm.io team so we know something is wrong?
 It would only contain this error code: $error_code" "yes" ; then
       echo "Sending problem report..." >&2
-      local BEARER_TOKEN="ZiV1jbwHBPfpIjF3LNFv9-glp53F7KcsvVvljgKxQAL"
-      local API_ENDPOINT="https://api.oasis.sandstorm.io/api"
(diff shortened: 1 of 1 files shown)


zenhack (approved):

ocdtrekkie:
@kentonv Does this change need to be manually deployed when we merge this?

ocdtrekkie:
I've modified this commit to remove the /v1.php, as Sandstorm Error Collector 1.0.1 now doesn't require that, and I may want to change out the stack in the error collector for a non-PHP stack at a later time. I tested the modified app accepts commands both with and without the trailing slash, as well as still with v1.php of course.

kentonv:
`install.sh` changes are automatically deployed as part of the release process.

PR #3683, opened 2022-12-27
Clarify email error message

jdougan:
Prior message implied the publicId was a grain name, not a grain mail id, and left it unclear where the id was coming from.

### DIFF
--- a/shell/imports/server/drivers/mail.js
+++ b/shell/imports/server/drivers/mail.js
@@ -164,7 +164,7 @@ Meteor.startup(function () {
                 } else {
                   // TODO(someday): We really ought to rig things up so that the 'RCPT TO' SMTP command
                   // fails in this case, by adding an onRcptTo() callback.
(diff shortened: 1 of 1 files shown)


ocdtrekkie (approved):
This is probably an improvement, though it may be preferable to explicitly state the id is the publicId, not a grain id.

zenhack:
Yeah I think making it we're talking about the grain's publicId, not the grain id, is worth doing.

jdougan:
Ok, I've made the change in my fork, do I have to so anything to update the PR?

ocdtrekkie (approved):

ocdtrekkie:
> Ok, I've made the change in my fork, do I have to so anything to update the PR?

Nope. The PR includes any commits on the same branch.

zenhack (approved):

kentonv:
Thanks!

PR #3684, opened 2022-12-30
fix(sec): upgrade mkdocs to 1.3.0

pen4:
### What happened？
There are 1 security vulnerabilities found in mkdocs 1.0.4
- [MPS-2022-6935](https://www.oscs1024.com/hd/MPS-2022-6935)

### What did I do？
Upgrade mkdocs from 1.0.4 to 1.3.0 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS
Signed-off-by:pen4<948453219@qq.com>


zenhack:
@ocdtrekkie, am I misremembering that there was some specific reason we were still on an old version of this package?

If not I'm fine with updating it, though it's not a security problem, since this is only used to build the docs and is never exposed to untrusted input (or even used by sandstorm at runtime).

ocdtrekkie:
I think it was just some combination of bitrot between the software, theme, custom changes to theme, and since-removed graphviz plugin. There's no reason not to get it updated, but we'll need to fix whatever breaks.

ocdtrekkie:
I pushed the docbuild.tar output from the CI job to https://u6zrcv7tixdig69dh0s0.ocdhost.sandcats.io/

This appears to work, except for loading some icons. Since the docs build is now broken without an mkdocs update (or a Python 3.8.x pin), I feel like we should merge this PR, and then work on fixing the rest.

zenhack (approved):

PR #3686, opened 2023-01-25
Replace WordPress app key

ocdtrekkie:
It's been a half decade since WordPress on Sandstorm was updated, I am not inclined to chase people down for keys who may or may not have retained them that long. @JamborJan was the last keyholder, but a Sandstorm-the-company team member or two may have also held it.

The pending update I have does *not* bump the WordPress version forward at this time, but does clean up some package bitrot, updates the base box slightly, and most importantly, adds Ian's powerbox-http-proxy so that WordPress can grab media when importing a backup. (It's likely this will make many other plugins usable inside the sandbox, and very likely, allow updates for non-read-only plugins.)

I flubbed line 123 here, I will force-push the branch in a couple minutes.

### DIFF
--- a/src/sandstorm/appid-replacements.capnp
+++ b/src/sandstorm/appid-replacements.capnp
@@ -110,7 +110,16 @@ const appIdReplacementList :List(AppIdReplacement) = [
   # See also [this discussion](https://github.com/wekan/wekan/issues/640) about the future
   # governance of Wekan.
 
-# ---- end Wekan entry ----
+  # ---- end Wekan entry ----
+  
+  # ---- WordPress entry ----
+
+  (original = "aax9j672p6z8n7nyupzvj2nmumeqd4upa0f7mgu8gprwmy53x04h",
+   replacement = "zp00qqrha6gp8515ju84ys91syyq8ezv77g9s1r0t3urz92uzmqh"),
+  # The original key was held by Jan Jambor and David Renshaw. The app was left unmaintained from
+  # 2018 to 2023, so Jacob Weisz replaced the key rather than trying to track down the original.
+
+  # ---- end WordPress entry ----
 
   # Add your entry here!
 ];


src/sandstorm/appid-replacements.capnp
```diff
@@ -110,7 +110,17 @@ const appIdReplacementList :List(AppIdReplacement) = [
...
 
-# ---- end Wekan entry ----
+  # ---- end Wekan entry ----
+  
+  # ---- WordPress entry ----
+
+  (original = "aax9j672p6z8n7nyupzvj2nmumeqd4upa0f7mgu8gprwmy53x04h",
+   replacement = "zp00qqrha6gp8515ju84ys91syyq8ezv77g9s1r0t3urz92uzmqh"),
+  # The original key is held by Jan Jambor and likely members of the original Sandstorm team. It's
+  # been half a decade since this app was updated, so I don't want to block on requesting a key
```
zenhack, line 120: Nit:

- Prose that mentions relative spans of time without reference to when it was written is likely to bitrot; maybe mention the date this change is being made, so future readers are not wondering when "now" is.
- Similarly, I try to avoid using the first person without making it clear who the author of the comment is; I would either add your name in parens or rephrase this to be more of a group decision (which it will be, if we all decide this should be merged).

zenhack (approved):
LGTM, Note that per the "Things you should NOT do" section, once this is merged you'll want to revert the change you made to appId in wordpress's pkgdef.

ocdtrekkie:
Ooh. Good catch. I was trying to remember which ID to use when signing, guess the old one too then.

ocdtrekkie:
Where does it say that? Oddly enough, draw.io is published with the a3w key in the pkgdef, and Wekan is published with the m86 key in the pkgdef. Which makes me think it doesn't actually matter which key is in the pkgdef?

zenhack:
It's in `appid-replacements.capnp`; just ctrl+F. I haven't looked at the implementation, but it wouldn't shock me if that just "happend" to work.

ocdtrekkie:
Ah there it is. I was using the little expand on GitHub to look up from this PR and didn't go far enough. I suppose I should fix it on draw.io to be consistent with the spec even though it works as a practical matter.

PR #3687, opened 2023-01-25
Allow sending emails to current user instead of grain owner

rs22:
I'm using Sandstorm's ability to send and receive emails in custom apps (to do some task automation). When sharing a grain with other people however, it's only possible for the grain to send mails to the grain owner, instead of the currently logged in user.

This PR changes this, and looks up the email address of the currently logged in user of the session.

This is also respected from the mail sending code: 
https://github.com/rs22/sandstorm/blob/send_mails_to_current_user/shell/imports/server/drivers/mail.js#L240
(Another useful addition would be to allow sending emails to other people who have access to the same grain.)

### DIFF
--- a/shell/imports/server/hack-session.js
+++ b/shell/imports/server/hack-session.js
@@ -372,20 +372,21 @@ class HackSessionContextImpl extends SessionContextImpl {
     // Get the user's e-mail address.
     //
     // Must be called in a Meteor context.
-
-    const grain = globalDb.collections.grains.findOne(this.grainId, { fields: { userId: 1 } });
-
-    const user = Meteor.users.findOne({_id: grain.userId});
-
-    const email = _.findWhere(SandstormDb.getUserEmails(user), { primary: true });
-
     const result = {};
-    if (email) {
-      result.address = email.email;
-    }
 
-    if (user.profile.name) {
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
Interesting! I don't know of a lot of people using mail support, so this is an area of code that isn't frequently looked at. Ian will probably be able to determine if this is safe to merge though.

zenhack:
This is a breaking change; the comments in hack-session.capnp are quite clear that this is supposed to return the address for the grain owner; it's quite possible this will break the expectations of some other private app that uses this (I don't know of any public ones that do). I don't think we should change this.

Probably the right thing to do is to do a powerbox request for `EmailAddress`, and add support to the platform for fulfilling that with the user's address.

rs22:
Understood. 

Would you be open to adding another method to hack-session.capnp returning the current user and lifting the restriction in the mail driver a bit so that it allows sending emails to both the grain owner and the current user? That would accomodate my use case.

kentonv:
This is really something that the user needs to opt into. Normally if I open someone's Sandstorm grain I don't expect I am revealing my email address to them or giving them permission to email me. So I agree with @zenhack that a powerbox request is needed here.

PR #3689, opened 2023-02-27
Fix #3688

### LINKED ISSUE #3688: "App market" shortcut leads to invalid URL
On any given app page, I noticed on both Alpha and my personal server, the "App Market" link, which should take you to the "store page" for the app, just gives a 404 on your own server. If your app page is at https://example.sandcats.io/apps/ztxnreej8a0pqn97ccfhuckswjkrmsaavn85n5sgzk30tv2k4y9h and you click App Market, it tries to load https://example.sandcats.io/app/ztxnreej8a0pqn97ccfhuckswjkrmsaavn85n5sgzk30tv2k4y9h (note the no "s" in "app"), instead of pointing at the app market server.

zenhack:
There were three problems:

- The database property wasn't actually available here, so the query crashed when run, meaning appMarketHost was being filled in with an empty string.
- The conditional was backwards, and so if the setting was absent this would then crash on the following line.
- There was an extra trailing slash on the default URL -- which apps.sandstorm.io doesn't seem to canonicalize away, so even after fixing the above we were greeted with a blank page.

### DIFF
--- a/shell/imports/client/apps/app-details-client.js
+++ b/shell/imports/client/apps/app-details-client.js
@@ -138,11 +138,12 @@ Template.sandstormAppDetails.helpers({
   },
 
   appMarketHost: function() {
-    let host = "https://apps.sandstorm.io/";
+    let host = "https://apps.sandstorm.io";
     const ref = Template.instance().data;
-    const db = ref._db;
+    // Pull the database from our parent context:
+    const db = Template.parentData(1)._db;
     const appMarket = db.collections.settings.findOne({ _id: "appMarketUrl" });
-    if (!appMarket) {
+    if (appMarket) {
       host = appMarket.value;
     }
     return host;


ocdtrekkie (approved):

ocdtrekkie:
FYI, the "fix X" magical linking in GitHub doesn't work if it's in the title of the PR and not the body. I've linked the issue this closes.

Also, I think we should cc: @gischer informationally on this one, once we get around to releasing this, they'll want to probably check it still works well in their case.

zenhack:
The commit should still close it when it lands I think.

gischer:
Thanks for the heads-up.  I'm a bit surprised I hadn't noticed this.  But after enough examination, it occurred to me that maybe I had, but got it confounded with some other issue, which I addressed, and never got back to checking this.

This is why I try to write this stuff down.

PR #3700, opened 2023-08-03
Add ability to run build action manually

ocdtrekkie:
I think we need this to be able to trigger the action manually: https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow


ocdtrekkie:
I am merging because this is CI only, and can only be used to run workflow files that are on the default branch (aka, the ones we've approved).

PR #3704, opened 2024-01-13
Update github actions to latest

wwahammy:
The Github actions used by Sandstorm are quite out of a date. Some of the versions are being deprecated; it's generally good practice to update them as appropriate.


ocdtrekkie:
upload-artifact is the only one I see a breaking change for and it shouldn't affect us, so hopefully this shouldn't break much/anything. We expect tests to fail currently, but we should get a good build artifact in roughly the same time as previous runs.

ocdtrekkie (approved):

PR #3705, opened 2024-01-13
Build on Ubuntu 20.04

wwahammy:
It doesn't seem possible for the github actions to build Sandstorm on Ubuntu 22.04.  This adjusts the build machine to use Ubuntu 20.04 so that the build succeeds/ gets further along.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,7 @@ env:
   CCACHE_DIR: ${{ github.workspace }}/.ccache"
 jobs:
   build_and_test:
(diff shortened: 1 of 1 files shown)


ocdtrekkie (approved):

ocdtrekkie:
I'd obviously like to figure out why this is and fix it, but I definitely prefer having a working build than not.

ocdtrekkie:
Adding myself a handy label for the moment for new tech debt we incur, as it might be helpful for revisiting later.

PR #3707, opened 2024-01-13
Add skipLibCheck so we stop checking node_modules package definitions

wwahammy:
skipLibCheck skips checking type declaration files. This should fix typescript type checking in the build

### DIFF
--- a/shell/tsconfig.json
+++ b/shell/tsconfig.json
@@ -30,7 +30,8 @@
     "resolveJsonModule": true,
     "types": ["node"],
     "esModuleInterop": true,
-    "preserveSymlinks": true
+    "preserveSymlinks": true,
+    "skipLibCheck": true
   },
   "exclude": [
     "./node_modules/**",


ocdtrekkie:
How does this differ from #3668?

wwahammy:
> How does this differ from #3668?

strictNullChecks is a safety feature; generally you'd want to keep that if you can because it reduces the risk of various errors.

As I've always understood it, all of the strictness rules are important but one issue is how they apply to any typings you have downloaded. After all, you might not be able to trust that every dependency followed the same compilation rules. The only situation where you could if you had a pretty strong control over all of your dependencies. Generally though, on most projects you don't; you're getting dependencies from all over. skipLibCheck allows you to opt libraries out of the these strictness rules while making sure all of your code meets that standard. Generally, that's how projects I've worked on are configured.

ocdtrekkie:
Do you want to sync this branch so we can test it with the other fixes included?

wwahammy:
@ocdtrekkie ts-check passes now :partying_face:

ocdtrekkie:
Closing this one, as I merged 3708 into the `dev` branch, which includes this commit.

PR #3708, opened 2024-01-14
Update @typescript-eslint/parser for compatibility with used typescript

wwahammy:
When I run `npm run lint` I received an warning saying @typescript-eslint/typescript-estree (a dependency of @typescript-eslint/parser) did not support the version of Typescript used in the project. 

This updates @typescript-eslint/parser to a more modern version without touching eslint itself (since I don't know what the consequences are of that)

--

I rebased this on #3707 so I can prove that the lint process still runs. I'm happy to rearrange if preferred.

### DIFF
--- a/shell/npm-shrinkwrap.json
+++ b/shell/npm-shrinkwrap.json
@@ -1,117 +1,248 @@
 {
   "name": "sandstorm-shell",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 3,
   "requires": true,
-  "dependencies": {
-    "@babel/code-frame": {
+  "packages": {
+    "": {
+      "name": "sandstorm-shell",
+      "version": "1.0.0",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@babel/runtime": "^7.20.7",
+        "@root/acme": "^3.1.0",
+        "@root/csr": "^0.8.1",
+        "@root/keypairs": "^0.9.0",
+        "@root/pem": "^1.0.4",
+        "@types/fibers": "^3.1.1",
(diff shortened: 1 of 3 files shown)


ocdtrekkie:
This is probably approaching a place I should check with @kentonv if he is okay with me merging things here for now or if we should move to a fork or maybe dev branch now. Merging docs and test infrastructure code falls generally within "I did not break the official build" area, but as we exit Actions code I want to check if he would like the master branch here to remain as-is.

There's pros and cons to transferring the repo or forking it (things like stars, watchers, issues, etc.) but I do not really want to explore that until we are approaching an actual release plan which needs to factor in the database migration.

ocdtrekkie:
I think the most appealing strategy to me personally would be to create a dev branch in this repo and merge new changes that impact the production build into that branch (we might need to adjust Actions on master to also build_and_test from it), which can be sort of a beta build of the fork. Later, when we release/fork/etc. we could rename that branch (probably to 'main' to reflect current standards, and set it as the default branch for sandstorm-org.

ocdtrekkie:
As a note, we currently expect the linter to find 773 warnings: https://github.com/sandstorm-io/sandstorm/blob/master/.github/workflows/build.yml#L74

Ideally, we should not allow it to increment upwards. However, considering the linter logs from last time the linter actually ran is likely lost to time, it may be okay to increment it. That will allow the lint step of the Action to pass.

I also edited this PR to target the `dev` branch, which is currently even with `master`.

ocdtrekkie:
@wwahammy Looks like this got us further, but tests didn't run because of the lint error, so now that I've forgiven the lint error we got test failures!

PR #3710, opened 2024-02-18
Ignore .vscode

wwahammy:
Anyone using VSCode with the project will have their local VSCode config settings added at `.vscode/`. Since these are generally unique to the individual and customized, it's easy to unintentionally add these to a commit unless they're gitignored. So we're doing that here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,5 @@ tests/tmp-sandstorm
 app-index.spk
 test-app.spk
 deps/llvm-build
+
+.vscode


ocdtrekkie:
This seems sensible, but is there any way to include it at the root and include any given .vscode directory one might have? This particular gitignore is in a fairly specific place, but this improvement presumably makes sense globally.

kpreid:
> is there any way to include it at the root and include any given .vscode directory one might have?

This is the default behavior for names with no `/`. From [git docs](https://git-scm.com/docs/gitignore):

> If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular `.gitignore` file itself. Otherwise the pattern may also match at any level below the `.gitignore` level.

ocdtrekkie:
@kpreid I knew there was one behavior or another for it, I just don't play with it often enough.

So this change just needs to be moved to the root gitignore.

ocdtrekkie:
Or I misread the commit entirely and shouldn't review PRs on my phone.

ocdtrekkie (approved):

ocdtrekkie:
Merged (from a desktop!)

PR #3718, opened 2024-10-16
Add funding-manifest-urls check

ocdtrekkie:
This is a non-code change to indicate we have provenance and authority to submit funding requests for the project. To be clear, this refers to funding for the Sandstorm Community project, and will be implemented as a fork, not as a direct update to the original repository.

### DIFF
--- a/.well-known/funding-manifest-urls
+++ b/.well-known/funding-manifest-urls
@@ -1,0 +1,1 @@
+https://sandstorm.org/funding.json


jdougan:
Does this mean there are plans regarding the future of Sandstorm?

ocdtrekkie:
We certainly have plans, but we do not have funding for those plans.

PR #3719, opened 2024-11-15
Docs: Update ephemeral subdomains, remove Oasis link

ocdtrekkie:
Generally most users will see a ui- prefix in their ephemeral subdomains, so I added this to some here to make this obvious. Note that if someone has a wildcard host config like alpha-*.example.com, they might have a ephemeral domain that looks like alpha-ui-45sda64dsfsd464ggdsg5.example.com, but that is uncommon enough I am not worried about clarifying it here.

Also, people cannot try this on Oasis.

### DIFF
--- a/docs/developing/path.md
+++ b/docs/developing/path.md
@@ -19,10 +19,10 @@ option.
 
 * Example grain URL: https://sandstorm.example.com/grain/TPeYUde5rioE5keWM
 
-* Example ephemeral domain URL: https://96cab9109791f1254002ac1f857ecee7.sandstorm.example.com/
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
@orblivion Can you review this PR?

orblivion:
Change looks fine, but why "most" users? Why not all or none? Was this documentation _wrong_ before this fix?

orblivion:
:+1: to merge though

ocdtrekkie:
Yeah, if you tamper with the wildcard config you may have more things around the URL but there should always be a ui- in it for a grain view as of several years ago, and I just don't think this page was updated to reflect that.

PR #3721, opened 2024-11-21
Docs: Add Ubuntu 24.04 fix instructions

ocdtrekkie:
We've had a bunch of people have weird reports of issues with Ubuntu 24.04, and so far this has been the solution to all of them. I think we should codify it in the documentation page.

@Michael-S found this solution, and deserves all the credit.

### DIFF
--- a/docs/administering/install-troubleshooting.md
+++ b/docs/administering/install-troubleshooting.md
@@ -132,7 +132,7 @@ This should resolve your problems. If not, please get in touch.
 ## Are grains not starting on Ubuntu 24.04 or later?
 
 If your wildcard DNS is configured correctly, you should see the app icons correctly on the apps tab. If
-grains are not starting, and you are on Ubuntu 24.04 or later, AppArmor may be restricting unpriviliged
+grains are not starting, and you are on Ubuntu 24.04 or later, AppArmor may be restricting unprivileged
 user namespaces. You can run the following commands to correct this:
 
 ```bash
@@ -140,6 +140,9 @@ sudo sh -c 'echo "kernel.apparmor_restrict_unprivileged_userns = 0" > /etc/sysct
 sudo sysctl --system
 sudo service sandstorm restart
 ```
+
+As of this writing, this restriction is very uncommon in Linux distributions, but you may wish to consult
+updated guidance. An AppArmor profile for Sandstorm would be a welcome contribution.
 
 ## Did you disable outside collaborators before attempting to configure Google or GitHub login methods?
 
--- a/docs/administering/install-troubleshooting.md
+++ b/docs/administering/install-troubleshooting.md
@@ -129,6 +129,18 @@ sudo service sandstorm restart
 
 This should resolve your problems. If not, please get in touch.
 
+## Are grains not starting on Ubuntu 24.04 or later?
+
+If your wildcard DNS is configured correctly, you should see the app icons correctly on the apps tab. If
+grains are not starting, and you are on Ubuntu 24.04 or later, AppArmor may be restricting unpriviliged
+user namespaces. You can run the following commands to correct this:
+
+```bash
+sudo sh -c 'echo "kernel.apparmor_restrict_unprivileged_userns = 0" > /etc/sysctl.d/sandstorm-userns.conf'
+sudo sysctl --system
+sudo service sandstorm restart
+```
+
 ## Did you disable outside collaborators before attempting to configure Google or GitHub login methods?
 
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
@orblivion Can you review this PR?

Michael-S:
Thanks for the nice comment, but I just found other web pages when I had the issue:
https://askubuntu.com/questions/1511654/how-to-make-balena-etcher-and-other-appimages-work-in-ubuntu-24-04 ( https://web.archive.org/web/20241121143246/https://askubuntu.com/questions/1511654/how-to-make-balena-etcher-and-other-appimages-work-in-ubuntu-24-04 )
https://ubuntu.com/blog/ubuntu-23-10-restricted-unprivileged-user-namespaces ( https://web.archive.org/web/20241120223537/https://ubuntu.com/blog/ubuntu-23-10-restricted-unprivileged-user-namespaces )

So real credit to Archisman Panigrahi on Ask Ubuntu.

And I'm not qualified to judge the security implications of this fix myself, but the same Archisman that gave the fix wrote a comment section "About security concerns" about it here: https://askubuntu.com/questions/1511854/how-to-permanently-disable-ubuntus-new-apparmor-user-namespace-creation-restric/1511983#1511983 ( https://web.archive.org/web/20241121143424/https://askubuntu.com/questions/1511854/how-to-permanently-disable-ubuntus-new-apparmor-user-namespace-creation-restric/1511983#1511983 )

docs/administering/install-troubleshooting.md
```diff
@@ -129,6 +129,18 @@ sudo service sandstorm restart
...
 This should resolve your problems. If not, please get in touch.
 
+## Are grains not starting on Ubuntu 24.04 or later?
+
+If your wildcard DNS is configured correctly, you should see the app icons correctly on the apps tab. If
+grains are not starting, and you are on Ubuntu 24.04 or later, AppArmor may be restricting unpriviliged
+user namespaces. You can run the following commands to correct this:
+
+```bash
+sudo echo "kernel.apparmor_restrict_unprivileged_userns = 0" > /etc/sysctl.d/99-userns.conf
```
orblivion, line 139: Did you actually run this successfully? I think the `sudo` powers wear off when you hit the `>` token:

```
sudo echo "test" > /etc/test-test-test
bash: /etc/test-test-test: Permission denied
```

orblivion:
Maybe wouldn't hurt to link to that security implications page from this solution.

docs/administering/install-troubleshooting.md
orblivion: Also how comfortable are you with `99-userns` being unique enough? That line (once permissions are fixed) would overwrite anything the user might have had there if they happened to have the same filename.

ocdtrekkie: I can do some testing with our VPS demo. Maybe we should change it to sandstorm-userns.conf or something so it's extremely unique?

orblivion:
Testing as far as the permissions error? If that's what you mean, you can
confirm what I mean on any Linux system.

If "sandstorm" is in the name I would think that's plenty unique. And it
signals that it's supposed to be a new file specific to what we're doing.

ocdtrekkie:
I am sure you are right, but clearly I should go do this on an Ubuntu 24.04 system and make sure I end up with working instructions before I update this PR. =) The other troubleshooting steps on the page didn't assume we already had root, so I figure we should not assume that as well.

ocdtrekkie:
So... my test box I cannot not be root. :|

Research tells me `sudo sh -c 'echo "kernel.apparmor_restrict_unprivileged_userns = 0" > /etc/sysctl.d/sandstorm-userns.conf'` should work?

@orblivion Can you confirm that?

orblivion:
The command works, though don't forget the 99 (I don't know what it's there for but they all seem to have numbers)

ocdtrekkie:
I read that it could be anything ending with .conf, and I *did* test with this filename on Ubuntu 24.04. I do not know why they start with numbers traditionally either.

ocdtrekkie:
Does this updated note suffice for mergeability?

Apparently the number scheme it's just a convenience for ordering: https://unix.stackexchange.com/questions/570770/what-does-the-number-in-sysctl-configuration-file-denote

But in the case of using "99-" it serves the same purpose as just naming the file starting with "sandstorm-"... it's processed last alphabetically. IMHO, I'd rather be fully descriptive in the filename, i.e. sandstorm-userns over 99-sandstorm or 99-userns which is less clear what it's for or what it does.

orblivion:
+1

PR #3722, opened 2024-12-01
Add container-build tool

wwahammy:


ocdtrekkie:
Hey, sorry I did not ever see the description on this PR! All the notifications about pushes to it were before the description was added!

So adding an additional/separate action is probably "no big deal" from a merging standpoint, if it doesn't affect production code.

My biggest concern would be the change to find-meteor-dev-bundle.sh which I would probably only want to change on our `dev` branch at this time.

My second question would be if adding a Dockerfile to the root would lead people to believe this is used in our Docker install instructions, or a correct way to deploy Sandstorm in production. If these tools, including container-build.py are meant for testing, I feel like they should live in the /tests directory.

PR #3723, opened 2024-12-15
Update documentation links to point to the sandstorm.org domain instead of .io

neuroradiology:
I hope these commits to the old repository are welcome. If they get pushed to the site then it will help users get to where the active development is taking place.

These link updates also include one change to the dev usergroup, which is now on dev.io instead of Google groups.

If these types of updates are welcome then I will try to spend more time updating documentation links. I'm not a professional dev, so I hope my pull requests are being done correctly; I welcome feedback if there's something I can improve.

### DIFF
--- a/docs/developing.md
+++ b/docs/developing.md
@@ -3,8 +3,8 @@ apps on top of Sandstorm, package existing apps, and understand how apps and San
 
 <div class="developer-next-steps">
 <a class="next-step tutorial" href="../vagrant-spk/packaging-tutorial/">New? Try tutorial</a>
-<a class="next-step why" href="https://sandstorm.io/developer">Why Sandstorm</a>
-<a class="next-step discussion" href="https://groups.google.com/d/forum/sandstorm-dev">Community Q&amp;A</a>
+<a class="next-step why" href="https://sandstorm.org/developer">Why Sandstorm</a>
+<a class="next-step discussion" href="https://groups.io/g/sandstorm-dev-group/topics">Community Q&amp;A</a>
 <a class="next-step app-market" href="https://apps.sandstorm.io/">See App Market</a>
 <a class="next-step live-chat" href="https://kiwiirc.com/client/irc.libera.chat/?channel=#sandstorm">Live chat via IRC</a>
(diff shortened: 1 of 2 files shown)


ocdtrekkie:
Updates are definitely welcome, and I think it probably makes sense to point to updated community resources (the groups.io in particular).

Generally speaking I've avoided updating links on .io sites to point to .org to maintain the separation between the original Kenton-owned infrastructure that is still running and the community organization. But now that the sandstorm.io homepage itself describes itself as historical and points to sandstorm.org, it is probably reasonable to update the links in docs, which is used for non-historical purposes.

docs/developing.md
```diff
@@ -46,8 +46,8 @@ The Sandstorm ecosystem is full of people who want to promote your app, give you
 it.
 
 - **Getting help**: [Community feedback and Q&A](https://groups.google.com/d/forum/sandstorm-dev) | [Real-time IRC chat on libera.chat](https://kiwiirc.com/client/irc.libera.chat/?channel=#sandstorm) | [Watch presentations on the Sandstorm YouTube channel](https://www.youtube.com/channel/UC8xKZRW86Fa9W00uAppBXXg)
-- **Publicity**: [Give a meetup/conference talk about your app](https://sandstorm.io/news/2015-12-17-community-talks) | [Public demo service for all Sandstorm apps](https://sandstorm.io/news/2015-02-06-app-demo)
-- **Read more**: [All community resources](https://sandstorm.io/community)
+- **Publicity**: [Share news about a meetup/conference talk about your app](https://sandstorm.org/news/) | [Public demo service for all Sandstorm apps](https://sandstorm.io/news/2015-02-06-app-demo)
```
ocdtrekkie, line 49: I am not sure I would just point to /news here. And there's a second .io link on this line. Note that the .org site does include all the original blog posts, so you could just update the .io to .org for both.

ocdtrekkie (changes requested):

ocdtrekkie (approved):
I'm still a little iffy on that line 49 change eliminating a specific blog post, but we also probably need to get to the point we have an updated thing about promoting Sandstorm anyways. I'm gonna go with it for now.

ocdtrekkie:
@neuroradiology If you have any interest and the time, we could chat about some of the things we are looking to do with the docs in the future. Most of it is pretty code-light, more on rearrangement and such.

PR #3725, opened 2024-12-23
install script - trivial updates to fix io => org where possible, and remove broken link

neuroradiology:
There were a few places where the domain name change from sandstorm.io could be updated to sandstorm.org

(notably: *not* the documentation, at present)

There was also a link to an obsolete (2014) blog post, which was updated to a discussion *regarding* the post that took place on hacker news.


ocdtrekkie:
So this is probably somewhere we cannot yet make changes. We do not host anything at docs.sandstorm.org yet, and the install script is still installing the final sandstorm.io release of Sandstorm.

PR #3727, opened 2025-01-02
Add icons/package-lock.json

wwahammy:
For unclear reasons, the icons/package-lock.json file was missing. It bizarrely gets generated on every new build but wasn't event included. It makes sense to just include it since package-lock files should always be committed.


ocdtrekkie:
I am not sure why this wasn't committed (though I believe it is not the only package-lock.json that is not committed in this repo), but if we are going to address this it would be against the dev branch and with the current/newest dependencies, so I am going to close this one out for now.

PR #3731, opened 2025-05-13
Docs: Note experimental app market

ocdtrekkie:
@neuroradiology pointed out this was not documented well. Considering we recommend people try apps in experimental before we publish them, and developers may want to make sure their app looks good on the market view, I think this would be a good place to point it out as well.


neuroradiology:
Reviewed and agree.
Ok to use as-is, in my opinion.

It may also be appropriate to point out that you can upload an app to your own sandstorm server *without* publishing it to the app market (with the "upload" button on the install page). That subtlety was lost on me initially.

PR #3732, opened 2025-05-19
CI: Use latest Ubuntu

ocdtrekkie:
Actions no longer runs ubuntu-20.04, and we know why Sandstorm wasn't working on ubuntu-latest now, so this PR applies that fix from our documentation and then uses the latest runner.


ocdtrekkie:
Well, that didn't work. (Though I expect to still need this for tests.)

ocdtrekkie:
I am guessing we need to go to capnproto-latest to fix this, there are build fixed in the file that fails for newer C++ versions.

@kentonv indicated the blocker we pinned capnproto for is probably no longer an issue. Though we should test that.

finkiki:
Too bad the junkie isn't there anymore

ocdtrekkie:
Bah, changing the base branch is messy.

My plan here, I think: Instead of trying to fix the build on master, I am going to do so against the `dev` branch, and update deps including capnp, and see where we get/if it builds. *If* it builds, then I'll deploy it somewhere to test with and see how it handles existing Sandstorm apps. If it doesn't build... well, that's some more stuff we probably need to fix in addition to meteor stuff.

PR #3742, opened 2026-02-01
Docs: Sandcats recovery for existing servers

ocdtrekkie:
This documents the existence of the recover-sandcats.sh script @mnutt contributed. In trying to determine where to put a section for it, I discovered the docs already mention the potential to need to regenerate these... while providing a much more painful recovery plan involving spinning up a temporary VM just to run the install script. This is better.


neuroradiology (approved):
This recovery functionality is so valuable. Thanks to everyone involved in the pull request, the testing, and the documentation updates.

I support this PR.

PR #3746, opened 2026-02-05
Update docs dependencies

ocdtrekkie:
I am going to let docs-ci test if this breaks anything. There are basically no security vulnerabilities we care about because we static publish everything, but if the update works we might as well silence the alerts.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,3 @@
-jinja2==3.0.3
-markdown==3.3.7
-mkdocs==1.3.0
+jinja2==3.1.6
+markdown==3.10.1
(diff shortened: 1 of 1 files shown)


neuroradiology (commented):
I support this pull request.

I opened the built version of the docs. Aside from seeing several areas where we still need to update the domain from .io to .org (which is a separate issue), the build dependencies seem to have preserved functionality, and I see no errors.

ocdtrekkie:
Yeah, this is on master and deploys to docs.sandstorm.io still. Renaming everything for docs will need to come with us hosting docs on sandstorm.org.

PR #3747, opened 2026-02-05
Fix blog link, home page on about

ocdtrekkie:
This is not an extensive "find everywhere Sandstorm.io is so-branded, especially since we haven't actually replaced the infrastructure yet, but I figured I'd remove the prominent .io link on the About page. Turns out... the most prominent URL on our About page is broken.

blog.sandstorm.io redirects to sandstorm.io/news now, so all of these URLs become sandstorm.io/news/news, and hence 404. Figured I'd go ahead and point them at our version of the site while I'm at it.


neuroradiology (approved):
Approved. All the translations! I didn't even think about those; well-spotted.

PR #3751, opened 2026-03-07
Proposal: relax csp to allow web workers

mnutt:
This is opened as a draft for discussion only.

Back in 2020 it came to light that allowing Service Workers would allow them to modify response headers and potentially allow external requests to leak. The [fix](https://github.com/sandstorm-io/sandstorm/issues/3538) was `worker-src 'none'`. Unfortunately this had the side effect of also disallowing Web Workers.

Web Workers can be really useful for a lot of applications that need to do anything moderately complex on the front end. In my case, I'm trying to get a LibreOffice app to work but it requires a web worker. I could disable CSP at the server level but that doesn't feel like the right solution.

Web Workers (to my knowledge) don't have the same potential security issues as Service Workers, but there is no way to target just Service Workers. [(link)](https://lists.w3.org/Archives/Public/public-webrtc-logs/2021Jun/0298.html) [(link)](https://github.com/w3c/webappsec-csp/issues/638)

In the meantime, I think we should be able to protect against Service Workers by instead limiting all worker loads to `'self'` and then rejecting the actual Service Worker http requests. I'm not 100% sure of this solution, but it seems to work from what I can tell?

### DIFF
--- a/shell/imports/server/pre-meteor.js
+++ b/shell/imports/server/pre-meteor.js
@@ -56,6 +56,14 @@ function writeErrorResponse(res, err) {
   res.end(err.htmlMessage || err.message);
 }
 
+function isServiceWorkerScriptRequest(req) {
+  // Service worker registrations fetch script URLs with a dedicated header.
+  if ((req.headers["service-worker"] || "").toLowerCase().trim() === "script") return true;
+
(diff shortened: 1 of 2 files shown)


ocdtrekkie:
I'm not opposed conceptually but CSP is a realm I have truly minimal expertise.

PR #3753, opened 2026-03-31
Meteor 2.16 upgrade deps

mnutt:
This upgrades as many dependencies as possible, trying to rid ourselves of dependabot warnings.

There were two acme dependencies (namecheap and cloudflare) which had problematic sub-dependencies and no new version. I had to fork those in order to upgrade their sub-dependencies.

This re-enables the ip-networking tests; the issue was that the tests hit build.sandstorm.io and the response had changed. It now uses example.com which is typically stable but we may eventually want to use something we control.


.gitignore
```diff
@@ -81,3 +81,6 @@ tests/tmp-sandstorm
 app-index.spk
 test-app.spk
 deps/llvm-build
+package-lock.json
+shell/package-lock.json
+tests/package-lock.json
```
ocdtrekkie: Aren't package-lock files something we should generally be aiming to commit?

ocdtrekkie:
I do think otherwise I am feeling generally pretty good about this PR.

I am thinking of that fact we should probably implement a cooldown check on packages. npm has a setting now but it might be too new for where we are at. There's a GitHub Action though that could just check our PRs: https://nesbitt.io/2026/03/04/package-managers-need-to-cool-down.html

There was another malicious package compromise this week, and now that we're actually updating our Node packages, we probably should look at protecting ourselves there.

ocdtrekkie:
I am noticing c93020930edb459514076584f814422160988b51 deletes the npm-shrinkwrap file, but I guess that's apparently substantially identical to package-lock in our case? (I presume we do not need to support very old npm versions and we do not publish Sandstorm as a package in the NPM registry.)

From my understanding, I am on board here, I just want to be explicit about what we are changing.

ocdtrekkie (approved):

ocdtrekkie:
Ah that actually broke the build!

https://github.com/sandstorm-io/sandstorm/blob/master/Makefile#L287

ocdtrekkie:
Unsure why the CI run has been going for three hours but I cancelled it and reran it, let's see if it's a transient failure.

ragnar-howler (commented):
🤖 **Automated PR Review Complete**

This PR has been scanned for:
- ✅ Common security issues
- ✅ Code style consistency  
- ✅ Test coverage indicators
- ✅ Documentation completeness

**Stats:** +11221 -5869 in 23 files

Please address any automated checks above and request human review when ready.

troyjfarrell:
I won't say that I understood everything in this, but I did learn a bit about the Sandstorm shell.  Thanks for this PR!

PR #3757, opened 2026-04-18
Add migration script

ocdtrekkie:
So this script starts with a bunch of functions and features lifted straight from install.sh. But most of the "business" of install.sh has been removed and replaced with validating Sandstorm is installed (we should see a "sandstorm" command on the system), relaunching as root if necessary, asking the user to validate that they have backed up their install, and then downloading and applying the update file manually. (This requires unsetting and resetting the update channel in the Sandstorm config file.)

I intend to host a copy of this at https://install.sandstorm.org/migrate.sh in the future, so it should be curl|bashable from there once we get that far.

Note that at present, if you run this on an existing updated Sandstorm server, it should largely do nothing except turn off your update channel, because it will download the latest Sandstorm release (currently Kenton's), and apply it, and then fail when running migrate-mongo, and hence never re-set your update channel. I also have obviously not yet changed the GPG key to use for verification.

### DIFF
--- a/migrate.sh
+++ b/migrate.sh
@@ -1,0 +1,478 @@
+#! /bin/bash
+
+# This script updates your Sandstorm.io server to the new version
+# built by the Sandstorm Community project. This migration script
+# will download the latest release of the community version, then
+# update your Sandstorm instance and migrate the database to a
+# newer version of MongoDB.
+#
+# This script downloads and installs binaries. This means that to use this
+# script, you need to trust that the authors are not evil, or you must use
+# an isolated machine or VM. Of course, since the Sandstorm authors'
+# identities are widely known, if they did try to do anything evil, you
+# could easily get them arrested. That said, if you'd rather install from
+# 100% auditable source code, please check out the Github repository instead.
+#
+# All downloads occur over HTTPS from Sandstorm's servers and are further
+# verified using PGP.
+
+if test -z "$BASH_VERSION"; then
+  echo "Please run this script using bash, not sh or any other shell." >&2
+  exit 1
+fi
+
+# We wrap the entire script in a big function which we only call at the very end, in order to
+# protect against the possibility of the connection dying mid-script. This protects us against
+# the problem described in this blog post:
+#   http://blog.existentialize.com/dont-pipe-to-your-shell.html
+_() {
+
+set -euo pipefail
+
+# Declare an array so that we can capture the original arguments.
+declare -a ORIGINAL_ARGS
+
+USE_DEFAULTS="no"
+DEFAULT_UPDATE_CHANNEL="dev"
+
+# Allow the environment to override curl's User-Agent parameter. We
+# use this to distinguish probably-actual-users installing Sandstorm
+# from the automated test suite, which invokes the install script with
+# this environment variable set.
+CURL_USER_AGENT="${CURL_USER_AGENT:-sandstorm-migrate-script}"
+
+# Define I/O helper functions.
+error() {
+  if [ $# != 0 ]; then
+    echo -en '\e[0;31m' >&2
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
Two things I'm inclined to think about yet:

Should we be checking if we have a reasonable amount of free space available, and if so, how much is reasonable? Does the existing migrate-mongo command already do this?

Many other install script issues can report back error codes to our reporting grain on Alpha. Should we be capturing when Mongo Migrate fails, and do we have a good way to do that?

ragnar-howler (commented):
🤖 **Automated PR Review Complete**

This PR has been scanned for:
- ✅ Common security issues
- ✅ Code style consistency  
- ✅ Test coverage indicators
- ✅ Documentation completeness

**Stats:** +478 -0 in 1 files

Please address any automated checks above and request human review when ready.

ocdtrekkie:
So IIRC @mnutt said he saw a ~450 MB mongo drop to ~250 MB on a restart. In my case, I had a ~363 MB mongo, and it dropped to about ~225 MB on restart. The build for the dev branch is actually also fairly hefty (231 MB) because of all of the extra mongo binaries we have to ship, which have grown a lot since the old versions Sandstorm used to host. I am not positive what order we download, unpack, backup, migrate, etc. but I feel like we might have an extra copy of the mongo database and at least one copy of the new release during this process.

But I feel like if we put a sanity check in to say have 1 GB of free space on the disk your Sandstorm server is on in order to migrate, that is both likely to work for most users and also likely to cover a lot of servers being able to migrate successfully.

migrate.sh
```diff
@@ -0,0 +1,501 @@
...
+# built by the Sandstorm Community project. This migration script
+# will download the latest release of the community version, then
+# update your Sandstorm instance and migrate the database to a
+# newer version of MongoDB.
+#
+# This script downloads and installs binaries. This means that to use this
+# script, you need to trust that the authors are not evil, or you must use
+# an isolated machine or VM. Of course, since the Sandstorm authors'
+# identities are widely known, if they did try to do anything evil, you
+# could easily get them arrested. That said, if you'd rather install from
```
neuroradiology, line 13: Suggest removing

> Of course, since the Sandstorm authors'
> identities are widely known, if they did try to do anything evil, you
> could easily get them arrested. That said,` 

The security call-out is important but I discussion of arresting the authors is weird.

ocdtrekkie: Hah, that is a carry-over from the install script, but I do agree we can remove it.

migrate.sh
```diff
@@ -0,0 +1,501 @@
...
+assert_sandstorm_installed() {
+  if ! which sandstorm > /dev/null 2>&1; then
+    fail "E_SANDSTORM_NOT_INSTALLED" \
+      "Sandstorm does not appear to be installed on this system. This script migrates an existing Sandstorm installation and cannot be run on a system without Sandstorm."
+  fi
+}
+
+assert_running_as_root() {
+  if [ "$(id -u)" != "0" ]; then
+    rerun_script_as_root
```
neuroradiology, line 294: Is this our preferred behavior? 

If the script has not been run as root then the script is automatically escalating itself. A root password will still be required, but I don't love that it happens automatically.

I would prefer failure, since it's only necessary to for the user to enter the command again as root (trivial).

However if this functionality is necessary, then my opinion is that it warrants a confirmation prompt. Failing to do so isn't the end of the world (the password is needed, after all), but it just seems like the most congenial behavior.

neuroradiology (approved):
Migration script reviewed. Ergonomic comments provided, but I approve the PR.

neuroradiology (approved):
Reviewed and approved.

PR #3760, opened 2026-05-07
Passkey auth

mnutt:


RyanSinger:
Built on top of this PR: #3762 adds passkey-only account creation using an identifier-first flow (email input + server-side routing to auth or registration). New users can now create an account with just a passkey, no Google sign-in required.

PR #3762, opened 2026-05-09
Passkey account creation: identifier-first flow

RyanSinger:
## Summary

Builds on #3760. Adds passkey-only account creation so users can sign up without needing Google or another provider first.

- Replaces bare "with Passkey" button with email input + "Continue with Passkey" form
- Server routes to authentication or registration based on whether the email has a passkey
- New users get a "Create a passkey" prompt (Touch ID, etc.) instead of the confusing "insert security key" dialog
- Returning users get a "Use your passkey" prompt
- Adds WebAuthn conditional UI (passkeys appear in browser autofill)
- Stores email on passkey credential for server-side routing lookup

## Test plan

- [ ] New user: enter email, click Continue with Passkey, complete registration, verify account created
- [ ] Returning user: enter same email, verify authentication (not registration)
- [ ] Conditional UI: click email field, verify passkey appears in autofill dropdown
- [ ] Error cases: invalid email, cancelled ceremony, HTTP context
- [ ] Existing account settings passkey management still works


ocdtrekkie:
So one key issue we would have here is that we need to either: Require/establish an alternative authentication method, which could just be email since it's collected here, or establish a method for an administrator to provide a reset or recovery path.

Currently Sandstorm administrators cannot change users' authentication methods at all, so if they have only a passkey and lose their passkey, their account would be unrecoverable.

PR #3763, opened 2026-05-09
Ensure update is always run before apt-get

mnutt:
Fixes an issue where our cache of apt locations differs from the apt repository's data.

PR #3766, opened 2026-05-09
Add SERVER_RUNTIME environment variable

mnutt:
This defaults to `classic` and is readable by Sandstorm grains as `SERVER_RUNTIME=classic`. It can only be modified by changing the constant in the source.


mnutt:
this will pass when https://github.com/sandstorm-io/sandstorm/pull/3763 is merged and this is rebased.

ocdtrekkie (approved):
I love that the actual code change here is a mere two lines.

I have also tested this in vagrant-spk using the CI build, which is pretty cool.

PR #3768, opened 2026-05-14
Get test apps from sandstorm.org

ocdtrekkie:
I think this will have tests pull all test apps from dl.sandstorm.org (hosted on Cloudflare R2) instead of depending on either the Sandstorm.io website or Sandstorm Alpha. This means tests should not break in a way we need Kenton to fix.


neuroradiology (approved):
Outstanding to get these testapps centralized and organized!

Approved.

PR #3770, opened 2026-05-29
Fix #3699: powerbox dialog: allow long URLs to wrap instead of trunca…

### LINKED ISSUE #3699: Lengthy Powerbox requests cannot be read in full
On TTRSS, I have been getting a Powerbox prompt that looks like this:
image

Hovering over it does nothing, which is how I expected I would see the rest of the URL. Luckily, I was able to find which feed was requesting it as well as the full URL in the debug log, since TTRSS and Ian's proxy both log things effectively there. But this isn't a great pattern, especially when the actual domain is off the end of the view entirely.

I think this should either be a multiline display (preferably), or show the full URL on hover.

berbad:
Long URLs in the Powerbox request dialog were being truncated with an ellipsis. This CSS change allows them to wrap onto multiple lines so the full URL is visible.
Fixes #3699

### DIFF
--- a/shell/client/styles/_powerbox.scss
+++ b/shell/client/styles/_powerbox.scss
@@ -64,12 +64,14 @@ body>.popup.request>.frame-container>.frame {
       display: block;
       vertical-align: middle;
       width: 100%;
-      height: 32px;
-      line-height: 32px;
+      min-height: 32px;
+      line-height: 1.4;
       background-size: 24px 24px;
       background-position: 4px 4px;
       background-repeat: no-repeat;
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
Thank you! Can you make these changes against the `dev` branch? We are going to be building our next release from there, and it has significant other improvements to the codebase. (If I change the base branch in GitHub, it tries to pull every other difference between the branches, which is now significant.)

berbad:
Closing in favor of #3773, which rebases this same fix onto the `dev` branch as requested.

PR #3771, opened 2026-05-30
Default to mongo 7 on new installs

mnutt:
I spent so much time testing upgrading an existing install I had an oversight in new installs not starting on MongoDB 7 by default.

The installer already has a check to make sure it isn't installing over an existing installation and we rely on that to not accidentally have it mark an existing install as upgraded.


ocdtrekkie:
Hah, tests were happy with it for fresh install but that's because this was done in tests. Amazing.

ocdtrekkie (approved):

PR #3772, opened 2026-06-03
installer adds apparmor profile on ubuntu 24.04

mnutt:
I tried this out in a Ubuntu 24.04 vm and it seemed to work fine, allowing me to start grains with userns. I can confirm that it installs and loads the apparmor profile, which is the same one we manually install in the CI build.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,7 +40,7 @@ jobs:
             which go || (sudo apt-get update && sudo apt-get install golang-go)
 
             sudo apt-get update && sudo apt-get install -y --no-install-recommends \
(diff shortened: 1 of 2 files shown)


ocdtrekkie:
Shouldn't we be able to remove setting up the app armor profile in the test and allow the Sandstorm install script to do it?

mnutt:
The issue is actually that we never actually _run_ the installer in the tests, we just use `make` to run the tests.

ocdtrekkie:
Fair enough! (Though unfortunate.)

I think this is pretty unobjectionable and you and CI are happy with it, so I am going to merge it.

ocdtrekkie (approved):

PR #3773, opened 2026-06-03
Fix #3699: powerbox dialog: allow long URLs to wrap instead of truncating

### LINKED ISSUE #3699: Lengthy Powerbox requests cannot be read in full

berbad:
Long URLs in the Powerbox request dialog were being truncated with an ellipsis. This CSS change allows them to wrap onto multiple lines so the full URL is visible.

Fixes #3699

This is a rebase of #3770 onto `dev` (as requested), so the diff is limited to the single `_powerbox.scss` change.

### DIFF
--- a/shell/client/styles/_powerbox.scss
+++ b/shell/client/styles/_powerbox.scss
@@ -64,12 +64,14 @@ body>.popup.request>.frame-container>.frame {
       display: block;
       vertical-align: middle;
       width: 100%;
-      height: 32px;
-      line-height: 32px;
+      min-height: 32px;
+      line-height: 1.4;
       background-size: 24px 24px;
       background-position: 4px 4px;
       background-repeat: no-repeat;
-      padding-left: 32px;
+      padding: 4px 4px 4px 32px;
+      word-break: break-all;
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
Thanks! As the person who found this bug annoying, I am excited to get this PR tested! I think the CI failure was transient/unrelated, so I am rerunning it now.

ocdtrekkie:
This looks good to me. All of the changes are confined to the "powerbox card" style, so I don't think it can affect anywhere else in the UI.

This was a test using a current Sandstorm build:
originalbehavior

And this was a test using the build artifact from this PR:
newbehavior

mnutt (approved):
Nice, it's a security improvement too.

PR #3774, opened 2026-06-03
Cherrypicking from master

ocdtrekkie:
So we have some conflicts between dev and master that have cropped up, and in most cases I don't care: Most of the commits to master have been solely for the docs, which we have moved off to a separate repo. (Eventually we can probably delete the entire docs tree from the dev branch, but I wanted to settle the differences between the branches first. There are a couple metadata tweaks here we should keep, plus the recover-sandcats script which I don't want to lose when we eventually are the default branch.

This PR grabs the few useful commits from master and cherrypicks them onto dev.

### DIFF
--- a/.well-known/funding-manifest-urls
+++ b/.well-known/funding-manifest-urls
@@ -1,0 +1,1 @@
+https://sandstorm.org/funding.json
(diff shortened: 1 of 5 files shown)


ocdtrekkie:
Actually, these are already merged commits, and also noncode, and I'm just gonna merge them here.

PR #3775, opened 2026-06-04
Update changelog for build 0.309

ocdtrekkie:
I think we are feature-locked for this release, so here is the changelog I expect to ship with release 309. The only outstanding changes to come are the update and signature checks and the migration script. The date will need to be updated on release day.

@mnutt and @berbad are particularly cited for contributions in this release, please let me know if there are any noteworthy or substantial changes we should yet mention here. (Obviously more detail on migration, trustworthiness of the upgrade, testing, etc. will go into the blog post, so this is much more succinct.)


neuroradiology (approved):
Reviewed and approved

PR #3776, opened 2026-06-05
Use docs,dl,install.sandstorm.org

ocdtrekkie:
Most of this only affects comments, links in UI, etc. but I went ahead and also updated the update check URL to look at our .org resources as well here. (see run-bundle.c++ for the meat there) This should be a pretty trivial one to review, just a lot of routine find-and-replace, but I figured I should get it filed in.


neuroradiology (approved):
Reviewed and approved.

All changes relate to sandstorm.io => sandstorm.org, and appear correct/appropriate.

PR #3779, opened 2026-06-17
Replace release GPG key

ocdtrekkie:
This PR replaces the signing key checked for by the install and migrate scripts to our new release key. Note, this does not include the update-sig which is an ed25519 key, nor does this address the documented "GPG-verified install" process documented here: https://docs.sandstorm.org/en/latest/install/#option-3-pgp-verified-install as it talks about keys being rotated monthly and such and mention the previous team members. We can address this later.

After this is merged, I will update install.sandstorm.org and dl.sandstorm.org with the updated install script and the updated signature file for v308.


orblivion (approved):
Confirmed that the public keys and fingerprints in this PR match each other, and that they allow me to verify a signature made by our shared Yubikey.

troyjfarrell (approved):
I have verified that the keys and fingerprints match my copy of the signing key.

